PR #7, opened 2022-09-28
setup basic husky, lint-staged and github actions

### LINKED ISSUE #5: Husky + Lint-staged + GH Actions
We did something similar on scaffold-eth a few weeks ago. Also check the scaffold-eth-typescript version of it.

https://github.com/scaffold-eth/scaffold-eth/pull/864

Should be pretty similar, but there might be some differences since we are using a different stack.

technophile-04:
### Description : 
Setted up a basic husky, lint-staged and GitHub actions workflow, The `pre-commit` and `gh-actions` are currently scoped to frontend only. The workflow is very similar to https://github.com/scaffold-eth/scaffold-eth/pull/864. 

### Quick Summary of the current workflow 
****Local flow :**** 
If there are any warnings in the front end you will be able to commit those changes locally but will get a warning while committing. 
Screenshot 2022-09-29 at 12 17 45 AM

****Github flow:**** 
If you push to the remote repository with warnings in the frontend, gh-actions will fail. 

### Design choice 
In root `package.json` I previously tried setting up `lint-staged` like this : 
```javascript
"lint-staged": {
    "packages/frontend/**/*.{ts,tsx}": ["eslint --fix", "prettier --write"]
  }
```
and was getting this error : 
Screenshot 2022-09-28 at 11 37 56 PM

Just putting it out here so that we don't miss out on something, Currently, we are using it similar to `scaffold-eth-typescript` [configuration](https://github.com/scaffold-eth/scaffold-eth-typescript/blob/8a1e540c0007a955a99498efa6e71c6101324834/package.json#L90) ..I feel they both are almost the same and the current solution also works fine. 

1) Should we create a separate `.lintstagedrc.json` in the root dir? 
2) Should use `next:lint` with `lint-staged` like  [here](https://nextjs.org/docs/basic-features/eslint#lint-staged), I tried finding some solutions with this approach but didn't get any good one, If we want this, I will dig in more deep.

### ToDo 
- [ ] Setup `eslint` and `solhint` and `prettier` for hardhat. Consider `hardhat` warnings and merge both the flows .
- [ ] Accordingly set ghActions for both hardhat and frontend
- [ ] Setup a PR CI in GitHub 

Would really like it if someone mentions the priority in which we should go 🙌. 
(fixes #5 )


carletex:
Thanks for the detailed PR :)

Check the review comments: there are some issues. Maybe we can also take a different approach with next lint:

> 1. Should we create a separate .lintstagedrc.json in the root dir?
> 2. Should use next:lint with lint-staged like [here](https://nextjs.org/docs/basic-features/eslint#lint-staged), I tried finding some solutions with this approach but didn't get any good one, If we want this, I will dig in more deep.

If you can make next:lint with lint-staged... I think that would be ideal. It seems to be the most "integrated" way of doing it. You might be running into issues because we are in a monorepo? "Root folder" might be the "frontend" folder.

I found this example (not monorepo) but it might help you.

https://amanhimself.dev/blog/setup-nextjs-project-with-eslint-prettier-husky-lint-staged/

Happy to help if you need!

carletex:
PS: @technophile-04 You have write access to the repo, so feel free to create branches here for the PRs. No need to use your own repo if you don't want to!

.lintstagedrc.js
```diff
@@ -0,0 +1,10 @@
+const path = require('path');
+
+const buildEslintCommand = (filenames) =>
+	`yarn next:lint --fix --file ${filenames
+		.map((f) => path.relative(path.join('packages','frontend'), f))
+		.join(' --file ')}`;
+
+module.exports = {
+	'packages/frontend/**/*.{ts,tsx}': [buildEslintCommand],
+};
```
carletex, line 10: This has a weird formatting. I think its using tabs instead of space.

Also not following the prettier rules. (double quote, etc)

package.json
```diff
@@ -13,7 +13,13 @@
     "deploy": "yarn workspace @se-2/hardhat deploy",
     "hardhat:test": "yarn workspace @se-2/hardhat test",
     "start": "yarn workspace @se-2/frontend dev",
-    "lint": "yarn workspace @se-2/frontend lint"
+    "next:lint": "yarn workspace @se-2/frontend lint",
```
carletex, line 16: We can add next:format also here!

packages/frontend/package.json
```diff
@@ -6,7 +6,8 @@
     "dev": "next dev",
     "build": "next build",
     "start": "next start",
-    "lint": "next lint"
+    "lint": "next lint",
+    "format": "yarn prettier -w"
```
carletex, line 10: If I run this task directly (entering into the folder and running "yarn format") nothing happens.

I guess it's because we are not specifying the sources files? From scaffold-eth: `"prettier": "npx prettier --write . '!(node_module|build)/**/*'"`

Should be similar. But I think we don't need "yarn" nor "npx" and can run prettier directly

carletex (commented):
Great work getting the next:link to work!! Way cleaner :)

Left some comment with some issues.

Thanks Shiv!

.lintstagedrc.js
```diff
@@ -0,0 +1,10 @@
+const path = require('path');
+
+const buildEslintCommand = (filenames) =>
+	`yarn next:lint --fix --file ${filenames
```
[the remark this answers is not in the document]
carletex: My bad! It's fixing it. Closing this :D

carletex (approved):
Added a couple of tweaks.

Ready to merge! Good work :)

Issue #1, opened 2022-09-26
Create a burner wallet implementation that works with wagmi and rainbow


ShravanSunder:
## Overview
- [x] create a burner wallet connector for wagmi
- [x] add burner connector to rainbow wallet
- todo: #9

## Details
Based on config values
- should log into burner wallet in the background on startup for localhost
- should disable burner wallet for other networks
- config to disable burner wallet entirely
- config to disable auto login to burner on localhost

carletex:
@ShravanSunder  Still not sure about how the `scaffold.config.json` would work. But some thoughts:

- How this co-exist or how is it different from a .env file?
- I think it should be `js` not `json`. So we can have some JS logic (like getting env vars there)
- I think it should live in the root
- It should be there and work out of the box, no need to generate it and no need to configure anything if you don't want to.

ShravanSunder:
@carletex  some thoughts
1. we can assume defaults and use that without generating it.  I"m also good with ts file.  
2. env variables are different than config.  env variables are keys that are based on environment.  So you can have API keys that are not checked in and based on prod, staging, dev env.  They allow layering.  This will be good for api keys
3. I think we can load env variables in the config.  
4. config should be fore things like the burner wallet settings.  they are not env variable based and impact runtime.

ShravanSunder:
@carletex here's a video overview, its ready for PR:  https://www.dropbox.com/s/f4trel929a2jeou/2022.09.30-21.07-Brave%20Browser%20Scaffod-eth%20App.mp4?dl=0

I'll do the config as a seperate PR i think.  better to split things up.  

this is the burner config so far
```
export type TAutoConnect = {
  /**
   * Enable the burner wallet.  If this is disabled, burner wallet is entierly disabled
   */
  enableBurnerWallet: boolean;
  /**
   * Always autoConnectToBurner to burner wallet on page load.
   */
  alwaysAutoConnectToBurnerOnLoad: boolean;
  /**
   * If user is not connected to any wallet:  On reload, connect to burner wallet
   */
  connectToBurnerIfDisconnectedOnLoad: boolean;
  /**
   * Auto connect: If the user was connected into a wallet before, on page reload reconnect automatically
   */
  allowAutoConnect: boolean;
};
```

ShravanSunder:
was thinking to simplify the options.  @carletex 
```ts
export type TAutoConnect = {
  /**
   * Enable the burner wallet.  If this is disabled, burner wallet is entierly disabled
   */
[...]

PR #8, opened 2022-09-29
#1 Burner Wallet

ShravanSunder:
@carletex here's a video overview, its ready for PR:  https://www.dropbox.com/s/f4trel929a2jeou/2022.09.30-21.07-Brave%20Browser%20Scaffod-eth%20App.mp4?dl=0

I'll do the config as a seperate PR i think.  better to split things up.  

this is the burner config so far, please see video for details.  Let me know if anything needs improving
```ts
export type TAutoConnect = {
  /**
   * Enable the burner wallet.  If this is disabled, burner wallet is entierly disabled
   */
  enableBurnerWallet: boolean;
  /**
   * Always autoConnectToBurner to burner wallet on page load.
   */
  alwaysAutoConnectToBurnerOnLoad: boolean;
  /**
   * If user is not connected to any wallet:  On reload, connect to burner wallet
   */
  connectToBurnerIfDisconnectedOnLoad: boolean;
  /**
   * Auto connect: If the user was connected into a wallet before, on page reload reconnect automatically
   */
  allowAutoConnect: boolean;
};
```


packages/frontend/components/hooks/useAutoConnect.ts
```diff
@@ -0,0 +1,92 @@
...
+import { useEffectOnce, useLocalStorage } from "usehooks-ts";
+import { burnerWalletId, defaultBurnerChainId } from "~~/web3/wagmi-burner";
+import { useBurnerWallet } from "~~/components/hooks/useBurnerWallet";
+
+// const supportedRpcUrls = [
+//   chain.localhost.rpcUrls.default,
+//   chain.localhost.rpcUrls.default.replace("127.0.0.1", "localhost"),
+//   chain.hardhat.rpcUrls.default,
+//   chain.localhost.rpcUrls.default.replace("127.0.0.1", "localhost"),
+// ];
```
carletex, line 12: Remove this?

packages/frontend/components/hooks/useAutoConnect.ts
```diff
@@ -0,0 +1,92 @@
...
+ * @returns
+ */
+const getInitialConnector = (
+  config: TAutoConnect,
+  previousWalletId: string,
+  connectors: Connector<any, any, any>[],
+): { connector: Connector | undefined; chainId?: number } | undefined => {
+  const allowBurner = config.enableBurnerWallet;
+
+  if (!!!previousWalletId) {
```
carletex, line 43: I think using only 1 bang should be enough :D

packages/frontend/components/hooks/useAutoConnect.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  useEffect(() => {
+    if (accountState.isConnected) {
+      // user is connected, set walletName
+      setWalletId(accountState.connector?.id ?? "");
+    } else {
+      // user has disconnected, reset walletName
+      setWalletId("");
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
carletex, line 78: Why disabling this?

`[accountState, setWalletId]` as deps cause any issues?

packages/frontend/pages/index.tsx
```diff
@@ -34,7 +55,9 @@ const Home: NextPage = () => {
         <Address address="0xd8da6bf26964af9d7eed9e03e53415d37aa96045" />
 
         <p>
-          <button className="btn btn-primary">Daisy UI Button</button>
+          <button className="btn btn-primary" onClick={() => tempTest.onClick()}>
```
carletex, line 58: The tx with the text contract doesn't seem to be working for me. What do I need to do?

image

The unknown account is my burner wallet address

packages/frontend/tsconfig.json
```diff
@@ -13,7 +13,10 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "incremental": true
+    "incremental": true,
+    "paths": {
```
carletex, line 17: What does this do?

carletex (commented):
Hey @ShravanSunder 

Left some review comments / questions.

I also pushed a couple of commits cleaning some stuff. If you need to do some tweaks to config stuff (nextjs, lint, etc), could you add a PR comment on the given code line explaining why is it needed? If not, some changes might seem random changes / personal preferences tweaks.

packages/frontend/components/useTempTestContract.tsx
```diff
@@ -0,0 +1,38 @@
...
+    addressOrName: testContractAddress,
+    contractInterface: tempContract.abi,
+    functionName: "purpose",
+  });
+
+  useEffect(() => {
+    if (cr.isSuccess) {
+      console.log(cr.data);
+    }
+  }, [cr.status, cw.status]);
```
carletex, line 32: would `[cr, cw]` as deps work? Rn getting a missing dependencies error.

packages/frontend/components/hooks/useAutoConnect.ts
ShravanSunder: actually you need 3.  !!! for a string means true if null, undefined or empty

packages/frontend/pages/index.tsx
ShravanSunder: you have a local chain running?

packages/frontend/tsconfig.json
ShravanSunder: paths allows you not to use relative path.  The root of the dir can be accessable with ~~/pages
rather than ../../pages etc
This also allows typescript to correct and validate imports when you move files.

packages/frontend/components/useTempTestContract.tsx
ShravanSunder: objects might change.  They are passed by reference, its generally not a good idea to use reference based variables for dependncies when you can use values

packages/frontend/pages/index.tsx
ShravanSunder: hmm i'll look into this

packages/frontend/tsconfig.json
carletex: Make sense. Thanks

packages/frontend/components/useTempTestContract.tsx
carletex: They shouldn't change in this case. But i'm fine with leaving it as it is. 

But you need to add the eslint-disable then.

packages/frontend/pages/index.tsx
ShravanSunder: i pushed some changes @carletex  There was a bug.
Also a reminder, before it works, it needs yarn chain, and eth for the burner account.

ShravanSunder:
> Hey @ShravanSunder
> 
> Left some review comments / questions.
> 
> I also pushed a couple of commits cleaning some stuff. If you need to do some tweaks to config stuff (nextjs, lint, etc), could you add a PR comment on the given code line explaining why is it needed? If not, some changes might seem random changes / personal preferences tweaks.

I've pushed with the commit comments.   Sorry i mistook pr comments for commit comments.  can add pr comments to if you like.

packages/frontend/.eslintrc.json
```diff
@@ -1,6 +1,12 @@
...
   "extends": ["next/core-web-vitals", "plugin:prettier/recommended"],
   "rules": {
-    "no-unused-vars": "error"
+    "no-unused-vars": "error",
+    "prettier/prettier": [
+      "warn",
+      {
+        "endOfLine": "auto"
+      }
+    ]
```
carletex, line 10: Ok, I see you put some comments in the commits :D It's a bit better if you add the comments here (easy to review!)

packages/frontend/pages/index.tsx
carletex: Ok, now it works

carletex (commented):
@ShravanSunder thanks for the updates. Check review comments.

A couple of things more:

- When ever I reload the page I get a different burner wallet address. the localstorage logic not working?

- Now I'm getting this errors when loading the page:

image

image

Maybe we are making a call before some data available?

packages/frontend/.eslintrc.json
ShravanSunder: its better if its warn, as it'll autocorrect without interfering with users.  I think all eslint, prettier autcorrect should be warns as that stops them from interfering with the user (but the rules are still enforced).

.vscode/settings.json
```diff
@@ -0,0 +1,34 @@
+{
```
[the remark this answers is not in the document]
ShravanSunder: that's fair.  in scaffold-eth-typescript, i used visual code settings to provide defaults as most beginners start with that (from stats).

ShravanSunder:
> @ShravanSunder thanks for the updates. Check review comments.
> 
> A couple of things more:
> 
> * When ever I reload the page I get a different burner wallet address. the localstorage logic not working?
> * Now I'm getting this errors when loading the page:
> 
> image
> 
> image
> 
> Maybe we are making a call before some data available?

fix the issue with local storage.

The errors are because of tempContract, it needs contract to be deployed and is just for our current testing.  you have to change the address of tempContract to the deployed contract.

.vscode/settings.json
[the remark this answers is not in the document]
carletex: Yep, I see your point.

But following the simple & minimal approach.... when in doubt, we should go with the thing that add less stuff to the repo :)

carletex (approved):

PR #11, opened 2022-10-04
Feature/  address input component

Naim-Bijapure:
basic address input component 

https://user-images.githubusercontent.com/22323693/193857210-347ae79e-9be1-4a7c-a1e6-491233e13a99.mp4


packages/frontend/components/scaffold-eth/AddressInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+  new Promise(resolve => setTimeout(() => resolve(true), time));
+
+/**
+ Address input component with ens name resolution
+*/
+
+const AddressInput = ({ value, onChange, placeholder }: IAddressInput): any => {
+  const { data: ensData, isLoading } = useEnsAddress({
+    name: isENS(value) ? value : "",
+  });
```
carletex, line 23: We want to use `chainId: 1` here to always use ENS resolution on mainnet

carletex, line 23: We want to use cache time also. It can be a large number.

packages/frontend/components/scaffold-eth/AddressInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+export const Sleep = async (time: number): Promise<any> =>
+  new Promise(resolve => setTimeout(() => resolve(true), time));
+
+/**
+ Address input component with ens name resolution
+*/
+
+const AddressInput = ({ value, onChange, placeholder }: IAddressInput): any => {
+  const { data: ensData, isLoading } = useEnsAddress({
+    name: isENS(value) ? value : "",
```
carletex, line 22: A cleaner way is to use the enabled prop.
```
{
  name: value,
  enabled: isENS(value)
}
```

packages/frontend/components/scaffold-eth/AddressInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+const isENS = (address = "") => address.endsWith(".eth") || address.endsWith(".xyz");
+
+export const Sleep = async (time: number): Promise<any> =>
+  new Promise(resolve => setTimeout(() => resolve(true), time));
+
+/**
+ Address input component with ens name resolution
+*/
+
+const AddressInput = ({ value, onChange, placeholder }: IAddressInput): any => {
```
carletex, line 20: I think this components need to have it internal state. We don't want to force the "user" of this component to define an state to use this component.

packages/frontend/components/scaffold-eth/AddressInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+      await Sleep(100);
+      onChange(ensData);
+    }
+
+    if (ensData === null && onChange) {
+      // add a little delay on load
+      await Sleep(100);
+      onChange("");
+    }
+  }, [ensData, onChange]);
```
carletex, line 42: Don't understand the need of this.

We just need the useEffect when we get the ENS resolution.

packages/frontend/components/scaffold-eth/AddressInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  useEffect(() => {
+    loadEnsAddress();
+  }, [ensData, onChange, loadEnsAddress]);
+
+  return (
+    <>
+      <div className={`form-control`}>
+        <label className="input-group input-group-sm">
+          <input
```
carletex, line 52: If you introduce a ens that doens't resolve (e.g. sadsadsadsa.eth) the input gets cleaned.

It should be show in red or something, but not removing the input data.

packages/frontend/pages/index.tsx
```diff
@@ -13,6 +15,7 @@ const tempAutoConnectConfig: TAutoConnect = {
 };
 
 const Home: NextPage = () => {
+  const [address, setAddress] = useState<string>("");
```
carletex, line 18: See the other comment on not needing this external state.

carletex (changes requested):

packages/frontend/components/scaffold-eth/AddressInput.tsx
Naim-Bijapure: ok moved the states to internal. but should we give the input result with onChange or not? 
using onChange to give ens resolved back let me know its ok or not.

carletex: Good question!

It's tricky because it'll depend on the use we give to this component. But yeah, I think we should the value back to the parent component.

We have two options here:
1. **onChange**. Send back the raw input of the user (+ the resolved address if he wrote an ENS)
2. **onSuccess**: Send back only when we have a VALID address (because the user pasted the address directly, or because the ENS resolution was succesfulll)

What do you think?

Also, let's not worry too much, we can adapt in the future depending on the needs.

carletex (approved):

PR #12, opened 2022-10-04
 Title: update:balance component basic ui

Naim-Bijapure:
basic balance component

[Screencast from 04-10-22 06:15:14 PM IST.webm](https://user-images.githubusercontent.com/22323693/193857374-f032c4c8-1589-4c3c-82d2-a8acd60f9166.webm)


packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+type BalanceProps = {
+  address: string;
+  price: number;
+  minimized?: boolean;
+  wrapperClasses?: string;
+};
+/**----------------------
+ * Display balance of an address
+ * ---------------------*/
```
carletex, line 12: Can you use the same "comment style" as in Address.tsx? Something like:

```js
/**
 Display balance of an address
**/

Don't care with style we use, but let's try to be consistent.

File: packages/frontend/components/scaffold-eth/Balance.tsx [hunk 2]
```diff
@@ -0,0 +1,80 @@
...
+
+export default function Balance({ address, price, minimized, wrapperClasses }: BalanceProps) {
+  const [isEthBalance, setIsEthBalance] = useState<boolean>(true);
```
carletex, line 15: No need to use the <boolean> type if it can be inferred from the default value

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
...
+    addressOrName: address,
+  });
+
+  let displayAddress = address?.slice(0, 5) + "..." + address?.slice(-4);
+
+  const onToggleBalance = () => {
+    setIsEthBalance(!isEthBalance);
+  };
+
+  //   on loading screen or no address supplied
```
carletex, line 31: There is an extra space there.

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <div className="animate-pulse flex space-x-4">
+        <div className="rounded-md bg-slate-300 h-6 w-6"></div>
+        <div className="flex items-center space-y-6">
+          <div className="h-2 w-28 bg-slate-300 rounded"></div>
+        </div>
+      </div>
+    );
+  }
+
+  //on fetch balance error
```
carletex, line 43: There is no space after //.

Also consider that we might not need these kind of comments. if (isError) is self explanatory :P Doesn't bother me if you want to leave it tho

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
+import { useState } from "react";
+import { useBalance } from "wagmi";
+
+type BalanceProps = {
+  address: string;
+  price: number;
+  minimized?: boolean;
```
carletex, line 7: I'd remove this option for now.

We just want to show the balance for a given address in a nice format + the eth/usd conversion (I know we are leaving this for now).

If showing the address it's something we need in the future, we'll add it!

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <div
+      className={` border-2 border-gray-400 rounded-xl p-2 flex flex-col items-center max-w-fit cursor-pointer  ${wrapperClasses}`}
+      onClick={onToggleBalance}
+    >
+      {minimized === false && <div className="text-[10px] opacity-70">{displayAddress}</div>}
+
+      {/* display  eth or doller balance  */}
+      <div className="w-full flex items-center  justify-center">
+        {isEthBalance === true && (
```
carletex, line 64: this can be just {isEthBalance &&

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <>
+            <span>{balanceData?.formatted}</span>
+            <span className="text-xs font-bold m-1">ETH</span>
+          </>
+        )}
+
+        {isEthBalance === false && (
```
carletex, line 71: this can be just {!isEthBalance &&

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
...
+export default function Balance({ address, price, minimized, wrapperClasses }: BalanceProps) {
+  const [isEthBalance, setIsEthBalance] = useState<boolean>(true);
+
+  const {
+    data: balanceData,
+    isError,
+    isLoading,
+  } = useBalance({
+    addressOrName: address,
+  });
```
carletex, line 23: This doesn't refresh and we want it to (like in SE-1)

We need to explore wagmi options for this hook: watch / cache time / stale time.

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  return (
+    <div
+      className={` border-2 border-gray-400 rounded-xl p-2 flex flex-col items-center max-w-fit cursor-pointer  ${wrapperClasses}`}
+      onClick={onToggleBalance}
+    >
+      {minimized === false && <div className="text-[10px] opacity-70">{displayAddress}</div>}
+
+      {/* display  eth or doller balance  */}
```
carletex, line 62: Extra spacing and typo on "dollar".

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <div className="w-full flex items-center  justify-center">
+        {isEthBalance === true && (
+          <>
+            <span>{balanceData?.formatted}</span>
```
carletex, line 66: We just want to show 4 decimals. rn:

image

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
...
+            <span className="text-xs font-bold m-1">ETH</span>
+          </>
+        )}
+
+        {isEthBalance === false && (
+          <>
+            <span>{(Number(balanceData?.formatted) * price).toFixed(4)}</span>
```
carletex, line 73: A better approach would be to use `'balanceData.value` which is a BigNumber and use BigNumber operations (like mul), instead of converting a string to a number.

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
...
+            <span className="text-xs font-bold m-1">$</span>
```
carletex, line 74: $ should go before the number

packages/frontend/components/scaffold-eth/Balance.tsx
carletex: But let's just rewrite this as a ternary:

{isEthBalance ? "..." : "..."}

packages/frontend/components/scaffold-eth/Balance.tsx
carletex: But let's just rewrite this as a ternary:

{isEthBalance ? "..." : "..."}

packages/frontend/pages/_app.tsx
```diff
@@ -5,6 +5,7 @@ import { alchemyProvider } from "wagmi/providers/alchemy";
 import { publicProvider } from "wagmi/providers/public";
 import "../styles/globals.css";
 import "@rainbow-me/rainbowkit/styles.css";
+import { useState, useEffect } from "react";
```
carletex, line 8: These are not being used.

packages/frontend/styles/globals.css
```diff
@@ -13,3 +13,12 @@ h4 {
...
 }
+
+/* temprory classes for bordering on ui */
+.bd-red {
+  border: 1px solid red;
+}
+
+.bd-blue {
+  border: 1px solid blue;
+}
```
carletex, line 24: These are not being used.

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,80 @@
+import { useState } from "react";
```
carletex, line 1: Rendering this components it gives an hydration error. You can check by refreshing the eth components page

This happens because of the wagmi cache. To avoid it, you need to use an extra useState. Check `Address.tsx`

image

carletex (commented):
Hey Naim, check all the review comments. You can mark them as resolved as you fix them.

Also, check why the GH actions linting test failed. With the changes that Shiv implemented (husky, etc) you should have seen the errors when commiting (maybe you need to run yarn install to set up husky?)

packages/frontend/components/scaffold-eth/Balance.tsx
carletex: Also, $ value show 2 decimals.

Naim-Bijapure: the multiplication have done this way is it ok?
1. ` <span>{ethers.utils.formatEther(balanceData?.value.mul(price).toString() as BigNumberish)}</span>`


also for 2 decimal values, for $  we have to wrap this inside Number

2.            ```
 <span>
              {Number(ethers.utils.formatEther(balanceData?.value.mul(price).toString() as BigNumberish)).toFixed(2)}
            </span>
```

let me know which 1 is ok or if there any other way

File: packages/frontend/components/scaffold-eth/Balance.tsx (hunk 15 above)
Comment (Naim-Bijapure) [reply packages/frontend/components/scaffold-eth/Balance.tsx:1]: for this have another solution like this 
`  const [mounted, setMounted] = useState(false);

  // default nextjs hydrtion issue managed
  useEffect(() => setMounted(true), []); // at init only
  if (!mounted) {
    return null;
  }

`

if we add this code inside _app.jsx then hydration issue will resolve . 
and we don't have to add useState pattern for wagmi  on every component . 

let me know if it is ok? 
will raise another pr for this fix

File: packages/frontend/styles/globals.css (hunk 14 above)
Comment (Naim-Bijapure) [reply packages/frontend/styles/globals.css:24]: sry this is for debugging borders on ui i use this to quickly add borders to check layouts etc.  
Sometimes I  forget to remove it. 

can we keep it for now until we go fully production release ??
because i have to remove it every time i push. 
on final build we can remove it.  if you don't mind ? 
whatever you say

Comment (carletex) [reply packages/frontend/styles/globals.css:24]: I think we shouldn't be adding debug stuff into the repo.

You can use "git update-index --assume-unchanged" if you want to do this.

File: packages/frontend/components/scaffold-eth/Balance.tsx (hunk 8 above)
Comment (Naim-Bijapure) [reply packages/frontend/components/scaffold-eth/Balance.tsx:25]: have used these options. 
     watch: true,
    cacheTime: 5_000,

File: packages/frontend/styles/globals.css (hunk 14 above)
Comment (Naim-Bijapure) [reply packages/frontend/styles/globals.css:24]: sure. let  me try this git hack

File: packages/frontend/components/scaffold-eth/Balance.tsx [hunk 16]
```diff
@@ -0,0 +1,81 @@
+import { BigNumberish, ethers } from "ethers";
+import { useEffect, useState } from "react";
+import { useBalance } from "wagmi";
+
+type BalanceProps = {
+  address: string;
+  price: number;
+  wrapperClasses?: string;
```
carletex, line 8: Let's remove this for now. I think we could be useful in the future, but we don't have the need right now.

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [balance, setBalance] = useState<any>();
+
+  const {
+    data: fetchBalanceData,
+    isError,
+    isLoading,
+  } = useBalance({
+    addressOrName: address,
+    watch: true,
+    cacheTime: 5_000,
```
carletex, line 26: You can add a ToDo here (or even better, create an issue) to talk about this value.

This shouldn't be used if we are in the local chain (so we get the value updated super fast when prototyping on localhost). For other networks, we should a sensible default (like 30 secs) in a config file, where people can change it.

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,81 @@
...
+        className={`border-2 border-gray-400 rounded-xl p-2 flex flex-col items-center max-w-fit cursor-pointer  ${wrapperClasses}`}
+      >
+        <div className="text-warning text-xs">Error</div>
+      </div>
+    );
+  }
+
+  return (
+    <div
+      className={`border-2 border-gray-400 rounded-xl p-2 flex flex-col items-center max-w-fit cursor-pointer  ${wrapperClasses}`}
```
carletex, line 60: extra space before ${wrapperClasses}. We are removing it anyway.

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,81 @@
...
+      onClick={onToggleBalance}
+    >
+      {/* display  eth or dollar balance  */}
+      <div className="w-full flex items-center  justify-center">
```
carletex, line 64: extra space between before justify-center

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,81 @@
...
+        {isEthBalance ? (
+          <>
+            <span>{Number(balance?.formatted).toFixed(2)}</span>
```
carletex, line 67: This will be be just: balance.toFixed(2) when you update the useffect

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <span className="text-xs font-bold m-1">ETH</span>
+          </>
+        ) : (
+          <>
+            <span className="text-xs font-bold m-1">$</span>
+            <span>
+              {Number(ethers.utils.formatEther(balance?.value.mul(price).toString() as BigNumberish)).toFixed(2)}
```
carletex, line 74: This will also be different when we update the useEffect

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,81 @@
...
+    isError,
+    isLoading,
+  } = useBalance({
+    addressOrName: address,
+    watch: true,
+    cacheTime: 5_000,
+  });
+
+  const onToggleBalance = () => {
+    setIsEthBalance(!isEthBalance);
```
carletex, line 30: Since we are using this pattern (to avoid hydration) lwe can just save the Number here

packages/frontend/pages/EthComponents.tsx
```diff
@@ -0,0 +1,13 @@
+import type { NextPage } from "next";
```
carletex, line 1: Let's not use this page. Let's put it in the home page, after the Address component.

IF there are conflicts in other PRs, we'll fix it.

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <div className="h-2 w-28 bg-slate-300 rounded"></div>
+        </div>
+      </div>
+    );
+  }
+
+  if (isError) {
+    return (
+      <div
+        className={`border-2 border-gray-400 rounded-xl p-2 flex flex-col items-center max-w-fit cursor-pointer  ${wrapperClasses}`}
```
carletex, line 51: extra space between before ${wrapperClasses}. We are removing it anyway.

carletex, line 51: extra space before ${wrapperClasses}. We are removing it anyway.

carletex, line 51: extra space before ${wrapperClasses}. We are removing it anyway.

carletex, line 51: extra space between before ${wrapperClasses}. We are removing it anyway.

carletex (changes requested):
Hey @Naim-Bijapure Left some comments.

Also, the yarn lock shouldn't have changed.

You can also back merge master into your branch. There are conflicts rn.

packages/frontend/components/scaffold-eth/Balance.tsx
```diff
@@ -0,0 +1,81 @@
...
+  wrapperClasses?: string;
+};
+
+/**
+ Display balance of an address
+*/
+
+export default function Balance({ address, price, wrapperClasses }: BalanceProps) {
+  const [isEthBalance, setIsEthBalance] = useState(true);
+  const [balance, setBalance] = useState<any>();
```
carletex, line 17: We don't need to do "any" here if we do the changes I requested on the useEffect

carletex (approved):
Added a couple of tweaks. Ready to merge!

PR #14, opened 2022-10-08
Hotfix/--hydration issue fix

Naim-Bijapure:
the issue. 
2022_October_08_13-08-16

no need to add  useEffect and useState patterns on every time we use wagmi hooks . 
instead inside _app.tsx have added mount init fix with useState . this will fix this hydration issue. 

adjusted Address.tsx and Balance.tsx component . 

app.tsx fix:

2022_October_08_12-52-42

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "hardhat:test": "yarn workspace @se-2/hardhat test",
     "start": "yarn workspace @se-2/frontend dev",
     "next:lint": "yarn workspace @se-2/frontend lint",
-    "next:format" : "yarn workspace @se-2/frontend format",
+    "next:format": "yarn workspace @se-2/frontend format",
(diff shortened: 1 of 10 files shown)


carletex:
Hey @Naim-Bijapure 

As I said in the other PR, you should branch for master on this hotfix.

Regarding your hydration fix: I think it's not a good idea. Yes, it fixes the hydration issue, but your completely remove the [pre-rendering](https://nextjs.org/docs/basic-features/pages#pre-rendering) capabilities that NextJS has.

You can check the pre-rendering by disabling the JS in your browser and you'll see that with your fix, it renders a page black (no pre-rendering at all).

Naim-Bijapure:
ohh ok if pre-redering is important then no use of this . will close this !

PR #16, opened 2022-10-14
 Title:hotfix:add eslint-typescript config for no-unused-var issue on ts

Naim-Bijapure:
added same fix https://github.com/scaffold-eth/se-2/pull/13 from master branch

### DIFF
--- a/packages/frontend/.eslintrc.json
+++ b/packages/frontend/.eslintrc.json
@@ -1,6 +1,9 @@
 {
-  "extends": ["next/core-web-vitals", "plugin:prettier/recommended"],
+  "parser": "@typescript-eslint/parser",
+  "extends": ["next/core-web-vitals", "plugin:prettier/recommended", "plugin:@typescript-eslint/recommended"],
   "rules": {
-    "no-unused-vars": "error"
+    "no-unused-vars": "off",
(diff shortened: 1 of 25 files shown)


packages/frontend/.eslintrc.json
```diff
@@ -1,6 +1,9 @@
 {
-  "extends": ["next/core-web-vitals", "plugin:prettier/recommended"],
+  "parser": "@typescript-eslint/parser",
+  "extends": ["next/core-web-vitals", "plugin:prettier/recommended", "plugin:@typescript-eslint/recommended"],
   "rules": {
-    "no-unused-vars": "error"
+    "no-unused-vars": "off",
```
carletex, line 5: I think we don't need this rule. Can you double check?

I removed that line, added this:

```
interface test {
  onChange?: (arg: string) => void;
}
```

And it works!

carletex (changes requested):

Naim-Bijapure:
@carletex  on this branch after merging master getting husky errors. 
as i can see lot of rules are failing from master due to "any" type used in many areas. 
we need to fix this type errors  on the master  after merging this branch

carletex (approved):

PR #27, opened 2022-10-19
added a faucet component

### LINKED ISSUE #19: Faucet Component
# Faucet Component

Create a Faucet Component that will give funds to the user when they are connected to the Burner Wallet. 

## Detailed Explanation

Create a new button that will only be displayed when the user is connected to the burner wallet. When a user clicks on the button it will send funds to the users burner wallet. 

The button will send 1 ETH, so it will not ask the user for an input for simplicity. In the future we can have the functionality for the user to choose the amount. For the initial feature lets make the button send 1 ETH by default. 

Here is a sample image for reference: (We can edit the name or have an icon for the faucet button)
Screen Shot 2022-10-14 at 11 38 25

## Links

As a reference we can use the [Faucet Component in Scaffold-eth](https://github.com/scaffold-eth/scaffold-eth/blob/5bdaf150c8007ba41abdd8c5763203a18baaa873/packages/react-app/src/components/Faucet.jsx)

technophile-04:
## Description : 
Added a basic faucet button that sends `1 ETH` to the connected address. This button is only visible if you are on a local hardhat chain. 

## Proposed Solution  : 
Created a utility function `getLocalProvider` , which gives you publicProviders configured in WAGMI. We need publicProvider to be more specific `localRpcProvider` for the faucet because it's default signer is the `0th` hardhat account which we will be using as `faucet`, and sending the transaction from. 

#### Reason to use separate function and not putting in `Faucet.tsx` itself : 
-  I tried digging into WAGMI and didn't find any direct function which provides you `publicProvider` directly. So you need to drill down a bit to get `publicProvider`

#### Reason to use WAGMI `publicProvider()` in `getPublicProvider` func and not using direclty  `new ethers.providers.StaticJsonRpcProvider()` :  
- It helped me make a more generic which supports other chains, and errors handled internally. 

### Tried Solution : 
1. Using `useProvider` hook:  It gives the provider which has the signer as a connected account in frontend, and we wanted to use `0th` account provided by hardhat which has initially `1000 ETH`. 

2. Using `usePrepareSendTransaction ` and `useSendTransaction ` again it sends transaction on behalf of account connected to frontend, I also tried finding is there a way to `connect` to differecnt acccount while `sendingTransaction` but didn't found anything good. Please let me know if we need to go with this approach will dig in more deeply.

## Suggestion Needed : 
Currently when you grab funds from the faucet everything works well but there seems to be some problem with `RainbowkitWallet` after grabbing the funds it does not update/refetch balance anytime soon, you need to referesh the page. (Demo attached below) 

Should we create a New issue for this? or try fixing it in this PR only? Like I tried finding a solution but didn't get anything useful.

Fixes #19 

https://user-images.githubusercontent.com/80153681/196798406-4f454d2c-26f1-4be0-8a1c-8d339d17ade9.mov

Open to any bad code quality/practices criticism, and will try my best to improve 🙌


packages/frontend/utils/getLocalProvider.ts
```diff
@@ -0,0 +1,17 @@
+import { Chain } from "wagmi";
```
carletex, line 1: Love the idea of this util.

Can we move it under `utils/scaffold-eth/getLocalProvider`? Trying to get all components / hooks / utils under the "scaffold-eth" folder.

packages/frontend/utils/getLocalProvider.ts
```diff
@@ -0,0 +1,17 @@
...
+import { publicProvider } from "wagmi/providers/public";
+
+/**
+  @dev Get the localProvider for given chain
+  @param chain : specific chain for localProvider
+  @return public provider for passed chain
+**/
+export default function getLocalProvider(chain: Chain) {
+  const chainProviderFunc = publicProvider();
```
carletex, line 10: I'm not a fan of "func", "obj" naming :D

Maybe `chainProviderLoader`?

packages/frontend/utils/getLocalProvider.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  const chainProviderObj = chainProviderFunc(chain);
```
carletex, line 12: Maybe `chainProviderData`?

packages/frontend/utils/getLocalProvider.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  const provider = chainProviderObj?.provider();
+
```
carletex, line 15: Personally, I'd only keep this last new line, and remove the others new lines above.

```ts
  const chainProviderFunc = publicProvider();
  const chainProviderObj = chainProviderFunc(chain);
  const provider = chainProviderObj?.provider();

  return provider;
```

carletex (changes requested):
Hey @technophile-04 thanks for this :) Left some review comments.

> Currently when you grab funds from the faucet everything works well but there seems to be some problem with RainbowkitWallet after grabbing the funds it does not update/refetch balance anytime soon, you need to referesh the page. (Demo attached below)

> Should we create a New issue for this? or try fixing it in this PR only? Like I tried finding a solution but didn't get anything useful.

Yep, we'll need to find a way to trigger that balance refresh with rainbow kit. Can you create a separate issue explaining the problem?

carletex (approved):

Issue #23, opened 2022-10-17
Zustand state management


carletex:
Create a basic zustand set up.

It can be a dummy store for now (in a `store.ts` file?), but let's set up the blueprint for future use.

ShravanSunder:
See #28

carletex:
Fixed on #28

PR #28, opened 2022-10-20
#23 create zustand store and examples.

ShravanSunder:
- create zustand store and services folder
- created example slices.  used example in tempContract
- Additionally, moved web3 folder into new services folder


packages/frontend/services/store/slices/anotherExampleTempState.ts
```diff
@@ -0,0 +1,16 @@
+import { TTempState as TAnotherTempState } from "./tempExampleState";
+
+export type TAnotherTempState2 = {
+  tempStuff: string;
+};
```
carletex, line 5: I guess we don't need that for this example?

Seems we are importing the Type from the other example

packages/frontend/services/store/store.ts
```diff
@@ -0,0 +1,36 @@
...
+  tempState: defaultTempState(),
+  setTempState: (newTempState: TTempState): void =>
+    set((state): TAppStore => {
+      state.tempState = newTempState;
+      return state;
+    }),
+});
+
+const createAnotherTempStateSlice: TAppSliceCreator<AnotherTempStateSlice> = set => ({
+  tempState: defaultTempState(),
```
carletex, line 21: Should it be "defaultTempState()" ??

packages/frontend/services/store/storeTypes.ts
```diff
@@ -0,0 +1,15 @@
+import { StateCreator } from "zustand";
+import { TempStateSlice } from "~~/services/store/slices/tempExampleState";
+import { AnotherTempStateSlice } from "~~/services/store/slices/anotherExampleTempState";
+
+/**
+ * The App store
+ * All your slices of state should be joined here
+ */
+
```
carletex, line 9: Remove this line? :D

carletex (changes requested):
This amazing @ShravanSunder great work!

I think reading the in-code comments that you wrote (+ when we have real examples) would be enough for people to understand how they should add new slices. Or do you think need to add a little README under `services/store`?

packages/frontend/services/store/slices/anotherExampleTempState.ts
ShravanSunder: i fixed it, was a mistake

ShravanSunder:
> This amazing @ShravanSunder great work!
> 
> I think reading the in-code comments that you wrote (+ when we have real examples) would be enough for people to understand how they should add new slices. Or do you think need to add a little README under `services/store`?

I added a readme!

carletex (approved):
lgtm!

Thanks for the Readme :)

PR #31, opened 2022-10-20
Update README.md

edakturk14:
Hey @carletex, I've updated the readme to have more details on what we're building w/se-2 and also included the deploy contracts section as we now have the functionality to deploy to different networks. Think it will be useful to keep the readme up to date as we add new functionality. thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,40 +1,73 @@
-# se-2
+# Scaffold-Eth 2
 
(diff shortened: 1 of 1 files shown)


carletex (approved):
Great updates @edakturk14 

Thanks!

PR #33, opened 2022-10-24
Account helper functions (generate / account)

carletex:
Took a different approach than SE1.

`yarn generate` will create/update the .env file with a private key.

For `yarn account` I'm using a more up to date (support with ES6 included :)) QR lib than SE1.

Also using hardhat scripts (instead of tasks)

P.S. We could also generate a mnemonic instead of a private key... but went for the private key, since hardhat.config was already using a private key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ ALCHEMY_API_KEY="",
 DEPLOYER_PRIVATE_KEY=""
 ```
 
+You can also generate a random account / private key with `yarn generate` (and check the generated account with `yarn account`)
+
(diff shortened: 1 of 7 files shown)


codenamejason:
@carletex looks like it generated someones used wallet with balances...

Screenshot 2022-10-25 at 10 57 49 PM

codenamejason (approved):

carletex:
Thanks @codenamejason !

The first screenshot is using account[0] (0xf39..) from hardhat!

edakturk14 (commented):
Looks awesome!

Issue #2, opened 2022-09-27
Contract UI component


carletex:
UI that adapts to the contract ABI (stored in packages/frontend/contracts/hardhat_contracts.json when you deploy)

It depends on a bunch of components and hooks. We'd figure out what we need as we go, creating issues as needed.

- [x] Address https://github.com/scaffold-eth/se-2/commit/8b327f4d647ca562ae104a699443b49bc2f0ebd0
- [x] AddressInput https://github.com/scaffold-eth/se-2/issues/3
- [x] Balance https://github.com/scaffold-eth/se-2/issues/4
- [ ] Transactor helper?

I'd do a hard refactor on the Contract component (to make it cleaner and easier to mantain). But as a reference we can use:
- scaffold-eth [Contract component](https://github.com/scaffold-eth/scaffold-eth/tree/master/packages/react-app/src/components/Contract)
- abi ninja [Contract component](https://github.com/carletex/abi.ninja/tree/master/packages/react-app/src/components/Contract). It has a better UX
- Naim's [debug-eth-contract](https://github.com/Naim-Bijapure/debug-eth-contract)

ShravanSunder:
@carletex We'l be using tailwind / headless ui and daisy for this component?

carletex:
MVP done on #36

PR #35, opened 2022-10-26
[WIP] Contract Component

technophile-04:
Setted up basic contract component for #2 . currently, it only shows pure/view functions which takes 0 inputs : 
Screenshot 2022-10-26 at 9 13 40 PM

### Flow : 
- `/debug` route render's the `Contract` component which takes in contractName as props. 
- In `Contract` component it gets currently connected `chainId` and uses `getGeneratedContract` function to get contract `abi` and `address` deployed at current user connected chain. 
- Rendering of variables is almost similar to eth-components [GenricComponent.tsx](https://github.com/scaffold-eth/eth-components/blob/next/src/ant/generic-contract/GenericContract.tsx)  and abi.ninjas [Contract.jsx](https://github.com/carletex/abi.ninja/blob/master/packages/react-app/src/components/Contract/index.jsx)

### Some issues : 
1) If we directly go to `/debug` on the browser the `chain` from `useNetwork` wagmi hook is coming to be undefined because of this we are unable to find a contract. 

Open to any suggestions be it variable names, code structure, or anything 🙌

### DIFF
--- a/packages/frontend/components/scaffold-eth/Contract/DisplayVariables.tsx
+++ b/packages/frontend/components/scaffold-eth/Contract/DisplayVariables.tsx
@@ -1,0 +1,49 @@
+import { ContractFunction } from "ethers";
+import { FunctionFragment } from "ethers/lib/utils";
+import React, { FC, SetStateAction, useCallback, useEffect, useState, Dispatch } from "react";
+// import { useIsMounted } from "usehooks-ts";
+import { tryToDisplay } from "./displayUtils";
+
+interface IDisplayVariableProps {
+  contractFunction: ContractFunction | undefined;
+  functionInfo: FunctionFragment;
+  refreshRequired: boolean;
+  setTriggerRefresh: Dispatch<SetStateAction<boolean>>;
+}
+
+// TODO Check for the need of useIsMounted
+
+export const DisplayVariable: FC<IDisplayVariableProps> = props => {
+  const [variable, setVariable] = useState("");
+  // const isMounted = useIsMounted();
+
+  const refresh = useCallback(async () => {
+    try {
+      if (props.contractFunction) {
+        const contractReturnValue = await props.contractFunction();
+        setVariable(contractReturnValue);
+        props.setTriggerRefresh(false);
+
+        // if (isMounted()) {
+        // }
+      }
+    } catch (e: any) {
+      console.log(e?.message);
+    }
+  }, [props]);
+
+  useEffect((): void => {
+    void refresh();
+  }, [refresh, props.refreshRequired, props.contractFunction]);
+
+  // TODO add refresh button for each variables
+  // TODO fix warning "<div> cannot appear descendant of <p>"(Reason : <Address> component)
+  return (
+    <div className="border-b-2 border-black space-y-1 pb-2">
+      <p className="text-black text-lg">{props.functionInfo.name}</p>
+      <p className="text-black font-normal">{tryToDisplay(variable)}</p>
+    </div>
+  );
+};
+
+export default DisplayVariable;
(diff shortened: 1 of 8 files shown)


packages/frontend/components/scaffold-eth/Contract/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  contractName: string;
+}
+
+/**
+ * TODO
+ * handle error
+ * handle types correctly
+ * handle loading state
+ * handle payable, public, functions
+ * ! fix the problem if you directly go to `/debug` we get : "chain is undefined"
```
carletex, line 17: Just pushed a change to master that should fix this

packages/frontend/components/scaffold-eth/Contract/DisplayVariables.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface IDisplayVariableProps {
+  contractFunction: ContractFunction | undefined;
+  functionInfo: FunctionFragment;
+  refreshRequired: boolean;
+  setTriggerRefresh: Dispatch<SetStateAction<boolean>>;
+}
+
+// TODO Check for the need of useIsMounted
+
+export const DisplayVariable: FC<IDisplayVariableProps> = props => {
```
carletex, line 16: I think we should avoid the use of "props", even if we have `IDisplayVariableProps`.

Let's have it listed {contractFunction, functionInfo...}

packages/frontend/components/scaffold-eth/Contract/DisplayVariables.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { tryToDisplay } from "./displayUtils";
+
+interface IDisplayVariableProps {
+  contractFunction: ContractFunction | undefined;
+  functionInfo: FunctionFragment;
+  refreshRequired: boolean;
+  setTriggerRefresh: Dispatch<SetStateAction<boolean>>;
+}
+
+// TODO Check for the need of useIsMounted
```
carletex, line 14: Is this because the hydration problem? In some case when you refresh the page, you get an hydration error (next server prerendering doesn't match the client)

I encounter this in some components (Like the address one) in case you want to take a look

packages/frontend/components/scaffold-eth/Contract/displayUtils.tsx
```diff
@@ -0,0 +1,38 @@
+import { TransactionResponse } from "@ethersproject/providers";
+import { formatUnits } from "@ethersproject/units";
+import { BigNumber } from "ethers";
+import React, { ReactElement } from "react";
+import Address from "../Address";
+
+export const tryToDisplay = (
+  thing: string | number | BigNumber | Record<string, any> | TransactionResponse | undefined,
```
carletex, line 8: let's call it "element" :D

packages/frontend/components/scaffold-eth/Contract/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const { address: contractAddress, abi: contractABI } = getGeneratedContract(chain?.id.toString(), props.contractName);
+
+  const contract = useContract({
+    addressOrName: contractAddress,
+    contractInterface: contractABI,
+    signerOrProvider: provider,
+  });
+
+  // maybe put this into utils files ?
```
carletex, line 32: Yep :)

carletex (commented):
Hey Shiv!

You can use a branch inside this repo (that way it's easy for me to push to it)

Left some initial comments. Also, I back merged master into your branch (it should fix the problem when going directly to /debug)

PR #36, opened 2022-10-28
Contract Component

technophile-04:
Setted up basic contract component for #2  . currently, it currently only shows pure/view functions which take 0 inputs : 
Screenshot 2022-10-26 at 9 13 40 PM

### Current Flow : 
- `/debug` route render's the `Contract` component which takes in contractName as props. 
- In `Contract` component it gets currently connected `chainId` and uses `getGeneratedContract` function to get contract `abi` and `address` deployed at current user connected chain. 
- Rendering of variables is almost similar to eth-components [GenricComponent.tsx](https://github.com/scaffold-eth/eth-components/blob/next/src/ant/generic-contract/GenericContract.tsx)  and abi.ninjas [Contract.jsx](https://github.com/carletex/abi.ninja/blob/master/packages/react-app/src/components/Contract/index.jsx)


technophile-04:
## Description :
- Added `ReadOnlyFunctionForm` component which displays read-only functions which accept more than or equal to 1 param.(Name can be improved 😅 )
- Refactored the `DisplayeVariables` component to use wagmi instead since it has better error handling and state management.

## Demo :

https://user-images.githubusercontent.com/80153681/199841942-df3c8ebc-8bb0-4c0a-8e61-f31ab0a56034.mov

packages/frontend/components/ErrorToast.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { FC } from "react";
+
+// TODO Remove this in future
+interface IErrorToast {
+  errorMessage: string;
+}
+
+const ErrorToast: FC<IErrorToast> = ({ errorMessage }) => {
```
technophile-04, line 8: Created a basic custom Toast component, but should we use a library to handle the toasts?  Because this component does not disappear after some time nor it has a close button

packages/frontend/components/scaffold-eth/Contract/DisplayVariables.tsx
```diff
@@ -0,0 +1,57 @@
+import { FunctionFragment } from "ethers/lib/utils";
+import React, { useState } from "react";
+import { useContractRead } from "wagmi";
+import ErrorToast from "~~/components/ErrorToast";
+import { tryToDisplay } from "./displayUtils";
+// import { ArrowPathRounded } from "@heroicons/react";
```
technophile-04, line 6: Getting this error `Cannot find module '@heroicons/react' or its corresponding type`

packages/frontend/components/scaffold-eth/Contract/InputUI.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+interface IInputUI {
+  setForm: Dispatch<SetStateAction<Record<string, any>>>;
+  form: Record<string, any>;
+  stateObjectKey: string;
+  paramType: ParamType;
+  functionFragment: FunctionFragment;
+}
+
+// TODO use AddressInput component instead
```
technophile-04, line 19: Currently this component uses `Address` component to show input address, but we should use the `AddressInput` component instead, we need to update `AddressInput` to take in `onChange` and `value` as prop similar to abi.ninja's [AddressInput.jsx](https://github.com/carletex/abi.ninja/blob/master/packages/react-app/src/components/AddressInput.jsx) component

packages/frontend/components/scaffold-eth/Contract/InputUI.tsx
```diff
@@ -0,0 +1,102 @@
...
+const InputUI: FC<IInputUI> = ({ setForm, form, stateObjectKey, paramType }) => {
+  let buttons: ReactElement = <></>;
+
+  // TODO maybe abstract each `div` as a small util Component, use switch case and cleanUp
```
technophile-04, line 23: Should we abstract each `div`'s inside the if condition and maybe be put it in a `utilsComponents` file to make it cleaner?

packages/frontend/components/scaffold-eth/Contract/utils.tsx
```diff
@@ -0,0 +1,86 @@
...
+      );
+    }
+    return null;
+  });
+};
+
+const getContractReadOnlyMethodsWithParams = (contract: Contract, contractMethodsAndVariables: FunctionFragment[]) => {
+  return contractMethodsAndVariables.map((fn, index) => {
+    const isQueryableWithParams =
+      (fn.stateMutability === "view" || fn.stateMutability === "pure") && fn.inputs.length > 0;
```
technophile-04, line 59: I hope this is the right way to get all the functions that are read-only and have params more than one :)

packages/frontend/components/scaffold-eth/Contract/ContractUI.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const { address: contractAddress, abi: contractABI } = getGeneratedContract(chain?.id.toString(), contractName);
+
+  const contract: Contract = useContract({
+    addressOrName: contractAddress,
+    contractInterface: contractABI,
+    signerOrProvider: provider,
+  });
+
+  const displayedContractFunctions = getAllContractFunctions(contract);
```
technophile-04, line 33: Should we remove this? and maybe use it internally in `getContractVariablesAndNoParamsReadMethods ` and `getContractReadOnlyMethodsWithParams ` functions?

carletex: I guess if we move it, we'd have to call it three times, right? Inside:

1. getContractVariablesAndNoParamsReadMethods
2. getContractReadOnlyMethodsWithParams
3. (to-do) getContractWriteMethods

In any case, shouldn't be too expensive to call that function 3 times.... so up to you if you want to move it for the sake of readability.

packages/frontend/components/scaffold-eth/Contract/utils.tsx
carletex: Looks good to me!

And for the (ToDo) `getContractWriteMethods` it should be:

`(fn.stateMutability !== "view" && fn.stateMutability !== "pure")`, right?

I think we don't need to differentiate between write methods with params AND write methods without params.

packages/frontend/components/scaffold-eth/Contract/InputUI.tsx
carletex: Yeah, cool! I talked about this with Naim while working on the Address component: https://github.com/scaffold-eth/se-2/pull/11#discussion_r999625592

> Also, let's not worry too much, we can adapt in the future depending on the needs

I think it's time to adapt :) I'm working on this. Will update.

packages/frontend/components/scaffold-eth/Contract/DisplayVariables.tsx
carletex: Fixed here 57790daf6d4c3568ed9f149f9af58cc87bf5b182

packages/frontend/components/ErrorToast.tsx
carletex: Let me think about this a little bit more. But I think we could implement something simple by ourselves without much trouble.

I'll do some testing, and If I see it gets too big... we might want to use a lib :P

packages/frontend/components/scaffold-eth/Contract/ReadOnlyFunctionForm.tsx
```diff
@@ -0,0 +1,79 @@
...
+  });
+  return initialForm;
+};
+
+interface IFunctionForm {
+  functionFragment: FunctionFragment;
+  contractAddress: string;
+}
+
+export const ReadOnlyFunctionForm: FC<IFunctionForm> = ({ functionFragment, contractAddress }) => {
```
carletex, line 23: Did you see my comments on a previous review about (not) using `FC`?

packages/frontend/components/scaffold-eth/Contract/InputUI.tsx
carletex: @technophile-04 Just updated the `<AddressInput>` component on 19c5e6bad0fb353aed4e79478fca75442e3fa728.

You can use it with something like:

```tsx
      <AddressInput
        name={stateObjectKey}
        value={form[stateObjectKey]}
        onChange={(value): void => {
          const formUpdate = { ...form };
          formUpdate[stateObjectKey] = value;
          setForm(formUpdate);
        }}
      />
```

Let me know if it works for you.

packages/frontend/components/scaffold-eth/Contract/InputUI.tsx
carletex: Yeah, I like the idea!

We could do a switch (instead of all if-elses):

```tsx
  switch (paramType.type) {
    case "bytes32":
      // call the util
      break;
  }
```

Regarding the utils... maybe we can have:

- **utilsContract** (currently named "utils")
- **utilsDisplay** (currently named "displayUtils")
- **utilsInputs** / **utilsComponents**... or whatever you like (depending if we only use it for the getting the "buttons" or algo the "inputs)

packages/frontend/components/scaffold-eth/Contract/ContractUI.tsx
technophile-04: Ohh yes, you are right will think about this

packages/frontend/components/scaffold-eth/Contract/ReadOnlyFunctionForm.tsx
technophile-04: Ohh really sorry, I missed it, I thought you were talking about destructing of props my bad

technophile-04:
## Description : 
- Added `WriteOnlyFunctionForm` component for write functions.
- Cleaned up the code and added comments 

## TODO : 
- [ ] find a better way  contract from `contracts` directory, currently, it's a hardcoded 
- [ ] use `useMemo` to optimize wherever possible to memoize functions/maps results
- [ ] Parse the txnResponse nicely.
- [ ] testing for different contract variables types.

## Demo 

https://user-images.githubusercontent.com/80153681/200908956-bcaa9de3-03bb-4231-a4da-a45fc29a8f8c.mov

packages/frontend/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -0,0 +1,165 @@
...
+): GeneratedContractType | undefined => {
+  if (!chainId || !contractName) {
+    return { address: "", abi: [] };
+  }
+
+  const contractsAtChain = ContractData[chainId as keyof typeof ContractData];
+  const contractsData = contractsAtChain[0]?.contracts;
+
+  return contractsData[contractName as keyof typeof contractsData];
+};
```
carletex, line 32: @technophile-04 Regarding "find a better way contract from contracts directory, currently, it's a hardcoded"

I think it's good enough as it's rn. This is how `hardhat_contracts.json` looks when you deploy a second contract in 31337:

image

I think `contractsAtChain[0]` is always going to return the right thing, since we are already inside the 31337 array.

I tweaked the name of the functions and fixed some issues with types.

packages/frontend/components/scaffold-eth/Contract/ContractUI.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  const contractVariablesDisplay = getContractVariablesAndNoParamsReadMethods(contract, displayedContractFunctions);
+  const contractMethodsDisplay = getContractReadOnlyMethodsWithParams(contract, displayedContractFunctions);
+  const contractWriteMethods = getContractWriteMethods(contract, displayedContractFunctions);
+
+  if (!contractAddress) {
+    return <p className="text-2xl">No Contract found !</p>;
```
carletex, line 47: I removed the `text-white` here because I wasn't seeing it while testing. I think you are using "dark mode" right?

Dark / light mode it's definitely something we have to implement on tailwind / daisy. We can create an issue for it!

technophile-04: Yup using dark mode, currently it's dark/white mode is based on the system theme

technophile-04:
A lot cleaner and looks good, plus the lib seems to be easy to customize too

packages/frontend/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+    // TODO Show more descriptive error message
+    if (inputError) {
+      const message = getParsedEthersError(inputError);
+      toast.error(message);
+    }
+
+    if (writeAsync) {
+      try {
+        await reactHotToast.promise(
+          writeAsync(),
```
technophile-04, line 61: used this as of now, maybe in the future wrap the `writeAsync` inside our own function, something similar to Transactor or more general which handles promise

technophile-04:
Hii @carletex, upgraded `wagmi` to latest version, also upgraded the `typescript` version as suggested in wagmi [docs](https://wagmi.sh/docs/migration-guide#07x-breaking-changes).

Currently, there are some issues with rainbow kit due latest version of wagmi : 
list of current bugs: https://github.com/rainbow-me/rainbowkit/issues/869

We can use the previous version of `wagmi`, it has some issues with types in `useContractWrite` which can bypassed/solved as suggested in this [discussion](https://github.com/wagmi-dev/wagmi/discussions/1286)

Please let me know if we should downgrade `wagmi` to 0.7.x ( x is latest recent working version)

carletex:
Hey @technophile-04 

I think we shouldn't update wagmi in this PR. It's already a big one + there are issues with 0.7 & 0.8.

We can think about upgrading when this is merged.. but there are more critical stuff, and also a bunch of things to update (Like NextJS if we want to do so...)

carletex (approved):
Made a few tweaks. Check the commits out! 

This looks good for a merge. Will be creating new issues with the ToDos.

Good job @technophile-04

packages/frontend/components/scaffold-eth/Contract/utilsComponents.tsx
```diff
@@ -0,0 +1,41 @@
...
+        value={txValue}
+        onChange={e => setTxValue(e.target.value)}
+      />
+      <button
+        style={{ cursor: "pointer" }}
+        onClick={async () => {
+          if (!txValue) return;
+
+          const floatValue = parseFloat(txValue);
+          if (floatValue) setTxValue("" + floatValue * 10 ** 18);
```
technophile-04, line 23: marking the culprit!

Issue #30, opened 2022-10-20
RanibowKit Button component not updating the balance


technophile-04:
## Description 
![Untitled design](https://user-images.githubusercontent.com/80153681/196988856-1513814e-5b6b-43d6-bc88-7b0f67ac359b.gif)

### Current Behaviour :
RainbowKit Button doesn't seem to update/re-fetch the connected address `Balance` after some interval. 
As in the above video when you grab funds from faucet(aka `0th` account of hardhat), the native scaffold-eth `Balance` component seems to show the updated balance of the `0th` account of hardhat, but the connected **address Balance** is not re-fetched/updated by **RainbowKit component**. 

### Expected Behaviour :
Similar to the native scaffold-eth `Balance` component, the RainbowKit component should also re-fetch the balance of the connected address after a small interval at least when you  are connected to the local hardhat chain

carletex:
@Naim-Bijapure made a fix attempt here: https://github.com/scaffold-eth/se-2/pull/32

Let's find out what is the best way to go about it.

As Shiv said:
> Similar to the native scaffold-eth-1 Balance component, the RainbowKit component should also re-fetch the balance of the connected address after a small interval at least when you are connected to the local hardhat chain

As we do in SE1:
- Local chain: fetch balance onBlock
- External chain: fetch balance every X seconds

We might also want to have  "refetch" function that you can call if you don't to wait the X secs.

carletex:
It seems that it's nos supported by default on Rainbow kit. See https://github.com/rainbow-me/rainbowkit/discussions/626

I tried a similar workaround as @Naim-Bijapure suggested:

```js
  const [connectKey, setConnectKey] = useState(0);

  useEffect(() => {
    const intervalId = setInterval(() => {
      setConnectKey(Math.floor(Math.random() * 1000));
    }, 2000);

    return () => clearInterval(intervalId);
  }, []);

  return (
      <ConnectButton
        // .....
        key={connectKey}
      />
  );
```

This works, but it creates a re-rendering blip that we don't want (only the second one is capture by the video):

https://user-images.githubusercontent.com/2486142/199526616-8f5b7e66-dcdc-4637-9dc0-d3ccea115046.mp4

I'm wondering if the solution is to implement a [Custom Connect Button](https://www.rainbowkit.com/docs/custom-connect-button)

carletex:
Fixed on #43

PR #43, opened 2022-11-09
 hotfix: rainbowkit faucet balance update

Naim-Bijapure:
issue: https://github.com/scaffold-eth/se-2/issues/30
used a custom connect button with an independent HeaderBalance component on the header which displays the updated balance. 

for now the UI is basic. 

 
[Screencast from 09-11-22 09:18:44 PM IST.webm](https://user-images.githubusercontent.com/22323693/200876443-8a3cabfa-91a1-487c-aefd-7352663ddb47.webm)


packages/frontend/components/Header.tsx
```diff
@@ -1,14 +1,40 @@
...
+    }
+  }, [fetchedBalanceData]);
+
+  return (
+    <span className="m-2">
+      {balance?.toFixed(2)} {balanceSymbol}
+    </span>
+  );
+};
+
```
carletex, line 37: Instead of creating this, lets use the component that we already have (`Balance`).

packages/frontend/components/Header.tsx
```diff
@@ -17,7 +43,7 @@ export default function Header() {
 
   return (
     <div className="mt-5 flex flex-col space-y-4 md:flex-row md:space-x-4 md:space-y-0 items-center justify-center">
-      <ConnectButton
+      {/* <ConnectButton
```
carletex, line 46: Remove this

packages/frontend/components/Header.tsx
```diff
@@ -27,7 +53,83 @@ export default function Header() {
...
           largeScreen: "icon",
         }}
-      />
+      /> */}
+
+      {/* custom connect button */}
+      <ConnectButton.Custom>
+        {({ account, chain, openAccountModal, openChainModal, openConnectModal, authenticationStatus, mounted }) => {
+          // Note: If your app doesn't use authentication, you
+          // can remove all 'authenticationStatus' checks
```
carletex, line 62: I noticed this custom button is coming from the rainbox kit examples: https://www.rainbowkit.com/docs/custom-connect-button

Could you remove the `authentication` part as they say? We are not going to use if for now.

packages/frontend/components/Header.tsx
```diff
@@ -27,7 +53,83 @@ export default function Header() {
...
+                            background: chain.iconBackground,
+                            width: 12,
+                            height: 12,
+                            borderRadius: 999,
+                            overflow: "hidden",
+                            marginRight: 4,
+                          }}
+                        >
+                          {chain.iconUrl && (
+                            <img
```
carletex, line 110: Check the linting. 
image

carletex (changes requested):
Thanks @Naim-Bijapure 

Like we talked on #30 this might be the only option for now. We lose the default theming but I guess that's an opportunity for us ;)

Left some review comments.

carletex (approved):

PR #44, opened 2022-11-14
hotfix:calling localhost node on live network

Naim-Bijapure:
issue: https://github.com/scaffold-eth/se-2/issues/38

hook useTempTestContract usage hardcoded local chain id to read the contract ( useContractRead hook) 
which causes to polling on 127.0.0.1 on live network when connected with metamask. 
in this pr update, check deployed contract from the hardhat_config.json file and skip the useContractRead. 

assuming useTempTestContract is a temporary file. 

2022_November_14_17-05-39


carletex:
Ooh @Naim-Bijapure thanks for finding this!!

Yep, this is a temp file so we don't need to fix it. We should remove `useTempTestContract` on https://github.com/scaffold-eth/se-2/pulls

PR #45, opened 2022-11-16
Setup sensible default pollingInterval 

### LINKED ISSUE #39: Watch on block / Polling time on the Balance component
We want to keep the balance from `<Balance>` updated (polling time VS watch on Block)

Reference PR from SE-1: https://github.com/scaffold-eth/scaffold-eth/pull/896

The approach we can take (similar to SE-1):
- 30 sec (will go to a config file) as polling time default for live networks
- onBlock for local network (so interactions with the local contract is fast)

We have this rn:

```js
  useBalance({
    addressOrName: address,
    watch: true,
    // ToDo: Read this value from config. Disabled for localhost.
    cacheTime: 300_000,
  });
```

Let's check & test all the available options in wagmi [useBalance](https://wagmi.sh/docs/hooks/useBalance) (watch, cacheTime, staleTime, tec..) to make this happen: 1 RPC call every 30 sec for live networks & 1 RPC local call every block

We'd be using this for other components/hooks so let's try to be a _bit_ generic.

technophile-04:
## Proposed solution : 
Added a `pollingInterval` to providers itself of `30sec`. This is only set if your are not on local hardhat chain,  using `.env.local` . Similar to se-1 [initialNetwork](https://github.com/scaffold-eth/scaffold-eth/blob/a3a2d6fc78c7b801c7ea453f035b1420389a94b5/packages/react-app/src/App.jsx#L57), here we are using `env.local` as source...maybe we can move to another config file as mentioned in #9

Fixes #39 

## Reason / Explanation for current approach :  

- It seems like wagmi does not have an option of `pollTime` similar to eth-hooks for each hooks (checkout small explanation about `pollTime` in se-1 sensible default issue [here](https://github.com/scaffold-eth/scaffold-eth/pull/896#issuecomment-1262712820)). 

- To stay in sync with chain, we need to pass [watch:true](https://0.6.x.wagmi.sh/docs/hooks/useBalance#watch-optional) as an option to each hook, and it under the hood uses `useBlockNumber` which in turn depends on `provider._pollingInterval`. 
  - checkout wagmi useBalance hook implementation code [here](https://github.com/wagmi-dev/wagmi/blob/257c6131ed494a628a0fa1c74405c15a954e9509/packages/react/src/hooks/accounts/useBalance.ts#L75).
- It was also suggested here :  https://github.com/wagmi-dev/wagmi/discussions/1252

**cacheTime vs staleTime** : 
It seems like wagmi follows same patters as [react-query](https://tanstack.com/query/v4/docs/overview) (it uses react-query under the hood)
| cacheTime      | staleTime |
| ----------- | ----------- |
| The duration until inactive queries will be removed from the cache. Queries transition to the inactive state as soon as there are no observers registered, so when all components which use that query have unmounted.| The duration until a query transitions from fresh to stale. As long as the query is fresh, data will always be read from the cache only - no network request will happen! If the query is stale (which per default is: instantly), you will still get data from the cache, but a background refetch can happen under certain conditions.|
|cacheTime determines for how long a certain response is supposed to stay in cache before it gets garbage collected.| staleTime determines for how long a certain response will still be considered fresh (or not stale), dismissing the need for a new request |

This above difference was taken from [here](https://stackoverflow.com/questions/72828361/what-is-staletime-cachetime-in-react-query). This article really nicely explains the difference b/w both [here](https://medium.com/doctolib/react-query-cachetime-vs-staletime-ec74defc483e#:~:text=More%20specifically%2C%20cacheTime%20determines%20for,need%20for%20a%20new%20request.)

Now coming to our case, inorder to keep values updated / in sync we need to mandatory pass `watch:true`, If we pass `cacheTime` along with `watch` it has no effect. 

```js
const {
    data: fetchedBalanceData,
    isError,
    isLoading,
  } = useBalance({
    addressOrName: address,
    watch: true,
    cacheTime: 30_000, // <-- No effect, can be removed
  });
```
No effect, since `watch` will be making `evm_blocknumber` call every `provider._pollingInterval` and if `blocknumber` is changed irrespective of weather the `cacheTime` is exceeded or not it will make `evm_getBalance` call. internal working of `useBalance` [here](https://github.com/wagmi-dev/wagmi/blob/257c6131ed494a628a0fa1c74405c15a954e9509/packages/react/src/hooks/accounts/useBalance.ts#L76). checkout `useEffect` in internal implementation. 

## Some points to note : 
I have commented out `webSocketProvider` option to wagmiClient : 
```diff
export const wagmiClient = createClient({
  autoConnect: false,
  connectors: wagmiConnectors,
  provider: appChains.provider,
- webSocketProvider: appChains.webSocketProvider,
+ // webSocketProvider: appChains.webSocketProvider,
});
```

**Explanation** : 
> note that WebSockets are much more intensive on your server resources, as they must manage and maintain the state for each client.**For this reason, many services may also charge additional fees for using their WebSocket endpoints.** - [ethers docs](https://docs.ethers.io/v5/api/providers/other/#WebSocketProvider)

> **If you provide a WebSocket provider, it will be used instead of polling in certain instances**. - [wagmi docs](https://0.6.x.wagmi.sh/docs/client#websocketprovider-optional)

Confused weather to keep it or not, If people use their own `alchemyApiKey` then its good to have because, it decreases the rpc calls as mentioned in wagmi docs. If they use our alchemyKey, I think it will get overwhelmed 

Below demo, demonstrates that automatically **only** `eth_getBalance` calls are made after `blocknumber` is changed, `eth_blocknumber` are handled by webSocketProvider internally and hence do not make request in networks tab.

ezgif-1-1feb8063cf

## Test Brach : 
Created a test branch to test, tinker and dive deeper :  [test-sensible-polling-time](https://github.com/scaffold-eth/se-2/tree/test-sensible-polling-time)

**Defaults**
- commented out all the unnecessary components/functions to avoid unknown requests. 
- pollingInterval is hardcoded to `30_000` for testing on mainnet
- `webSocketProvider` is also commented to see all calls

**created a sample example to test cacheTime and staleTime to better utilize wagmi caching :**
- Created a page `cacheDemo` which denotes how we can potentially utilize caching with `watch` option being true
- `cacheDemo` page uses `useBlocknumber`(This hook basically returns current block number). 

```js
  const { data: fetchBlocknumber, isLoading } = useBlockNumber({
    watch: true,
    cacheTime: 40_000,
    staleTime: 5_000,
  });
```
In the above code, we have used `cacheTime`, `staleTime` both with `watch` option.

So if we omit staleTime, useBlockNumber will make the request everytime component mounts.
If staleTime is being passed it will actually use the cache data the next time you visit the page in b/w `5sec`.

https://user-images.githubusercontent.com/80153681/202283892-c92a3ff0-584e-4929-a548-8805cf5ab0eb.mov

In the above video we can see that `eth_blocknumber` was not made at all, when we go second time on the page. 

**Conclusion**
I was planning to use this pattern for all the hooks(the `cacheTime` and `staleTime` can be optimized). Maybe something like this : 
```js
const { data: fetchBlocknumber, isLoading } = useBlockNumber({
    watch: true,
    ...getSensibleConfig(currentChain),
  });
```
Basically `getSinsibleConfig` will set some sensible values for `cacheTime`, `staleTime` depending on the chain choosen.

Please let me know if we should go ahead with the above pattern 🙏. 

## Alternate possible solution  : 
Maybe create our own [useOnRepetation](https://github.com/scaffold-eth/eth-ui/blob/releases/v2/src/useOnRepetition.ts) hook wrapper and pass it the [refetch](https://0.6.x.wagmi.sh/docs/hooks/useBalance#return-value) function, returned from wagmi hooks. If we don't want put `pollingInterval` and go with the approach followed in https://github.com/scaffold-eth/scaffold-eth/pull/896 i.e is passing`pollingTime` to each hook


packages/frontend/services/web3/wagmiConnectors.tsx
```diff
@@ -27,9 +27,19 @@ export const appChains = configureChains(
       // This is ours Alchemy's default API key.
       // You can get your own at https://dashboard.alchemyapi.io
       apiKey: "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF",
+      priority: 0,
     }),
-    publicProvider(),
+    publicProvider({ priority: 1 }),
   ],
+  {
+    stallTimeout: 3_000,
```
technophile-04, line 35: added a `stallTimeout` and `priority` to each provider, it helps fix the issue for 429 error too many request. link to discussion : https://github.com/wagmi-dev/wagmi/discussions/1146 and solution : https://github.com/wagmi-dev/wagmi/discussions/1146#discussioncomment-4112555

damianmarti:
Thanks for the long explanation! 

One question (since I'm new to the repo and I didn't see all the code yet): are we using the useBalance from wagmi?

packages/frontend/services/web3/wagmiConnectors.tsx
```diff
@@ -27,9 +27,19 @@ export const appChains = configureChains(
...
+    // Sets pollingInterval if using chain's other than local hardhat chain
+    ...(process.env.NEXT_PUBLIC_NETWORK_LOCALHOST !== "true"
+      ? {
+          pollingInterval: 30_000,
```
carletex, line 39: Should we have this as an env var?

packages/frontend/services/web3/wagmiConnectors.tsx
technophile-04: I tried experimenting with it thoroughly and here's the difference : 

To test it what I did was decrease the network speed to `slow 3G` before mainnet rpc calls are made and below are the results : 

### stallTimeout : 2_000
https://user-images.githubusercontent.com/80153681/203033490-6dcfee13-c32d-4d7e-8487-da18f0d82572.mov 

### stallTimeout : 3_000

https://user-images.githubusercontent.com/80153681/203033915-82fe172e-83de-4b65-b8e6-05bf820a3b28.mov

### Observations : 
- `wagmi` uses the second provider passed in the config array when there is no / pending response from the first RPC if the `stallTimeout` is overdue.  
- Now, since our second provider i.e mainnet publicProvider fails, the default behavior of `wagmi` is to try again the same request if it fails and thus it goes on making more requests with mainnet publiProvider (checkout first video)...

### Conclusion : 

- It seems that setting `stallTimeout : 3000` is the sweet spot because while testing for me alchemyProvider took max to max 3000ms for returning the response and thus it was not delegated to `publicProvider` but I think it might vary for others 
- The reason for not using time less than `3000ms` is sometime while testing (very rare) `alchmeyProvider` took some time for response(wihout decresing network speed) and it triggered `publicProvider` which resulted in getting 429's errors. Although it never happened when using `stallTimeout : 3000 ` . 


PS : I think `wagmi` has fixed this issue in their newer version. so in future if we upgrade it,  this will be gone more details [here](https://github.com/wagmi-dev/wagmi/discussions/1146#discussioncomment-4162077)

packages/frontend/.env.local.example
```diff
@@ -0,0 +1,2 @@
+NEXT_PUBLIC_ENABLE_TESTNETS=true
+NEXT_PUBLIC_NETWORK_LOCALHOST=true
```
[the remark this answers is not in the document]
technophile-04: Ohh yes, this is a great idea ! added `.env.development`

technophile-04:
> What would be the advantage of using this VS. watch: true + pollingTime in the chainConfiguration? The latter seems to be working fine. It's just about request when mounting components?

Yes, we are only saving a request while mounting....Even I think "watch: true + pollingTime in the chainConfiguration" is better instead of the extra configuration for `cacheTime` and `staleTime` which don't have that much effect

packages/frontend/services/web3/wagmiConnectors.tsx
carletex: Great explanation. Now I get it. Thanks :)

carletex (approved):
Great Job @technophile-04 

Pushed a tiny tweak. Ready to merge!

PR #50, opened 2022-11-25
Add loaded flag to getContractMethods

damianmarti:
Add loaded flag to getContractMethods on utilsContract to allow to check if the contract was loaded.

Now when there aren't any read or write method, the UI show a clear message.


carletex (approved):
Thanks @damianmarti !

I added the types for the new return value.

Issue #49, opened 2022-11-25
Wrong Network indicator


carletex:
We want to show a "Wrong Network" indicator when `NEXT_PUBLIC_NETWORK` doesn't match with the selected network on rainbow kit.

Ideally it'll have a "switch network" button that will switch it to `NEXT_PUBLIC_NETWORK`

carletex:
Fixed in #51

PR #51, opened 2022-11-26
update #49 :wrong network indicator ui with quick switch button

Naim-Bijapure:
image


carletex (changes requested):
This PR is giving me an infinite RPC call loop:

image

carletex:
@Naim-Bijapure Investigating now as it might be something that conflicts with master. Maybe localhost / hardhat?

carletex:
@Naim-Bijapure seems that the infinite loop is fixed!

But now, I'm not getting the warning.

I'm connected to 

image

And public network is

image

I'm not getting the warning & switch.

Issue #48, opened 2022-11-24
Transactor / Notifier


carletex:
In SE-1 we have the Transactor util that we use to wrap TXs to get UI feedback while the tx is beign sent / pending / mined. It's using `bnc-notify` from Blocknative under the hood.

For now (v0.1), I think we could implement our own using Toasts (as we do on the Contract component):

```ts
  await reactHotToast.promise(
    writeAsync(),
    {
      loading: "Mining transaction, Hold tight!",
      success: "Mined successfully !",
      error: "Error while processing the transaction",
    },
    {
      success: {
        icon: "🔥",
      },
    },
  );
```

We could extract this into a util/component that takes the promise as an argument. We could use it when sending a write tx (contract component, faucet, etc)

carletex:
Fixed on #55

PR #55, opened 2022-11-30
Basic Transactor

technophile-04:
### Description : 
Setted up a bare minimum transactor hook

### Reason to use a custom hook instead of a function : 
created a custom hook instead of [transactor function](https://github.com/scaffold-eth/eth-components/blob/next/src/functions/transactor.ts) because instead of making `signer` argument **mandatory** for the function we can leverage the `wagmi` hook (useSigner) . Also since we are using wagmi hooks for making `wrtie` transaction , if we go with `function` approach we need to get `signer` separately everytime and then pass it to `transactor` function.

example (keeping transactor as function) : 
```js
const signer = useSigner()
const {writeAsync} = useContractWrite({ ...config });
const tx = Transactor(signer);
await tx(writeAsync())
```

example (keeping transactor as hook) : 
```js
const {writeAsync} = useContractWrite({ ...config });
const tx = useTransactor();
await tx(writeAsync())
```

Please let me know if I should go with the custom hook approach or not.

Also in issue #48, it was mentioned to make `util/component` was a bit confused about it. Please let me know if the current (custom hook) approach is correct or I should go with other approaches 🙏 . 

Thanks so much!


carletex:
> Please let me know if I should go with the custom hook approach or not.

Love the hook approach, make more sense. On the issue I mentioned a util/component thinking about v0.1, but you are on v1.0, which is great! Thanks Shiv!

Lmk when is ready for review.

technophile-04:
## Description : 
Updated the transactor to properly show synced messages while doing transactions. 

## Approach : 
The new `useTransactor` hook is very similar to [Transactor.js](https://github.com/scaffold-eth/eth-components/blob/next/src/functions/transactor.ts) in scaffold-eth. It's just that we can leverage the wagmi hook (useSigner) instead of passing the signer every time. 

### Demo - successful transaction on goerli (please skip the part 00:16 to 0:35 forgot to fast-forward) : 

https://user-images.githubusercontent.com/80153681/206261643-f9b3dde4-0aa1-4ee0-b58d-8bc9818004fd.mov

### Demo - showing error : 

https://user-images.githubusercontent.com/80153681/206261942-c2ead7ca-c4d1-4b59-afdd-8e63bc37a33c.mov

### Important points to note : 
If you look at this [block of code](https://github.com/scaffold-eth/eth-components/blob/77fe6753a9b0db7659e59764f2658fa5b03b9a87/src/functions/transactor.ts#L173) in transactor.js of se-typescript. Their they are waiting for transaction confirmation and then calling a `callback` even I did the same but took a different approach of leveraging `transactionRes.wait()` to make sure the transaction is included in a block and confirmed, referenced from ethers doc [here](https://docs.ethers.io/v5/api/providers/types/#providers-TransactionResponse) checkout `.wait()`. My approch is at line `84` of  [useTransactor](https://github.com/scaffold-eth/se-2/pull/55/files#diff-d3067bcf2ebbb6ecc55334af30911bf013c7a6ecd03decf595fc11d878eab952). Please let me know if we should do it similar to se-typescipt, I tried making it clean and easy to understand but please let me know. 

### TODOs : 

https://user-images.githubusercontent.com/80153681/206269407-25017889-460a-4c65-be76-9f8911d40230.mov

If we see here the error overflows the toast. Should I fix it in this PR or should we make a new issue for creating a custom toast with a se-2 theme for all stuff. 

Also should we create an issue for [getParsedEthersError](https://github.com/scaffold-eth/se-2/blob/74169bba53ad2877511118ddbb4e9f97b58a2acf/packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx#L161) for parsing error more efficiently?

packages/nextjs/components/scaffold-eth/toast.tsx
```diff
@@ -18,6 +18,36 @@ function Toast(message: string, alertClass: toastClassType) {
   });
 }
 
+// TODO Remove this in future
+/**
+ * Custom Txn Toast
+ */
+function TxToast(blockExplorerLink: string) {
```
technophile-04, line 25: Really sorry for adding this here in this file, was not able to find a better place at that moment 😅, please let me know if I should move it somewhere else

packages/nextjs/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx
```diff
@@ -60,26 +61,18 @@ export const WriteOnlyFunctionForm = ({
...
+        await writeTxn(writeAsync());
         setRefreshDisplayVariables(prevState => !prevState);
       } catch (e: any) {
         const message = getParsedEthersError(e);
-        toast.error(message);
+        toast.error(message, {
+          style: {
+            wordBreak: "break-all",
+          },
+        });
```
carletex, line 75: This is using our custom toast, which only take 1 param. This shouldn't be working.

packages/nextjs/components/scaffold-eth/toast.tsx
carletex: We definitely need to think better about how we use the toasts. Rn it's a bit messy, using react-hot-toast directly & our custom version on top of it at the same time.

Can your create an issue for it? We can discuss there. There are a few options.

But this is great, we've been using toast for a while, and now we understand better our needs.

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const { data } = useSigner();
+  if (signer === undefined && data) {
+    signer = data;
+  }
+
+  const result: TTransactionFunc = async (tx, callback) => {
+    let toastId = null;
+    let transactionReceipt: TransactionReceipt | undefined;
+    let transactionResponse: SendTransactionResult | TransactionResponse | undefined;
+    if (signer) {
```
carletex, line 33: Instead of wrapping everything inside `if (signer)`, let's do a "guard clause".

Return if !signer. I think the code is more readable that way.

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
```diff
@@ -0,0 +1,108 @@
+import { TransactionRequest, TransactionResponse, TransactionReceipt } from "@ethersproject/abstract-provider";
+import { SendTransactionResult } from "@wagmi/core";
+import { BigNumber, ethers, Signer } from "ethers";
+import { Deferrable } from "ethers/lib/utils";
+import { useSigner } from "wagmi";
+import toast from "react-hot-toast";
+import { getParsedEthersError } from "~~/components/scaffold-eth/Contract/utilsContract";
+import { toast as customToast } from "~~/components/scaffold-eth/index";
+
+const DEBUG = true;
```
carletex, line 10: I know this is used a lot on SE-1, but let's not do it here for now.

We can use logs for errors (BTW, we should use `console.error()` instead of `console.log()`)

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        if (tx instanceof Promise) {
+          if (DEBUG) console.log("AWAITING TX", tx);
+
+          toastId = toast.loading("Awaiting for user confirmation");
+          const transactionRes = await tx;
+          toast.remove(toastId);
+
+          // TODO add etherscan url in loading toast
```
carletex, line 57: Yep, I think this is important. You usually want to check the TX while is pending.

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
```diff
@@ -0,0 +1,108 @@
...
+          toastId = toast.loading("Mining transaction, Hold tight!");
+          transactionReceipt = await transactionRes.wait();
+          toast.remove(toastId);
+
+          customToast.tx(`${etherscanTxUrl}${transactionRes.hash}`);
```
carletex, line 62: We have to do something for localhost.

image

lt goes to `localhost.etherscan.io`. 

In SE-1, I think there is "Local tx sent" or something like that.

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+        let etherscanNetwork = "";
+        if (networkName && chainId && chainId > 1) {
+          etherscanNetwork = networkName + ".";
+        }
+
+        let etherscanTxUrl = "https://" + etherscanNetwork + "etherscan.io/tx/";
+        if (chainId && chainId === 100) {
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          etherscanTxUrl = "https://blockscout.com/poa/xdai/tx/";
```
carletex, line 47: I think we can abstract this into a util. (`networks.ts`?)

It can have a `getBlockExplorerTxLink` method

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        if (tx instanceof Promise) {
```
carletex, line 50: I think we can make this if/else simpler (since there are kind of doing the same thing).

I wrote this as an example (you might need to adjust some stuff):

```ts
      let transactionRes;
      toastId = toast.loading("Awaiting for user confirmation");
      if (tx instanceof Promise) {
        // Tx is already prepared by the caller
        transactionRes = await tx;
      } else if (tx != null) {
        // Raw tx
        if (!tx.gasPrice) {
          tx.gasPrice = gasPrice || ethers.utils.parseUnits("4.1", "gwei");
        }
        if (!tx.gasLimit) {
          tx.gasLimit = BigNumber.from(ethers.utils.hexlify(120000));
        }
        transactionRes = await signer.sendTransaction(tx);
      }

      toast.remove(toastId);
      // TODO add etherscan url in loading toast
      toastId = toast.loading("Mining transaction, Hold tight!");
      transactionReceipt = await transactionRes.wait();
      toast.remove(toastId);
```

Also wrote some comments on the if/else to make it clearer.

carletex (commented):
Hey @technophile-04 

Tested it and works!

I left some review comments. Most of them trying to make the Transactor hook more simple & readable.

packages/nextjs/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx
technophile-04: Yes, sorry my bad

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
technophile-04: Yup, tysm for this 🙌

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
technophile-04: Lol, don't know why I forgot this every time :( ...Tysm 🙌

packages/nextjs/components/scaffold-eth/toast.tsx
technophile-04: Yes, I updated the custom Txn toast, for now, will create an issues for Toasts

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
technophile-04: with updated commits, I have added a condition to not show the link if you are on localhost...Please let me know If I should update it

packages/nextjs/components/Spinner.tsx
```diff
@@ -0,0 +1,22 @@
+import React from "react";
```
carletex, line 1: This is how I see it (no animation). 

image

Maybe you see it correctly in the forced dark mode?

Also the "X" it's too close. No big deal for now tho.

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
```diff
@@ -0,0 +1,90 @@
...
+  tx: Promise<SendTransactionResult> | Deferrable<TransactionRequest> | undefined,
+  callback?: ((_param: any) => void) | undefined,
+) => Promise<Record<string, any> | undefined>;
+
+export const useTransactor = (_signer?: Signer, gasPrice?: number): TTransactionFunc | undefined => {
+  /**
+   * ### Summary 
+    if signer is provided in params use it(this means dev wants to send a raw transaction)
+    else use the signer from connect wallet (mostly its an wagmi writeTxn)
+   */
```
carletex, line 21: Maybe we can move this above the function as a JSDoc comment. Something like:

```ts
/**
 * Runs TXs showing UI feedback.
 * @param _signer If signer is provided => dev wants to send a raw tx.
 * @param gasPrice
 */
```

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
```diff
@@ -0,0 +1,90 @@
...
+          tx.gasLimit = BigNumber.from(ethers.utils.hexlify(120000));
+        }
+
+        transactionResponse = await signer.sendTransaction(tx);
+      } else {
+        throw new Error("Incorrect transaction passed to transactor");
+      }
+      toast.remove(toastId);
+
+      const etherscanTxnURL = network ? getBlockExplorerTxLink(network, transactionResponse.hash) : "";
```
carletex, line 59: nitpick: maybe rename it to `blockExplorerTxURL`?

packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -0,0 +1,24 @@
...
+    return "";
+  }
+
+  let etherscanNetwork = "";
+  if (name && chainId && chainId > 1) {
+    etherscanNetwork = name + ".";
+  }
+
+  let etherscanBaseTxUrl = "https://" + etherscanNetwork + "etherscan.io/tx/";
+  if (chainId && chainId === 100) {
```
carletex, line 17: This can just be `if (chainId === 100)`. Same in line 12, we can remove the chainId condition.

carletex (changes requested):
Thanks for the updates Shiv!

Almost there, just added some little things

packages/nextjs/components/Spinner.tsx
technophile-04: In https://github.com/scaffold-eth/se-2/pull/55/commits/041925bd7d24f1d5c30719125d8393f9422ef973 I have just updated the color, could please check it once again? 
For me it's working 
Light : 
Screenshot 2022-12-09 at 5 24 47 PM
Dark : 
Screenshot 2022-12-09 at 5 24 10 PM


PS: Even I had the same problem at the very start, the in-between animation was not showing, but it was solved once I changed the color of  `fill-color` in `Spinner.jsx`. Please let me know if the problem still exist (The spinner code was take from [here](https://flowbite.com/docs/components/spinner/))

carletex: TW is not compiling the filling color for me (since it's dynamic).

https://stackoverflow.com/a/71068925

This is why I did this:

image

Not sure why it's working for you :D 

Anyway, I hardcoded it for now: 64a3624e5b9bc8596fe092211cd584dbde34b195

carletex (approved):

PR #56, opened 2022-11-30
hotfix: minor fix wronge network alert defined chain value

Naim-Bijapure:


carletex:
Hey @Naim-Bijapure I'm still having the issue:

To reproduce:
1. yarn start / chain / deploy
2. Go to localhost:3000 =>  I'm autoconnected to the burner wallet. Everything works fine.
3. Switch to goerli on .env.development
4. Go to the browser, I get this:
image. Can't do anything there.

I think this in localStorage is making it to auto connect:

image

Naim-Bijapure:
@carletex  got your point. 
look like the issue is because of hot reload maybe some cache issue. 

can you try this way. 
1. switch to goerli from hardhat in .env.development
2.  on ui clear local storage and refresh the page (start fresh ) . 
3. and check the above use case it will work. 

maybe when we switch network on dev env due to hot reloading from nextjs some caching mismatch happens with burner wallet. 
and when we start fresh by clearing local storage and reloading it will work as expected.

PR #57, opened 2022-12-01
Refetch display variables in  Debug ContractUI

technophile-04:
### Approach taken : 
```js
  const [refreshDisplayVariables, setRefreshDisplayVariables] = useState(false);
```
created a common state(`refreshDisplayVariables`) which is shared by `WriteOnlyFunctionForm` and `DisplayVariables` component . `WriteOnlyFunctionForm`  **changes the state** and based on state changes `DisplayVariables` uses [`refetch`](https://0.6.x.wagmi.sh/docs/hooks/useContractRead#return-value) function to refetch the variables again. 

### Demo : 

https://user-images.githubusercontent.com/80153681/205015942-f6b85592-cd07-407a-965e-c32cfbb03873.mov

### Alternate approaches that can be considered : 
1) using `watch : true`  : 
This is a simple solution we can use, its just that each variable will keep making `eth_blocknumber` calls and its a bit less responsive since will be making `eth_blocknumber` after `3sec` to get new data.
2) using `queryKey`(in future if wagmi make it available) : 
As mentioned #32 we can use `key` if wagmi exposes it in future (currently its not exposed even in new version 0.8). wagmi using useQuery internally...[check this out](https://github.dev/wagmi-dev/wagmi/blob/d21a868a2d6bf31df38ff80a261456e264e5d128/packages/react/src/hooks/contracts/useContractRead.ts#L155)

### Some useful ongoing discussions on wagmi: 
1) https://github.com/wagmi-dev/wagmi/discussions/1238

### Some discussions on `refetch` which uses the same patterns as wagmi : 
1) https://github.com/apollographql/react-apollo/issues/562

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/DisplayVariables.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/DisplayVariables.tsx
@@ -8,10 +8,10 @@ import { toast } from "~~/components/scaffold-eth";
 type TDisplayVariableProps = {
   functionFragment: FunctionFragment;
   contractAddress: string;
-  displayVariablesRefresh: boolean;
+  refreshDisplayVariables: boolean;
 };
 
-const DisplayVariable = ({ contractAddress, functionFragment, displayVariablesRefresh }: TDisplayVariableProps) => {
+const DisplayVariable = ({ contractAddress, functionFragment, refreshDisplayVariables }: TDisplayVariableProps) => {
   const {
     data: result,
     isFetching,
@@ -30,7 +30,7 @@ const DisplayVariable = ({ contractAddress, functionFragment, displayVariablesRe
     (async () => {
       await refetch();
     })();
-  }, [refetch, displayVariablesRefresh]);
+  }, [refetch, refreshDisplayVariables]);
 
   return (
     <div className="border-b-2 border-black space-y-1 pb-2">
--- a/packages/nextjs/components/scaffold-eth/Contract/DisplayVariables.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/DisplayVariables.tsx
@@ -1,5 +1,5 @@
 import { FunctionFragment } from "ethers/lib/utils";
-import React from "react";
+import React, { useEffect } from "react";
 import { useContractRead } from "wagmi";
 import { tryToDisplay } from "./utilsDisplay";
 import { ArrowPathIcon } from "@heroicons/react/24/outline";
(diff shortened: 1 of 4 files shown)


packages/nextjs/components/scaffold-eth/Contract/DisplayVariables.tsx
```diff
@@ -25,6 +26,12 @@ const DisplayVariable = ({ contractAddress, functionFragment }: TDisplayVariable
     },
   });
 
+  useEffect(() => {
+    (async () => {
+      await refetch();
+    })();
```
carletex, line 32: Since we are not doing anything with the response || catching errors, we can just call `refetch()` with no async/await

carletex (commented):
Nice approach! Left a minor comment.

I noticed a bug while testing this (I think the bug wasn't introduced in this PR):

https://user-images.githubusercontent.com/2486142/205110153-24d6c4ee-97e5-415c-a15d-f39555229805.mp4

A RPC request is made each time you write in the input. It seems that some state change / re-rendering is causing this. Can you check that out? In this PR or in another one, up to you.

Thanks Shiv!

technophile-04:
> A RPC request is made each time you write in the input. It seems that some state change / re-rendering is causing this. Can you check that out? In this PR or in another one, up to you.

Yes, I think its something to do with `usePrepareContractWrite`  as mentioned [here](https://0.6.x.wagmi.sh/docs/prepare-hooks/intro#ux-pitfalls-without-prepare-hooks). 

I also found this while tinkering : 
The RPC requests are made only if you are connected to the burner wallet but If you connect your own wallet(eg. Metamask) it doesn't seem to make those requests. 

demo : 

https://user-images.githubusercontent.com/80153681/205861027-d641225c-f6b3-4493-ad13-42e6e018e475.mov


Possible solution : 
I tried using `useContractWrite` without `usePrepareContractWrite`, by passing `mode: 'recklesslyUnprepared'` as mentioned in the doc [here](https://0.6.x.wagmi.sh/docs/hooks/useContractWrite#mode) and it does seem to work. 

I would love to tackle it into a separate PR and find any alternatives to this if possible since it is recommended in the docs to pair `useContractWrite` with `usePrepareContractWrite` [here](https://0.6.x.wagmi.sh/docs/prepare-hooks/intro#ux-pitfalls-without-prepare-hooks). Also need to check properly why it is happening only with the burner wallet and not when you connect with your own wallet. But please let me know if I should create a separate PR or tackle it into this PR itself.

carletex (approved):
Great job @technophile-04 

Yeah, we can tackle it in a separate PR (Could you create an issue for it?). Not super urgent (since only happens with the burner wallet on localhost), so you can take #55 first.

Merging this, thanks!!

PR #58, opened 2022-12-05
Save ETH price in Zustand state

### LINKED ISSUE #52: Save ETH price in Zustand state
We have a `useEthPrice` hook that fetches the USD price of ETH. The Balance component is the only one using the hook rn, but we'll use it in some other places (like a ETH/USD price indicator on the bottom left of the page.)

So:

- Lets fire the hook when the App starts and save it to the global state.
- Update the Balance component to read the value from the state
- For now, no need to do polling.

damianmarti:
And use the price from the state at the Balance component.

Closes #52


packages/nextjs/pages/_app.tsx
```diff
@@ -10,7 +10,13 @@ import { appChains } from "~~/services/web3/wagmiConnectors";
 import { wagmiClient } from "~~/services/web3/wagmiClient";
 import Header from "~~/components/Header";
 
+import { useAppStore } from "~~/services/store/store";
+import { useEthPrice } from "~~/hooks/scaffold-eth";
+
 const ScaffoldEthApp = ({ Component, pageProps }: AppProps) => {
+  const price = useEthPrice();
+  useAppStore(state => state.priceSlice.setPriceData({ price: price }));
```
damianmarti, line 18: I'm not sure if it's the best place to set it. Please let me know if there is a better place.

packages/nextjs/services/store/slices/priceSlice.ts
```diff
@@ -0,0 +1,25 @@
...
+  };
+};
+
+export type TPriceSlice = {
+  priceData: TPriceData;
+  setPriceData: (newPriceState: TPriceData) => void;
+};
+
+export const createPriceSlice: TAppSliceCreator<TPriceSlice> = set => ({
+  priceData: defaultPriceState(),
```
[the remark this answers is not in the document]
damianmarti: Thanks for the review!

I changed it to a number. If we need an object we can change it later.

packages/nextjs/pages/_app.tsx
damianmarti: Changed!

PR #70, opened 2022-12-13
Eth price footer

### LINKED ISSUE #63: Show ETH / USD price indicator on the bottom left of the page
image

carletex:
Fixes #63 

Few other things tackled in this PR:
- Fix issue setting the Zustand state (for eth price)
- Created sticky footer
- Narrow the navbar

### DIFF
--- a/packages/nextjs/services/store/slices/ethPriceSlice.ts
+++ b/packages/nextjs/services/store/slices/ethPriceSlice.ts
@@ -12,8 +12,13 @@ export type TEthPriceSlice = {
 export const createEthPriceSlice: TAppSliceCreator<TEthPriceSlice> = set => ({
   ethPrice: defaultEthPriceState(),
   setEthPrice: (newValue: number): void =>
-    set((state): TAppStore => {
-      state.ethPriceSlice.ethPrice = newValue;
-      return state;
-    }),
+    set(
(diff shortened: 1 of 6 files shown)


packages/nextjs/services/store/slices/ethPriceSlice.ts
```diff
@@ -12,8 +12,13 @@ export type TEthPriceSlice = {
 export const createEthPriceSlice: TAppSliceCreator<TEthPriceSlice> = set => ({
   ethPrice: defaultEthPriceState(),
   setEthPrice: (newValue: number): void =>
-    set((state): TAppStore => {
-      state.ethPriceSlice.ethPrice = newValue;
-      return state;
-    }),
+    set(
+      (state): TAppStore => ({
+        ...state,
```
damianmarti, line 17: @carletex I think that this is not needed since the set method merges by default the state -> https://github.com/pmndrs/zustand/blob/main/docs/guides/immutable-state-and-merging.md

damianmarti: Oh! Thanks for the long explanation!

If we refactor our slices to be at the same level then we have to worry about avoiding name collitions.

I think we can leave it now as you did it and then we can decide if there is a better solution.

Issue #65, opened 2022-12-12
Header: Mobile / RWD Styling


carletex:
We don't have any final design yet, but we can start by make it look like this (used [daisy examples](https://daisyui.com/components/navbar/)):

![image](https://user-images.githubusercontent.com/2486142/207102756-fc322a17-3910-4d8f-8a35-39176776f8a4.png)

carletex:
https://daisyui.com/components/drawer/#navbar-menu-for-desktop--sidebar-drawer-for-mobile

carletex:
Fixed on #71

PR #71, opened 2022-12-14
Mobile drawer / styling

simplemachine92:
A start for #65


carletex:
Hey @simplemachine92 . Thanks for the PR! Is this ready for review? You can tag me as the reviewer when ready.

@edakturk14 was assigned to the  #65 issue, but I think she hasn't started working on it yet, so all good. But next time check if the issue is taken before working on it.

simplemachine92:
My fault, I'll check in first next time. Also, this should be ready for review when

- [ ] Drawer closes on selection
- [ ] Drawer closes properly when hamburger is clicked
- [ ] Wallet info is displayed after connecting wallet

edakturk14:
hey @simplemachine92! no worries, thanks for providing the list & contributing to the project. I hadn't started working so it was good timing. Are you planning to continue to work on the points or should I continue with the issue?

simplemachine92:
Hey eda, I'll complete them, if that's okay. ty for having me

simplemachine92:
Hey @carletex, ready for review.

Screenshot from 2022-12-14 18-54-15

camdengrieh:
The mobile drawer looks nice, however the nav bar is looking like this where the window is not quite small enough...
image

packages/nextjs/components/Header.tsx
```diff
@@ -4,14 +4,28 @@ import Link from "next/link";
...
+import { BugAntIcon, SparklesIcon, HomeModernIcon, BeakerIcon } from "@heroicons/react/24/outline";
+
+function checkAndCloseDropDown(e: { currentTarget: any }) {
+  const targetEl = e.currentTarget;
+  if (targetEl && targetEl.matches(":focus")) {
+    setTimeout(function () {
+      targetEl.blur();
+    }, 0);
+  }
+}
```
carletex, line 16: This is not the way to this in React. For opening / closing the dropdown we can have a state.

packages/nextjs/components/Header.tsx
```diff
@@ -4,14 +4,28 @@ import Link from "next/link";
...
   const router = useRouter();
 
   return (
     <Link href={href} passHref>
       <a
+        onClick={() => {
+          if (document.activeElement instanceof HTMLElement) {
+            document.activeElement.blur();
+          }
+        }}
```
carletex, line 28: We won't need this either if we use state.

packages/nextjs/components/Header.tsx
```diff
@@ -54,8 +68,64 @@ export default function Header() {
...
           </li>
+          <li>
+            <NavLink href="/ninja">
+              <BeakerIcon className="h-4 w-4" />
+              Ninja
+            </NavLink>
+          </li>
         </ul>
       </div>
+      <div className="navbar bg-base-100">
```
carletex, line 79: This is a navbar inside a navbar, it should be the same navbar for all (where you show/hide elements depending on the resolution OR two separate sibling navbars)

Also since your aro not hiding it, it makes the menu items to wrap:
image

packages/nextjs/components/Header.tsx
```diff
@@ -54,8 +68,64 @@ export default function Header() {
...
+        <div className="navbar-start">
+          <div className="dropdown sm:hidden">
+            <label
+              tabIndex={0}
+              className="btn btn-ghost ml-1 hover:bg-secondary"
+              onMouseDown={e => {
```
carletex, line 85: When using the "state" way of doing things, this could be `onClick`

packages/nextjs/components/Header.tsx
```diff
@@ -54,8 +68,64 @@ export default function Header() {
...
+                checkAndCloseDropDown(e);
+              }}
+            >
+              <svg
```
carletex, line 89: Let's not inline SVG here.

We use heroicons

packages/nextjs/components/Header.tsx
```diff
@@ -54,8 +68,64 @@ export default function Header() {
...
+                  Debug Contracts
+                </NavLink>
+                <NavLink href="/example-ui">
+                  <SparklesIcon className="h-4 w-4" />
+                  Example UI
+                </NavLink>
+                <NavLink href="/ninja">
+                  <BeakerIcon className="h-4 w-4" />
+                  Ninja
+                </NavLink>
```
carletex, line 119: You are repeating the same links here, which it's not good DX (having to change in two places).

If it's the same content, we could move it to a component above. If we need a different markup, we could have an array with the data 

```js
const menuItems = [
 {
  label: "Home",
  link: "/"
}
//...
  
];
```

packages/nextjs/components/Header.tsx
```diff
@@ -54,8 +68,64 @@ export default function Header() {
...
+              </li>
+            </ul>
+          </div>
+          {/* Maybe we want a scaffold-eth link in navbar besides hamburg menu */}
+          {/*  <Link href="/" className="sm:hidden btn btn-ghost text-sm">
+            Scaffold-eth
+          </Link> */}
```
carletex, line 126: We'll see when we have the final design. For now you can remove it.

packages/nextjs/components/scaffold-eth/Contract/NinjaUI.tsx
```diff
@@ -0,0 +1,93 @@
+import { Contract } from "ethers";
```
carletex, line 1: Yeah, let's remove this. As in the [contributing guidelines](https://github.com/scaffold-eth/se-2/blob/master/CONTRIBUTING.md), we want to keep the core simple an minimal.

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
```diff
@@ -47,7 +49,7 @@ export default function RainbowKitCustomConnectButton() {
                 );
               }
 
-              if (chain.unsupported || chain.id !== definedChain?.id) {
+              if (chain.unsupported || (chain.id !== definedChain?.id && pathName != "/ninja")) {
```
carletex, line 52: No ninja stuff :)

packages/nextjs/pages/ninja.tsx
```diff
@@ -0,0 +1,96 @@
+import type { NextPage } from "next";
```
carletex, line 1: Remove this also from the PR

packages/nextjs/components/scaffold-eth/index.tsx
```diff
@@ -4,3 +4,4 @@ export { default as Balance } from "./Balance";
 export { default as AddressInput } from "./AddressInput";
 export { default as toast } from "./toast";
 export { default as ContractUI } from "./Contract/ContractUI";
+export { default as NContractUI } from "./Contract/NinjaUI";
```
carletex, line 7: Remove

packages/nextjs/components/Header.tsx
carletex: Doesn't look good here:
image

carletex: On a standard mobile, it looks like this:

image

carletex (changes requested):
Hey @simplemachine92 

Check the review comments. Keep in mind that we want to keep the core (main branch) simple & minimal. Also some stuff doesn't look good or doesn't follow the "React" way of doing things.

packages/nextjs/components/Header.tsx
```diff
@@ -54,8 +68,64 @@ export default function Header() {
               Example UI
             </NavLink>
           </li>
+          <li>
+            <NavLink href="/ninja">
+              <BeakerIcon className="h-4 w-4" />
+              Ninja
+            </NavLink>
+          </li>
```
carletex, line 76: Remove

packages/nextjs/components/Header.tsx
simplemachine92: New version: 
Screenshot from 2022-12-15 16-40-07

Doesnt overlap when resized
Screenshot from 2022-12-15 16-42-16

packages/nextjs/components/Header.tsx
simplemachine92: I've left this in atm b/c it works when separating Navlinks into their own component and doesn't break anything in my testing.

simplemachine92:
@carletex Thanks for the review, and sorry about the ninja stuff, I didn't mean to add that to the same branch.

I've followed your review as closely as I could, but `onClick={() => {
          if (document.activeElement instanceof HTMLElement) {
            document.activeElement.blur();
          }
        }}` seems to work best with Navlinks being in a separate component, lmk what you think.

technophile-04:
Hey, @camdengrieh tysm for the PR  🙌

### Some suggestions :  
**Unifying Drawer in Header itself** :

If you look at [daisy ui responsive navbar example](https://daisyui.com/components/navbar/#responsive-dropdown-menu-on-small-screen-center-menu-on-large-screen) they have done it in one component itself, in our case that would be `Header.jsx`. 

Reason : I feel `Drawer.tsx` is a pretty small component and also Drawer.tsx container has some unessary classes like `navbar`. please checkout out [responsive navbar daisy UI example](https://daisyui.com/components/navbar/#responsive-dropdown-menu-on-small-screen-center-menu-on-large-screen)

---

**Removing `NavLinks` component and adding `Navlink` in `Header.jsx`** : 

Reason : we don't `NavLinks` component since we can directly map over `menuItems` checkout below point and dynamically render `li` in map. We can declare `NavLink` component on  top of `Header.jsx` since it won't be used by any other components, and its small too. 

---

**Using array of objects to render Navlinks instead of hardcoding** : 

As @carletex mentioned in one of his review its better to maintain a data structure where we can add links in the future and it automatically updates in the UI without changing the jsx of `Header` component. 

eg : As Carlos mentioned : 
```js
const menuItems = [
   {
    label: "Home",
    link: "/",
    icon : null
  },
  {
    label: "Debug Contracts",
    link: "/debug",
    icon : <BugAntIcon className="h-4 w-4" />
  },
];

```  

checkout this [blog](https://www.freecodecamp.org/news/dynamic-navigation-in-nextjs/) for reference. 

Please let me know if you need any help, or have doubts would love to help. 

Pardon me If I missed something, happy to discuss  🙌

camdengrieh:
credit to @simplemachine92

packages/nextjs/components/scaffold-eth/Drawer.tsx
```diff
@@ -0,0 +1,41 @@
...
+export default function Drawer() {
+  const [isOpen, setIsOpen] = useState(Boolean);
+
+  return (
+    <div className="sm:hidden navbar bg-base-100 min-h-0 flex-shrink-0">
+      <div className="navbar-start">
+        <div className="dropdown">
+          <button
+            className="btn btn-ghost ml-1 hover:bg-secondary"
+            // I'm not sure if strictmode is causing this to have inconsistent behavior?
```
technophile-04, line 18: Are you talking about this ? 

https://user-images.githubusercontent.com/80153681/208154998-83f8364b-92af-4db4-8d9d-724804045c79.mov

As we can see, if we click outside the drawer its closed but the state `IsOpen` is still true and thats why we need to click two times on hamburger icon to open drawer nextime . 

To fix this we you can create a `ref` use it to keep track...checkout out [this](https://www.codingdeft.com/posts/react-on-click-outside/) for reference

packages/nextjs/components/scaffold-eth/Navlinks.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const router = useRouter();
+
+  return (
+    <Link href={href} passHref>
+      <a
+        onClick={() => {
+          if (document.activeElement instanceof HTMLElement) {
+            document.activeElement.blur();
+          }
+        }}
```
technophile-04, line 16: Also we can remove this and pass in `setIsOpen` as props to `NavLink` and use it here to close the drawer.

simplemachine92:
Thanks for the great notes @technophile-04. Lmk what you think about this revision!

technophile-04:
@simplemachine92 thanks so much, this is really great !!

Added few comments, we are almost there 🙌

packages/nextjs/components/Header.tsx
```diff
@@ -1,17 +1,53 @@
...
-import { BugAntIcon, SparklesIcon } from "@heroicons/react/24/outline";
+import { Bars3Icon, BugAntIcon, SparklesIcon } from "@heroicons/react/24/outline";
+import { useRouter } from "next/router";
+
+/**
+ * Defs for NavLinks
+ ** @note Are we sure we don't want this in a separate component?
+
+ ** @note I looked for a method to pull the routes from next useRouter, is it possible to auto-populate
+          this list?
```
technophile-04, line 14: > Are we sure we don't want this in a separate component?

Yup, as mentioned in the previous comment since `NavLink` component is small and will not be used by any other component we can keep it here : )

> I looked for a method to pull the routes from next useRouter, is it possible to auto-populate
    this list?

The returned [router object](https://nextjs.org/docs/api-reference/next/router#router-object) from `useRouter` don't seem to have any direct property/method to access all routes, but I might have missed something. 
Also, the current approach  looks good to me since we want to give options to add custom Icons too. 

PS : we can remove this comments : )

packages/nextjs/components/Header.tsx
```diff
@@ -20,14 +56,65 @@ const NavLink = ({ href, children }: { href: string; children: React.ReactNode }
...
+      {menuItems.map((item, index) => {
+        return (
+          <li key={index}>
+            <NavLink href={item.link} setIsOpen={setIsOpen}>
+              {item.icon}
+              {item.label}
+            </NavLink>
+          </li>
+        );
+      })}
```
technophile-04, line 71: maybe we can inline it in the Header itself, instead of creating a separate component for only mapping over `menuItems` ? but upto you  🙌

packages/nextjs/components/Header.tsx
```diff
@@ -20,14 +56,65 @@ const NavLink = ({ href, children }: { href: string; children: React.ReactNode }
...
+    </>
+  );
+};
+
+/* End of possible Navlinks Components */
```
technophile-04, line 76: we can remove this

packages/nextjs/components/Header.tsx
```diff
@@ -20,14 +56,65 @@ const NavLink = ({ href, children }: { href: string; children: React.ReactNode }
...
+
 /**
  * Site header
  */
 export default function Header() {
+  const [isOpen, setIsOpen] = useState(Boolean);
+  const ref = useRef<HTMLInputElement>(null);
```
technophile-04, line 83: Instead of `HTMLInputElement ` we can use `HTMLDivElement` since we are keeping reference of `Div`

packages/nextjs/components/Header.tsx
```diff
@@ -20,14 +56,65 @@ const NavLink = ({ href, children }: { href: string; children: React.ReactNode }
...
+
+  // Is this considered accessing the DOM? Is this okay?
+  useEffect(() => {
+    function handler(event: { target: any }) {
+      if (!ref.current?.contains(event.target)) {
+        setIsOpen(false);
+      }
+    }
+    window.addEventListener("click", handler);
+    return () => window.removeEventListener("click", handler);
```
technophile-04, line 93: >  Is this considered accessing the DOM? Is this okay?

I was unable to find a better way, but please fill free to implement it if you find a better approach 🙌

Any particular reason for using `window` here ? 

maybe we use `document.body.addEventListener("click", handler)`  to be more specific since event will hit document first. checkout [this](https://stackoverflow.com/a/59370985) for more info

packages/nextjs/components/Header.tsx
```diff
@@ -20,14 +56,65 @@ const NavLink = ({ href, children }: { href: string; children: React.ReactNode }
...
+
   return (
     <div className="navbar bg-base-100 min-h-0 flex-shrink-0">
       <div className="navbar-start">
-        <div className="flex items-center	gap-2 mx-4">
+        <div className="sm:hidden dropdown" ref={ref}>
+          <button
+            className="btn btn-ghost hover:bg-secondary"
+            onClick={() => {
+              setIsOpen(!isOpen);
```
technophile-04, line 103: It would be nice if we do : 
```diff
+ setIsOpen(prevIsOpenValue => !prevIsOpenValue)
- setIsOpen(!isOpen)
```
since we are more sure we will be getting more up-to-date value, just a suggestion

technophile-04:
Also @simplemachine92 can we handle this too, since the links start wrapping too soon and for some users it might be the default behavior since the width range at which its breaks is common : 

https://user-images.githubusercontent.com/80153681/208246657-5f75e3bc-e27d-43f0-be3b-36f54fea1351.mov

Thanks so much 🙌🙌

simplemachine92:
@technophile-04, here you are. lmk.

technophile-04:
This is great ! thanks @simplemachine92,  great job !!! . Added a few commits to fix some minor issues 🙌. 

@carletex would love it if you could review it once 🙌 

PS: Currently the navbar is not optimized for devices like tabs and Ipads, the nav links wrap on these devices. Should we show the drawer on these devices too? or do we want to show links at these screen widths? A bit confused about this

carletex:
Pushed a few tweaks. Merging this!

Thanks

PR #72, opened 2022-12-15
Add interval to the hook

camdengrieh:
The useEthPrice() Hook fetches the price from the ETH/DAI pool every 15 seconds, instead of at every point in time the side effect is being used. This can be adjusted accordingly in setInterval()


carletex:
Hey @camdengrieh  Thanks for the PR.

> Instead of at every point in time the side effect is being used

The call is being made only when the provider changes. So it's usually once or twice. The idea of introducing polling is to have an up to date price.

Also, your approach have a problem: it's waiting 15 seconds to make the first call, and we want that to happen immediately, and the poll every X seconds. (we have the `NEXT_PUBLIC_RPC_POLLING_INTERVAL` env var BTW)

technophile-04:
Hey @camdengrieh, since we are setting up the `setInterval` in the `useEffect` that has a dependency in our case `[provider]` we might possibly run into an issue of [memory leak](https://blog.devgenius.io/react-memory-leaks-what-why-and-how-to-clean-them-up-93606a07de84) if the provider changes. 

**Some possible solutions** : 
- Maybe creating another `useEffect` with empty dependency and setting up `setInterval` in there. 
- Another clean way will be utilizing [`useInterval`](https://usehooks-ts.com/react-hook/use-interval) since we have [`usehook-ts`](https://usehooks-ts.com/) already installed and they might have handled good edge-cases. 

----

**Summarizing the points @carletex mentioned and above point we have** : 
- [ ] Using `NEXT_PUBLIC_RPC_POLLING_INTERVAL` instead of hardcoding `15000` 
- [ ] Fetching immediately on initial render and the poll
- [ ] Handle memory leaks properly  

Pardon me if I have missed something, would love to discuss 🙌

camdengrieh:
Thanks @technophile-04 & @carletex for the feedback, this is all simply a learning experience for me. I made some changes considering the information you had given me. Could you please review and see if I am missing something.
I have:
- [x] Used `NEXT_PUBLIC_RPC_POLLING_INTERVAL` instead of hardcoding
- [x] Price should be fetching on Initial render?
- [x] Used the useInterval hook with an empty dependency

camdengrieh:
Also, would it not be more optimal to instantiate the `route` variable with the known pair address directly (0xa478c2975ab1ea89e8196811f51a7b7ade33eb11), instead of fetching the pair address etc.?

technophile-04:
Thanks, @camdengrieh, this is great !! 

> Also, would it not be more optimal to initialize the route variable with the known pair address directly (0xa478c2975ab1ea89e8196811f51a7b7ade33eb11), instead of fetching the pair address etc.? 

Actually I didn't get the question properly 😅, trying to reply from my understanding : 
```js
// we are not just fetching the address of the pair here but are actually fetching the whole Pair metadata which contains the current price and many more things.
const pair = await Fetcher.fetchPairData(DAI, WETH[DAI.chainId], provider);
```
checkout the [`fetchPairData`](https://docs.uniswap.org/sdk/v2/reference/fetcher#fetchpairdata) for more description and return type. 

Also the `Route` class from `@uniswap/sdk` expects the first argument as an array of type `Pair` and not just an address so we can't hardcode. 

Sorry If I missed something, but I think the current approach is fine 🙌...

Added some comments, we are almost ready to merge 🙌

packages/nextjs/hooks/scaffold-eth/useEthPrice.ts
```diff
@@ -28,5 +27,24 @@ export const useEthPrice = () => {
     fetchPriceFromUniswap();
   }, [provider]);
 
+  useInterval(
+    () => {
+      const fetchPriceFromUniswap = async () => {
+        try {
+          const DAI = new Token(1, "0x6B175474E89094C44Da98b954EedeAC495271d0F", 18);
+          const pair = await Fetcher.fetchPairData(DAI, WETH[1], provider);
```
technophile-04, line 35: instead of `WETH[1]` we can keep ` WETH[DAI.chainId]` as it was previously since it describes itself instead of putting a magic number `1` 🙌

packages/nextjs/hooks/scaffold-eth/useEthPrice.ts
```diff
@@ -15,10 +15,9 @@ export const useEthPrice = () => {
     const fetchPriceFromUniswap = async () => {
```
technophile-04, line 15: since we are using same `fetchPriceFromUniswap` twice, lets move it to `uitls/scaffold-eth` with name `fetchETHPriceFromUniswap` which returns the type of number (ethPrice) and takes in `provider` as argument

camdengrieh:
Oh I see, you cleared everything up for me in regards to fetching the price from Uniswap; I didn't notice that `pair` had to be typeOf Pair, when initialising `route`, but I understand now that it's more than just the address that needs to be provided...

camdengrieh:
Ready for review @technophile-04 

- [x] Reverted back to `WETH[DAI.chainId]`
- [x] Moved `fetchETHPriceFromUniswap` to `utils/scaffold-eth`  - returns price as typeOf number - takes `provider` as argument

technophile-04:
Tysm @camdengrieh! It looks way cleaner !

added some very small suggestions below, I think we are almost ready to merge 🙌

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
```diff
@@ -0,0 +1,14 @@
+import { Fetcher, Route, Token, WETH } from "@uniswap/sdk";
+import { Provider } from "@wagmi/core";
+
+export default async function fetchPriceFromUniswap(provider: Provider) {
```
technophile-04, line 4: can we export this function through `/utils/scaffold-eth/index.ts`? checkout out how we exported `getLocalProvider` from their

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
```diff
@@ -0,0 +1,14 @@
...
+  try {
+    const DAI = new Token(1, "0x6B175474E89094C44Da98b954EedeAC495271d0F", 18);
+    const pair = await Fetcher.fetchPairData(DAI, WETH[DAI.chainId], provider);
+    const route = new Route([pair], WETH[DAI.chainId]);
+    const price = parseFloat(route.midPrice.toSignificant(6));
+    return price;
+  } catch (error) {
+    console.log("useEthPrice - Error fetching ETH price from Uniswap: ", error);
```
technophile-04, line 12: maybe we can use `console.error` instead of `console.log` : 
```diff
- console.log("useEthPrice - Error fetching ETH price from Uniswap: ", error);
+ console.error("useEthPrice - Error fetching ETH price from Uniswap: ", error);
```

camdengrieh: Done

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
camdengrieh: Done

technophile-04:
Thanks @camdengrieh  🙌 !!

carletex (commented):
Great work @technophile-04 and @camdengrieh !

It works perfectly. I added a few tweaks (check the commits):

**Enable/Disable polling flag on useEthPrice**
In general, the defaults should be to don't hit the RPCs too hard. I added a enable/disable flag for the polling.

**Use asyn/await.**  
As a convention, we use async/await in the project (VS .then())

**Return price = 0 on error. Explicit TS type on return functions**
I think it's better if the util always return a number. Now it returns 0 when error (hope this is never the real value of ETH 😂)

@technophile-04 Can you review & test these changes?

camdengrieh:
Firstly, thanks for all the feedback, i'm learning a lot just from your feedback @technophile-04 & @carletex 

> **Return price = 0 on error. Explicit TS type on return functions** I think it's better if the util always return a number. Now it returns 0 when error (hope this is never the real value of ETH 😂)

Lol, my bad for this one, I don't know why I added this one

technophile-04:
Thanks, @carletex. Everything seems to work perfectly 🙌 . Just noticed a small thing while testing  : 

In the below video, what I did was block the RPC call to mainnet so that it errors out for testing and notice that the `price` div at footer dosen't show off with `0` as price : 

https://user-images.githubusercontent.com/80153681/208462276-cebf921a-76e9-442a-a89c-b29f75b26dc8.mov

It seems that at [`Footer.jsx`](https://github.com/scaffold-eth/se-2/blob/bdd3f1bbfa710b0ef0276377dea259120c1d0169/packages/nextjs/components/Footer.tsx#L15) we have this condition which shows the `price` div only when its > 0: 
```js 
 {ethPrice > 0 && (
          <div className="btn btn-primary btn-sm font-normal cursor-auto">
            <CurrencyDollarIcon className="text-black h-4 w-4 mr-0.5" />
            <span>{ethPrice}</span>
          </div>
 )}

```
update the condition to `ethPrice >= 0` at [de3c1f3](https://github.com/scaffold-eth/se-2/pull/72/commits/de3c1f3e14255ce9e2e33087ef28245f97bb80f1), please let me know if I should revert it 🙌. 

Other than that everything works perfectly and we are ready to merge :)

technophile-04 (approved):

carletex:
So happy you are learning from this @camdengrieh <3

Thanks for the review @technophile-04 Merging this!!!

damianmarti:
@technophile-04 @carletex sorry for the late comment. Maybe we can replace the price with a spinner if the ethPrice is zero. I don't want to see a zero eth price :-)

camdengrieh:
@damianmarti I thought we don't show anything until there is an ethPrice greater than zero now?

damianmarti:
@camdengrieh Oh! yes! Sorry!

PR #77, opened 2022-12-23
UI styles for debug contract

technophile-04:
### Setted up a basic layout for UI design : 
Screenshot 2022-12-23 at 5 31 21 PM


technophile-04:
### Description : 
Almost done with 0th iteration of the `debug contract` page : 

https://user-images.githubusercontent.com/80153681/209548802-42f33783-885f-4ba5-b541-9c469e150942.mov

### TODO's : 
- [ ] Show the contract address and it's balance
- [ ] Truncate addresses for `DisplayVariables` eg : 0x55b9CB0bCf56057010b9c471e7D42d60e1111EEa -> 0x55b9...EEa
- [ ] Cleanup and polish the UI 

### Wanted some suggestions: 
1) I have used daisy UI's [Collapse component](https://daisyui.com/components/collapse/) to show the read and write contract's function. Seems like daisy UI does this by using CSS only because of which in future if we need open/close state then it will be somewhat difficult to get. Please let me know if I should implement the collapsable manually with react state, happy to refactor the code :) 
eg: Where we might need open/close state is changing the `DownArrow` icons besides read and write buttons : 
Screenshot 2022-12-26 at 6 15 45 PM

2) While Implementing the design I found repeating the same class names for different div's, is it ok if [extract these classes](https://tailwindcss.com/docs/reusing-styles#extracting-classes-with-apply) and put them inside globas.css ?

3) Feel free to point out bad coding practices and stuff that can be done better...happy to refactor it again!

packages/hardhat/contracts/YourContract.sol
```diff
@@ -4,21 +4,39 @@ pragma solidity >=0.8.0 <0.9.0;
 import "hardhat/console.sol";
```
technophile-04, line 4: Revert the contract to our basic `YourContract.sol`, added this for testing

packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -28,3 +28,19 @@ export function getBlockExplorerTxLink(network: Network, txnHash: string) {
 
   return blockExplorerTxURL;
 }
+
+export function getNetworkColor(chainName: string | undefined) {
```
technophile-04, line 32: Please let me know what you think about it : 
used to diplay networks name with desired color : 
Screenshot 2022-12-27 at 5 38 17 PM

Should we add [`NETWORKS`](https://github.com/scaffold-eth/scaffold-eth/blob/923cff6aab3b52589a8f128dc42d142b7791e702/packages/react-app/src/constants.js#L26) similar to scaffold-eth v1 ?

carletex (changes requested):
Thanks for this @technophile-04. Looking great!!

Left a few review comments.

Answering to your suggestions:

> I have used daisy UI's [Collapse component](https://daisyui.com/components/collapse/) to show the read and write contract's function. Seems like daisy UI does this by using CSS only because of which in future if we need open/close state then it will be somewhat difficult to get. Please let me know if I should implement the collapsable manually with react state, happy to refactor the code :)
eg: Where we might need open/close state is changing the DownArrow icons besides read and write buttons :

It seems that Daisy has the `collapse-arrow` class. Could you try that? https://daisyui.com/components/collapse/#with-arrow-icon

BTW, **Can you make the content visible by default?** https://daisyui.com/components/collapse/#force-open

> If you look at Address input in the above video it doesn't seem to fit well since we are using the AddressInput component there.

Rn the AddressInput is not used anywhere else. So what if we tweak it as we need... and we can think about generalizing it OR abstract it in the future? let's see how that looks like!

> While Implementing the design I found repeating the same class names for different div's, is it ok if [extract these classes](https://tailwindcss.com/docs/reusing-styles#extracting-classes-with-apply) and put them inside globas.css ?

This is tricky! I'd say: If it's a group of classes that we are going to use in different component / pages, then @apply sounds good. But if it's for divs in the same component, I don't think it makes sense (for now at least!). We can always refactor later on.

> Feel free to point out bad coding practices and stuff that can be done better...happy to refactor it again!

Everything looks clean and minimal ;)

Thanks Shiv!

packages/nextjs/components/scaffold-eth/Contract/InputUI.tsx
```diff
@@ -37,7 +37,7 @@ const InputUI = ({ setForm, form, stateObjectKey, paramType }: TInputUIProps) =>
   }
 
   return (
-    <div className="flex space-x-2 items-end">
+    <div className="flex items-end border-2 border-gray-100 bg-base-200 rounded-xl text-primary/90 justify-between pr-3">
```
carletex, line 40: I feel that the placeholder color it's too strong: we can make the color (or opacity) lighter... or use another color, like the grey in the Address component?

image

(Not super important because a lot of this tweaks will come later on... but maybe we make it look better for now)

technophile-04:
Hey @carletex, tysm for the suggestions super helpful !!

Update PR with the requested changes 🙌

packages/nextjs/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx
```diff
@@ -89,11 +89,15 @@ export const WriteOnlyFunctionForm = ({
...
-    <div className="flex flex-col items-start space-y-2 border-b-2 border-black pb-2">
-      <p className="text-black my-0">{functionFragment.name}</p>
+    <div className="flex flex-col gap-3 text-gray-700">
+      <p className="font-medium my-0 break-words">{functionFragment.name}</p>
       {inputs}
       {functionFragment.payable ? <TxValueInput setTxValue={setTxValue} txValue={txValue} /> : null}
-      <button className={`btn btn-primary btn-sm ${isLoading && "loading"}`} onClick={handleWrite}>
+      <button
+        className={`btn btn-secondary rounded-3xl font-medium shadow-md shadow-secondary h-[2.8rem] min-h-[2.8rem]
+        text-sm normal-case self-end border border-gray-200 ${isLoading && "loading"}`}
```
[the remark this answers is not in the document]
technophile-04: Ohhh, will surely remember next time 🙌

carletex (approved):
Great work @technophile-04 🚀

Pushed a few tweaks & merging it!

PR #83, opened 2023-01-01
theme toggle button

### LINKED ISSUE #79: Light/ Dark mode switch
The dark mode colors (+ needed config) are already in `tailwind.config.js`.

Let's create a switch for light/dark mode

Info:
- https://daisyui.com/docs/themes/ (Daisy recomends https://github.com/saadeghi/theme-change. We can take a look)
- https://daisyui.com/components/toggle/

image
image

technophile-04:
### Description 
Setted up a basic toggle button : 

https://user-images.githubusercontent.com/80153681/210176539-4d93f3e8-a611-4f96-bbc4-ad47d86e70a4.mov

### Approach : 
Utilized [`useDarkMode` ](https://usehooks-ts.com/react-hook/use-dark-mode) hook provided by usehooks-ts library. 

### Reason for not using [`theme-change`](https://github.com/saadeghi/theme-change) : 
- I felt that theme-change could have been more useful if we have had more than two themes because for only light and dark theme we have already have `useDarkMode` which takes care off cases from persisting it in localstorage to grabbing the default OS theme and setting it based on that. 
- but if we plan to have more themes then theme-change seems best for this purpose. 

### TODO : 
- [x] Currently, we can see that in the demo there is only a static `SunIcon` make it dynamic 

Fixes #79

### DIFF
--- a/packages/nextjs/components/SwitchTheme.tsx
+++ b/packages/nextjs/components/SwitchTheme.tsx
@@ -1,0 +1,20 @@
+//SwitchTheme.tsx
+import React, { useEffect } from "react";
+import { useDarkMode } from "usehooks-ts";
+import { SunIcon } from "@heroicons/react/24/outline";
+const SwitchTheme = () => {
+  const { isDarkMode, toggle } = useDarkMode(false);
+
+  useEffect(() => {
+    const body = document.body;
+    body.setAttribute("data-theme", isDarkMode ? "scaffoldEthDark" : "scaffoldEth");
+  }, [isDarkMode]);
+
+  return (
+    <div className="flex space-x-2">
+      <input type="checkbox" className="toggle toggle-primary bg-primary" onChange={toggle} checked={isDarkMode} />
+      <SunIcon className="swap-off h-6 w-6" />
+    </div>
+  );
+};
+export default SwitchTheme;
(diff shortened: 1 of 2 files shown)


packages/nextjs/components/SwitchTheme.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <div className="flex space-x-2">
+      <input
+        id="theme-toggle"
+        type="checkbox"
+        className="toggle toggle-primary bg-primary"
+        onChange={toggle}
+        checked={isDarkMode}
+      />
+      {isMounted() && (
```
technophile-04, line 23: The reason for doing this is there is the inconsistent output we don't do it : 

**Case when there is inconsistent output**  : 
- select dark mode and close the browser
- now when you open it again since during the build of NEXT.js page localstorage is not available initially `isDarkMode` is `false` (in theory it should have been true because we choose dark mode earlier)

demo of inconsistent behavior(withou `isMounted()`): 

https://user-images.githubusercontent.com/80153681/210214863-5532bb17-0cbc-4799-ba33-1525e249591e.mov

Take a look at `Icons` ^

carletex: This is a good catch!

The `useIsMounted` could be useful for nextjs hydration errors!

carletex (approved):
Good stuff @technophile-04 

Pushed a tiny tweak & merging this.

Thanks!!

PR #84, opened 2023-01-02
Fix /debug page dark mode

technophile-04:
### Description : 
Previously the /debug was not compatible with dark mode and the UI was looking bad, tried fixing the color's also now utilizing colors from theme so they change dynamically.

Demo : 

https://user-images.githubusercontent.com/80153681/210276184-29a6f980-7701-46f2-a8c5-c34c8f98c0fa.mov

### DIFF
--- a/packages/nextjs/components/scaffold-eth/AddressInput.tsx
+++ b/packages/nextjs/components/scaffold-eth/AddressInput.tsx
@@ -52,7 +52,7 @@ const AddressInput = ({ value, name, placeholder, onChange }: TAddressInputProps
             name={name}
             type="text"
             placeholder={placeholder}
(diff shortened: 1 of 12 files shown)


carletex:
Now it looks waay better! Thanks @technophile-04 

I pushed a commit with some tiny fixes:
- Removed opacity on colors (except placeholders). We might need to tweak the colors on the config file, but we shouldn't use opacity to tweak the colors (except in maybe some cases: placeholders, disabled text, etc)
- Borders: in light mode it looked weird (border color matching the page bg)
Before
image
After
image
- Fix hardhat color on network file (extra 6 :D)


Once thing that we could explore in the future: use [daisy cards](https://daisyui.com/components/card/) instead of our custom classes.

Issue #59, opened 2022-12-06
RPC calls on keypress in /debug contracts page


technophile-04:
### Description : 
If we go to the `/debug` page and fill in Input for the write transaction, for each keypress there is an `eth_estimateGas` RPC call.

Below is the  video demonstrating the issue that was mentioned by Carlos in #57 : 

https://user-images.githubusercontent.com/2486142/205110153-24d6c4ee-97e5-415c-a15d-f39555229805.mp4

PS: The RPC requests are made only if you are connected to the burner wallet but If you connect your own wallet(eg. Metamask) it doesn't seem to make those requests. checkout out the demo [here](https://github.com/scaffold-eth/se-2/pull/57#issuecomment-1338977686) 

### Possible Cause : 
It may be `usePrepareContractWrite ` which is making this calls as mentioned in the wagmi doc [here](https://0.6.x.wagmi.sh/docs/prepare-hooks/intro#ux-pitfalls-without-prepare-hooks). 

### Possible solution : 
Using `mode: 'recklesslyUnprepared'` for `useContractWrite` hook. 

PS : There might be some alternatives.

technophile-04:
## Background / description : 

Feel free to skip and checkout _summary and suggestion needed_ directly 🙌

I tried debugging why there were no RPC calls made when you connect your wallet and turns out that any extension's(in our case its metamask) network requests are not shown in network tab, you need particularly inspect extensions network tab here is the [reference](https://stackoverflow.com/questions/50673373/how-to-inspect-network-traffic-from-chrome-extensions). 

Below is the video that shows metamask's internal RPC is used to make this request. The below video is in on localhost but I checked on goerli too and metamask's internal RPC is used (left side console is of metamask extenstion) : 

https://user-images.githubusercontent.com/80153681/208920943-a7185c67-29fa-4668-9b20-b4e5742bb3eb.mov

## Some small issues in current Contract component specifically with [writeOnlyFunctionForm.jsx](https://github.com/scaffold-eth/se-2/blob/main/packages/nextjs/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx) : 

[...]

Issue #80, opened 2022-12-30
Default Solidity Contract


carletex:
On SE-1 we have a basic [set purpose contract](https://github.com/scaffold-eth/scaffold-eth/blob/master/packages/hardhat/contracts/YourContract.sol) as the default solidity contract.

Let's discuss what would be good to have as the **default contract on SE-2**.

I feel that we could have:

- More types: address (owner?), uint (counter?), array / struct?
- Payable & non payable function
- Withdraw function?
- Extending some openzeppelin?
..

I think having this out of the box, make the /debug page richer. It will also be a key part of #81 

P.S: Right now, on the [SE-2 contract](https://github.com/scaffold-eth/scaffold-eth/blob/master/packages/hardhat/contracts/YourContract.sol) we have a few pointless stuff. (the intent was to the display all the posible types / cases on the debug page to do the UI/UX correctly)

damianmarti:
I think that maybe we can have 2 example contracts. One really simple, like the Purpose contract and another one with more features, to be able to have a more complete example page based on it. What do you think?

damianmarti:
The reason to have two contracts is to keep the first contract simple to new users.

technophile-04:
**Here are some of my contract ideas:**

1) v2 of Purpose contract : 
      - The idea is - In the beginning, you have your(owner) purpose set and once the contract is deployed people can change your purpose but there will be a minimum required amount to change the purpose.
      - Only the owner can withdraw the funds
      - Some data structures : mapping(address => purpose), event(address, purpose, amount sent), Counter(noOfTime purpose changed)
      - don't know if it makes sense 😅

2)  FundMe / By me a Coffee contract : 
      - fund me example [here](https://github.com/technophile-04/hardhat-codes/blob/main/hh-fund-me-ts/contracts/FundMe.sol)
      - By me a Coffee example [here](https://docs.alchemy.com/docs/how-to-build-buy-me-a-coffee-defi-dapp)
      - these are just some ideas and we can add/remove stuff accordingly
3) YourNFT contract from se-typescript : 
      - an example is [here](https://github.com/scaffold-eth/scaffold-eth-typescript/blob/release/packages/solidity-ts/contracts/YourNFT.sol)
  
4) A simpler and more robust version of the challenge-1-staking contract : 
      - It will be simple but we can have some more data structures like structs and all.
      
I agree with Damu that it shouldn't be a complex contract and must be intuitive and even a beginner should be able to grasp it in one go. 

[...]

PR #86, opened 2023-01-04
Basic draft for v2 of Purpose contract

technophile-04:
### Description
This is a basic draft for #80

### DIFF
--- a/packages/hardhat/contracts/YourContract.sol
+++ b/packages/hardhat/contracts/YourContract.sol
@@ -39,7 +39,7 @@ contract YourContract {
   function changePurpose(string memory newPurpose) public payable costs {
     addressToPurpose[msg.sender] = newPurpose;
     purpose = newPurpose;
-    emit PursposeChange(msg.sender, newPurpose, msg.value);
+    emit PurposeChange(msg.sender, newPurpose, msg.value);
   }
 
   /* ========== INTERNAL FUNCTIONS ========== */
--- a/packages/hardhat/contracts/YourContract.sol
+++ b/packages/hardhat/contracts/YourContract.sol
@@ -4,45 +4,61 @@ pragma solidity >=0.8.0 <0.9.0;
 import "hardhat/console.sol";
 
 contract YourContract {
-  /* ========== TYPE DECLARATIONS ========== */
 
-  /* ========== STATE VARIABLES ========== */
-  uint256 public immutable i_changePurposePrice;
-  string public purpose;
-  mapping(address => string) public addressToPurpose;
+    // State Variables
+    address public immutable owner;
+    string public purpose = "Building Unstoppable Apps!!!";
+    bool public premium = false;
+    uint256 public price = 0.001 ether;
+    uint256 public totalCounter = 0;
+    mapping(address => string) public addressToPurpose;
 
-  /* ========== EVENTS ========== */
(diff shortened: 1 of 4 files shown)


carletex:
Hey @technophile-04 great starting point!

I know this was still a Draft, but since I've been talking with Austin & Andrea about this, I took a look and committed some stuff.

I'm thinking in 3 things:
- Having a rich /debug page
- Using some of the most common stuff in solidity
- Having an Example UI that can play nice with the contract.

This is just a first try, ofc! Even if we merge this, I'm expecting to make a bunch of tweaks afterwards.

Can you test and see if everything works as expected? I noticed that the Balance (from the burner wallet AND the contract) is not updating.

carletex:
@technophile-04 Made another tweak to make the contract even simpler for now.

>  I noticed that the Balance (from the burner wallet AND the contract) is not updating.

I think this was just only a matter of not showing enough ETH decimals. Fixed it!

technophile-04:
### Description  
While testing found a bug that if you try to withdraw funds from the contract it gives "Transaction out of gas" error. 

### Steps to reproduce 
1) Make sure initially the balance of the contract is 0
2) Set purpose with some value of ETH
3) try to withdraw funds using the `withdraw` function and you will get the error.  

### Tried debugging and below is the explanation for issue
- Since we are using `usePrepareContractWrite` what it does is it sends `eth_estimateGas` call in background before hand when app renders. 
- "Transaction ran out of gas" this error only comes when initially the contract balance is 0 -> setPurpse -> withdraw, because `usePrepareContractWrite` has already made `eth_estimateGas` call for `withdraw` function. 
- Since initially the contract balance is 0 the `eth_estimateGas` returns lower value.
- Now, when you setPurpose and then try to `withdraw` because estimateGas was when contract balance was 0 and now since it has increased it fails
- If you refresh the page and then try to `withdraw` it works because now on app render since initial contract balance is > 0, the `eth_estimateGas` call returns higher value and `withdraw` transaction go through.


https://user-images.githubusercontent.com/80153681/211212928-420559ac-2bf8-4603-b4a6-bcbd9d0222b2.mov

### Approach taken 
Removed the `usePrepareContract` hook at [3740b03](https://github.com/scaffold-eth/se-2/pull/86/commits/3740b03b83c5b2e6e34f0f481154a2da0d298d5e) . 

### Reson for current approch 
- As mentioned in #59, even if we remove the `usePrepareContractWrite` hook the UX won't change because we are not using it properly and it's as good as not using it.
- found this to be a minimal and sweet solution :)

PS : If we go with the current approach, it will fix #59 too.

carletex:
@technophile-04 this is amazing, thank you!!

From #59 
> Ideally the send button should be disabled until we input correct values. But I also wonder if this is good DX, since while developing locally we want it to be raw and check directly by sending transaction if it fails.

I 100% agree with you here.

It's always scary to use something that reads "recklesslyUnprepared" haha, but I think there is a valid reason here. This solution is working (and fixing 2 issues at the same time!) so I'd say: let's go for it. 

There might a solution that recalculate the gas estimations, but let's not worry about that now. I don't think the "UX pitfalls" wagmi docs mention affect us in the /debug page. We'd have to think if we want to take the same approach when we do #88

carletex (approved):

Issue #88, opened 2023-01-06
Useful hooks/Utils for Contracts


carletex:
When creating your app in SE, it's super handy to have hooks that loads the deployed contracts (hardhat_contracts / external_contracts).

In SE-2 we are using wagmi, but I think we could benefit of having some layer (hooks / utils) on top of wagmi to load these deployed contracts.

In SE-1, we have:
- useContractLoader (load a contract read / write)
- useContractReader (load & watch a contract variable)

This is what wagmi provides:
![image](https://user-images.githubusercontent.com/2486142/210976236-27c497c4-3e01-44cf-b135-2db3b12fa41a.png)

In my mind, what I want to do as a dev is something like this (rough sketch for the sake of the example):

```ts
// useContractLoader (or whatever name) will load the contract (read=no signer connected, write=signer connected).
// No need to pass any provider/signer params, since we have the wagmi hooks holding the state.
const yourContract = useContractLoader("YourContract");
await tx(yourContract.setPurpose("Hola", {value: utils.parseEther("0.001")}))

// useContractReader (or whatever name) will watch (all the time in localhost, rpc polling time in other networks) a variable/s within the contract.
const { purpose, userPurposeCounter } = useContractReader("YourContract", {config});
// Config could be an array of the things we want to read (with its params). Wagmi has useContractReads for multiple reads at the same time. E.g config:
const config = {
 ["purpose", []]
 ["userPurposeCounter", ["0xaddresss"]]
}
```

We'll want to leverage as much as possible on the TS auto-completions for the contracts.

Naim-Bijapure:
@carletex  on the previous nextjs template have tried this . 
https://github.com/Naim-Bijapure/scaffold-eth-next/blob/main/packages/next-ts/hooks/useAppLoadContract.ts

created an useAppLoadContract hook which takes a contract name as an argument and returns its instance  and types for typescript autocomplete ( inspired from SE typescript). 

 there using useContract hook from wagmi to return contract instance. 
 also there need to configure contract abi files paths in a constant file first. 
 
what do you think? 
should I start digging into this?

technophile-04:
Really love the idea of "useful hooks/Utils for Contracts" much needed !!! 

I have been thinking and digging about it and here are some of my thoughts :

## Hooks idea : 
 ### 1) wrapper for `useContractRead` or `useContractWrite` : 
[...]

PR #93, opened 2023-01-16
utils and hooks for better DX

technophile-04:
### Description 
Added a basic `useScaffoldRead` hook (to start with draft). 

### wagmi 0.6.x type inference problem
It seems that wagmi 0.6.x does not support type inference. The type inference was introduced in [wagmi 0.7.x](https://wagmi.sh/react/migration-guide#07x-breaking-changes).

### Problems faced
- I tried upgrading the version of wagmi(not in this branch) and seems like for type inference we need to generate abi's in `.ts` file. Currently we are exporting all abi's in `generated` directory as `.json` file. 
	
  > Unfortunately [TypeScript doesn't support importing JSON as const]
(https://github.com/microsoft/TypeScript/issues/32063). If you have ABIs in `.json` files, you should convert the files to `.ts`, export the ABIs, and add const assertions to them. (We have future integrations planned for Foundry/Hardhat users to make this easier.) - wagmi docs

  also while exporting we need to export it `as const`. Here is some useful discussion around this topic https://github.com/wagmi-dev/wagmi/discussions/1084#discussioncomment-3891592

### Suggestions needed 
Now to get good auto-completion and take full advantage of the typescript as mentioned in #88, I think the following are steps that we need to take : 
a. Bump the version of wagmi
b. handle the exports of abi to .ts file
c. create utility hooks and utils

1.) **I feel 1 and 2 can be conquered parallelly but should we do it in this PR itself?**

2.) **would love to hear a view on hooks like `useYourContractRead` instead of `useScaffoldRead`?**
Reason: Felt a lot of resistance to get the correct types for abi but was not able to achieve it, because it is dynamic.Also maybe because of my limited knowledge of typescript. Would love it if someone could give it a try 🙌 .

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/index.ts
+++ b/packages/nextjs/hooks/scaffold-eth/index.ts
@@ -3,3 +3,4 @@ export * from "./useBurnerWallet";
 export * from "./useEthPrice";
 export * from "./useTransactor";
 export * from "./useOutsideClick";
+export * from "./useScaffoldRead";
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldRead.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldRead.ts
@@ -1,0 +1,25 @@
+import { useContractRead, useNetwork } from "wagmi";
+import { getDeployedContract } from "~~/components/scaffold-eth/Contract/utilsContract";
+
+const useScaffoldRead = (
+  contractName: string,
+  functionName: string,
+  readConfig?: Omit<Parameters<typeof useContractRead>[0], "functionName">,
+) => {
+  const { chain } = useNetwork();
+  const deployedContractData = getDeployedContract(chain?.id.toString(), contractName);
+  let contractAddress = "";
+  let contractABI = [];
+  if (deployedContractData) {
+    ({ address: contractAddress, abi: contractABI } = deployedContractData);
+  }
+
+  return useContractRead({
+    ...readConfig,
+    functionName,
+    addressOrName: contractAddress,
+    contractInterface: contractABI,
+  });
+};
+
+export default useScaffoldRead;
--- a/packages/nextjs/pages/demoExample.tsx
+++ b/packages/nextjs/pages/demoExample.tsx
@@ -1,0 +1,25 @@
+import type { NextPage } from "next";
+import Spinner from "~~/components/Spinner";
+import useScaffoldRead from "~~/hooks/scaffold-eth/useScaffoldRead";
+
+const DemoExample: NextPage = () => {
+  const { data, isLoading, refetch } = useScaffoldRead("YourContract", "purpose");
+  return (
+    <div className="flex flex-col items-center justify-center">
+      <h1 className="text-2xl my-5">Example UI</h1>
+      {isLoading ? <Spinner /> : <p>{data}</p>}
+      <button
+        disabled={isLoading}
+        className="btn btn-primary"
+        onClick={async () => {
+          console.log("Refetching...");
+          await refetch();
+        }}
+      >
+        Refetch
+      </button>
+    </div>
+  );
+};
+
+export default DemoExample;


carletex:
> 2.) would love to hear a view on hooks like useYourContractRead instead of useScaffoldRead?
Reason: Felt a lot of resistance to get the correct types for abi but was not able to achieve it, because it is dynamic.Also maybe because of my limited knowledge of typescript. Would love it if someone could give it a try raised_hands .

Maybe we could explore TS generics? https://www.typescriptlang.org/docs/handbook/2/generics.html

```typescript
useScaffoldRead<YourContract>(//...)
```

Also, I like `useScaffoldContractRead` than `useScaffoldRead`. What do you think?

carletex:
This is unlocked since #96!

technophile-04:
> I tried upgrading the version of wagmi(not in this branch) and seems like for type inference we need to generate abi's in .ts file. Currently we are exporting all abi's in generated directory as .json file.

Did some digging and came across some resources from which we can take inspiration : 
1. https://github.com/wagmi-dev/wagmi/issues/1090#issuecomment-1284716268
2. https://github.com/samkingco/sk-web3/blob/d7cd967383759300a9a96721aa49c592b8e4352f/packages/app/gen-abi.sh
3. https://github.com/anteqkois/tipdapp/blob/72bcd6d3e2bc55c41b13e919298fe5782c0bc495/contracts/scripts/deploy.ts#L45-L64

2 and 3 were posted here -> https://github.com/wagmi-dev/wagmi/issues/1106

carletex:
@technophile-04 For the type inference, maybe we can also check https://github.com/scaffold-eth/scaffold-eth-typescript (or even ask @ShravanSunder for guidance!)

carletex:
@technophile-04 Also, feel free to extract the typing inference to another issue.

We can start using the hooks in the example UI (without fancy type auto completion) and add it later. I think it's a must have (since we are using TS), but it's not needed to make the hooks work properly.

Just saying this because I don't want us to blocked too much on this!

technophile-04:
Yes @carletex, 100% agree with you will do it ASAP 🙌

technophile-04:
Yup @carletex makes sense and agree with will do it ASAP!

technophile-04:
Hii @carletex added v0.0.1 iteration of `useScaffoldContractRead` and `useScaffoldContractWrite` they are very loosely typed currently. 

Would love to know your thoughts on function signatures mainly on `useScaffoldContractWrite (contractName: string, functionName: string, args?: any[], value?: string)` kept `value?: string` because don't won't dev always struggling with making sure its `BigNumber | number` and handled it internally in `useScaffoldContractWrite` with `parseTxnValue` but my thinking maybe very one sided on both hooks and would love to know others thought/inputs also happy to revamp the whole thing again with others perspective

packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
```diff
@@ -0,0 +1,24 @@
+import { useContractRead, useNetwork } from "wagmi";
+import type { Abi } from "abitype";
+import { getDeployedContract } from "~~/components/scaffold-eth/Contract/utilsContract";
+
+/**
```
carletex, line 5: Can we add a Hook description here? Mentioning that it loads the deployed contracts automatically, etc.

packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
```diff
@@ -0,0 +1,25 @@
+import { useContractWrite, useNetwork, usePrepareContractWrite } from "wagmi";
+import { getDeployedContract } from "~~/components/scaffold-eth/Contract/utilsContract";
+import { parseTxnValue } from "~~/utils/scaffold-eth";
+
+/**
+ * @param contractName - deployed contract name
```
carletex, line 6: Hook description? :D

packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @param functionName - name of the function to be called
+ * @param args - arguments for the function
+ * @param value - value that will be send with transaction
```
carletex, line 9: I think this is "will be **sent**"

packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
```diff
@@ -0,0 +1,24 @@
...
+  readConfig?: Parameters<typeof useContractRead>[0],
+) => {
+  const { chain } = useNetwork();
+  const deployedContractData = getDeployedContract(chain?.id.toString(), contractName);
+
+  return useContractRead({
+    ...readConfig,
+    functionName,
+    address: deployedContractData?.address,
+    abi: deployedContractData?.abi as Abi,
```
carletex, line 22: Can we make some watch / polling magic here?

Some ideas:
- `watch: true` when on hardhat
- Polling time in live networks.

Some inspiration from `useEthPrice`:
```js
const pollingTime = process.env.NEXT_PUBLIC_RPC_POLLING_INTERVAL
  ? parseInt(process.env.NEXT_PUBLIC_RPC_POLLING_INTERVAL)
  : 30_000;
```

carletex (changes requested):
Hey @technophile-04 

Amazing stuff :)

Like you said, this is v0.0.1. I'd say, lets not try to make it perfect for now: as we use the hooks we'll learn what's needed and what's not.

I think `value` as a string & handling the conversion inside the hook is a good idea.

I left some minor comments. Let's merge after we fix those little things.

Thanks Shiv!!

packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
technophile-04: Just added `wath:true` at [0a446cf](https://github.com/scaffold-eth/se-2/pull/93/commits/0a446cff3746939be2b995a4e560bf22aa1a6649) because we have already configured  pollingInterval for `watch:true`  at https://github.com/scaffold-eth/se-2/pull/45 🙌

packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
technophile-04: Added a description at [99a09a0](https://github.com/scaffold-eth/se-2/pull/93/commits/99a09a078672505e9211a16162248254aeca5fcb) feel free to request a rewrite 🙌

packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
carletex: True that!!! I forgot. Thanks :)

carletex:
You didn't re-request the review 🔄... 

image

But I think this is ready! :D Really good job Shiv. Thanks

PR #94, opened 2023-01-16
Grab funds tooltip

rin-st:
daisyui doesn't support complex tooltips, like bottom-left, only centered bottom or centered left. So I placed the tooltip to the left (only when the balance is 0)
image

bottom variant is bad:
image

Possible solutions if you want the bottom tooltip:
- move the faucet button to the left
- use something like `react-popper` https://popper.js.org/react-popper/v2/

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useWalletBalance.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useWalletBalance.ts
@@ -1,0 +1,30 @@
+import { useCallback, useEffect, useState } from "react";
+import { useBalance } from "wagmi";
+import { useAppStore } from "~~/services/store/store";
+
+export function useWalletBalance(address?: string) {
+  const [isEthBalance, setIsEthBalance] = useState(true);
+  const [balance, setBalance] = useState<number | null>(null);
+  const price = useAppStore(state => state.ethPriceSlice.ethPrice);
+
+  const {
+    data: fetchedBalanceData,
+    isError,
+    isLoading,
+  } = useBalance({
+    addressOrName: address,
+    watch: true,
+  });
+
+  const onToggleBalance = useCallback(() => {
+    setIsEthBalance(!isEthBalance);
+  }, [isEthBalance]);
+
+  useEffect(() => {
+    if (fetchedBalanceData?.formatted) {
+      setBalance(Number(fetchedBalanceData.formatted));
+    }
+  }, [fetchedBalanceData]);
+
+  return { balance, price, isError, isLoading, onToggleBalance, isEthBalance };
+}
--- a/packages/nextjs/components/scaffold-eth/Balance.tsx
+++ b/packages/nextjs/components/scaffold-eth/Balance.tsx
@@ -1,6 +1,4 @@
-import { useEffect, useState } from "react";
-import { useBalance } from "wagmi";
-import { useAppStore } from "~~/services/store/store";
+import { useWalletBalance } from "~~/hooks/scaffold-eth/useWalletBalance";
 
 type TBalanceProps = {
   address: string;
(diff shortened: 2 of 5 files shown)


packages/nextjs/hooks/scaffold-eth/useWalletBalance.ts
```diff
@@ -0,0 +1,30 @@
+import { useCallback, useEffect, useState } from "react";
+import { useBalance } from "wagmi";
+import { useAppStore } from "~~/services/store/store";
+
+export function useWalletBalance(address?: string) {
```
carletex, line 5: I like the idea of extracting this functionality as a hook!

The name... what about `useAddressBalance`, or `useAccountBalance`? Seems more appropriate than Wallet.

carletex, line 5: Why the address arg is optional?

carletex (changes requested):
Hey @rin-st 

Thanks for this! Happy to see you are interested on collaborating on SE-2 ;)

I left a couple of comments in the code review.

Also, a couple things:

**1. Tooltip always open**

Maybe I wasn't clean enough on the issues (sorry!), but we wan't the tooltip to show all the time (not only on hover) when we are in localhost & balance is 0. You can use the `tooltip-open` class

**2. Tooltip position**
> daisyui doesn't support complex tooltips, like bottom-left, only centered bottom or centered left. So I placed the tooltip to the left (only when the balance is 0)

For this, you can use the daisy `tooltip-bottom` class and add the needed Tailwind classes to look like we want. I managed to do this adding some tw classes (`before:left-auto before:right-0 before:transform-none before:content-[attr(data-tip)]`)

rinat

You can also make the tooltip tail (the triangle indicator) a bit bigger, overriding the `--tooltip-tail` var. (6px looks good, as in the screenshot)

rin-st:
Hey @carletex, thanks for the comments!
1. fixed
2. fixed. Overriding `--tooltip-tail` and `--tooltip-offset` doesn't work for me, so I changed styles by tw classes. It seems there are only a few different vars that can be overridden in daisy-ui. https://daisyui.com/docs/themes/#-5:~:text=There%20are%20a-,few%20optional%20CSS%20variables%20that%20you%20can%20use%20in%20daisyUI%20themes,-to%20customize%20design
image

packages/nextjs/hooks/scaffold-eth/useWalletBalance.ts
rin-st: updated to `useAccountBalance`

rin-st: Because `wagmi`'s `useAccount` can return an undefined address and `useBalance` is ok with using an undefined address. So it works correct

carletex: I saw that it was working properly, but from the DX perspective is a bit confusing using the hook `useAccountBalance` that has a non-mandatory `address` arg.

But yeah, we need it to be this way to play nicely with wagmi.

Thanks!

carletex (changes requested):
Thanks for the updates! We are almost there:

> 2. fixed. Overriding --tooltip-tail and --tooltip-offset doesn't work for me, so I changed styles by tw classes. It seems there are only a few different vars that can be overridden in daisy-ui. See [#CSS variables in daisyUI themes](https://daisyui.com/docs/themes/#-5)

I don't like the magic number (42) in the TW classes. You can add custom styles to a theme: [Method 1](https://daisyui.com/docs/themes/#-6), [Method 2](https://daisyui.com/docs/themes/#-8).

I prefer Method 2 I think. For both themes.

```js
   //....
  "--rounded-btn": "9999rem",

 // Add this for both themes.
  ".tooltip": {
    "--tooltip-tail": "6px",
  },
```

You only need to add `--tooltip-tail`, since `--tooltip-offset` is calculated based on the the tail.

rin-st:
Updated, thanks. Strange, why I can't change the variable globally, but can change it in CSS class?
Checked daisy code - that variables are defined inside `.tooltip` class so it override global vars 🤷‍♂️ .
I don't like magic numbers too (and Tailwind JIT mode), but CSS is CSS 😄 , it's full of magic numbers. Of course we don't need it after `--tooltip-tail` works

packages/nextjs/hooks/scaffold-eth/useWalletBalance.ts
rin-st: it's much better than redundant `if-else`'s, problems with conditional hooks, and so on :)

carletex (approved):
Great work @rin-st 

Merging this!

PR #96, opened 2023-01-17
Bump up the version of wagmi to 0.10.x

technophile-04:
Updated all the files which were using wagmi, making them compatible with the new version of wagmi. Also updated Rainbowkit to latest version since its depended on wagmi and upgrading wagmi broke the old version of rainbowkit.


packages/nextjs/utils/scaffold-eth/parseAddressTo0x.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * @dev Converts an address string to the form of `0x${string}`
+ * @param {string} address - address to be parsed
+ * @return string of the form `0x${string}`
+ */
+export default function parseAddressTo0x(address: string | undefined): `0x${string}` | undefined {
```
technophile-04, line 6: Added this function because now wagmi only accept address of type `0x{string}` .

- Feel free to suggest a better name 
- also was a bit uncomfortable for creating a whole new file for one function, should we create one general file that contains utils that are **pure**,  or should we directly add them in `index.ts` of `utlis/scaffol-eth` ?

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -47,31 +47,33 @@ const getAllContractFunctions = (contract: Contract): FunctionFragment[] => {
...
+                  key={`DV_${fn.name}_${index}`}
+                  functionFragment={fn}
+                  contractAddress={contract.address}
+                  refreshDisplayVariables={refreshDisplayVariables}
+                />
+              );
+            }
+            return null;
+          })
+          .filter(n => n),
```
technophile-04, line 74: I think this is redundant, should we remove this? (I might have missed something)

was added in this PR: https://github.com/scaffold-eth/se-2/pull/50

packages/nextjs/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx
```diff
@@ -44,13 +44,13 @@ export const WriteOnlyFunctionForm = ({
...
-    addressOrName: contractAddress,
+    address: parseAddressTo0x(contractAddress),
     functionName: functionFragment.name,
-    contractInterface: [functionFragment],
+    abi: [functionFragment],
     args: keys.map(key => form[key]),
     mode: "recklesslyUnprepared",
     overrides: {
-      value: txValue,
+      value: parseTxnValue(txValue),
```
technophile-04, line 53: ### Please take a look at `parseTxnValue` function. 

### Reasong for adding : 
In new version of wagmi `value` is of type - `BigNumber | number | undefined`

packages/nextjs/utils/scaffold-eth/parseTxnValue.ts
```diff
@@ -0,0 +1,22 @@
+import { BigNumber } from "ethers";
+
+/**
+ * @dev utility to parse string to wagmi's PayableOverrides.value.Returns BigNumber if its convertible or returns -1 if there is error or invalid string
+ * @param  str - any string
+ * @return {BigNumber | number | undefined}
+ */
+export default function parseTxnValue(str: string): BigNumber | number | undefined {
```
technophile-04, line 8: ### Working : 
- It returns `-1` denoting that the value passed as a string cannot be parsed by BigNumber. 
- Due to this  whenever a user enters eg : `fdsffdfsf` he gets an error displayed as 
> invalid hex string (argument="value", value="-0x01", code=INVALID_ARGUMENT, version=bytes/5.7.0)
In the above error, we can see its telling `value=-0x01` which is kind of misguidance but didn't find a better a way to do it 😅

Also feel free to suggest a better name

technophile-04:
Bug which I noticed and there is also an open issue https://github.com/rainbow-me/rainbowkit/issues/837 .

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
rin-st: hey @technophile-04! `map` can return some `null`s and `filter` removes them. If you don't need `null`s, you need that filter

packages/nextjs/utils/scaffold-eth/parseAddressTo0x.ts
rin-st: but how address can be without 0x?) for me it's ok to create a single file for helper functions, it's easier to find it.

packages/nextjs/utils/scaffold-eth/parseAddressTo0x.ts
```diff
@@ -0,0 +1,13 @@
...
+  if (!address) return undefined;
+  if (address.startsWith("0x")) {
+    return `0x${address.substring(2)}`;
```
rin-st, line 9: return address

packages/nextjs/utils/scaffold-eth/parseAddressTo0x.ts
rin-st: also, you can create type something like `0xAddress = 0x{string}` and use it.

packages/nextjs/components/scaffold-eth/Contract/utilsComponents.tsx
```diff
@@ -18,18 +19,29 @@ const TxValueInput = ({ setTxValue, txValue }: { setTxValue: Dispatch<SetStateAc
         <button
           className="cursor-pointer text-xl font-semibold pt-1 text-accent"
           onClick={async () => {
-            if (!txValue) return;
-            const floatValue = parseFloat(txValue);
-            if (floatValue) setTxValue("" + floatValue * 10 ** 18);
+            try {
+              const floatValue = parseFloat(txValue);
+              if (floatValue) {
+                setTxValue("" + floatValue * 10 ** 18);
```
rin-st, line 25: String(floatValue * 10 ** 18). Explicit is better than implicit

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -47,31 +47,33 @@ const getAllContractFunctions = (contract: Contract): FunctionFragment[] => {
...
-            />
-          );
-        }
-        return null;
-      })
-      .filter(n => n),
-  };
+  return contract
+    ? {
+        loaded: true,
```
rin-st, line 56: why do we need that flag?

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -47,31 +47,33 @@ const getAllContractFunctions = (contract: Contract): FunctionFragment[] => {
...
+        methods: contractMethodsAndVariables
+          .map((fn, index) => {
+            const isQueryableWithNoParams =
+              (fn.stateMutability === "view" || fn.stateMutability === "pure") && fn.inputs.length === 0;
+            if (isQueryableWithNoParams) {
+              return (
+                // DV -> DisplayVariables
```
rin-st, line 63: remove comment?

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -47,31 +47,33 @@ const getAllContractFunctions = (contract: Contract): FunctionFragment[] => {
...
+                <DisplayVariable
+                  key={`DV_${fn.name}_${index}`}
```
rin-st, line 65: maybe rename it to DisplayVariables_... and remove comment?

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -82,29 +84,31 @@ const getContractVariablesAndNoParamsReadMethods = (
...
+        loaded: true,
+        methods: contractMethodsAndVariables
+          .map((fn, index) => {
+            const isQueryableWithParams =
+              (fn.stateMutability === "view" || fn.stateMutability === "pure") && fn.inputs.length > 0;
+            if (isQueryableWithParams) {
+              return (
+                // FFR -> FunctionFormRead
+                <ReadOnlyFunctionForm
+                  key={`FFR_${fn.name}_${index}`}
```
rin-st, line 101: and here

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -115,30 +119,32 @@ const getContractReadOnlyMethodsWithParams = (
...
+    ? {
+        loaded: true,
+        methods: contractMethodsAndVariables
+          .map((fn, index) => {
+            const isWriteableFunction = fn.stateMutability !== "view" && fn.stateMutability !== "pure";
+            if (isWriteableFunction) {
+              // FFW -> FunctionFormWrite
+              return (
+                <WriteOnlyFunctionForm
+                  key={`FFW_${fn.name}_${index}`}
```
rin-st, line 136: and here

packages/nextjs/components/scaffold-eth/Contract/utilsDisplay.tsx
```diff
@@ -4,7 +4,7 @@ import { BigNumber, ethers } from "ethers";
 import React, { Dispatch, ReactElement, SetStateAction } from "react";
 import { Address } from "~~/components/scaffold-eth";
 
-type displayContentType = string | number | BigNumber | Record<string, any> | TransactionResponse | undefined;
+type displayContentType = string | number | BigNumber | Record<string, any> | TransactionResponse | undefined | unknown;
```
rin-st, line 7: DisplayContentType

packages/nextjs/services/web3/wagmi-burner/BurnerConnector.ts
```diff
@@ -78,15 +80,15 @@ export class BurnerConnector extends Connector<StaticJsonRpcProvider, BurnerConn
     return Promise.resolve();
   }
 
-  async getAccount(): Promise<string> {
+  async getAccount(): Promise<`0x${string}`> {
```
rin-st, line 83: it seems it can be undefined too

packages/nextjs/utils/scaffold-eth/parseTxnValue.ts
rin-st: probably you can use https://docs.ethers.org/v5/api/utils/bignumber/#BigNumberish as a param instead of a string

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
technophile-04: Ohh yes! missed it :), Thanks🙌

packages/nextjs/utils/scaffold-eth/parseTxnValue.ts
technophile-04: Actually, I wanted to pass values that come from input field that's why kept it string, check https://github.com/scaffold-eth/se-2/pull/96#discussion_r1073185982, but happy to reconsider it :)

packages/nextjs/utils/scaffold-eth/parseAddressTo0x.ts
technophile-04: I tried doing but since the return type is `0x{string}` we can't return just address because the param `address` is of type `string`

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
technophile-04: check https://github.com/scaffold-eth/se-2/pull/50

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
technophile-04: Yup, make sense but I feel let's do it in a different PR 🙌

packages/nextjs/services/web3/wagmi-burner/BurnerConnector.ts
technophile-04: Thanks, handled this at [54999ee](https://github.com/scaffold-eth/se-2/pull/96/commits/54999ee7e063e97f1cb5be66c58f6c0d6d595a44)

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
technophile-04: check https://github.com/scaffold-eth/se-2/pull/96#discussion_r1080773581

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
technophile-04: check https://github.com/scaffold-eth/se-2/pull/96#discussion_r1080773581

packages/nextjs/components/scaffold-eth/Contract/utilsDisplay.tsx
technophile-04: check https://github.com/scaffold-eth/se-2/pull/96#discussion_r1080773581

packages/nextjs/utils/scaffold-eth/parseAddressTo0x.ts
technophile-04: would love to hear file name suggestion 😅, Thanks 🙌

rin-st: `convertTo0xAddress`?

packages/nextjs/utils/scaffold-eth/parseAddressTo0x.ts
rin-st: return address as 0x${string}  :)

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
rin-st: strange variable, it's always true. Will check it later

packages/nextjs/utils/scaffold-eth/parseTxnValue.ts
rin-st: maybe return undefined instead of -1 and check it before sending tx? Have you tried to send tx with -1 overrides value? Ideally, we need to check the value and if it's wrong show an error in the form and disable the button. Probably we need to create issue for that.

packages/nextjs/components/scaffold-eth/Contract/utilsDisplay.tsx
rin-st: It seems it's already changed

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
rin-st: fixed here https://github.com/scaffold-eth/se-2/pull/100

packages/nextjs/components/scaffold-eth/Contract/utilsComponents.tsx
rin-st: fyi https://github.com/scaffold-eth/se-2/pull/101

packages/nextjs/utils/scaffold-eth/parseTxnValue.ts
rin-st: https://github.com/scaffold-eth/se-2/pull/101 changed tx value input, so txValue is always correct BigNumber or undefined, so it seems you don't need this function anymore

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
rin-st: https://github.com/scaffold-eth/se-2/pull/100/commits/964bc20b8ca09fcae0b88743942acbf89442c8b4

carletex:
I think we should focus now on this PR and not merging anything else until this is done.

Great job @technophile-04 . And thanks for the review @rin-st 

Let's first **fix the conflicts** (I merged some stuff from Rinat, maybe I should've waited - my bad :/)

The only think that I don't like is having to use the `parseAddressTo0x` when using wagmi hooks. Don't you all think is bad DX? Not sure if there is another option: Maybe `import { Address } from "abitype";` as they do in wagmi... and use that type instead of `string`. Not sure of the implications yet, but would love to hear your thoughts! 

Thanks!!

rin-st:
I completely agree with @carletex, great job @technophile-04! Sorry for the many comments, I'm trying to do my best and hope you're not angry at me :). I resolved conflicts with `main` branch since I made most of it.

Please check if we need `parseTxnValue`. 

Regarding `parseAddressTo0x`, how address can be without `0x` at the start, why do we need that function?

Thanks!

carletex:
Hey, just found this: https://github.com/wagmi-dev/wagmi/issues/1330#issuecomment-1320317099.

OP has the same feeling as me (having to use "as" or "parseTo0x" is a pita). What if we try the solution they propose? cc @technophile-04 @rin-st

technophile-04:
> Hey, just found this: https://github.com/wagmi-dev/wagmi/issues/1330#issuecomment-1320317099.

@carletex  love this, so sorry I missed it. I would vote on using it because I don't want dev's always stuck at such strict typing and also ether.js types have an address as type `string`

rin-st:
In other circumstances, I'd prefer to use `Address`, because it would be easy to find for example that an empty string is in the address variable or it's broken for some reason. As I understand, to use it we need just one `as` at `getDeployedContract` return, and rename `Address` component. 

But yes, everyone who'll work with addresses will need to use that type, so it's worse for dx. So I agree, it's a good solution.
@carletex

rin-st:
@technophile-04 regarding decimals, it's wei, so it should be integers :). I thought about adding wei/gwei/eth radio buttons near that input, so we'll need decimals here. Not sure it will be easy

carletex:
> The value input field is not accepting decimal point numbers.I think it was added in https://github.com/scaffold-eth/se-2/pull/101 @rin-st could you please take a look also happy to help and discuss some approaches raised_hands

> @technophile-04 regarding decimals, it's wei, so it should be integers :). I thought about adding wei/gwei/eth radio buttons near that input, so we'll need decimals here. Not sure it will be easy

I was just tinkering with Austin and found this issue: Yes, it's wei so no decimals... **but** the way people use that input modifier is by writing something like "0.25" and clicking the button to convert it to wei. Also we feel that people don't use the "#" operator so we might remove it for now. I'm gonna to create an issue for it.

rin-st:
Makes sense. Probably it will be easier to have that radio buttons wei/gwei/eth but we should discuss it in a new issue

regarding `Address` let's use https://github.com/wagmi-dev/wagmi/issues/1330#issuecomment-1320317099 for now. If we'll have problems with it we change it

carletex:
Just created https://github.com/scaffold-eth/se-2/issues/103 with the simple posible solution. We can iterate later.

Regarding the Address type: let's override it to string. We should not forget that SE-2 will be used by a lot newcomers, so most of the time easier is better!

technophile-04:
Hey guys re-seted the history and removed the `prarseAddressTo0x` util 

**Approach taken** : 
- reseted the history to [088886a](https://github.com/scaffold-eth/se-2/pull/96/commits/088886a2716353de845e2d722914fe4f0d62cb88)
- added `abitype` to devDependencies of !! _**monorepo**_  !! 
- remove the use of `parseAddressTo0x` util and remove the util too 

**Reason for the approach and problems** : 
- Don't know why for me adding `abitype` to `devDependency` of `nextjs` directory is not working that's why added it directly to _**monorepo**_ (Even I feel we shouldn't add such thing to monorepos's dependency, would love if someone could give this a try might be some problem with my vsCode)  
- used `disable-next-line no-unused-vars` for `global.d.ts`, I think there is some problem with eslint :  checkout https://github.com/typescript-eslint/typescript-eslint/issues/2287
- **IMP: if you pull the changes and go to `BurnerConnector.ts` you will find some typescript errors because in some places AdressType still refers to `0x{string}` even after adding the declaration file.**

Just didn't want to block this PR at the moment and would love other's approach since current one is not at all great, I am exploring on this and will update if found something

packages/nextjs/utils/scaffold-eth/parseTxnValue.ts
carletex: Since merging  https://github.com/scaffold-eth/se-2/pull/104 I think we need to use this util again.

I added it here: https://github.com/scaffold-eth/se-2/pull/96/commits/2b3de3b15e37d5af6f8341b25ed0c42077121e36

I agree that we can handle this better, but let's do it in another issue.

technophile-04:
> For me (I use Intellij) it works without adding abitype as a devDependency anywhere (I deleted all the node_modules folders && yarn install). The global.d.ts does the trick for me. 

@carletex  yes ! this did work, Tysm 🙌

- Removed the `abitype` from dependency  @[c07983c](https://github.com/scaffold-eth/se-2/pull/96/commits/c07983c7455833d4a28725d070d04ceb750ea423)
- added [@ts-expect-error](https://timmousk.com/blog/typescript-ignore-next-line/#ts-ignore-vs-ts-expect-error) to ignore the errors for now @[c67afb6](https://github.com/scaffold-eth/se-2/pull/96/commits/c67afb687e4c8db2e4375389f2b0f6d4b32f2340), feel free to change the comment message
- lastly just wanted to mention this https://github.com/rainbow-me/rainbowkit/issues/837 so that we don't get confused in future

technophile-04:
> used disable-next-line no-unused-vars for global.d.ts, I think there is some problem with eslint

Fixed @[0a0e780](https://github.com/scaffold-eth/se-2/pull/96/commits/0a0e780d2d17b565094ef5e5191fc0f21ee5a8b7), It was suggested here https://github.com/technophile-04/se-2/pull/1/files, also added `types` dir where we can put all types, please let me know what do you think ? also happy to revamp it 🙌  

> IMP: if you pull the changes and go to BurnerConnector.ts you will find some typescript errors because in some places AdressType still refers to 0x{string} even after adding the declaration file

There was some problem with wagmi versioning check out https://github.com/wagmi-dev/wagmi/issues/1712 which was fixed at the new version in wagmi (0.10.15) also now no need for @ts-expect-error :) 

Also changed typescript version to `4.7.4` from `4.7.2` because its mentioned in [wagmi](https://wagmi.sh/react/typescript) docs also was suggested here->  https://github.com/wagmi-dev/wagmi/issues/1712#issuecomment-1402385271

I think we are ready to merge, @carletex  what do you think? only one bug is remaining https://github.com/rainbow-me/rainbowkit/issues/837

carletex:
Amazing! I think we finally have it!!

> I think we are ready to merge, @carletex what do you think? only one bug is remaining https://github.com/rainbow-me/rainbowkit/issues/837

Yeah, this bug is annoying and has been open for 3 months. Can you open an issue in SE-2? So it's documented in our repo. Maybe we can wait a bit and if not trying to roll our custom fix.

PR #97, opened 2023-01-18
fix: remove faucet button jump

rin-st:
before:
https://user-images.githubusercontent.com/25638585/213171302-fe3397e2-8398-4bc3-8667-99a982fac9a4.mov

after:
https://user-images.githubusercontent.com/25638585/213171037-1650423d-e512-4be3-872b-1d76e8fd28eb.mov


carletex:
Hey @rin-st Thanks!

Yep, that jump was annoying :D

I wanted to keep the faucet button rounded so I pushed a commit. Not a fan of using !, but it was a decent way to give the spinner the same width / height and no-margin.

Issue #98, opened 2023-01-18
Fix dark theme


rin-st:
<img width="1328" alt="Screenshot 2023-01-18 at 16 43 18" src="https://user-images.githubusercontent.com/25638585/213183763-645652ab-6f7d-4bcc-8100-3dc931c3f912.png">

PR #99, opened 2023-01-18
update dark theme

rin-st:
changed the description text color to default, because we don't have a gray color variable for themes. For example, we can use `neutral` color, but not sure. Also, I don't like the way to calculate the color for networks, but for now it at least looks better.

before:
Screenshot 2023-01-18 at 16 43 18
after:
image


carletex (approved):
Great stuff @rin-st !

Pushed a tiny thing. Merging!

Fixes #98

PR #100, opened 2023-01-19
Small adjustments

rin-st:


carletex (approved):
Hey @rin-st 

We have to be careful with this kind of PRs, since it touches multiple unrelated files and it's hard to test/review. IMO it's better to fix these little things when working on a related PR that touches a given file / functionality.

In any case, all these tweaks you made are great and make the code better. Thanks!

PR #101, opened 2023-01-19
fix: tx value input

rin-st:
updated tx value input, so now it works correctly with big numbers. Shows an "Invalid number" error message if the input isn't correct.

before:
https://user-images.githubusercontent.com/25638585/213478844-5b52f1d8-2baf-429f-8f3b-95c86324c833.mov

after:
https://user-images.githubusercontent.com/25638585/213479173-dad94ba4-4555-485f-af20-433de5cd2f7c.mov


carletex (approved):
Nice!! Thanks @rin-st

PR #104, opened 2023-01-21
Fix/tx value operators

rin-st:
- removed # button
- added the possibility to write decimal numbers
- added correct multiplying by 10^18
- tooltip

demo with **10^3** multiplier to see how it works:

https://user-images.githubusercontent.com/25638585/213860428-5d8c7d4f-9274-4d5e-9548-8b226fd5575d.mov

probably should be merged after https://github.com/scaffold-eth/se-2/pull/96

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/utilsComponents.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/utilsComponents.tsx
@@ -1,42 +1,75 @@
-import { BigNumber } from "ethers";
 import { Dispatch, SetStateAction } from "react";
 import { toast } from "~~/utils/scaffold-eth";
+
+const NUMBER_REGEX = /^\d+\.?\d*$/;
+
+/**
+ * Removes redundant leading and trailing zeros from stringified number
+ * @param num stringified number
+ */
+const removeRedundantZeroes = (num: string) => {
+  return (
+    num
+      // leading
+      .replace(/^0+(0\.|[1-9]+)/, "$1")
+      // trailing
+      .replace(/(\.[0-9]*[1-9])0+$|\.0*$/, "$1")
+  );
+};
+
+/**
+ * Multiplies stringified number to positive power of 10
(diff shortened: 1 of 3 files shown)


packages/nextjs/components/scaffold-eth/Contract/utilsComponents.tsx
```diff
@@ -1,60 +1,81 @@
...
+  if (numberParts.length === 1) {
+    result = `${num}${"0".repeat(powerOf10)}`;
+  } else if (numberParts[1] && powerOf10 < numberParts[1].length) {
+    result = `${numberParts[0]}${numberParts[1].slice(0, powerOf10)}.${numberParts[1].slice(powerOf10)}`;
+  } else {
+    result = `${numberParts[0]}${numberParts[1]}${"0".repeat(powerOf10 - numberParts[1].length)}`;
+  }
+
+  return removeRedundantZeroes(result);
+};
```
carletex, line 41: Hey @rin-st ,

I'm trying to understand the benefits of adding this complexity 

VS

Converting the string to float + doing 10 ** 18 + back to string as it was before. Specially now that we won't have this issue anymore: https://github.com/scaffold-eth/se-2/pull/101 

What do you think?

rin-st: https://user-images.githubusercontent.com/25638585/214067007-0dd09909-d892-41f9-ae13-7f903b547d7e.mov

1) JS doesn't always correctly calculates values, especially big numbers
2) you can't send tx when tx value is something like 1e+36

carletex: I appreciate your thoroughness ;) Even if not super important in this page.

Added some comments to `multiplyStringifiedNumberToPowerOf10` (it's not very friendly to read :D) and changed the error handling a bit (too many toast XD).

Good job.

PR #107, opened 2023-01-23
Check TS typings on lint-staged and GH actions

carletex:
Right now the only way I see type errors is with my IDE.

I added a `yarn next:check-types` to be able to check those.

- Added it to lint-staged: I didn't find a way to check only specific files with TSC (as we do with the linting).
- Added it to the GH Action also.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -33,5 +33,8 @@ jobs:
       - name: Run nextjs lint
         run: yarn next:lint --max-warnings=0
 
+      - name: Check typings on nextjs
+        run: yarn next:check-types
+
       - name: Run hardhat lint
         run: yarn hardhat:lint --max-warnings=0
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -5,12 +5,17 @@ const buildNextEslintCommand = (filenames) =>
     .map((f) => path.relative(path.join("packages", "nextjs"), f))
     .join(" --file ")}`;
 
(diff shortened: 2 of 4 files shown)


carletex:
It seems that this is not ignoring `node_modules`

technophile-04:
> Added it to lint-staged: I didn't find a way to check only specific files with TSC (as we do with the linting).

First I thought it would be nice if we type-check only staged changes and was finding some solutions to overcome it until I came across this [comment](https://dev.to/fatedx/comment/22l09). And now it makes sense to let tsc go through all the files so I think the current approach is nice also if its get too annoying for the developer he can use `@ts-ignore` since `ban-ts-comment` is off now. 

But would love to know your thoughts on this, if we plan to go other way(type-checking only staged files) then we can use below resources/packages to solve this problem : 
1. [tsc-files](https://github.com/gustavopch/tsc-files) package, article explaining use of it [here](https://dev.to/samueldjones/run-a-typescript-type-check-in-your-pre-commit-hook-using-lint-staged-husky-30id)
2. Stackoverflow discussion [solution](https://stackoverflow.com/a/60950355)

technophile-04 (approved):

carletex:
> I came across this [comment](https://dev.to/fatedx/comment/22l09).

I like that and it makes a lot of sense.

Let's keep it for now, and if at some point the checks on lint-staged take a lot of time... maybe we can run that command only on the GH action.

Thanks Shiv!

PR #109, opened 2023-01-24
Fix bug for collapsible of writeFunction

### LINKED ISSUE #108: Small bug on`/debug` page
### Description : 
It happens only happens with write functions. 

https://user-images.githubusercontent.com/80153681/214323567-0b7600dc-eccd-4541-a52b-bcc022519e9d.mov

technophile-04:
### Demo 

https://user-images.githubusercontent.com/80153681/214339549-c313aaa1-8baa-4119-8834-399f5a58ba6f.mov

Fixes #108

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
@@ -90,7 +90,7 @@ const ContractUI = ({ contractName }: TContractUIProps) => {
                 <p className="my-0 text-sm">Write</p>
               </div>
             </div>
-            <div className="collapse-content py-3 px-4 min-h-12 transition-all duration-200 overflow-visible">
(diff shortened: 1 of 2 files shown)


carletex:
Thanks @technophile-04 !

1. Not sure if we'll keep the collapsible, I don't see much value to it. Also it's a bit weird that when you fold it you see the name of the first method. But for now, let's merge this fix!

2. Pushed https://github.com/scaffold-eth/se-2/pull/109/commits/b9af9b82b89c663a74ad4df6c17224817bf4c095 to remove the magic number (-50px): It's just a nitpick haha But it's a magic number since it depends on the size of the tooltip content. I positioned using `right-[-10px]`, which is not a magic number! since 10px is relative to the container, and doesn't care about the size of the tooltip content) + I add some more ugly classes (love you Tailwind!).

PR #112, opened 2023-01-26
Handle error when hardhat_contracts.json doesnt exist

carletex:
In `utilsContract.tsx` we had:

```js
// ToDo. Handle when this doesn't exist?
import ContractData from "~~/generated/hardhat_contracts.json";
```

If people follow the command order specified in the README (chain, deploy, start) this error will never happen. But I feel that we should handle this error, in case the `hardhat_contract.json` is not there.

I tried a few approaches, and this one feels like the most simple one. 
It gives a warning in the console, but not in the UI. UI only gives an error if the file exists and you delete (which seems an edge use case :D).

Can you double check @technophile-04 ?
Happy to consider another approach.

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
@@ -1,6 +1,4 @@
 import { FunctionFragment } from "ethers/lib/utils";
-// ToDo. Handle when this doesn't exist?
-import ContractData from "~~/generated/hardhat_contracts.json";
 import { Contract, utils } from "ethers";
 import DisplayVariable from "~~/components/scaffold-eth/Contract/DisplayVariables";
 import { ReadOnlyFunctionForm } from "./ReadOnlyFunctionForm";
@@ -22,6 +20,13 @@ const getDeployedContract = (
   contractName: string | undefined | null,
 ): GeneratedContractType | undefined => {
   if (!chainId || !contractName) {
+    return;
+  }
+
+  let ContractData;
+  try {
+    ContractData = require("~~/generated/hardhat_contracts.json");
+  } catch (e) {
     return;
   }
 


technophile-04:
Nice, this seems like a fair solution

In the future, we can also explore https://beta.nextjs.org/docs/routing/error-handling for more descriptive errors and guide developers correctly 🙌

technophile-04 (approved):

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -25,6 +23,13 @@ const getDeployedContract = (
...
   }
 
+  let ContractData;
+  try {
+    ContractData = require("~~/generated/hardhat_contracts.json");
+  } catch (e) {
+    return;
+  }
+
   const contractsAtChain = ContractData[chainId as keyof typeof ContractData];
```
rin-st, line 33: it will break here if ContractData isn't set

technophile-04: But I think if the interpreter reaches this line then its sure that ContractData will be set, even Typescript does not complain but I might have missed something 🙌

rin-st: no, ts thinks it has `any` type

technophile-04: Yeah, but I don't think it will be `undefined` but I am sure I am missing something

technophile-04:
Just saw this on se-typescript while doing `yarn start` and found it helpful we can explore something like this in future : 
Screenshot 2023-01-27 at 11 17 29 PM

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
carletex: So, what's the verdict here? :D

If the file doesn't exist an error will be thrown, catched and the function will return, so line 33 won't be executed. Right?

The idea is to avoid showing an error when the builder forgot to run "yarn deploy".

rin-st: My bad, I didn't see `return` in the catch block, and thought a function will continue it's execution. Sorry 🤷‍♂️

rin-st (approved):

carletex:
Thanks all!

> Just saw this on se-typescript while doing yarn start and found it helpful we can explore something like this in future
> We can probably add that message to the front end after or instead of the "No Contract found!" message

Yep, like both.

Also the "No contract found" it's so ugly positioned right now :rofl:

PR #115, opened 2023-01-27
Feat/example UI

rin-st:
https://user-images.githubusercontent.com/25638585/215157520-9f352599-9a92-4a69-bf4a-09d9775fbd30.mov


rin-st:
Todos:
1) download fonts
2) black switch button isn't centered on figma

packages/nextjs/tailwind.config.js
```diff
@@ -3,6 +3,18 @@ module.exports = {
...
+    extend: {
+      backgroundImage: {
+        gradient: "url('/assets/gradient-bg.png')",
+      },
+      transitionDuration: {
+        2000: "2000ms",
+        3000: "3000ms",
+      },
+    },
+  },
```
carletex, line 16: Since we are only using this in ExampleUI, I feel this pollutes the config file a bit.

Let's use this values directly on TW. `bg-[url('')]` && `duration-[2000ms]`

packages/nextjs/components/ExampleUi/ContractInteraction/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const ContractInteraction = () => {
+  return (
+    <div className="flex bg-base-300 relative">
+      <DiamondIcon className="absolute top-24" />
+      <CopyIcon className="absolute bottom-0 left-36" />
+      <HareIcon className="absolute right-0 bottom-24" />
+      <div className="flex flex-col w-full mx-20">
+        <UiDescription />
+        <PurposeSetter />
```
carletex, line 15: Personally, I think I would have inline both of these here. I guess I tend to split components a bit less. What is your mental model for it?

Don't change it! Just wanted to say it here so we can discuss and align better.

packages/nextjs/components/ExampleUi/ContractInteraction/PurposeSetter.tsx
```diff
@@ -0,0 +1,29 @@
...
+      <span className="text-6xl text-black">Set a Purpose_</span>
+
+      <div className="mt-8 flex gap-5 items-center">
+        <input
+          type="text"
+          placeholder="Write your purpose here"
+          className="input w-full px-5 bg-gradient bg-[length:100%_100%] border border-black text-2xl placeholder-white"
+        />
+        <div className="flex rounded-full border border-black p-1 flex-shrink-0">
+          <div className="flex rounded-full border-2 border-black p-1">
```
carletex, line 15: `border-primary` instead of `border-black`?

packages/nextjs/components/ExampleUi/ContractInteraction/CopyIcon.tsx
```diff
@@ -0,0 +1,28 @@
+type CopyIconProps = {
+  className: string;
+};
```
carletex, line 3: nitpick: Maybe we inline this directly? `({ className }: { className: string }) `.

having the type seems a bit verbose

packages/nextjs/components/ExampleUi/ContractInteraction/CopyIcon.tsx
```diff
@@ -0,0 +1,28 @@
+type CopyIconProps = {
```
carletex, line 1: Maybe we can move this icons into a `icons` or `assets` folder? Just for the sake of having a cleaner folder content.

Not sure if right below `ExampleUi` or `ExampleUi/ContractInteraction`.

carletex (changes requested):
This is great stuff @rin-st!!!

> download fonts

Andrea is finding a hosted font (google fonts probably) so we don't need to host it in the project.

> black switch button isn't centered on figma

What do you mean with this? 

---

I left some minor comments in the code. 

Also there is some confusion on how the screen "rolling text" works. I think the video is a bit confusing. The idea is that the text moves in an infinite loop (until you change the direction VS. changing direction when it reach the edges). I think it's called "marquee". There was a [HTML marquee](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/marquee) element, but it's deprecated.

I think we could implement it ourselves. Some vanillaJS inspiration.
https://codepen.io/pprakash/pen/oNxNeQE

But also, not sure how this element should work now:
image.

Maybe it can refill and unfill the same way as it's now... and when you switch, it does the same but in the opposite direction (starting from the right side).

Open to suggestions... and also I'm asking Andrea.

Thanks!

rin-st:
Thanks @carletex!
> Andrea is finding a hosted font (google fonts probably) so we don't need to host it in the project.

Got it, but keep in mind that downloaded fonts are faster and independent from host servers. We can change it in future if needed

> black switch button isn't centered on figma
> What do you mean with this?

The button png isn't symmetric, so it's not positioned properly. Asked Andrea in pm and she changes it.

That codepen example with `setInterval(...,0)` is too heavy, need to think about how to implement that. Ideally, it should be no JS but I can't see a solution for now

Also, she noticed some bugs in styling, I'll change them in the next commits

packages/nextjs/components/ExampleUi/ContractInteraction/index.tsx
rin-st: It's easier to locate code/errors in smaller components. But to be honest, I thought it would be more code in that components :)

packages/nextjs/components/ExampleUi/ContractInteraction/PurposeSetter.tsx
rin-st: it's black on designs :)

carletex: `border-primary` will be black, right? As defined on the `exampleUi` theme in `tailwind.config.js`.

It's not a big deal, but I try to use theme colors instead of hardcoded ones.

rin-st:
I didn't see your message but found that `react-fast-marquee` too. I used it, but my head almost exploded even with it 😄 . I hope it's better now

https://user-images.githubusercontent.com/25638585/215553117-2d39ec96-0a49-4f80-b1b9-b3601d2578b5.mov

rin-st:
added new font, changed slider at the bottom (only from left to right now). Fought with react-fast-marquee important styles and google fonts. Dunno why gfonts doesn't work for me, downloaded it and now it works.
https://user-images.githubusercontent.com/25638585/215591017-7ae1da2a-fc9e-49fb-acac-626a2ed11e65.mov

carletex:
Great stuff @rin-st 

Merging this!

PR #117, opened 2023-01-27
Using Graph Client to query the Lens API

schmidsi:
This PR is not intended to be merged. I just played around with the possibility of including Graph Client into this repo that queries the [Lens API](https://api.lens.dev) (not the Lens Subgraph).

It has some rough edges, but it works. Something like this could give a nice separation of concerns and the ability for plugins.


technophile-04:
Tysm Simon !! This will be a great additional plugin for CLI (#251)🙌 

Its already referenced in #315 so we can come back here later too, closing this for now but really appreciate and thanks for this !!

PR #124, opened 2023-02-01
Real data on ExampleUI

carletex:
Since we merged https://github.com/scaffold-eth/se-2/pull/93 and https://github.com/scaffold-eth/se-2/pull/115, I wanted to put them together, getting real data on example UI + setting the purpose from there.

I noticed a few things with our hooks / utils that I wrote code comments (I'll add some review comments now).

Like we said before, the best way to fine tune our hooks / utils is using them!


packages/nextjs/components/ExampleUi/ContractData/index.tsx
```diff
@@ -1,17 +1,25 @@
...
 
   const containerRef = useRef<HTMLDivElement>(null);
   const purposeRef = useRef<HTMLDivElement>(null);
 
+  // I guess this is related to: https://github.com/scaffold-eth/se-2/issues/116
+  // Ideally I'd like data to be the right type based on the function name / variable name we are calling.
+  // @ts-expect-error
+  const { data: totalCounter }: { data: BigNumber } = useScaffoldContractRead("YourContract", "totalCounter");
+  // @ts-expect-error
+  const { data: currentPurpose }: { data: string } = useScaffoldContractRead("YourContract", "purpose");
```
carletex, line 21: I guess this is related to: https://github.com/scaffold-eth/se-2/issues/116

Ideally I'd like the data to be the right type based on the function name / variable name we are calling.

packages/nextjs/components/ExampleUi/ContractInteraction/PurposeSetter.tsx
```diff
@@ -1,6 +1,23 @@
...
 import { ArrowSmallRightIcon } from "@heroicons/react/24/outline";
+import { useScaffoldContractWrite, useTransactor } from "~~/hooks/scaffold-eth";
+import { getParsedEthersError } from "~~/components/scaffold-eth/Contract/utilsContract";
+import { toast } from "~~/utils/scaffold-eth";
 
 export const PurposeSetter = () => {
+  const [newPurpose, setNewPurpose] = useState("");
+
+  // Could we move useTransactor inside useScaffoldContractWrite?
+  const writeTxn = useTransactor();
```
carletex, line 11: Could we move useTransactor inside useScaffoldContractWrite?

packages/nextjs/components/ExampleUi/ContractInteraction/PurposeSetter.tsx
```diff
@@ -1,6 +1,23 @@
...
+  const { writeAsync, isLoading } = useScaffoldContractWrite(
+    "YourContract",
+    "setPurpose",
+    [newPurpose],
+    // Since this hook is going to be "high-level" maybe we should accept the value in ETH (not wei)?
+    // Passing "0.01" seems more high-level-dx-friendly.
+    "10000000000000000",
```
carletex, line 18: Since this hook is going to be "high-level" maybe we should accept the value in ETH (not wei)?

Passing "0.01" seems more high-level-dx-friendly.

Thoughts?

packages/nextjs/components/ExampleUi/ContractInteraction/PurposeSetter.tsx
```diff
@@ -10,11 +27,31 @@ export const PurposeSetter = () => {
...
-              Send <ArrowSmallRightIcon className="w-3 h-3 mt-0.5" />
+            <button
+              className={`btn btn-primary rounded-full capitalize font-normal font-white w-24 flex items-center gap-1 hover:gap-2 transition-all tracking-widest ${
+                isLoading ? "loading" : ""
+              }`}
+              onClick={async () => {
+                if (writeAsync && writeTxn) {
+                  try {
+                    // If the contract is not deployed this "works". I get a "mined successfully" msg.
+                    await writeTxn(writeAsync());
```
carletex, line 42: If the contract is not deployed this "works". I get a "mined successfully" msg.

Maybe we are not handling well errors in the write tx

packages/nextjs/components/ExampleUi/ContractInteraction/PurposeSetter.tsx
carletex: Haven't thought about the limitations of doing so, but doing it will also remove the need of doing `if (writeAsync && writeTxn) {` when calling the write TX

technophile-04: `if (writeAsync && writeTxn) { ` yeah this is a bit annoying and we definitely need to abstract it, also I see many people many making mistake here `await writeTxn(writeAsync())` they might forget to call `writeAsync` and it's better to handle it in `useScaffolContractWrite`

### Suggestion : 
Maybe we can use `useTransactor` inside `useScaffoldContractWrite` and wrap wagmi's `writeAsync` and then pass it to the user of the hook? 

Umm regarding inlining `useTransactor` in `useScaffoldContractWrite`, like `useTransactor` can also be handled to send raw txn's like in Faucet.tsx https://github.com/scaffold-eth/se-2/blob/0bf3a09a3b47017ad18e5ee4cf79ce83f4fc9628/packages/nextjs/components/scaffold-eth/Faucet.tsx#L29

Like I really feel we need to handle this `if (writeAsync && writeTxn) { ` internally and we can may be use the `suggestion` mentioned above instead of putting in `useTransactor` in `useScaffoldContractWrite` but happy to take either approach 🙌

packages/nextjs/components/ExampleUi/ContractInteraction/PurposeSetter.tsx
technophile-04: Ohh yes !! this shouldn't happen, Looking into it 🙌

packages/nextjs/components/ExampleUi/ContractInteraction/PurposeSetter.tsx
rin-st: > Umm regarding inlining useTransactor in useScaffoldContractWrite, like useTransactor can also be handled to send raw txn's like in Faucet.tsx

I think it's better to use `useTransactor` and that `if` inside `useScaffoldContractWrite`, but also use `useTransactor` for raw transactions. Or maybe create a different hook `useRawTransactor` for raws

technophile-04: > Or maybe create a different hook `useRawTransactor` for raws
Even this is a great idea! if we go with `useRawTransactor` then we can eliminate `useTransactor` 🙌

carletex: Great ideas!

I tried this: https://github.com/scaffold-eth/se-2/pull/124/commits/e09915c72b43951c318629619b47c1d2ca130f3a

It was the shortest path possible. I love the `onClick={writeAsync}` use for the hook user :D

What do you all think?

packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
```diff
@@ -22,5 +25,23 @@ export const useScaffoldContractWrite = (contractName: string, functionName: str
...
+  const sendContractWriteTx = async () => {
+    if (wagmiContractWrite.writeAsync && writeTx) {
+      try {
+        // If the contract is not deployed this "works". I get a "mined successfully" msg.
+        await writeTx(wagmiContractWrite.writeAsync());
+      } catch (e: any) {
+        const message = getParsedEthersError(e);
+        toast.error(message);
+      }
+    }
```
rin-st, line 39: else show toast, so user will undertand why nothing happens

packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
```diff
@@ -22,5 +25,23 @@ export const useScaffoldContractWrite = (contractName: string, functionName: str
     },
   });
 
-  return useContractWrite(config);
+  const wagmiContractWrite = useContractWrite(config);
+
+  const sendContractWriteTx = async () => {
+    if (wagmiContractWrite.writeAsync && writeTx) {
```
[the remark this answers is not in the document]
carletex: Like this idea and the toast suggestions!

We have more granular errors msgs now: https://github.com/scaffold-eth/se-2/pull/124/commits/0f26f690d5281f26e5b0dd3155b5c88d4487ce19

packages/nextjs/components/ExampleUi/ContractInteraction/PurposeSetter.tsx
carletex: Created https://github.com/scaffold-eth/se-2/issues/129

carletex:
Merging this! Still some improvements to make to the hooks, but we'll make them on separate PRs

cc @rin-st Check https://github.com/scaffold-eth/se-2/pull/124/commits/27f7a13bb89220b850ec99d8416d3273534c23f9. I'll explain on https://github.com/scaffold-eth/se-2/issues/118

PR #126, opened 2023-02-01
feat: upgrade to next 13

rin-st:
Everything works fine :)


technophile-04:
Thanks @rin-st, tested it out and everything works like a charm!!

Just noticed a very small warning in the console(not that important though) : 
Screenshot 2023-02-02 at 2 24 55 PM

technophile-04 (approved):

rin-st:
Thank you, @technophile-04 ! I didn't notice it. Updated pr, also changed font import. We still have a warning from `react-blockies` but they didn't approve prs for years 🤷‍♂️

carletex:
As discussed on DMs with @rin-st we are removing the NextJS recommended method to import a font, since we don't want to self-host it for a couple of font use in the example page.

For the main SE2 font (rn it's the default) we might want to use the Next recommended method. https://nextjs.org/docs/basic-features/font-optimization

Thanks al!!

Issue #127, opened 2023-02-02
accept the value in ETH for the `useScaffoldContractWrite` hook


technophile-04:

carletex:
Fixed on #128

Issue #125, opened 2023-02-01
Improve new useScaffoldContract* hooks


technophile-04:
The main aim of this issue is to keep track of all the improvements/bugs mentioned/noticed as we move towards `beta-launch` and improve it. Some of these TODOs will also be handled by #116 

- [ ] #116 
- [x] #127

carletex:
@technophile-04 I think we can just create single issues whenever we find something to fix regarding hooks. Thanks for this!

PR #128, opened 2023-02-02
Accept the value in ETH for the useScaffoldContractWrite

technophile-04:
This aim of this PR is to handle some tasks of #125. Maybe we can merge this PR when we handle 2-3 tasks of #125 don't want to clutter github with lots of PR


technophile-04:
[5f6c018](https://github.com/scaffold-eth/se-2/pull/128/commits/5f6c01829b9bd03bf440fb0c0e2abfa8fb68ff97) fixes #127

carletex:
@technophile-04 feel free to create a single PRs for each task. I love atomics & small PRs :D

packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
```diff
@@ -18,7 +19,7 @@ export const useScaffoldContractWrite = (contractName: string, functionName: str
     args,
     functionName,
     overrides: {
-      value: value ? parseTxnValue(value) : undefined,
+      value: value ? utils.parseEther("1").mul(parseTxnValue(value)) : undefined,
```
carletex: image

This was giving me an error when trying a decimal value.

I think this should be enough c9897b02a56cda27ea7fbb7c9f026956102ff08c. Maybe I'm missing an use case where it breaks... but we can fix later if we encounter an issue.

carletex (approved):

PR #130, opened 2023-02-05
Show error if contract is not present while sending txn

technophile-04:
## Reason for current bug. 
There are two cases : 
1. You cloned se-2 freshly and forgot to do `yarn deploy` after `yarn chain`, in this situation you don't have `generated` dir and if you try to send transatiion it will revert because of this condition  
```js
if (!deployedContractData) {
      toast.error("Target Contract is not defined");
      return;
}
``` 
2. The `genrated ` directory is already present but after you tinker with contract you forgot to do `yarn deploy` after `yarn chain`, this is the case where transation are going through even though we havn't deployed the contract.

I think the reason for this is that both `ethers` and `wagmi` doesn't check if the address you are calling is a contract or not and since we are calling an EOA with some `data` filled in `transation object` it dosen't have any effect and does not revert, hence we are getting `Minned Successfully !` and our balance is also getting deducted because its like calling EOA with `value` as "0.1ETH" and `data` as function signature and args which doesn't make sense but goes through without reverting (feel free to correct me If I have missed something or wrong 🙌 ) . 

## Reason for the approach taken
Currently looking at blockchain to see whats stored at `deployedContractData.address`, if its `0x0` then its high possibility that its not a contract. 

Saying  **"__Possibility__"** because checkout this https://github.com/ethers-io/ethers.js/issues/494#issuecomment-484567506

Also I saw this approach was taken in`se-1` and `eth-hooks` too  checkout : 
https://github.com/scaffold-eth/scaffold-eth/blob/e4afffc7e3d6c2bb575240b4de830703e8925548/packages/react-app/src/components/Contract/index.jsx#L69 internal working of `useContractExistsAtAddress` [here](https://github.com/scaffold-eth/eth-ui/blob/18df2143db4776813d734b74bde925bf2fd00ca9/packages/eth-hooks/src/hooks/useContractExistsAtAddress.ts#L41). 

Also was suggested here : https://github.com/ethers-io/ethers.js/discussions/3084#discussioncomment-2954385

Instead of handling the error at `useTransactor.tsx`, handled it inside `senContractWriteTx` function inside `useScaffoldContractWrite` hook because I think `useTransactor` seems like `dumb` hook which dosen't have context of whats its calling it just calls, to handle error inside `useTransactor` we need to pass it more arguments. 

Last thing currently we are showing error like this : 
Screenshot 2023-02-06 at 2 07 09 AM

For this I have createad a new component inside `packages/nextjs/components/scaffold-eth` because tailwind classes dosen't work inside `packages/nextjs/hooks/scaffold-eth` since its path not included in `tailwind.config.js`. 

Should we inline it inside `useScaffoldContractWrite` and add path to `tailwind.config.js` ? Also I think current approach is not minimal of showing error with component so fee free to point it out and will add a simple message asap instead of creating new components  🙌


technophile-04:
Also should we change below message to : 
https://github.com/scaffold-eth/se-2/blob/49553af6a7316d1b71cf0f51d09ad5e875a438c6/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts#L33

"Target Contract is not defined, did you forgot to do `yarn deploy` ?"

packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.tsx
```diff
@@ -34,9 +36,15 @@ export const useScaffoldContractWrite = (contractName: string, functionName: str
       return;
     }
 
+    // Looking at blockchain to see whats stored at `deployedContractData.address`, if its `0x0` then high possibility that its not a contract
+    const code = await provider.getCode(deployedContractData.address);
+    if (code === "0x") {
+      toast.error(<NoContractAddressError address={deployedContractData.address} />);
+      return;
+    }
+
```
carletex, line 45: Check the ToDo comment I wrote last week here: https://github.com/scaffold-eth/se-2/blob/main/packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx#L36

I feel it makes more sense to have this logic inside `getDeployedContract`.

If we do that, we could have the same error message for two cases you mentioned... since are both the same (contract is not deployed.)

```ts
    if (!deployedContractData) {
      toast.error("....");
      return;
    }
```

packages/nextjs/components/scaffold-eth/NoContractAddressError.tsx
```diff
@@ -0,0 +1,13 @@
...
+  return (
+    <>
+      <Address address={address} />
+      <p className="break-normal my-1">
+        Above address dosen&apos;t seems like an contract address did you forgot to do
+        <br /> `yarn deploy` ?
+      </p>
+    </>
+  );
+}
```
carletex, line 13: This is super specific for our use case (it's a toast error when using useScaffoldContractWrite). Not sure if we'll be using the component somewhere else, so let's not optimize early.

I think we can inline it in `useScaffoldContractWrite`: directly on the toast message, or in a component above.

carletex (changes requested):
Nice approach @technophile-04 

Left a couple of comments!

packages/nextjs/components/scaffold-eth/NoContractAddressError.tsx
carletex: Ok, sorry I didn't read this!!!

> For this I have createad a new component inside packages/nextjs/components/scaffold-eth because tailwind classes dosen't work inside packages/nextjs/hooks/scaffold-eth since its path not included in tailwind.config.js.
> Should we inline it inside useScaffoldContractWrite and add path to tailwind.config.js ? Also I think current approach is not minimal of showing error with component so fee free to point it out and will add a simple message asap instead of creating new components raised_hands

Yesss, let's add it to the config.

Love to see that we share the same mindset.

packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.tsx
```diff
@@ -34,9 +36,15 @@ export const useScaffoldContractWrite = (contractName: string, functionName: str
       return;
     }
 
+    // Looking at blockchain to see whats stored at `deployedContractData.address`, if its `0x0` then high possibility that its not a contract
+    const code = await provider.getCode(deployedContractData.address);
+    if (code === "0x") {
```
rin-st, line 41: is it '0x' or '0x0' like in the comment?

packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.tsx
technophile-04: > I feel it makes more sense to have this logic inside getDeployedContract.

Yup agree 💯, replaced `getDeployedContract` util with [useDeployedContractInfo](https://github.com/scaffold-eth/se-2/pull/130/commits/ad8ee98a7ba5af6ca146e15255a3f736babd9c21) hook so that we levearage wagmi's usefull hooks internally like `useProvider` instead of everytime creating it and then passing it to `getDeployedContract` util. 


PS : feel free to change the hook name and also happy to consider any approach

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
```diff
@@ -0,0 +1,44 @@
...
+        // Looking at blockchain to see whats stored at `deployedContractData.address`, if its `0x` then high possibility that its not a contract
+        const code = await provider.getCode(contractsData.address);
+        if (code === "0x") {
+          setDeployedContractData(undefined);
+          throw new Error();
+          return;
+        }
+        setDeployedContractData(contractsData?.[contractName as keyof typeof contractsData]);
+      } catch (e) {
+        toast.error("Target contract not found, did your forgot `yarn deploy` ?");
```
technophile-04, line 37: Showing the toast error inside here itself so that dev's can figure it out early : 

https://user-images.githubusercontent.com/80153681/217036461-8908fc63-1647-45f1-ad9d-a0ec1549cca9.mov

but this might bit annoying and happy to remove it

packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.tsx
technophile-04: Tysm @rin-st, I think in hex `0x` and `0x0` are same but might be wrong so fixed the comment at [f326e6e](https://github.com/scaffold-eth/se-2/pull/130/commits/f326e6ed4b9f4455de22d52d3f350d43025dbfd9), also while comparing its compared with `0x` as suggested here : https://github.com/ethers-io/ethers.js/discussions/3084#discussioncomment-2954385

packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.tsx
carletex: Love using a hook!

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
carletex: I removed, I think it was too much haha

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
```diff
@@ -0,0 +1,50 @@
...
+  useEffect(() => {
+    (async () => {
+      let ContractData;
+      try {
+        ContractData = require("~~/generated/hardhat_contracts.json");
+        const contractsAtChain = ContractData[chainId as keyof typeof ContractData];
+        const contractsData = contractsAtChain?.[0]?.contracts;
+
+        // Looking at blockchain to see whats stored at `deployedContractData.address`, if its `0x0` then high possibility that its not a contract
+        const code = await provider.getCode(contractsData.address);
```
carletex: This wasn't getting the right address. I fixed it on https://github.com/scaffold-eth/se-2/pull/130/commits/4c93cb75162bb9529ada46d8f2bbb554e0ba7fb8

With some other minor tweaks

carletex (commented):
@technophile-04 I pushed a fix and some minor tweaks.

Ready to merge!! Thanks Shiv! ;)

PR #131, opened 2023-02-05
Add surge for deployment of nextjs

technophile-04:
### Description / Flow
The current flow is very similar to se-1 : 
1. `yarn build`
This runs `next build && next export`, the reason for `next export` is for static files creation because it seems like surge only supports static file deployments. 
2. `yarn surge`
This run `surge out`  here `out` is directory outputted by `next build && next export` and surge deploys things inside `out` directory. 

### IMP 
I have used ` unoptimized: true` for images in next.config.js  because since we are using `next export` i.e building static pages in this Nextjs dosen't support image optimization with `next/image` if we don't do it we get the following error : 
https://stackoverflow.com/questions/65487914/error-image-optimization-using-next-js-default-loader-is-not-compatible-with-n

### Some research 
With surge, we won't be able to leverage the best abilities of Next.js like Image, creating api's etc and did some research what would be some other options that go well with Next.js  : 
1. [Vercel Cli](https://vercel.com/docs/cli)
2. [Netlify Cli](https://www.npmjs.com/package/netlify-cli)
3. [Layer0 Cli](https://docs.layer0.co/guides/cli)
4. [Fly cli](https://fly.io/docs/languages-and-frameworks/nextjs/)

What do you all think? feel free to suggest more option also happy refactor current approach 🙌 

cc @carletex @rin-st

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,8 @@
     "hardhat:lint-staged": "yarn workspace @se-2/hardhat lint-staged",
     "hardhat:test": "yarn workspace @se-2/hardhat test",
     "start": "yarn workspace @se-2/nextjs dev",
+    "build": "yarn workspace @se-2/nextjs build",
+    "surge": "yarn workspace @se-2/nextjs surge",
     "next:lint": "yarn workspace @se-2/nextjs lint",
     "next:format": "yarn workspace @se-2/nextjs format",
     "next:check-types": "yarn workspace @se-2/nextjs check-types",
--- a/packages/nextjs/next.config.js
+++ b/packages/nextjs/next.config.js
@@ -3,6 +3,9 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
+  images: {
+    unoptimized: true,
+  },
 };
 
 module.exports = nextConfig;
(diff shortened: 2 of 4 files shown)


carletex:
> With surge, we won't be able to leverage the best abilities of Next.js like Image, creating api's etc and did some research what would be some other options that go well with Next.js

Yep, we mentioned this on the GH Issue. I guess Vercel is the top option for NextJS projects. I haven't use their CLI, I always configure the deploy connecting github from the Vercel UI (and then any push to main is deployed automatically).

I think it's something we could explore!

@technophile-04 Could you check the differences (specially on DX / simplicity) between Surge CLI vs Vercel CLI? I'll be interested to see the steps you need to take (if you don't have an account on any of them) to deploy: going to the UI? confirm email?

- If Vercel CLI allows to not having to create an account (in the UI) and you can deploy directly from the CLI the first time you use it... maybe we could consider switching from Surge to Vercel.
- If not, we could offer both options (or just having Surge and telling people about options in the DOCS)

In any case, we have this PR that it's working and we can merge any time :) Thanks Shiv!!!

carletex:
Closing this for now since we merged #133

PR #132, opened 2023-02-05
Fix network names in env's and add reference to supported chains

technophile-04:
### Description 
It seems we are another chain name like `homestead` instead of `mainnet` (might be due to wagmi latest update) which causes the below problem : 

https://user-images.githubusercontent.com/80153681/216852150-a30f6a43-d536-4c96-b683-061de16d6e01.mov

Updated the lookup comment chain name in `env` and also added a link to supported chains from wagmi docs


carletex (approved):
Thanks @technophile-04 !!

PR #133, opened 2023-02-07
deploy nextjs using vercel cli

technophile-04:
## Flow : 
1. You run `yarn vercel`
Once you run `yarn vercel` it tells you to login with either of the below approaches (please ignore the command `yarn next:deploy` the ss is of before chaining the name) . 

Screenshot 2023-02-07 at 2 18 40 PM

**IMP** ->  The flow of login : 

1. Continue with Github 
Its like an OAuth authentication. if you select this option it will tell you to authorzie `vercel` on your github and once you accept it the CLI will automatically take you to next step. If you have already authorized to vercel on github only a vercel window pops up saying verified. 

2. Continue with email 
If you enter an email that is not registered with vercel it tells you to first signup with vercel 
Screenshot 2023-02-07 at 3 16 14 PM

2. After login you get to choose some options and add project name(this all can skipped with [--yes](https://vercel.com/docs/cli/deploy#yes) ) and once done your project is deployed: 
Screenshot 2023-02-07 at 2 24 19 PM

## Once the project is deployed, how do I make some changes and re-deploy to the same URL? 
- Whenever you run `yarn vercel` for the first time it creates a `.vercel` directory which has `.json` file which links your directory to vercel project. 
- Now after fixing some bug/making changes you run `yarn vercel` it will re-deploy your site to a preview URL
Screenshot 2023-02-07 at 2 40 11 PM
Preview URL - This not the actual deployement, its the palce where you can test some things if all your changes working. 
To deploy to same URL you need to run `yarn vercel --prod`

My opinion : 
I loved the DX of Vercel, like the login part is one-time hassle and not much of big pain if you do it through github(most of the people use) for some people those who haven't signed up with vercel and not using github instead going with email as signup have to provide phone-number(even this is one time for creating an account on vercel). 

checkout deployed app on goerli - https://test-deploy-iota.vercel.app/

Just wanted to put the `unpacked size` of this [`vercel`](https://www.npmjs.com/package/vercel) package is 8.04MB

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,16 @@ Run the command below to deploy the smart contract to the target network. Make s
 yarn deploy --network network_name
 ```
 
+## Deploying your NextJS App
+
+Run `yarn vercel` and follow the steps to deploy to Vercel. Once you log in (email, github, etc), the default options should work. It'll give you a public URL.
+
+If you want to redeploy to the same production URL you can run `yarn vercel --prod`. If you omit the `--prod` flag it will deploy it to a preview/test URL.
+
+**Make sure your `.env.production` file has the values you need.**
+
+**Hint**: We recommend connecting the project GitHub repo to Vercel so you the gets automatically deployed when pushing to `main`
+
 ## Contributing to Scaffold-Eth 2
 
 We welcome contributions to Scaffold-Eth 2!
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "next:check-types": "yarn workspace @se-2/nextjs check-types",
     "postinstall": "husky install",
     "precommit": "lint-staged",
-    "next:deploy" : "yarn workspace @se-2/nextjs deploy"
+    "vercel" : "yarn workspace @se-2/nextjs vercel"
   },
   "packageManager": "yarn@3.2.3",
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "next:check-types": "yarn workspace @se-2/nextjs check-types",
     "postinstall": "husky install",
     "precommit": "lint-staged",
-    "vercel" : "yarn workspace @se-2/nextjs vercel"
+    "vercel": "yarn workspace @se-2/nextjs vercel"
   },
   "packageManager": "yarn@3.2.3",
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "next:format": "yarn workspace @se-2/nextjs format",
     "next:check-types": "yarn workspace @se-2/nextjs check-types",
     "postinstall": "husky install",
-    "precommit": "lint-staged"
+    "precommit": "lint-staged",
+    "next:deploy" : "yarn workspace @se-2/nextjs deploy"
   },
   "packageManager": "yarn@3.2.3",
   "devDependencies": {
--- a/packages/nextjs/.env.production
+++ b/packages/nextjs/.env.production
@@ -1,4 +1,6 @@
+# The network where your DApp lives in.
+# mainnet / hardhat / polygon / optimism / arbitrum / goerli
 # Check supported chains - https://wagmi.sh/core/chains#supported-chains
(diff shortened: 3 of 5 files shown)


technophile-04:
Some issues which I found **Not related to this PR** : 
1. if you go to https://test-deploy-iota.vercel.app/debug you see `NO contract found` and you need to connect your wallet particularly to goerli to see the contract and even interact with it. This issue seems like an `beta-launch` issue I think the reason might be using of `useNetwork` and `useProvider` which depends on connected wallet rpc i might be wrong but looking into this 
2. I think we should keep it `true` here https://github.com/scaffold-eth/se-2/blob/5a715b28e6bf3c82500a03384154f76ef9ced2ed/packages/nextjs/.env.production#L2 because while deploying frontend project the env's are picked from `env.production` and most of the people will deploy to testnets. 
Saying this because lol I faced this and spend some time finding it out, if you keep `NEXT_PUBLIC_ENABLE_TESTNETS=false` and just change `NEXT_PUBLIC_NETWORK=goerli` in `env.production` it has this wired bug in deployed app : 
Screenshot 2023-02-07 at 2 43 48 PM
where it tells to switch to goerli even though you are on goerli already

rin-st:
Good job, Shiv!

> you see NO contract found and you need to connect your wallet particularly to goerli to see the contract and even interact with it

Probably it's because you deployed it to Goerli, so you can't see it from other network?

> I think we should keep it true here

Agree. Interesting, why it's false? 

> NEXT_PUBLIC_ENABLE_TESTNETS=false and just change NEXT_PUBLIC_NETWORK=goerli in env.production it has this wired bug in the deployed app

as I understand it's because `goerli` isn't added to appChains in this case. We need to add comments to `.env` that `NEXT_PUBLIC_NETWORK` should be in `appChains` (and comments for other vars too)
https://github.com/scaffold-eth/se-2/blob/main/packages/nextjs/services/web3/wagmiConnectors.tsx#L28

Also, I found that `ContractData` component is broken for the wrong network, so I'll need to fix it, thanks!

technophile-04:
Thanks so much Rinat 🙌

> Probably it's because you deployed it to Goerli, so you can't see it from other network?
Yes, but i think we need it to be shown even if you are not connected/on wrong network for this we use our Alchmey provider and if your not connected instead of `send` it show `connect wallet` on `/debug` page 

 > as I understand it's because goerli isn't added to appChains in this case. We need to add comments to .env that NEXT_PUBLIC_NETWORK should be in appChains (and comments for other vars too)
 💯 💯

carletex (approved):
Amazing stuff @technophile-04 thanks!

> I loved the DX of Vercel, like the login part is one-time hassle and not much of big pain if you do it through github(most of the people use) for some people those who haven't signed up with vercel and not using github instead going with email as signup have to provide phone-number(even this is one time for creating an account on vercel).

I tested it and I feel the same. Also Vercel support all NextJS capabilities (not like surge). So let's roll with it!

> if you go to https://test-deploy-iota.vercel.app/debug you see NO contract found and you need to connect your wallet particularly to goerli to see the contract and even interact with it. This issue seems like an beta-launch issue I think the reason might be using of useNetwork and useProvider which depends on connected wallet rpc i might be wrong but looking into this

Yep! I saw you created an issue/PR. I'll comment there. Thanks

> I think we should keep it true here (NEXT_PUBLIC_ENABLE_TESTNETS)
[se-2/packages/nextjs/.env.production](https://github.com/scaffold-eth/se-2/blob/5a715b28e6bf3c82500a03384154f76ef9ced2ed/packages/nextjs/.env.production#L2)

I changed it for now. I think we should take another approach for this, since the network switcher is not something that most people use (you usually deploy your dApp to work on a single network): hardhat when developing + goerli/mainnet on prod. Maybe instead of NEXT_PUBLIC_ENABLE_TESTNETS we can have a NEXT_PUBLIC_ENABLE_NETWORK_SWITCH (defaults fo false). Anyway, I'll think more about this and will create an issue for it.

I also added some (v0.1) documentation to the README.

PR #135, opened 2023-02-08
fix display of displayVariables on wrong network

technophile-04:
TODO : 
- [x] Make reads from contract available on the UI 
- [ ] Find a better way for `const chain = NETWORKS[process.env.NEXT_PUBLIC_NETWORK as string] as TChainAttributes` or maybe handle error if chainName is wrong in `env`
- [x] Show `connect wallet` instread of `send` if disconnected
- [x] Test and handle errors properly 

Demo : 

https://user-images.githubusercontent.com/80153681/217884539-034b60e7-8017-4903-93e7-c67a27453d33.mov


packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -161,3 +162,20 @@ export function getBlockExplorerTxLink(network: Network, txnHash: string) {
 
 export const getNetworkDetailsByChainId = (chainId: number) =>
   Object.values(NETWORKS).find(val => val.chainId === chainId);
+
+/**
+ * @dev returns the metadata of the chain like id, nativeCurrency etc specified in .env file or logs an error in console if chain mentioned in .env is unsupported
+ */
+export const getConfiguredChainFromENV = () => {
```
technophile-04, line 169: >  Find a better way for `const chain = NETWORKS[process.env.NEXT_PUBLIC_NETWORK as string] as TChainAttributes` or maybe handle error if chainName is wrong in `env`
 
 To handle this and show useful error in console came up with this function `getConfiguredChainFromENV `  please let me your thoughts, happy to take any other approaches 🙌 
 
 **Benefits** 
-  this functions gets you full metadata of chain configured in env along with autocompletion 
- logs potential error info if you misspelled chain or provide and unsupported chain 

I also tried exploring typed env but since `NEXT_PUBLIC` will be client side so didn't find any good solutions also we loose the advantage of getting whole chain metadata like id, name, ... etc which is helpful and typed

packages/nextjs/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx
```diff
@@ -31,6 +32,9 @@ export const WriteOnlyFunctionForm = ({
 }: TWriteOnlyFunctionFormProps) => {
   const [form, setForm] = useState<Record<string, any>>(() => getInitialFormState(functionFragment));
   const [txValue, setTxValue] = useState("");
+  const { isConnected } = useAccount();
```
technophile-04, line 35: Just noticed a small bug : 
If you try to disconnect from `/debug` page the modal doesn't disconnects you and also don't close 

https://user-images.githubusercontent.com/80153681/217891401-b265f494-490f-4b89-8b22-c22466eadcbd.mov

Tried debugging and found out that `useAccount` is causing this issue, removing it everything works fine. 

I think this will be fixed in https://github.com/wagmi-dev/wagmi/pull/1679
as mentioned by Carlos in https://github.com/scaffold-eth/se-2/issues/110#issuecomment-1422961150

packages/nextjs/utils/scaffold-eth/networks.ts
carletex: Good stuff! I renamed it 5fef117244169b4405cb6ec1b2176e3edc7fcb02 to hide implementation details on the function name

Just a silly nitpick I had, sorry haha

packages/nextjs/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx
carletex: Are we sure this is the same issue?

I get the not closing modal part (Should be fixed when this is merged https://github.com/rainbow-me/rainbowkit/pull/1000)

But the not-disconnecting part is more concerning.

I tried some stuff and it seems to be an issue inside `RainbowKitCustomConnectButton` (If I remove it, it works). Maybe the autoconnect?

Since the error is a bit annoying and I want to merge this PR I did a (temp?) fix for it: 75ec8de2e2646e81d4ee24342c0187fa2af9a2e8

carletex (approved):
Good stuff Shiv!

I left some comments and made some tweaks.

Also added it to the useScaffoldContractRead / useScaffoldContractWrite.

PR #145, opened 2023-02-09
fix: empty contract data

rin-st:
updates contract data block when there is no greeting
before:
image

after: 
image


carletex:
Thanks!

PR #147, opened 2023-02-09
feat: autocomplete for useScaffoldContractRead

rin-st:


rin-st:
Experimented with types for our hooks. Autocompletion works, but for it I added some questionable things :D
- created a script for generating ts Abi's from hardhat-generated json
- added that script to `deploy` so on every deploy ts abi's will be created too

return type still doesn't work, I need to copy lot's of types from 'abitype' or 'wagmi' need to investigate if it worth to continue this pr.

image

sverps:
This looks nice! I've had a look at trying to get the return type to work, but couldn't get the inferred type to pass to the wrapping hook. So might need a complex custom type that gets the return type from the abi.

This feels like a step in the right direction to me.

sverps:
I've had another run and (as you already said) using mainly some abitype stuff, got the return types working with the following approach:
```
import { useContractRead } from "wagmi";
import { Abi, AbiParametersToPrimitiveTypes, ExtractAbiFunctionNames, ExtractAbiFunction } from "abitype";

type GeneratedContract<TAbi> = {
  readonly address: string;
  readonly abi: TAbi;
};

type AbiReturnParameters<TAbi extends Abi, TFunctionName extends string> = ExtractAbiFunction<
  TAbi,
  TFunctionName
>["outputs"];
type AbiReturnType<TAbi extends Abi, TFunctionName extends string> = AbiParametersToPrimitiveTypes<
  AbiReturnParameters<TAbi, TFunctionName>
>[0];

/**
 * @dev wrapper for wagmi's useContractRead hook which loads in deployed contract contract abi, address automatically
 * @param contractData - generated contract data
 * @param functionName - name of the function to be called
 * @param readConfig   - wagmi configurations
 */
export const useScaffoldContractRead = <TAbi extends Abi, TFunctionName extends ExtractAbiFunctionNames<TAbi>>(
  contractData: GeneratedContract<TAbi>,
  functionName: TFunctionName,
  readConfig?: Parameters<typeof useContractRead>[0],
) => {
  return useContractRead({
    functionName,
    address: contractData.address,
    abi: contractData.abi,
    watch: true,
    ...readConfig,
  }) as Omit<ReturnType<typeof useContractRead>, "data" | "refetch"> & {
    data: AbiReturnType<TAbi, TFunctionName>;
    refetch: (options?: {
      throwOnError: boolean;
      cancelRefetch: boolean;
    }) => Promise<AbiReturnType<TAbi, TFunctionName>>;
  };
};
```

rin-st:
Wow, awesome! Hope I'll check it tomorrow

technophile-04:
Thaks so much Samuel this is really awesome 🤩!!! 

I have resolved the conflicts with `main` and updated this branch with second solution provided by @sverps so that its easy iterate and review. 


https://user-images.githubusercontent.com/80153681/219936748-df6402dc-b48a-4dc3-87f5-20bd6cd618e5.mov

**All the credits goes to @sverps  and @rin-st 🔥**


----

### Some known issues with current approach 
- Currently the UI breaks if you forgot to do `yarn deploy` and `generated` folder is not present 
Screenshot 2023-02-19 at 1 53 19 PM
- Github actions fails

---- 
### Suggestions needed 
I was wondering if we should have `scaffold.config.ts`  as single source of truth for configuring you Dapp's network. 

Reason - since `process.env` is treated as a runtime value we have hardtime figuring out the actual literal value because of which are not able to take full advantage of type inference. Also checkout demo vido there in second part I have hardcoded `TARGET_NETWORK` instead of getting it from `.env` or from our `utils/hooks`. 

We can even take advantage of this in other hooks eg `useDeployedContractInfo` : 
Screenshot 2023-02-19 at 2 12 49 PM

Disadvantage I can think of - We are being depenedent on `generated` file alot and if the developer fails to do it whole app will break. 

---- 

I am testing it properly and will be updating it with more known errors, would love to others thoughts. Also really sorry for pushing to this branch itself and feel free to revert it to [d37ba9d](https://github.com/scaffold-eth/se-2/pull/147/commits/d37ba9d7b687e7b98a0ab48d77c8ca1d78c6b366) just wanted to get started and iterate more on this 


Again all thanks to @sverps  and @rin-st  for their great work 🙏

rin-st:
for now I don't see other solutions, than
> Github actions fails

remove `generated` folder from gitignore, since typings rely on `generated/hardhat_contracts` file

> Currently the UI breaks if you forgot to do yarn deploy and generated folder is not present

change `start` script to
```
"start": "yarn deploy && yarn workspace @se-2/nextjs dev",
```

thoughts? @carletex @technophile-04 @sverps

carletex:
This is amazing!!

Testing this now and seeing how we can solve the existing issues. Will post my feedback here soon.

Thanks all!

carletex:
While I'm playing with this, some thoughts:

. It'll take some time to get this right, but let's keep exploring this approach.
- wagmi/cli as @technophile-04 suggested, might be a good solution when everything works :D It's super new, so let's see how it evolves
- _(from my fuzzy brain)_ I still think there should be a way to tweak `useDeployedContractInfo` / or passing a TS generic (with the ABI?) somewhere to get the types for the contract data. Anyway, it's no clear in my head right yet haha. Converting JSON to TS, seems to be needed.

**The biggest DX issue that we have right now:**
b
a

Since `data` in unknown we get the `ReactNode` error. It won't happen if data is `any`. Can we make that happen as v0.1? Creating an issue for it. https://github.com/scaffold-eth/se-2/issues/189

If we fix that, we won't have autocompletions / type inference, but at least you can use data freely.

> remove generated folder from gitignore, since typings rely on generated/hardhat_contracts file

The best solution would be that the generation happens on the build step, i.e. running the hardhat node + deploy on build.

Thanks all! I'll keep iterating on this.

sverps:
It makes sense to have the type definitions centralized inside a `useDeployedContractInfo` hook, and then inferred from it in the hooks that use it.

I'll have a crack at it and see what works :+1:

rin-st:
implemented in https://github.com/scaffold-eth/se-2/pull/202

PR #155, opened 2023-02-13
rename toast to notification

### LINKED ISSUE #138: Hiding toast from "react-hot-toast"
image

This is a minor/non-important issue, but improves DX.

 It would be nice to be able to "hide" for the builder `toast` from ""react-hot-toast". It's a bit confusing. Not sure if there is a way to hide that dependecy.

Or maybe we can just rename our "toast" to something else: "alert", "notification"?

technophile-04:
### Description 
renamed `toast` to the notification as discussed in #138

I ran `yarn next:lint` to check if anything broke and everything seems to work fine also tested manually

Fixes #138


carletex (approved):
Looks good!

PR #157, opened 2023-02-13
Feat/ens reverse lookup

rin-st:
Added resolving from address to ens + avatar.

Possibly there are more than one ens for address. Showing ens resolved by wagmi.

Fixed autocomplete styles

https://user-images.githubusercontent.com/25638585/218565427-3e4d585f-88d5-4fb7-bab8-7d405c9e7cfd.mov

### DIFF
--- a/packages/nextjs/tailwind.config.js
+++ b/packages/nextjs/tailwind.config.js
@@ -30,6 +30,19 @@ module.exports = {
           ".tooltip": {
             "--tooltip-tail": "6px",
           },
+
+          [`input:-webkit-autofill,
+            input:-webkit-autofill:hover,
+            input:-webkit-autofill:focus,
+            textarea:-webkit-autofill,
+            textarea:-webkit-autofill:hover,
+            textarea:-webkit-autofill:focus,
+            select:-webkit-autofill,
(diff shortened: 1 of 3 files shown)


technophile-04:
Looks great !! tested it out and works well 🙌

packages/nextjs/tailwind.config.js
```diff
@@ -30,6 +30,19 @@ module.exports = {
           ".tooltip": {
             "--tooltip-tail": "6px",
           },
+
+          [`input:-webkit-autofill,
```
technophile-04, line 34: small nitpick(maybe we can ignore it) : 

main branch : 

https://user-images.githubusercontent.com/80153681/218668808-5ca6c102-e6b9-4cc1-9797-4c0e7223d321.mov

This branch : 

https://user-images.githubusercontent.com/80153681/218668840-77cce80e-9001-44d7-bb1c-72afd4e8b624.mov

I see the only differece is while selecting atuocomplete it highlights input with different color, once the autocomplete is selected both branch looks same(just didn't wanted to ovveride defaults)

please don't remove this yet. Let's hear from @carletex first, I don't think its much of an issue even if we keep it

PR #158, opened 2023-02-13
fix(input): dark mode address input background should be aligned with regular input

sverps:
In dark mode, the address input has a different background color than the regular input. This small fix aligns both to use the expected dark background.


carletex (approved):
Nice catch, thanks @sverps!!

I did some tweaks also

Before:
image

After:
image

PR #159, opened 2023-02-14
feat: address => ens

rin-st:
https://user-images.githubusercontent.com/25638585/218752772-15712487-14e4-477a-8093-5ddd2dd16264.mov


carletex (approved):
Working great, thanks @rin-st !

I pushed 0e901238f734d30ab0dcdd016ffc213bbb3a80d8 to fix not displaying the result when is "0" (it shows in the video that Rinat posted on the PR). We want it to be like this: 

image

It was introduced here https://github.com/scaffold-eth/se-2/pull/156 
Is there an use case where we need that condition? @technophile-04

rin-st:
Thanks!

@technophile-04 added that condition to show `false` result

carletex:
Oh, I see. Thank! It was omitting the 0's tho.

I'll check https://github.com/scaffold-eth/se-2/pull/164 soon

PR #160, opened 2023-02-14
Fix/autofill

rin-st:
before:
https://user-images.githubusercontent.com/25638585/218755028-74659c4a-cc2c-422d-a7ff-d639837c5586.mov
https://user-images.githubusercontent.com/25638585/218755065-9483462c-5407-4859-bbbb-3ef320f634a5.mov

after:
https://user-images.githubusercontent.com/25638585/218755115-867cd2ac-0acc-4cc2-b5f1-3f91ea795b6c.mov
https://user-images.githubusercontent.com/25638585/218755149-5a3d6894-db7e-4f51-b8ab-244dc08cbd74.mov

### DIFF
--- a/packages/nextjs/components/scaffold-eth/AddressInput.tsx
+++ b/packages/nextjs/components/scaffold-eth/AddressInput.tsx
@@ -84,7 +84,7 @@ const AddressInput = ({ value, name, placeholder, onChange }: TAddressInputProps
             name={name}
             type="text"
             placeholder={placeholder}
-            className={`input input-ghost focus:outline-none focus:bg-transparent focus:text-gray-400 h-[2.2rem] min-h-[2.2rem] border w-full font-medium placeholder:text-accent/50 text-gray-400 grow ${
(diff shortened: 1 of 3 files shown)


carletex:
I think I'm missing what is happening here haha

Is it just the color? That when it gets autocompleted it uses a different one?

Is there a way to disabled autofill? We'd definitely want to do that if possible. A bunch of times people have been doxxed while doing SE1 demos, specially with the "AddressInput" component

carletex:
> It's a color, background color, borders, and border radius. For some reason, I thought we need that autofill, so it would be easier to enter the same data every time. If we can just disable autofill, it's much easier

Yep, we don't want autofill. We've seen a lot of people doxxing their private house address by using SE1 components (sometimes in a live stream).

I think disabling it wasn't trivial at the time, because I remember people trying to do without success (some browsers still auto filling if it was enabled in their settings).

But yeah, let's disable it!

rin-st:
Reverted changes and disabled autofill. Bug for browsers still exists https://caniuse.com/input-autocomplete-onoff, but I think browser autocomplete for addresses/wallets won't work for inputs with names like `userGreetingCounter_input_0__address_address`

PR #161, opened 2023-02-14
feat: contract changes feedback

rin-st:
Added animation hook + basic animations
https://user-images.githubusercontent.com/25638585/218795086-4e30ef09-64f0-4803-be78-7e4c1e2cc80c.mov

### DIFF
--- a/packages/nextjs/components/ExampleUi/ContractData.tsx
+++ b/packages/nextjs/components/ExampleUi/ContractData.tsx
@@ -2,6 +2,7 @@ import { useEffect, useRef, useState } from "react";
 import Marquee from "react-fast-marquee";
 import { useScaffoldContractRead } from "~~/hooks/scaffold-eth";
 import { BigNumber } from "ethers";
+import { useAnimationConfig } from "~~/hooks/scaffold-eth/useAnimationConfig";
 
 const MARQUEE_PERIOD_IN_SEC = 5;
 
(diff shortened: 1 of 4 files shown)


rin-st:
updated hook, removed animation on first load

technophile-04:
Looks awesome !! really like the hook approach

#### Small issue which I noticed : 
If we switch to dark mode it dosen't show up clearly (maybe we can use `bg-accent` instead of `bg-primary`  : 

https://user-images.githubusercontent.com/80153681/218824160-a85b7426-078f-49c8-a475-3536d00b3a4a.mov


#### Question : 
Should we highlight the changes if the users sets the same value second time ? example if initial value is "Hello" and user sets same greeting "Hello" should we highlight the same ? 


#### My suggestion : 
Instead of showing underlined highlight, maybe we highlight the whole content 

https://user-images.githubusercontent.com/80153681/218825406-95d7541d-e214-44c1-b524-63684ff1e868.mov

Like even the underlined highlight looks decent no problem with it 🙌

carletex:
This is amazing @rin-st looks great!

> Instead of showing underlined highlight, maybe we highlight the whole content

I like this from @technophile-04 a lot!

> Should we highlight the changes if the users sets the same value second time ? example if initial value is "Hello" and user sets same greeting "Hello" should we highlight the same ?

Not a big deal in any case. What ever is easier/default (for now)

Thanks!

rin-st:
Thanks!
> Instead of showing underlined highlight, maybe we highlight the whole content

Awesome idea! Updated
> Should we highlight the changes if the users sets the same value second time ? example if initial value is "Hello" and user sets same greeting "Hello" should we highlight the same ?

> Not a big deal in any case. What ever is easier/default (for now)

I think it shouldn't change. And it's easier, so I didn't change it :)

https://user-images.githubusercontent.com/25638585/218834488-cd7f7987-594c-47e2-b205-597c78625cdc.mov

carletex (approved):
Love the animations and the new hook, great job!

I played a bit with it an added a zoom in/out animation for the example UI :D

https://user-images.githubusercontent.com/2486142/219019874-275c265d-8e31-4cdb-8017-537c3f9f63ef.mp4

rin-st:
Awesome, thanks!

Note:
```diff
- const { showAnimation: showCurrentGreetingAnimation } = useAnimationConfig(currentGreeting);
```

Keep in mind, since you removed animation dependency from the current greeting, it now works when the total counter changes, even when the greeting doesn't change. Display variables component show animation only for changed values

carletex:
Yep, and I think it's ok since the counter will always change and we are making a animation for the whole screen.

Issue #142, opened 2023-02-09
useScaffoldEventsRead hook


carletex:
A way to read past events from a given contract, and listening to new events.

We can see what wagmi offers, and create a hook on top of it (as in useScaffoldContractRead & useScaffoldContractWrite)

KcPele:
It should keep track of past event and listens for new event that will be added to the past event. 
- should it listen for any event from the smart contract or accept the event name and keep track of that particular one?

KcPele:
#162

PR #162, opened 2023-02-15
useScaffoldEventsRead hook

KcPele:
useScaffoldEventsRead  is a custom hook built on top of wagmi useContractEvent

### DIFF
--- a/packages/nextjs/components/ExampleUi/ContractInteraction.tsx
+++ b/packages/nextjs/components/ExampleUi/ContractInteraction.tsx
@@ -11,7 +11,7 @@ export default function ContractInteraction() {
 
   const { writeAsync, isLoading } = useScaffoldContractWrite("YourContract", "setGreeting", [newGreeting], "0.01");
 
-  const eventContract = useScaffoldEventsRead("YourContract", "GreetingChange", (...args: any) => {
+  const eventContract = useScaffoldEventsRead("YourContract", "GreetingChange", (...args) => {
     console.log(args);
   });
   console.log(eventContract);
--- a/packages/nextjs/components/ExampleUi/ContractInteraction.tsx
+++ b/packages/nextjs/components/ExampleUi/ContractInteraction.tsx
@@ -3,13 +3,18 @@ import { DiamondIcon } from "./assets/DiamondIcon";
 import { HareIcon } from "./assets/HareIcon";
 import { ArrowSmallRightIcon, XMarkIcon } from "@heroicons/react/24/outline";
 import { useState } from "react";
-import { useScaffoldContractWrite } from "~~/hooks/scaffold-eth";
+import { useScaffoldContractWrite, useScaffoldEventsRead } from "~~/hooks/scaffold-eth";
 
 export default function ContractInteraction() {
   const [visible, setVisible] = useState(true);
   const [newGreeting, setNewGreeting] = useState("");
 
   const { writeAsync, isLoading } = useScaffoldContractWrite("YourContract", "setGreeting", [newGreeting], "0.01");
+
+  const eventContract = useScaffoldEventsRead("YourContract", "GreetingChange", (...args: any) => {
+    console.log(args);
+  });
+  console.log(eventContract);
 
   return (
     <div className="flex bg-base-300 relative pb-10">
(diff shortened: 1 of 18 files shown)


technophile-04:
Tysm @KcPele  🙌 , did some minor changes : 

1. made `callbackListener` type similar to wagm's [`listener`](https://github.com/wagmi-dev/wagmi/blob/cb707f01cbdcc62a70cf5c8a162d77948d6b6a56/packages/react/src/hooks/contracts/useContractEvent.ts#L79) instead of any at [d51dfdd](https://github.com/scaffold-eth/se-2/pull/162/commits/d51dfdd570d457cb913f436fc8b1a2b0e4888956)
2. Now we don't need to mention arguments types as `any` while passing callback checkout -> [84ba542](https://github.com/scaffold-eth/se-2/pull/162/commits/84ba5421bf288ab13effac6dd14657222c073449)
3. Renamed hook inlined with other `useScaffoldContract*` hooks -> [5115448](https://github.com/scaffold-eth/se-2/pull/162/commits/511544859b9ec93d012689c47f15ca1354d5984f)

technophile-04 (approved):

packages/nextjs/components/ExampleUi/ContractInteraction.tsx
```diff
@@ -3,14 +3,18 @@ import { DiamondIcon } from "./assets/DiamondIcon";
 import { HareIcon } from "./assets/HareIcon";
 import { ArrowSmallRightIcon, XMarkIcon } from "@heroicons/react/24/outline";
 import { useState } from "react";
-import { useScaffoldContractWrite } from "~~/hooks/scaffold-eth";
+import { useScaffoldContractWrite, useScaffoldContractEvent } from "~~/hooks/scaffold-eth";
```
carletex, line 6: Minor thing but: maybe we could move this to `ContractData.tsx`, since it's a read op?

carletex (commented):
Thanks @KcPele & @technophile-04 

From the issue description https://github.com/scaffold-eth/se-2/issues/142:
> A way to read **past events** from a given contract, and listening to new events.

I think that this implementation only listens to new events, while the user is on the page. Which is super handy! But _I think_ we also want to be able to retrieve past events. e.g. we want to list all the token transfers.

SE1 does it. But I think wagmi doesn't have a hook for it. Maybe it's too heavy and people are supposed to use another approach for it (like subgraphs)?

In any case, we can discuss on the existing issue and merge this.

Maybe change the hook name to `useScaffoldEventSubscriber` or similar?

Thoughts? Thanks!!

packages/nextjs/components/ExampleUi/ContractInteraction.tsx
rin-st: hmm, why do we need it? Maybe remove it for now?

technophile-04:
> I think that this implementation only listens to new events, while the user is on the page. Which is super handy! But I think we also want to be able to retrieve past events. e.g. we want to list all the token transfers.

Yeah, really sorry didn't read #142 properly


> Maybe change the hook name to useScaffoldEventSubscriber or similar?

Yup, this name makes more sense updated at [c53f357](https://github.com/scaffold-eth/se-2/pull/162/commits/c53f35786886b239b803b3f97b9c01573476836d)

packages/nextjs/components/ExampleUi/ContractInteraction.tsx
technophile-04: > Minor thing but: maybe we could move this to ContractData.tsx, since it's a read op?

💯, moved it at [25cda29](https://github.com/scaffold-eth/se-2/pull/162/commits/25cda2943095ac547fe0ecbc32cbed1ec9e0737b)

> hmm, why do we need it? Maybe remove it for now?
Yup makes sense too

rin-st (approved):

packages/nextjs/components/ExampleUi/ContractInteraction.tsx
KcPele: Listening for all the events will be very tricky except if we build a mini-event listener and make use of other packages or since the graph will still be added to this build, we should just make it simpler and easy to use for anyone making use of it.

KcPele:
@technophile-04 thanks for the changes

packages/nextjs/components/ExampleUi/ContractInteraction.tsx
carletex: > Listening for all the events will be very tricky except if we build a mini-event listener and make use of other packages or since the graph will still be added to this build, we should just make it simpler and easy to use for anyone making use of it.

SE-1 (with eth-hooks) does it. You can specify an start block when calling the `useEventListener` hook.

abasedape:
i was just working on this but decided to check! well done guys ill need to pull this down on local to check it out

PR #163, opened 2023-02-15
No need to check if transactor is undefined

carletex:
Something that came up while hacking with Austin.

Rn, when you use the transactor, you need to wrap it (or return in a guard clause) to check that it's not undefined like:

```tsx
<button
  className="btn btn-primary"
  onClick={() => {
    if (!tx) return;
    tx({
      to: contract?.address,
      value: ethers.utils.parseEther(0.01),
    });
  }}
>
  Send!
</button>
```

If not, you get a:

image

Not the best DX.

I think we are always returning a Promise, so the check is meaningless. 

Leave it here for review in case I'm missing something.


rin-st:
Yep,`useTransactor` always return a function. It seems someone forgot to remove return type of `useTransactor` after changing a hook.

Also, it's much easier to write `tx?.(...)` instead of 
```
if (!tx) return;
tx(...)
```

PR #164, opened 2023-02-15
fix: show literal false value returned by contract

technophile-04:
### Description
It seems if the contract read function returns `false` then its not show in the UI 

#### Before 

https://user-images.githubusercontent.com/80153681/219032901-2152254f-6e49-413c-9b51-3c0d475f437e.mov

### After 

https://user-images.githubusercontent.com/80153681/219033862-fa27f9d6-edce-42b1-9ab1-0163a94e74c5.mov

### Approach 
used looseEqual to check only for null and undefined, checkout -> https://stackoverflow.com/a/15992131

cc @carletex @rin-st, would be happy to refactor if I missed some cases 🙌

### DIFF
--- a/packages/hardhat/contracts/YourContract.sol
+++ b/packages/hardhat/contracts/YourContract.sol
@@ -3,6 +3,7 @@ pragma solidity >=0.8.0 <0.9.0;
 
 // Useful for debugging. Remove when deploying to a live network.
 import "hardhat/console.sol";
+
 // Use openzeppelin to inherit battle-tested implementations (ERC20, ERC721, etc)
 // import "@openzeppelin/contracts/access/Ownable.sol";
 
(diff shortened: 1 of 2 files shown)


packages/nextjs/components/scaffold-eth/Contract/ReadOnlyFunctionForm.tsx
```diff
@@ -59,7 +59,7 @@ export const ReadOnlyFunctionForm = ({ functionFragment, contractAddress }: TRea
       {inputs}
       <div className="flex justify-between gap-2">
         <div className="flex-grow">
-          {result ? (
+          {result != null ? (
```
carletex, line 62: Do we need the ternary?

Or could it just be:

```ts
          {result != null && (
            <span className="block bg-secondary rounded-3xl text-sm px-4 py-1.5">
              <strong>Result</strong>: {displayTxResult(result)}
            </span>
          )}
```

carletex:
Hey @technophile-04 

left a minor comment, and also noticed a bug that seems to be coming from the Wagmi cache (not related to this PR):

https://user-images.githubusercontent.com/2486142/219074319-423ca661-ea91-4eae-98b3-c12e550d4b19.mp4

The result is appearing without clicking the "read" button. And as you can see in this video the value is cached:

https://user-images.githubusercontent.com/2486142/219074910-c93639e0-5e1c-4b5d-aff9-19e261c2cf3f.mp4

We can definitely create a new issue/PR for this, just wanted to mention it.

Thanks!!

packages/nextjs/components/scaffold-eth/Contract/ReadOnlyFunctionForm.tsx
rin-st: or 
```
{result ?? (<span className="block bg-secondary rounded-3xl text-sm px-4 py-1.5">
    <strong>Result</strong>: {displayTxResult(result)}
  </span>
)}
```

rin-st: I'd prefer `result !== null && result !== undefined` so we and new devs don't need to remember what is `==` to `null`. Or add a link to StackOverflow as comment before this line. For example, if I would meet that line, I'd think any falsy value is `==` to  `null`

technophile-04: Ye, thanks for this 🙌 fixed at [3fd4b58](https://github.com/scaffold-eth/se-2/pull/164/commits/3fd4b58e7c70737a760f65fd03201f62a17bc5a3). According to me using `&&` won't cause a problem also tested it works fine. 

> Therefore, if the condition is true, the element right after && will appear in the output. If it is false, React will ignore and skip it.

react [docs ^ ](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator), since we are comparing so this will be either false or true and won't cause any issue

packages/hardhat/contracts/YourContract.sol
```diff
@@ -3,6 +3,7 @@ pragma solidity >=0.8.0 <0.9.0;
```
technophile-04, line 3: please revert this 🙏

technophile-04:
> The result is appearing without clicking the "read" button. And as you can see in this video the value is cached
> We can definitely create a new issue/PR for this, just wanted to mention it.

Yeah this is really annoying and would love to handle it in a different PR since it will be easy there to add an explanation of the approach taken

packages/nextjs/components/scaffold-eth/Contract/ReadOnlyFunctionForm.tsx
technophile-04: > I'd prefer result !== null && result !== undefined so we and new devs don't need to remember what is == to null. Or add a link to StackOverflow as comment before this line. For example, if I would meet that line, I'd think any falsy value is == to null

Added a comment for same at [215b851](https://github.com/scaffold-eth/se-2/pull/164/commits/215b8511edf3b64311d1f4f98f88ff8dc22a49bb)

carletex: > Added a comment for same at [215b851](https://github.com/scaffold-eth/se-2/pull/164/commits/215b8511edf3b64311d1f4f98f88ff8dc22a49bb)

Sorry, I was on my way to edit it also haha. I kind of prefer the explicit option `result !== null && result !== undefined`. Not as pretty. Not a big deal... just a nitpick :D

> According to me using && won't cause a problem also tested it works fine.

Yep, I think this is only tricky when you use template literals (${}), where `&&` would add "false" to the text.

carletex:
Thanks all!!

PR #154, opened 2023-02-12
Create local faucet

technophile-04:
Created up a bare minimum faucet modal(The modal also closes if you click outside it, forgot to mention in demo) : 

https://user-images.githubusercontent.com/80153681/218315141-e7556c06-9c10-484e-9617-f9f61faf5edd.mov

TODO : 
Screenshot 2023-02-12 at 7 24 49 PM
Currently, the above 3 options are remaining, was wondering if I should conquer them in this PR itself?

PR #166, opened 2023-02-15
fix UI break if local chain not running 

technophile-04:
### Desc 
I think this was introduced in #154 while listing account. If you forgot to do `yarn start` the UI break with the below error
Screenshot 2023-02-16 at 12 26 13 AM

### Approach : 
Added a very basic error, feel free to suggest a different error message because the current one might seem too specific also maybe we can silent error. 

### More suggestion 
I was also wondering if we should do `notification.error("Signer not found")` here:  https://github.com/scaffold-eth/se-2/blob/f2fde1e04a02defc111d12776288996da232b9b6/packages/nextjs/hooks/scaffold-eth/useTransactor.tsx#L44

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Faucet.tsx
+++ b/packages/nextjs/components/scaffold-eth/Faucet.tsx
@@ -30,9 +30,14 @@ export default function Faucet() {
 
   useEffect(() => {
     const getFaucetAddress = async () => {
-      if (provider) {
-        const accounts = await provider.listAccounts();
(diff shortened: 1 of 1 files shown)


carletex (approved):
Nice catch @technophile-04 , thanks!

I've pushed a commit with some tweaks:

- Hint "yarn chain" on the error
- Changed console.log to console.error
- Removed the line from console errors. It's misleading since it's usually wrong (we always make changes to a file, and it'll be a PITA to update those lines)
- Added the signer error that you suggested.

PR #170, opened 2023-02-16
fix(multiply string): make the button hitbox larger to improve UX

sverps:
The asterisk button to multiply the value by 10^18 was very narrow and not always easy to click. This fix improves accessibility by increasing its "hitbox".


carletex:
Hey @sverps 

Thanks for this, love the bigger hitbox.

A few things regarding PRs:
- We usually follow the process: Issue + discussion (if needed) + PR. This way we make sure we are aligned and the changes you want to make are needed / wanted, and you are not missing some context.
- Adding screenshots/videos before & after (specially for UI stuff) makes the review easier.

PR #171, opened 2023-02-16
feat(nav): make top nav sticky for easier access on mobile

sverps:
For improved UX (especially on mobile), I think the header should be sticky in order to always have easy access to navigation and address functionalities.

### DIFF
--- a/packages/nextjs/components/Header.tsx
+++ b/packages/nextjs/components/Header.tsx
@@ -56,7 +56,7 @@ export default function Header() {
   );
 
   return (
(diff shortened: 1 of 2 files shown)


carletex:
Cool @sverps This might make sense. We are currently working on the header with Andrea (the designer), specially on the wallet buttons.

Could you add some shadows?

The transition with the page content looks a bit rough.

image

sverps:
I added the same shadow that the other components use:

image

carletex:
After merging, I realized that this was applying to all screen sizes, not only mobile/tablet. We don't want that.

Fixed here 059af5452cff03c4ed6e28dbd67e77b446e8feaf

Issue #123, opened 2023-02-01
Update fixed elements on mobile


rin-st:
Right now, fixed elements look strange, and we need to think about what to do with them.

https://user-images.githubusercontent.com/25638585/216045432-2abacc06-8a05-4d42-b9eb-77bc9f20a795.mov

Simplest solutions:
1. Add background to theme-switcher
2. Move elements to the footer and make them non-fixed.

sverps:
I think adding a background is the better option. It keeps those buttons easily accessible and doesn't give up a lot of screen real estate (since the buttons were already making this area not really useful for the things behind it).

PR #172, opened 2023-02-16
fix(footer): align left footer buttons with right theme switcher

sverps:
This adds a background to the fixed footer elements, but only on mobile. #123

### DIFF
--- a/packages/nextjs/components/Footer.tsx
+++ b/packages/nextjs/components/Footer.tsx
@@ -15,21 +15,24 @@ export default function Footer() {
   const configuredNetwork = getTargetNetwork();
 
   return (
-    <div className="min-h-0 p-5 flex justify-between items-center flex-col sm:flex-row gap-4">
+    <div className="min-h-0 p-5 mb-11 lg:mb-0">
       <div>
-        <div className="flex space-x-2 fixed z-10 m-4 bottom-0 left-0">
-          {ethPrice > 0 && (
-            <div className="btn btn-primary btn-sm font-normal cursor-auto">
-              <CurrencyDollarIcon className="h-4 w-4 mr-0.5" />
-              <span>{ethPrice}</span>
-            </div>
-          )}
-          {configuredNetwork.id === hardhat.id && <Faucet />}
+        <div className="fixed flex justify-between items-center w-full z-20 p-4 bottom-0 left-0 pointer-events-none">
+          <div className="flex space-x-2 pointer-events-auto">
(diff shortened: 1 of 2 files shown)


carletex:
This is related to https://github.com/scaffold-eth/se-2/issues/123

I'd say: 

- Let's remove the background: Not a fan of the magic div with a magic number for the height.
- Use `md` instead of `lg`
- `mb-10` should be enough (instead of `mb-16`)

A couple of more fixes (existing errors):
- Could you vertically center the button with the theme switcher?
image

- Make the "Fork me...." to wrap.
image

Thanks @sverps !!

sverps:
Thanks for the remarks :+1: 
I'll clean it up a bit when I get some more time.

sverps:
Removed the background:
image

As for the wrapping of the "Fork me", even on devices as small as 360px, it still renders decent, so not sure it really needs wrapping (and I couldn't get flex-wrap to work for some reason... :disappointed: )

packages/nextjs/components/Footer.tsx
```diff
@@ -15,9 +15,9 @@ export default function Footer() {
   const configuredNetwork = getTargetNetwork();
 
   return (
-    <div className="min-h-0 p-5 flex justify-between items-center flex-col sm:flex-row gap-4">
+    <div className="min-h-0 p-5 flex justify-between items-center flex-col sm:flex-row gap-4 mb-10 lg:mb-0">
       <div>
-        <div className="flex space-x-2 fixed z-10 m-4 bottom-0 left-0">
+        <div className="flex items-center space-x-2 fixed z-20 h-16 bottom-0 left-4">
```
carletex, line 20: We shouldn't use magic numbers (like h-16, left-4)

Height should be a result of the element size + paddings. And we shouldn't move elements with "left / right" props (only `0`)

carletex (changes requested):
left a code review comment. Thanks

sverps:
Had to put all fixed elements inside the same fixed parent (rather than positioning them separately) in order to be able to align them.

The pointer-events need to be passed through the parent container to access the footer links that are rendered at a lower zIndex.

Additionally, `mb-11` looked like a slightly better choice than `mb-10`:
image

carletex:
Great job @sverps 

Oh, since we are "squash&merging" you don't need to push force to the branch to keep the commit history clean. It will also avoid me having tu pull force / delete branch.

Thanks!

PR #173, opened 2023-02-16
fix(spacing): decrease some paddings on mobile

sverps:
A fix to make the chain and contract variables display full width on mobile (to match the read and write components that precede it).
Additionally, small padding improvements for mobile to better use limited screen real estate.


carletex:
I think that was mostly fixed here: https://github.com/scaffold-eth/se-2/commit/185c33469115ecbd8d97b5c8c5c04d9160c1a439

Also check (general PR comments) https://github.com/scaffold-eth/se-2/pull/170#issuecomment-1434477027

Thanks!!

PR #174, opened 2023-02-16
this looks better to me but I submit it for sand garden review

austintgriffith:
🙇‍♂️ i deleted the network and fluff text and moved the static display to the left of the contract 

image

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
@@ -1,6 +1,6 @@
 import { Contract } from "ethers";
 import { useMemo, useState } from "react";
-import { useContract, useNetwork, useProvider } from "wagmi";
+import { useContract, useProvider } from "wagmi";
 import {
   getAllContractFunctions,
   getContractReadOnlyMethodsWithParams,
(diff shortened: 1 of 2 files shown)


rin-st:
lgtm! added small styles fix for mobile

Screenshot 2023-02-17 at 13 33 12

packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
```diff
@@ -21,14 +20,13 @@ type TContractUIProps = {
  * ToDo. Handle loading state
  **/
 const ContractUI = ({ contractName }: TContractUIProps) => {
-  const { chain } = useNetwork();
+  //const { chain } = useNetwork();
```
carletex, line 23: Saving this for later use? :D @austintgriffith

carletex:
> i deleted the network and fluff text and moved the static display to the left of the contract

I like the column arrangement better, with the info/read data on the left. The whole thing is not centered, but I think I like it.

- The network info: I think that's handy and should be there. I put it below the contract info in a smaller text
- The fluff text was suppose to give some info to people who land into the page and don't know what it's going. I added some text to the bottom.


Let's keep iterating.

image

PR #176, opened 2023-02-17
Error toast of faucet on deployement networks

technophile-04:
### Description 
When you configure `env` network other than `hardhat` we get the following error  on the first render: 

Screenshot 2023-02-17 at 10 06 45 AM

### Reason 
Since `getFaucetAddress` is used in `useEffect` and is called on render irrespective of chain configured in `env` https://github.com/scaffold-eth/se-2/blob/2a61bc87ac80f1e2de9dd6cad0cc10eebf2703e6/packages/nextjs/components/scaffold-eth/Faucet.tsx#L48

### Approach 
Added a guard at `Footer.jsx` level, since this seems the most minimal and clean

### another approach 
Instead of adding guard at `Footer.jsx` level, we can have it before calling the function https://github.com/scaffold-eth/se-2/blob/2a61bc87ac80f1e2de9dd6cad0cc10eebf2703e6/packages/nextjs/components/scaffold-eth/Faucet.tsx#L48

### DIFF
--- a/packages/nextjs/components/Footer.tsx
+++ b/packages/nextjs/components/Footer.tsx
@@ -4,12 +4,15 @@ import { useAppStore } from "~~/services/store/store";
 import { HeartIcon } from "@heroicons/react/24/outline";
 import SwitchTheme from "./SwitchTheme";
 import { Faucet } from "~~/components/scaffold-eth";
+import { getTargetNetwork } from "~~/utils/scaffold-eth";
+import { hardhat } from "wagmi/chains";
 
 /**
  * Site footer
  */
 export default function Footer() {
   const ethPrice = useAppStore(state => state.ethPrice);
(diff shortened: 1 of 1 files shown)


rin-st (approved):

carletex (approved):
Thanks @technophile-04 ! This fixes the issue.

But I noticed two things (that we might want to solve in separate issues)

1. Too many notifications on load :D
image. Not sure if this should be handled on the caller, or directly on the notification util (saving a list of currently showed notifications, and don't show if a notification with the same content is there?)

2. If `NEXT_PUBLIC_NETWORK` is a invalid network, it defaults to hardhat (because of `getTargetNetwork`). I get the same "Cannot connect to local provider" which is confusing. Maybe we should show a notification error in `getTargetNetwork` (alongside the the console.error) and return undefined.

PR #177, opened 2023-02-17
fix: wagmi cached result on `/debug` page

technophile-04:
### Approach 
Maintained our own state instead of relying on wagm

### Reason for the approach taken 
I tried different [configurations](https://wagmi.sh/react/hooks/useContractRead#configuration) which wamit's `useContractRead` hook provides but none were helpful and it makes sense. 

The current approach seemed minimal and makes more sense since we want to show data only when users hit's `Read` button.

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/ReadOnlyFunctionForm.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/ReadOnlyFunctionForm.tsx
@@ -42,7 +42,10 @@ export const ReadOnlyFunctionForm = ({ functionFragment, contractAddress }: TRea
     return (
       <InputUI
         key={key}
-        setForm={setForm}
+        setForm={updatedFormValue => {
+          setResult(null);
+          setForm(updatedFormValue);
+        }}
         form={form}
         stateObjectKey={key}
         paramType={input}
--- a/packages/nextjs/components/scaffold-eth/Contract/ReadOnlyFunctionForm.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/ReadOnlyFunctionForm.tsx
@@ -22,7 +22,7 @@ type TReadOnlyFunctionFormProps = {
 
 export const ReadOnlyFunctionForm = ({ functionFragment, contractAddress }: TReadOnlyFunctionFormProps) => {
   const [form, setForm] = useState<Record<string, any>>(() => getInitialFormState(functionFragment));
-  const [result, setResult] = useState<unknown>();
+  const [result, setResult] = useState<unknown>(undefined);
   const configuredChain = getTargetNetwork();
 
   const { isFetching, refetch } = useContractRead({
@@ -43,7 +43,7 @@ export const ReadOnlyFunctionForm = ({ functionFragment, contractAddress }: TRea
       <InputUI
         key={key}
         setForm={updatedFormValue => {
-          setResult(null);
+          setResult(undefined);
           setForm(updatedFormValue);
         }}
         form={form}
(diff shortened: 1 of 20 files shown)


carletex:
Hey @technophile-04 I like the approach. Not sure why the cache options from wagmi don't work.

One thing that I'm not sure about: Should we remove the result when you modify the input? (checking when "form" changes in a useEffect) Rn it stays until you click read again.

What do you think?

https://user-images.githubusercontent.com/2486142/219621745-7605e3cd-3623-480e-8826-e494fffbbf22.mp4

packages/nextjs/components/scaffold-eth/Contract/ReadOnlyFunctionForm.tsx
```diff
@@ -44,7 +42,10 @@ export const ReadOnlyFunctionForm = ({ functionFragment, contractAddress }: TRea
     return (
       <InputUI
         key={key}
-        setForm={setForm}
+        setForm={updatedFormValue => {
+          setResult(null);
```
rin-st, line 46: nitpick: add `if (result !== null)` for this `setResult` since it can add renders.
And I think it should be the default value, `undefined`, or set default value `null` to `useState`

carletex: >  add if (result !== null) for this setResult since it can add renders.

I think when you have multiple state changes (in the same event loop cycle i.e. same handler), React batches them so it doesn't add any extra renders.

rin-st: Yes, especially in such a small component. But it's not a 100% stable thing 🤷‍♂️ . Not critical at all :)

rin-st (approved):

packages/nextjs/components/scaffold-eth/Contract/ReadOnlyFunctionForm.tsx
technophile-04: > I think it should be the default value, undefined, or set default value null to useState

Since we were setting the value that's why setted it to null but fixed it at [54e2b64](https://github.com/scaffold-eth/se-2/pull/177/commits/54e2b64d78c16772d5bd7ea814883a2afe9adbcd) thanks 🙌

rin-st: Thanks! Added a small fix (default state was already undefined) https://github.com/scaffold-eth/se-2/pull/177/commits/09e0af11a3098ab08bc3b2adc93fe51213e220f1

jxom:
Setting `cacheTime` to `0` should disable the cache for `useContractRead`. Does this not work?

technophile-04:
> Setting cacheTime to 0 should disable the cache for useContractRead. Does this not work?

Tysm @jxom for the suggestion 🙌, yeah no luck setting `cacheTime : 0` it still picks up the value from somewhere. Btw we have `enabled : false` because we are using controlled input and were concered about rpc calls + wanted to show ouput only on `Read` button clicked. 


https://user-images.githubusercontent.com/80153681/219941281-c8661ebc-3c5c-4ff7-bbd7-4506b357147e.mov

The abovr video demonstrates two case : 

1) cacheTime : 0 and enabled : false
Ideally in this case `result` shouldn't been stored somewhere and should not give pre-calculated value since we have `cache` disabled. 

2) cacheTime : 0 and enabled : true
In this case `cache` is  ~~ _**diabled**_~~  and we are allowing request to go throught and ~~_**update the cahe**__~~ isn't this contradictory ? Since we do not have any `cache` to update ? but in video we see that for this case the value is updated. 

Please pardon me if I am wrong somewhere or misinterpreted something wrong 🙏, Thanks though for the suggestion and review  🙌

carletex:
Thanks @technophile-04 

Since saving the result in the state makes this work as expected, let's merge it until we figure out why disabling wagmi cache doesn't work.

Thanks for chiming in @jxom! Let us know if we are missing something.

PR #179, opened 2023-02-17
fix: balance view

rin-st:
Fixes balance view, it's broken for some resolutions, and now it's clear what that value means. Also changed "Network:" text color.

before:
Screenshot 2023-02-17 at 18 28 46

after:
Screenshot 2023-02-17 at 18 28 25

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
@@ -63,14 +63,17 @@ const ContractUI = ({ contractName }: TContractUIProps) => {
       <div className="col-span-1 flex flex-col">
         <div className="bg-base-100 border-base-300 border shadow-md shadow-secondary rounded-3xl px-8 mb-6 space-y-1 py-4">
           <div className="flex">
-            <div className="flex gap-1">
+            <div className="flex flex-col gap-1">
               <Address address={contractAddress} />
-              <Balance address={contractAddress} />
+              <div className="flex gap-1 items-center">
+                <span className="font-bold text-sm">Balance:</span>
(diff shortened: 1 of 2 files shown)


rin-st:
Btw, why do we need that empty space on the right side? Maybe remove it?

packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
```diff
@@ -63,14 +63,17 @@ const ContractUI = ({ contractName }: TContractUIProps) => {
...
+                <span className="font-bold text-sm">Balance:</span>
+                <Balance address={contractAddress} className="px-0 h-1.5 min-h-[0.375rem]" />
+              </div>
             </div>
           </div>
           {chain && (
-            <p className="my-0 text-sm" style={{ color: networkColor }}>
-              <span className="font-bold">Network</span> : {chain.name}
+            <p className="my-0 text-sm">
+              <span className="font-bold">Network</span>: <span style={{ color: networkColor }}>{chain.name}</span>
```
carletex, line 76: Just noticed this also. Thanks! ;)

carletex:
> Btw, why do we need that empty space on the right side? Maybe remove it?

Yep, I mentioned here. https://github.com/scaffold-eth/se-2/pull/174#issuecomment-1434442355

We could:
- Make the input area bigger 
- Center the whole thing with the current size

but IDK, I kind of like it how it's rn. Not being centered gives more weight to the input area.

In any case, lets keep like this, and we'll do a UI/UX review when the core is completed.

Same thing with the components: I feel that we need to standardize them (like the optional `className` that you added here).

But we'll have a better view when the core is completed and we use it.

Issue #168, opened 2023-02-16
Supported networks


carletex:
Rn we have:

![image](https://user-images.githubusercontent.com/2486142/219417259-777d5262-dd8e-4e89-a44d-f787328e2f48.png)

And we have the network switcher:
![image](https://user-images.githubusercontent.com/2486142/219417825-381e4982-f3df-4f42-ae1c-b8a817bc3a59.png)

That allows you to switch:

![image](https://user-images.githubusercontent.com/2486142/219417927-002cdb6c-2cf2-42a6-8be3-cf5b106fa4f3.png)

---

I think this shouldn't be the default, since **most** people build a dApp for a given network, not many.

My idea:

- Let's have the `NEXT_PUBLIC_NETWORK` in the `env` files be the only supported network for the app (you could have hardhat for development, goerli for staging and maninnet for production)
- Remove the network switch.

In the future, we could support multiple (comma-separated) network values on `NEXT_PUBLIC_NETWORK`, but for now, let's focus on the bigger use case.

Thoughts? @technophile-04 @rin-st

rin-st:
You're right, agree.

What do you think about using Sepolia instead of Goerli? It is recommended for dapps development [Eth docs](https://ethereum.org/en/developers/docs/networks/#which-testnet-should-i-use)

carletex:
> What do you think about using Sepolia instead of Goerli? It is recommended for dapps development [Eth docs](https://ethereum.org/en/developers/docs/networks/#which-testnet-should-i-use)

The issue is that Alchemy doesn't support sepolia rn, and we are
![photo1676912211](https://user-images.githubusercontent.com/2486142/220166365-596a9652-dc19-4d5f-9b2d-34d7fddace07.jpeg)

---

Austin is talking with them rn. So let's see if they have a plan to support Sepolia in the near future. If not, we might need to switch providers.

PR #182, opened 2023-02-17
HOTFIX: Mainnet working when selected as target network

carletex:
Because of the naming of mainnet in wagmi/rainbowkit (homestead), when `NEXT_PUBLIC_NETWORK` was "mainnet" it wasn't working. Using `homestead` on `NEXT_PUBLIC_NETWORK` wasn't working either (because our logic).

photo1676659213
photo1676659144

---

This is kind of time sensitive so I created this hotfix. Merging it and hoping not to break anything :angel: 

We need to think how to handle networks better and define a source of truth for networks (maybe our `networks.ts`?). 

Also refactor things like this (and use our `getTargetNetwork`):

image

This is also related to #168

Let's review this next week @technophile-04 @rin-st

### DIFF
--- a/packages/nextjs/utils/scaffold-eth/networks.ts
+++ b/packages/nextjs/utils/scaffold-eth/networks.ts
@@ -1,133 +1,132 @@
 import { Network } from "@ethersproject/networks";
-import { appChains } from "~~/services/web3/wagmiConnectors";
 
 export type TChainAttributes = {
   name: string;
   // color | [lightThemeColor, darkThemeColor]
   color: string | [string, string];
-  chainId: number;
+  id: number;
 };
 
 export const NETWORKS: Record<string, TChainAttributes> = {
-  localhost: {
-    name: "localhost",
+  hardhat: {
+    name: "hardhat",
     color: ["#666666", "#bbbbbb"],
-    chainId: 31337,
+    id: 31337,
   },
   mainnet: {
     name: "mainnet",
     color: "#ff8b9e",
-    chainId: 1,
+    id: 1,
   },
   goerli: {
     name: "goerli",
     color: "#0975F6",
-    chainId: 5,
+    id: 5,
   },
   gnosis: {
     name: "gnosis",
     color: "#48a9a6",
-    chainId: 100,
+    id: 100,
   },
   polygon: {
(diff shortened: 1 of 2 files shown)


packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -160,21 +159,19 @@ export function getBlockExplorerTxLink(network: Network, txnHash: string) {
...
+export const getNetworkDetailsByChainId = (id: number) => Object.values(NETWORKS).find(val => val.id === id);
 
 /**
  * @dev Returns the target network metadata or defaults to hardhat if the network is unsupported/undefined
  */
 export const getTargetNetwork = () => {
   const network = process.env.NEXT_PUBLIC_NETWORK;
-  const { chains } = appChains;
-  const configuredChain = chains.find(chain => chain.network === network);
+  const configuredChain = NETWORKS[network ?? "hardhat"];
```
rin-st, line 169: Note: It will be Hardhat when there is no `NEXT_PUBLIC_NETWORK` in env, but `undefined` when the network is `''`, (`NEXT_PUBLIC_NETWORK=` in `.env`)

rin-st:
> Also refactor things like this (and use our getTargetNetwork):

https://github.com/scaffold-eth/se-2/pull/185

packages/nextjs/utils/scaffold-eth/networks.ts
technophile-04: ++, but it will be set again to `hardhat` if it's `undefined`....but maybe we refactor to make it simple

technophile-04:
> We need to think how to handle networks better and define a source of truth for networks (maybe our networks.ts?).

Yes, this has always been a confusion...just checked the main branch and everything works  great 🤞

packages/nextjs/utils/scaffold-eth/networks.ts
rin-st: Yep, and it will not show console error in the first case, but it will show it in second

technophile-04: yes ! 🙌, Fix at #187

PR #185, opened 2023-02-18
Refactor connect button network

rin-st:
Refactor mentioned here https://github.com/scaffold-eth/se-2/pull/182#issue-1589860816

### DIFF
--- a/packages/nextjs/utils/scaffold-eth/useNetworkColor.ts
+++ b/packages/nextjs/utils/scaffold-eth/useNetworkColor.ts
@@ -1,10 +1,9 @@
-import { getNetworkDetailsByChainId } from "./networks";
+import { getTargetNetwork } from "./networks";
 import { useDarkMode } from "usehooks-ts";
 
-export function useNetworkColor(chainId?: number) {
+export function useNetworkColor() {
   const { isDarkMode } = useDarkMode();
-  const networkDetails = chainId && getNetworkDetailsByChainId(chainId);
-  if (!networkDetails) return "#666666";
+  const networkDetails = getTargetNetwork();
 
   return Array.isArray(networkDetails.color)
     ? isDarkMode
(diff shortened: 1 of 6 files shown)


technophile-04:
LGTM 🙌!

packages/nextjs/utils/scaffold-eth/useNetworkColor.ts
```diff
@@ -1,10 +1,9 @@
-import { getNetworkDetailsByChainId } from "./networks";
+import { getTargetNetwork } from "./networks";
 import { useDarkMode } from "usehooks-ts";
 
-export function useNetworkColor(chainId?: number) {
+export function useNetworkColor() {
```
technophile-04, line 4: we can mention a short JsDoc comment 

maybe - "gives you the color of the current configured/target network"

rin-st: Thanks, @technophile-04 ! Added JsDoc. But chainId is still removed since we don't need it for now and it intersects with `getTargetNetwork` change https://github.com/scaffold-eth/se-2/pull/187. Let's add it when we need it.

carletex: Added the comment + moved the "util" to the "hook" folder (since I think it should be there - it's using a hook inside it)

> also was wondering if should we accept chainId ?

I think this could be useful at some point. But let's add it if we end needing it.

carletex: Sorry, didn't see your comment @rin-st 

fixing it now

packages/nextjs/utils/scaffold-eth/useNetworkColor.ts
```diff
@@ -1,6 +1,10 @@
 import { getTargetNetwork } from "./networks";
 import { useDarkMode } from "usehooks-ts";
 
+/**
+ * @dev Gives you the color of the current configured/target network
+ * @return {string | [string, string]} network color/colors
```
carletex, line 6: I think this never returns an array, right?

packages/nextjs/utils/scaffold-eth/useNetworkColor.ts
carletex: Sorry for the push force :D

packages/nextjs/utils/scaffold-eth/useNetworkColor.ts
rin-st: oops, you're right, thanks :)

PR #187, opened 2023-02-18
handle edge-case for getTargetNetwork

technophile-04:
### Desc 
There are very less chances that dev might delete `NEXT_PUBLIC_NETWORK` from `env` but if he does so then he won't get any error in the console and network will be set to `hardhat` without letting him know 

Originally discussed here -> https://github.com/scaffold-eth/se-2/pull/182/files#r1111042625

### DIFF
--- a/packages/nextjs/utils/scaffold-eth/networks.ts
+++ b/packages/nextjs/utils/scaffold-eth/networks.ts
@@ -166,14 +166,13 @@ export const getNetworkDetailsByChainId = (id: number) => Object.values(NETWORKS
  */
 export const getTargetNetwork = () => {
   const network = process.env.NEXT_PUBLIC_NETWORK;
-  const configuredChain = NETWORKS[network ?? "hardhat"];
 
-  if (!configuredChain) {
+  if (!network || !NETWORKS[network]) {
     // If error defaults to hardhat local network
     console.error("Network name misspelled or unsupported network used in process.env");
     const hardhatChain = NETWORKS["hardhat"];
     return hardhatChain;
   }
 
-  return configuredChain;
+  return NETWORKS[network];
 };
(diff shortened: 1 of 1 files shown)


packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -166,14 +166,13 @@ export const getNetworkDetailsByChainId = (id: number) => Object.values(NETWORKS
  */
 export const getTargetNetwork = () => {
   const network = process.env.NEXT_PUBLIC_NETWORK;
-  const configuredChain = NETWORKS[network ?? "hardhat"];
 
-  if (!configuredChain) {
+  if (!network || !NETWORKS[network]) {
     // If error defaults to hardhat local network
     console.error("Network name misspelled or unsupported network used in process.env");
```
technophile-04, line 172: Also should mention in the end of console.error that the "target network is set to hardhat" OR "using default network as hardhat"

rin-st: And probably change console.error to `Network name is not set, misspelled or unsupported network used in process.env`

packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -166,14 +166,13 @@ export const getNetworkDetailsByChainId = (id: number) => Object.values(NETWORKS
  */
 export const getTargetNetwork = () => {
```
rin-st, line 167: Try to change the value in `.env` to invalid. All the console becomes red. (It was before your changes too). I think we need to calculate the value once and save it to the global state/context. (probably another pr, but not sure)

carletex (approved):
Thanks!!

PR #188, opened 2023-02-20
Supported networks

### LINKED ISSUE #168: Supported networks

carletex:
Fixes #168 

- We had a mix between our own defined networks (in `networks.ts`) and wagmi networks. In this PR, we are using wagmi networks as the source of truth and using our `NETWORKS` (now `NETWORKS_EXTRA_DATA`) as a way to adding extra metadata to some networks (right now it's only the color, but in the future, it could be: displayName, BlockExplorer, etc)

- Wagmi/Rainbowkit available networks are now only the target app network (`NEXT_PUBLIC_NETWORK`) and mainnet (we need it for ENS resolution, Get ETH price, etc)

- Removed the network switcher (as discussed in #168), but we now show the target network below the balance (thanks Andrea for this!)
image

Would love to hear your thoughts on this @rin-st @technophile-04 and make sure that I'm not missing anything.

### DIFF
--- a/packages/nextjs/.env.development
+++ b/packages/nextjs/.env.development
@@ -1,6 +1,5 @@
 # The network where your DApp lives in.
-# mainnet / hardhat / polygon / optimism / arbitrum / goerli
+# mainnet / hardhat / polygon / optimism / arbitrum / goerli / sepolia
 # Check supported chains - https://wagmi.sh/core/chains#supported-chains
 NEXT_PUBLIC_NETWORK=hardhat
-NEXT_PUBLIC_ENABLE_TESTNETS=true
 NEXT_PUBLIC_RPC_POLLING_INTERVAL=30000
--- a/packages/nextjs/.env.production
+++ b/packages/nextjs/.env.production
@@ -1,6 +1,5 @@
 # The network where your DApp lives in.
-# mainnet / hardhat / polygon / optimism / arbitrum / goerli
+# mainnet / hardhat / polygon / optimism / arbitrum / goerli / sepolia
 # Check supported chains - https://wagmi.sh/core/chains#supported-chains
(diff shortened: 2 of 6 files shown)


technophile-04:
Really like the refactor !! , Like that now `getTargetNetwork` now sets the configured chains 🙌 much more integrated , `NETWORKS_EXTRA_DATA` have keys as id's instead of names much needed and really goes well with wagmi `chains` !! 

Also tested it out both locally and goerli and everything works great !! 

Just a small nitpick and we can tackle it in new PR , On mobile devices we are not showing current connected network 
|Main | This branch | 
|-----|-------------|
|Screenshot 2023-02-21 at 5 26 31 PM| Screenshot 2023-02-21 at 5 27 28 PM|

Functionally and code wise it really looks more simplified and focused from main

technophile-04 (approved):

rin-st (approved):

carletex:
Thanks for the review @technophile-04 !

I did a lil tweak on 98aa10db11f3ecf6fc37c6a20997ed1e6bd0c8b6 to show the balance + network on mobile. I think we were hiding it because it didn't fit when we had the network switcher. It might look a bit tight on super small devices (only on hardhat, since we have the faucet button)

image

PR #190, opened 2023-02-20
wrap data with `displayTxResultAsText` util

### LINKED ISSUE #189: Make { data } for useScaffoldContractRead to be "any"
Until we get the autocompletion / type inference right. For now we'll avoid:

b
a

Check https://github.com/scaffold-eth/se-2/pull/147#issuecomment-1437373694

technophile-04:
### Approach 
Wrapped `data` with `displayTxResultAsText`, this util seems great for this work : ) 

Would love to hear your thoughts @carletex @rin-st on this approach 

PS : this is just draft will test properly and re-request for review

Fixes  #189

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/utilsDisplay.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/utilsDisplay.tsx
@@ -41,7 +41,7 @@ export const displayTxResult = (
   return JSON.stringify(displayContent, null, 2);
 };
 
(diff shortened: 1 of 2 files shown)


carletex:
Noiiice haha <3

The only «problem» I see is that we are transforming the result, and people might expect the raw value (the bignumber, or whatever).

Maybe we can type assert `data` with all the posible returning types from a contract read?? "BigNumber | string |....". Just thinking out loud

carletex:
Closing this in favor of https://github.com/scaffold-eth/se-2/pull/192

Thanks Shiv!

PR #191, opened 2023-02-20
update github workflow to run yarn chain and yarn deploy

technophile-04:
Since we are relying on `generate` in this repo, it will be a good idea to run `yarn chain` & `yarn deploy` before linting to avoid github-actions failing due to `generated` dir not found or error with types. 

Currently, as such, there is no need for this PR but we can probably merge this when we are almost on verge of merging #147. 

To test this I have updated my `master` branch here is the ci/cd result -> https://github.com/technophile-04/se-2/actions/runs/4226777193/jobs/7340581893


carletex:
Merging this since #202 is going to be merged soon.

PR #192, opened 2023-02-20
fix(types): allow generically set return type for contract read hooks

### LINKED ISSUE #189: Make { data } for useScaffoldContractRead to be "any"

sverps:
Fixes #189 

I propose keeping it rather simple and clean in a first step. The user should know what the return type is (since he'll use it). He can choose to use the return value as `any`, or he can provide the type if he wants better type-safety.


carletex:
Love this @sverps 

This is exactly what we need.

Thanks!!

technophile-04:
Love this approach !!

PR #194, opened 2023-02-21
"yarn fork" command for mainnet forking

carletex:
As in SE1, we want to be able to run our local hardhat node with mainnet forking.

Don't feel super happy about putting the API Key on package.json, even if we already exposed it in `hardhat.config.js`. It's a default alchemy key we have for SE1.

It's always about finding the balance between: good practices & easy-to-use/no-config

Thoughts? Any other approaches?

image

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   "scripts": {
     "account": "yarn workspace @se-2/hardhat account",
     "chain": "yarn workspace @se-2/hardhat chain",
+    "fork": "yarn workspace @se-2/hardhat fork",
     "deploy": "yarn workspace @se-2/hardhat deploy",
     "compile": "yarn workspace @se-2/hardhat compile",
     "generate": "yarn workspace @se-2/hardhat generate",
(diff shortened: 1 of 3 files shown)


rin-st (approved):

rin-st:
we can add another `.env` variable, like `FORKING_ENABLED` and use it in forking [config](https://hardhat.org/hardhat-network/docs/reference#forking) like
```
networks: {
  hardhat: {
    forking: {
      url: `https://eth-mainnet.alchemyapi.io/v2/${providerApiKey}`,
      enabled: process.env.FORKING_ENABLED === 'true'
    }
  }
}
```

and script will be much simpler too, something like
`FORKING_ENABLED=true hardhat node --network hardhat --no-deploy`

technophile-04:
Love @rin-st suggestions, since if the developer configures `ALCHEMY_API_KEY` we will be using his api key...other wise he needs to update in package.json too

carletex:
Amazing @rin-st !! Thanks for finding that enabled prop.

Fixed in c1efe45aba85e53a23808d798d2702d6830995db

Issue #141, opened 2023-02-09
TS / Lint-staged for begginers


carletex:
We have a nice lint-staged / precommit workflow (check linting + TS types) + GH Action before merging to `main` that is pretty handy for us.

BUT, we have to keep in mind that this is a starter kit and some of the people that will use SE2 will be beginners, and fighting with TS types is daunting. 

So we need to find a way to disable this for people who choose to do so: Maybe it's just something in the README (instructions on how to remove/disable it). It could also be disabled by default.... and you have to enable (from an .env file?).

Opening the discussion :D

carletex:
Another cool idea:

We could try having a message on top of this (when committing)

![image](https://user-images.githubusercontent.com/2486142/217774666-57151c5c-6ff4-4ee4-8daa-823bb1bb0f90.png)

Like: "If you want to remove this checks, please remove ........."

rin-st:
Since it's TS project, beginners will need to fight with TS anyway, so they can just use `any` for types. If we disable lint-staged, the app can break in some places, and it will be hard to find where and why.

I believe beginners will use mostly our components, and we have to create good docs/comments there to understand how it works. If they'll create their own components, they should know React/TS 🤷‍♂️ . Regarding things like `BigNumber` we also should add info how to work with it. Disabling types more likely will add errors

Also, as I understand GH actions work for PRs to master, but our main branch is `main` so it's disabled. Am I right?

technophile-04:
> Since it's TS project, beginners will need to fight with TS anyway, so they can just use any for types. If we disable lint-staged, the app can break in some places, and it will be hard to find where and why.

💯💯 agree and make complete sense let's not disable it by default instead I think it would helpfull if we mention it somewhere on how to diable it because while hacking in hackathon for 1 day with other teammates sometimes it might get annoying for begginers in typescript hacking with this repo 🙌

I feel the best place will be where Carlos mentioned in his 2nd comment https://github.com/scaffold-eth/se-2/issues/141#issuecomment-1423894870. 

Some command to bypass pre-commit hook : 
1. git commit -m -n "commit message"  - `-n` flag is shorthand for `--no-verify`
2. Comment out `yarn lint-staged --verbose` in `.husky/pre-commit` file

carletex:
We just merged #195

I think we just need to make another PR to update the README with (maybe in a whole new section? We can ask for help to @edakturk14, when we are clear about it):
- Tell people about `yarn vercel:yolo`
[...]

PR #195, opened 2023-02-21
add option to ignore linting and types error during deploy

technophile-04:
#### Description
Had a discussion with @carletex about it... sometimes beginners might have a hard time figuring out the correct types / linting errors and it would be handy if we have an option that lets you yolo app quickly

By default its `false` but if someone want to share a quick demo/poc deployed `url` it will be nice. 

#### Approach take : 
Added a flag to `.env.production` which by default is `false` and to ignore errors you run `yarn vercel:yolo` command which sets this flag to `true` or manually set  the flag to `true` and run `yarn vercel`. 

#### Demo : 
Screenshot 2023-02-21 at 3 33 21 PM

^ Here is the SS, vercel skipping both types and linting


carletex:
We've created a monster. YOLO!! haha

Uncle bob, don't judge us.

We don't need to forget that we are building a tool that (if everything goes well) a lot of people will use to onboard on Ethereum, and some of them won't have TS knowledge, so we want to make their life easier at the beginning, and not let them get frustrated. It's about to find Balance!

This is related to #141 (fixes some of the issues).

packages/nextjs/next.config.js
```diff
@@ -3,6 +3,12 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
+  typescript: {
+    ignoreBuildErrors: process.env.NEXT_PUBLIC_IGNORE_BUILD_ERROR === "true" ? true : false,
```
rin-st, line 7: you can do it without ternary :) `ignoreBuildErrors: process.env.NEXT_PUBLIC_IGNORE_BUILD_ERROR === "true"`

packages/nextjs/next.config.js
```diff
@@ -3,6 +3,12 @@
...
+  },
+  eslint: {
+    ignoreDuringBuilds: process.env.NEXT_PUBLIC_IGNORE_BUILD_ERROR === "true" ? true : false,
```
rin-st, line 10: and here

packages/nextjs/next.config.js
technophile-04: Trust me I actually spent 4-5 min trying to figure out the best way to write this and this didn't clicked 😂😂, Thanks @rin-st looks better fixed at #197

PR #196, opened 2023-02-21
"yarn verify": Etherscan verify

carletex:
As in SE1, we want to be able to verify our deployed contracts.

`yarn verify --network <>`

Same dilemma as in #194 (good practices & easy-to-use/no-config


packages/hardhat/hardhat.config.ts
```diff
@@ -10,8 +10,11 @@ const providerApiKey = process.env.ALCHEMY_API_KEY ?? "oKxs-03sij-U_N0iOlrSsZFr2
 // If not set, it uses the hardhat account 0 private key.
 const deployerPrivateKey =
   process.env.DEPLOYER_PRIVATE_KEY ?? "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
+// If not set, it uses ours Etherscan default API key.
+const etherscanApiKey = process.env.ETHERSCAN_API_KEY ?? "DNXJA8RX2Q3VZ4URQIWP7Z68CJXQZSC6AW";
```
technophile-04, line 14: Instead of using `??` maybe we should use `ternary` because I see people renaming `.env.example` -> `.env` and just fill in `DEPLOYER_PRIVATE_KEY` something like this : 
```shell
# This is a template for the environment variables
# To use this template, copy this file, rename it .env, and fill in the values
ALCHEMY_API_KEY=
DEPLOYER_PRIVATE_KEY=abfdsfsdfsafsafsadfsafsdf
ETHERSCAN_API_KEY=
``` 
and since we have used `??` the final `const etherscanApiKey` will be `empty string` instead of our default key and this might result in extra debugging for the developer

rin-st: or use `||`

carletex: Thanks! 0b37578a8e56f280b9a9bb30493e7feb3afc1f57

damianmarti:
Love all the libraries updated at SE-2!!! I spent hard times trying to verify some contracts with the oldie scaffold-eth that use etherscan-verify (for example, it doesn't support the libraries param, so I had to use hardhat verify). 

I really appreciate all the hard work and all the improvements you all are doing at SE-2!!!

PR #197, opened 2023-02-21
use direct === for ignoreError instead of ternary

technophile-04:
### Description 
This feels lot better and minimal 

Originally suggested here -> https://github.com/scaffold-eth/se-2/pull/195#discussion_r1112993857


rin-st (approved):

carletex:
<3

Issue #116, opened 2023-01-27
Contracts type Inference / autocompletions (& custom hooks vs wagmi hooks)


technophile-04:
Currently, we are trying to create some utilities/hooks  as Carlos mentioned in #88 : 
> When creating your app in SE, it's super handy to have hooks that loads the deployed contracts (hardhat_contracts / external_contracts).
In SE-2 we are using wagmi, but I think we could benefit of having some layer (hooks / utils) on top of wagmi to load these deployed contracts.

A basic iteration of `useScaffoldContractRead` and `useScaffoldContractWrite` has been created at #93 but it's very loosely typed....The reason for creating this utility is if you use wagmi's `useContractRead/Write` you always need to pass in `abi` and `address` but we can handle it internally. 

What we want to achieve is something like this -> https://twitter.com/wagmi_sh/status/1580591788611338240 with the above hooks. 

We can leverage typescript generics as mentioned in this comment https://github.com/scaffold-eth/se-2/pull/93#issuecomment-1383935227 and even [`useContractRead`](https://github.com/wagmi-dev/wagmi/blob/6f7dacfe5e09e01eb8c2c73ed33db9ce2e2f1223/packages/react/src/hooks/contracts/useContractRead.ts#L75) accepts generics. 

Some resources / usefull lookups : 
1) Problem faced while implementing it : https://github.com/scaffold-eth/se-2/pull/93#issue-1534650952
2) https://github.com/scaffold-eth/se-2/pull/93#issuecomment-1403947795
3) https://github.com/scaffold-eth/se-2/pull/93#issuecomment-1403996177

carletex:
Not sure if useful, but there is a `typechain-types` folder generated inside the `hardhat` package when you compile/deploy a contract.

technophile-04:
I have been exploring different ways to conquer this and here is my research : 
- The way wagmi handles the autocompletes is using their `ABIType` packagage and they recommend to use `abi as const` or inline the abi checkout -> https://wagmi.sh/react/typescript#type-inference

I think the approach taken by @rin-st in #147  seems correct to me, although the developer needs to pass in the `abi` manually to `useScaffoldContractRead`, I feel this is the only way to get autocompletion for `functionName` as shown by Rinat in https://github.com/scaffold-eth/se-2/pull/147#issuecomment-1424839258

[...]

PR #200, opened 2023-02-21
Implement wagmi-cli

technophile-04:
### Description : 
We can use [@wagmi-cli](https://wagmi.sh/cli/getting-started) to generate react hooks based on ABI's and it also some different plugins which can help us in future. 

### To use [feat/wagmi-cli](https://github.com/scaffold-eth/se-2/tree/feat/wagmi-cli) branch :
1. checkout to branch  [`feat/wagmi-cli`](https://github.com/scaffold-eth/se-2/tree/feat/wagmi-cli)
2. run `yarn install`
3. `yarn chain`
4. `yarn deploy`
5. `yarn wagmi generate` to generate hooks inside `packages/nextjs/generated`
6. `yarn start` and checkout http://localhost:3000/example-ui

Original discussion #116

### DIFF
--- a/packages/hardhat/package.json
+++ b/packages/hardhat/package.json
@@ -29,6 +29,7 @@
     "@types/qrcode": "^1",
     "@typescript-eslint/eslint-plugin": "latest",
     "@typescript-eslint/parser": "latest",
+    "@wagmi/cli": "^0.1.9",
     "chai": "^4.3.6",
     "eslint": "^8.26.0",
     "eslint-config-prettier": "^8.5.0",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
     "postinstall": "husky install",
     "precommit": "lint-staged",
     "vercel": "yarn workspace @se-2/nextjs vercel",
-    "vercel:yolo": "yarn workspace @se-2/nextjs vercel:yolo"
+    "vercel:yolo": "yarn workspace @se-2/nextjs vercel:yolo",
+    "wagmi": "yarn workspace @se-2/hardhat wagmi"
   },
   "packageManager": "yarn@3.2.3",
   "devDependencies": {
--- a/packages/hardhat/wagmi.config.ts
+++ b/packages/hardhat/wagmi.config.ts
@@ -1,0 +1,13 @@
+import { defineConfig } from "@wagmi/cli";
+import { hardhat, react } from "@wagmi/cli/plugins";
+
+export default defineConfig({
+  out: "../nextjs/generated/contractHooks.ts",
+  contracts: [],
+  plugins: [
+    hardhat({
+      project: "./",
+    }),
+    react(),
+  ],
+});
--- a/packages/nextjs/components/ExampleUi/ContractData.tsx
+++ b/packages/nextjs/components/ExampleUi/ContractData.tsx
@@ -22,10 +22,12 @@ export default function ContractData() {
 
   const { data: totalCounter } = useYourContractTotalCounter({
     address: deployedContract?.address,
+    watch: true,
   });
 
   const { data: currentGreeting, isLoading: isGreetingLoading } = useYourContractGreeting({
     address: deployedContract?.address,
+    watch: true,
   });
 
   useYourContractGreetingChangeEvent({
--- a/packages/nextjs/components/ExampleUi/ContractData.tsx
+++ b/packages/nextjs/components/ExampleUi/ContractData.tsx
@@ -1,8 +1,12 @@
 import { useEffect, useRef, useState } from "react";
 import Marquee from "react-fast-marquee";
-import { useScaffoldContractRead, useScaffoldEventSubscriber } from "~~/hooks/scaffold-eth";
-import { BigNumber } from "ethers";
+import { useDeployedContractInfo } from "~~/hooks/scaffold-eth";
 import { useAnimationConfig } from "~~/hooks/scaffold-eth/useAnimationConfig";
+import {
+  useYourContractGreeting,
+  useYourContractGreetingChangeEvent,
+  useYourContractTotalCounter,
+} from "~~/generated/contractHooks";
 
 const MARQUEE_PERIOD_IN_SEC = 5;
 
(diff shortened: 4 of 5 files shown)


technophile-04:
### Update - 
Added a script that generates `wagmi.config.ts` file when you run `yarn wagmi generate`. 

Now you no more need to pass an address every time you use hooks generated from wagmi. 

<table>
<tr>
<td> Before </td> <td> After </td>
</tr>
<tr>
<td>

```js
 const { data: deployedContract } = useDeployedContractInfo("YourContract");
  const { data: totalCounter } = useYourContractTotalCounter({
    address: deployedContract?.address,
  });
```

</td>
<td>
    
```js
 const { data: totalCounter } = useYourContractTotalCounter({
    watch: true,
  });
```
</td>
</tr>
</table>

rin-st:
@technophile-04 Looks awesome!

packages/hardhat/package.json
```diff
@@ -11,7 +11,8 @@
     "lint": "eslint --config ./.eslintrc.json --ignore-path ./.eslintignore ./*.ts ./deploy/**/*.ts ./scripts/**/*.ts ./test/**/*.ts",
     "lint-staged": "eslint --config ./.eslintrc.json --ignore-path ./.eslintignore",
     "test": "REPORT_GAS=true hardhat test --network hardhat",
-    "verify": "hardhat etherscan-verify"
+    "verify": "hardhat etherscan-verify",
+    "wagmi" : "hardhat run scripts/generateWagmiConfig.ts && wagmi"
```
rin-st, line 15: maybe create `wagmi-generate` script?

rin-st: You can also add `prettier --write wagmi.config.ts` after running script, so the config will look better

packages/hardhat/scripts/generateWagmiConfig.ts
```diff
@@ -0,0 +1,64 @@
...
+    `import { defineConfig } from "@wagmi/cli";
+      import { hardhat, react } from "@wagmi/cli/plugins";
+      
+      export default defineConfig({
+        out: "../nextjs/generated/contractHooks.ts",
+        contracts: [],
+        plugins: [
+          hardhat({
+            project: "./",
+            deployments: ${util.inspect(result)}
```
rin-st, line 51: why utils, not JSON.stringify?

packages/hardhat/package.json
technophile-04: Tysm @rin-st, always love your suggestions 🙌

updated at [ae4b077](https://github.com/scaffold-eth/se-2/pull/200/commits/ae4b077b03f74d702365c3003f81642fec4e4072)

packages/hardhat/scripts/generateWagmiConfig.ts
technophile-04: Thanks actually on initially I had issue with JOSN.stringfy so was using `uits` and  forgot to change it 

update at [eb16cbd](https://github.com/scaffold-eth/se-2/pull/200/commits/eb16cbd540f3b9f814118b32b9220c4c32e09cb2)

packages/hardhat/scripts/generateWagmiConfig.ts
```diff
@@ -0,0 +1,63 @@
...
+    `import { defineConfig } from "@wagmi/cli";
+      import { hardhat, react } from "@wagmi/cli/plugins";
+      
+      export default defineConfig({
+        out: "../nextjs/generated/contractHooks.ts",
+        contracts: [],
+        plugins: [
+          hardhat({
+            project: "./",
+            deployments: ${JSON.stringify(result, null, 2)}
```
roninjin10, line 50: This config generation feels unnecessary.   Could it just import the deployment artifacts directly in the wagmi config?

technophile-04: Good catch, Thanks !! Update it at [ba65263](https://github.com/scaffold-eth/se-2/pull/200/commits/ba65263479ddb4ceece10fd6e9c79f9675afac42). 

Although had to use the `getParsedDeploymentsForWagmiConfig` function to generate [`deployments`](https://wagmi.sh/cli/plugins/hardhat#deployments-optional) options since we don't want user to pass in `contractAddress` to generated `hooks` and it should be infered internally checout -> https://github.com/scaffold-eth/se-2/pull/200#issuecomment-1440507844

carletex:
Should we close this @technophile-04 ? And maybe keep the branch.

technophile-04:
>  Should we close this @technophile-04 ? And maybe keep the branch.

Yup, we can close this for now, 

I was planning to quest on this branch for better support of external contracts : 

### Initial Iteration : ( `yarn generateMyAbis` ): 

This command will help you generate/fetch Abi's from etherscan or custom abi's and put it inside `external_contracts.ts` / `external_abi.ts` appended with `as const` so that you get **`wagmi's`** autocompletion. 

```ts
// wagmi.config.ts

export default defineConfig({
  plugins: [
    etherscan({
      apiKey: process.env.ETHERSCAN_API_KEY,
      contracts: [
        {
          name: 'DAI',
          address: '0x6b175474e89094c44da98b954eedeac495271d0f',
        },
      ],
    }),
  ],
})
```

Because I feel this will really increase DX and you no more need to go to etherscan and copy abi's,  just put in contract address and it will generated abi's ready to use with `wagmi's` autocompletion. 

I think this will be pretty easy to accomplish. 

----

### Final Iteration : 

I know we can support external contract with our custom hooks but the process to do so is a bit hectic where you copy the abi and paste it inside `hardhat_contract.ts` also the data structure of how we are storing `abi's` is bit hard to grasp and people might get confused where to put external contract ABI. 

I think it would be really great if we have a single source of truth through which contract ABIs are generated also people will get an option to generate ABI's from etherscan, sourcify , custom ABI, etc. 

And you can utilize all these generated ABIs from our `custom hooks` with auto-completion. 


--- 

Just thinking out loud and this may be one of the approaches to how we support external contracts with our custom hooks. 

I also think the current `main branch` approach  is also not too bad for supporting external contracts with custom hooks if we have reference/docs/ mentioned in README.md on how to do so  🙌

PR #202, opened 2023-02-22
fix(types): better types and autocomplete for custom contract hooks

### LINKED ISSUE #116: Contracts type Inference / autocompletions (& custom hooks vs wagmi hooks)

sverps:
Fixes #116

# Implemented types throughout all the hooks.

## useScaffoldEventSubscriber

autocompletes contractName and eventName
infers correct type for event inputs in the callback

## useScaffoldContractWrite

autocompletes contractName and function name (only allows write functions)

## useScaffoldContractRead

autocompletes contractName and function name (only allows read functions)
proper return types

## useDeployedContractInfo

autocompletes contractName
Returns typed abi and address

# TODO

Source of truth for chain id

@carletex @rin-st @technophile-04


rin-st:
Hey! Is this pr updates https://github.com/scaffold-eth/se-2/pull/147? Why didn't you add last changes there?

sverps:
> Hey! Is this pr updates #147? Why didn't you add last changes there?

Yes, this builds off of those changes, but when I created this PR, I wasn't added to the repo yet, so couldn't push to that other branch.

packages/nextjs/hooks/scaffold-eth/contract.types.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export type Chain = keyof typeof contracts;
+
+type Contracts = typeof contracts[DefaultChain][0]["contracts"];
+
+export type ContractName = keyof Contracts;
+
+export type Contract<TContractName extends ContractName> = Contracts[TContractName];
+
+type InferContractAbi<TContract> = TContract extends { abi: infer O } ? O : never;
```
rin-st, line 14: I tried to understand why is zero here. What is it does? Probably it's better to change that `O` type

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
```diff
@@ -1,52 +1,52 @@
...
  * @dev use this hook to get a deployed contract from `yarn deploy` generated files.
  * @param contractName - name of deployed contract
  * @returns {GeneratedContractType | undefined} object containing contract address and abi or undefined if contract is not found
  */
-export const useDeployedContractInfo = (contractName: string | undefined | null) => {
+export const useDeployedContractInfo = <TContractName extends ContractName>(contractName: TContractName) => {
   const configuredChain = getTargetNetwork();
-  const [deployedContractData, setDeployedContractData] = useState<undefined | GeneratedContractType>(undefined);
-  const [isLoading, setIsLoading] = useState(true);
+  const deployedContract = useMemo(
```
rin-st, line 14: it seems you don't need useMemo here

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
```diff
@@ -1,52 +1,52 @@
...
-        if (code === "0x" || !contractsData || !(contractName in contractsData)) {
-          return;
-        }
-        setDeployedContractData(contractsData[contractName]);
-      } catch (e) {
-        // Contract not deployed or file doesn't exist.
-        setDeployedContractData(undefined);
-      } finally {
-        setIsLoading(false);
+      if (!mounted) {
```
rin-st, line 30: not sure how mounted can be `false` here

sverps: If the hook is used in a component that gets mounted and unmounted before the async call resolves (e.g. user navigates to another page).

The pattern with the `mounted` is often used when a `useEffect` needs to call async functionality to avoid doing state changes on unmounted components.

In short, it's not always technically necessary, but it's good practice, and it ensures the console doesn't get random warnings like the following:

> Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
sverps: I think you're right. Since it's simply accessing an object in a const, it's reference shouldn't change between renders. Well spotted :+1:

packages/nextjs/hooks/scaffold-eth/contract.types.ts
sverps: It's just a generic type that infers it's type from the abi. I'll improve the name a bit.

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
rin-st: I understood you did that because of that warning, but I didn't meet that pattern before. At least I don't remember. Nice one, thanks :)

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
```diff
@@ -1,52 +1,52 @@
...
+        return;
+      }
+      // If contract code is `0x` => no contract deployed on that address
+      if (code === "0x" || !deployedContract) {
```
technophile-04, line 30: I think `!deployedContract` condition should be put before `line 28` because it error's out when if people forgot to do `yarn deploy` I think we discussed on other PR that app should not break if user forgots it not able to find it right now 😅. 

Also I think maybe can update this file in a different PR 🤔? Since it has been used at different places we need to test it properly which increases the overhead of testing other stuff of this PR, 

btw I love the changes but need to test it properly, Thanks 🙌

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
carletex: Didn't know about the pattern either! I'll try not to overuse it now haha

https://stackoverflow.com/a/60907638

technophile-04: PS : we have great util hook alerady for this instead of implementing on our own -> https://usehooks-ts.com/react-hook/use-is-mounted

sverps: Ah cool, for consistency, I guess it's cleaner to use the hook. Reduces boilerplate :)

I'll update it

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
sverps: You're right about the check needing to happen before :+1: 

As for other places that use this hook, they've all been checked and updated in this PR (since the other hooks need this as well). Not sure if splitting it in a different PR is needed, since these changes require testing all the same things anyway, I believe.

technophile-04: Yup makes sense 🙌

carletex:
@sverps Could you fix conflicts here?

Will love to start testing, and I think #231 will be ready soon.

Thanks!

carletex:
More conflicts, sorry :/

- [ ] Conflicts
- [ ] Update `export type DefaultChain = "31337";` with the `scaffoldConfig.targetNetwork.id` value.
- [ ] I think we also need https://github.com/scaffold-eth/se-2/pull/191, right?

Anything else?

I'll do a full review when ready!

Thanks

sverps:
Making progress on this, but found a small issue that is a bit tricky to resolve. Currently, typescript doesn't show any error when you pick a functionName that needs arguments in the read or write hook, but don't provide them.

```typescript
useScaffoldContractWrite("YourContract", "setGreeting"); // Doesn't show TS error
```

 I think it would be nice if typescript tells the user that he's missing the arguments.

This became a bit tricky to resolve (especially due to the optional config in the read hook and the optional value param in the write hook, and the missing optional config for the write hook).

Here are a couple of options on how to proceed:

[fix/116-type-inference-option-1](https://github.com/sverps/se-2/tree/fix/116-type-inference-option-1)
- Continues on the API style that we had for the hooks (i.e. first param is the contract name, second is the function name, ...)
- Requires some additional logic to resolve the optional arguments to their proper value. When in the read hook, no arguments were required, but you wanted to provide a config param, you needed to pass an empty array for args to have it work properly.
- Typescript errors are indicating what is wrong, but the actual errors are a bit cryptic.

[fix/116-type-inference-option-2](https://github.com/sverps/se-2/tree/fix/116-type-inference-option-2)
- Clear typescript errors that indicate in a more understandable way what is missing or wrong
- API is closer to wagmi API, which makes a possible transition to using wagmi hooks directly a bit lower.
- Need to pass in an object into the hooks to make their implementation and type cleaner

Option 3: Make function arguments always required, with the downside of often having to pass an empty array (especially for read functions).

So please check out those two branches and look at the `ContractData.tsx` file. There you can see how the typescript errors are shown. There you'll get a better understanding of the pros and cons of both options.

carletex:
Will review it later today. Just wanted to un-draft it to see if the GH Actions worked. It did! :D

packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
```diff
@@ -10,23 +11,30 @@ import scaffoldConfig from "~~/scaffold.config";
...
+    address: deployedContract?.address,
+    abi: deployedContract?.abi,
     watch: true,
-    args,
+    args: args as unknown[],
     ...readConfig,
-  }) as Omit<ReturnType<typeof useContractRead>, "data"> & {
-    data: TReturn;
+  }) as Omit<ReturnType<typeof useContractRead>, "data" | "refetch"> & {
+    data: AbiFunctionReturnType<ContractAbi, TFunctionName>;
```
technophile-04, line 34: I think we should also union  it with undefined : 
```ts
data: AbiFunctionReturnType<ContractAbi, TFunctionName> | undefined;
```

Reason : 
It will be `undefined` initially and may give false information to the developer in compile time and he will get a runtime error if he won't add check for `undefined`. 

Even wagmi has it : 
Screenshot 2023-03-20 at 8 14 19 PM


Sorry, I know its not right time to review it 😅 since it's still in the `draft` stage. But just noticed while playing with hooks and wanted to mention it so that I don't forget it 🙌

packages/nextjs/pages/debug.tsx
```diff
@@ -1,17 +1,14 @@
...
-  const contractNames = useDeployedContractNames();
-  const [selectedContract, setSelectedContract] = useState<string>();
-
-  useEffect(() => {
-    if (!selectedContract && contractNames.length) {
-      setSelectedContract(contractNames[0]);
-    }
-  }, [contractNames, selectedContract]);
+  const contractsData = contracts[scaffoldConfig.targetNetwork.id as unknown as keyof typeof contracts]?.[0]?.contracts;
+  const contractNames = contractsData ? (Object.keys(contractsData) as ContractName[]) : [];
```
carletex, line 10: (Nitpick alert :D)

This is a bit rough to read. I see 3 options:

1. Just a comment: like `// Gets all the deployed contract in the target network....`
2. Have a util (above `Debug` not exported): like `getDeployedContractsInTargetNetwork()`
3. Same as 2, but in a util file, exported.

I'd personally go with 1 or 2.

carletex:
This is looking really good. Great job @sverps !! Great reviews & hindsight @technophile-04 @rin-st <3

I feel that what Shiv is suggesting is the way to go if all the type hinting / auto-completions / errors work. Basically the same as using wagmi hooks but with some extra magic :)

packages/nextjs/pages/debug.tsx
sverps: Yeah, agreed. I'll move it to a util file and clean it up a bit :+1:

sverps:
Thanks for the suggestion @technophile-04, I like this option even more :+1: 

One thing I'm still a bit unsure about is whether or not the `args` should be strongly typed, or allow for a tupple of the needed type and `undefined` (what I did in the last commit). The reason being that a user might use this hook in combination with some form, so the initial values of the form might not be valid `args` values. Will need to play a bit more with it to better understand the implications.

Maybe that's the reason why `wagmi` makes the `args` optional, in order for the dev to code it in a way that he can prepare the args tupple before providing it...

technophile-04:
Even I am confused a bit, although I tend to like the current state where we are doing union with `undefined` for each index inside a tuple because I think it makes hooks easy to use and also will increase DX for beginners too. 

But would love to hear @carletex and @rin-st thoughts on this, I have also created `test` branch -> https://github.com/scaffold-eth/se-2/tree/fix/116-type-inference-test where I have put `wagmi's` `useContractRead` and `useContractWrite` below our `useScaffoldRead/write` to compare for Errors / DX and find out edge cases if we are missing in our hooks compared to `wagmi's` native hooks

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
```diff
@@ -1,54 +1,47 @@
...
-type GeneratedContractType = {
-  address: string;
-  abi: any[];
-};
+import scaffoldConfig from "~~/scaffold.config";
 
 /**
  * @dev use this hook to get a deployed contract from `yarn deploy` generated files.
  * @param contractName - name of deployed contract
  * @returns {GeneratedContractType | undefined} object containing contract address and abi or undefined if contract is not found
```
carletex, line 11: This `GeneratedContractType` is not there anymore. Should we just remove the return here?

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
```diff
@@ -1,54 +1,47 @@
 import { useEffect, useState } from "react";
+import contracts from "../../generated/hardhat_contracts";
```
carletex: Use ~~

a4d41cc2b7d41133affa576e3da7d3c10637a2a1

carletex:
Just read your last comment @technophile-04 , sorry!

> Even I am confused a bit, although I tend to like the current state where we are doing union with undefined for each index inside a tuple because I think it makes hooks easy to use and also will increase DX for beginners too.

I agree!

> But would love to hear @carletex and @rin-st thoughts on this, I have also created test branch -> https://github.com/scaffold-eth/se-2/tree/fix/116-type-inference-test where I have put wagmi's useContractRead and useContractWrite below our useScaffoldRead/write to compare for Errors / DX and find out edge cases if we are missing in our hooks compared to wagmi's native hooks

Thanks!!! I'll take a look. 

Stil going to merge, so we can start testing. I have a couple of builds lined up and can't wait to use the new hooks :) Using them would be the best way to see if need to make some tweaks.

PR #204, opened 2023-02-23
wrong network display + switch

KcPele:
Screenshot from 2023-02-23 11-16-14
When it is connected to other networks

Screenshot from 2023-02-23 11-16-41

when it is connected to mainnet

### DIFF
--- a/packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
+++ b/packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
@@ -2,7 +2,6 @@ import { ConnectButton } from "@rainbow-me/rainbowkit";
 import { ChevronDownIcon } from "@heroicons/react/24/solid";
 import { TAutoConnect, useAutoConnect, useNetworkColor } from "~~/hooks/scaffold-eth";
 import Balance from "~~/components/scaffold-eth/Balance";
-
 import { BlockieAvatar } from "~~/components/scaffold-eth";
 import { getTargetNetwork } from "~~/utils/scaffold-eth";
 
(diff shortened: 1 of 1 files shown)


packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
```diff
@@ -43,15 +38,21 @@ export default function RainbowKitCustomConnectButton() {
...
-                return (
-                  <div className="rounded-md shadow-lg p-2">
-                    <span className="text-error mr-2">Wrong network selected - ({chain.name})</span>
-                    <span className="text-primary mr-2">Switch network to</span>
-                    <button className="btn btn-xs btn-primary btn-outline" onClick={onSwitchNetwork}>
-                      {configuredChain.name}
-                    </button>
-                  </div>
-                );
+                if (chain.id === 1) {
```
carletex, line 41: Why are you adding this check? @KcPele

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
```diff
@@ -2,7 +2,7 @@ import { ConnectButton } from "@rainbow-me/rainbowkit";
 import { ChevronDownIcon } from "@heroicons/react/24/solid";
 import { TAutoConnect, useAutoConnect, useNetworkColor } from "~~/hooks/scaffold-eth";
 import Balance from "~~/components/scaffold-eth/Balance";
-import { useSwitchNetwork } from "wagmi";
+
```
carletex, line 5: Remove this line :)

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
```diff
@@ -43,15 +38,21 @@ export default function RainbowKitCustomConnectButton() {
...
-                    <button className="btn btn-xs btn-primary btn-outline" onClick={onSwitchNetwork}>
-                      {configuredChain.name}
-                    </button>
-                  </div>
-                );
+                if (chain.id === 1) {
+                  return (
+                    <div className="rounded-md shadow-lg p-2">
+                      <button
+                        className="text-white flex items-center gap-1 rounded-[10px] bg-red-500 px-4 py-2"
```
carletex, line 45: The button has a weird border around it.

image

carletex: Also we should use daisy colors defined on `tailwind.config.js` and the `btn` classes that we are using in other buttons

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
KcPele: its because when it's on mainnet it does not show the wrong network button. But it works fine on any other network. i can take it off and it will still work fine

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
KcPele: ok i will take that off

carletex:
@KcPele thanks!

I pushed a fix with the things I was mentioning on the code review.

This is still a bit confusing Mainnet Ethereum is enabled in wagmi because of the ENS / ETH price mainnet resolution, but the app won't allow you to use it.

image

I added the target network next to the button, so hopefully is more clear:

image

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
KcPele: which class contains the specific color. i have checked btn-error and btn-warning, the colors are different

PR #206, opened 2023-02-23
fix(form): improved conversion buttons

### LINKED ISSUE #203: Improve input form conversions
The following issues were encountered:
- Converting uint to ether throws an error when the form is empty.
- Converting uint to ether throws error when pressing the asterix multiple times
- Conversion button hitboxes are too small (see also #170 )

image

sverps:
Fixes #203


carletex:
Looks good @sverps Thanks!

I also centered it vertically on 4a5c2147cdb025c2e5005fe4358cee6a3945d030

b
image
a
image

PR #208, opened 2023-02-23
Add section to disable types/error checks in README

technophile-04:
#### Description 
Created a basic iteration of Disabling types/error checks section 

for `Disabling Github Workflow` part I suggested the removal of `.github` directory but maybe can suggest two ways : 
1.) Simple way : delete `.github` 2) Disable workflow only for push: which suggest modifying of `lint.yaml` file. 

Currently, the `warning` seems a bit harsh, maybe remove it or change it? also, we can have a different titles for the section 

Feel free to suggest any changes be it revamping of the whole thing again 🙌

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,24 +95,31 @@ If you want to redeploy to the same production URL you can run `yarn vercel --pr
 **Hint**: We recommend connecting the project GitHub repo to Vercel so you the gets automatically deployed when pushing to `main`
 
 ---
-## ⚠️ Disabling type/error checks 
-> **Warning**
-> We don't encourage to use the below options
+## ⚠️ Disabling type & linting error checks 
+> **Hint**
+> Typescript helps you catch errors at compile time, which can save time and improve code quality, but can be challenging for those who are new to the language or who are used to the more dynamic nature of JavaScript. Below are the steps to disable type & lint check at different levels  
 ### Disabling commit checks
-Move to `.husky/pre-commit` file and comment out `yarn lint-staged --verbose`
+We run `pre-commit` [git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) which lints the staged files and don't let you commit if there is an linting error. 
+
+To disable this, got to `.husky/pre-commit` file and comment out `yarn lint-staged --verbose`
+
 ```diff
 - yarn lint-staged --verbose 
 + # yarn lint-staged --verbose
 ```
 
 ### Deploying to Vercel without any checks 
-To ignore all types / error checks while deploying, use : 
+Vercel by default runs types and lint checks while developing `build` and deployment fails if there is an types or lint error. 
+
+To ignore types and lint error checks while deploying, use : 
 ```shell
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -94,6 +94,27 @@ If you want to redeploy to the same production URL you can run `yarn vercel --pr
 
 **Hint**: We recommend connecting the project GitHub repo to Vercel so you the gets automatically deployed when pushing to `main`
 
+---
+## ⚠️ Disabling type/error checks
```
carletex, line 98: Maybe something like "Disabling types and linting errors checks"?

README.md
```diff
@@ -94,6 +94,27 @@ If you want to redeploy to the same production URL you can run `yarn vercel --pr
 
 **Hint**: We recommend connecting the project GitHub repo to Vercel so you the gets automatically deployed when pushing to `main`
 
+---
+## ⚠️ Disabling type/error checks 
+> **Warning**
+> We don't encourage to use the below options
```
carletex, line 100: Instead of this, we might want to be like:

"Typescript is great blablabla..... but is can be daunting for newcomers..... so here we show you how you can disable.....blabla... until you get familiar with it.

README.md
```diff
@@ -94,6 +94,27 @@ If you want to redeploy to the same production URL you can run `yarn vercel --pr
...
+### Disabling commit checks
+Move to `.husky/pre-commit` file and comment out `yarn lint-staged --verbose`
```
carletex, line 102: Explain a bit what commit checks are:

"when you commit it'll automatically run checks... blabla... and it will fail if it find any type/linting error. You can disabled it by:"

README.md
```diff
@@ -94,6 +94,27 @@ If you want to redeploy to the same production URL you can run `yarn vercel --pr
...
+```diff
+- yarn lint-staged --verbose 
++ # yarn lint-staged --verbose
+```
+
+### Deploying to Vercel without any checks 
+To ignore all types / error checks while deploying, use :
```
carletex, line 109: Same as my previous comment:

"when you want to push your app to vercel, it'll automatically run checks... blabla... and the build will fail if it find any type/linting error. You can disabled it by:"

README.md
```diff
@@ -94,6 +94,27 @@ If you want to redeploy to the same production URL you can run `yarn vercel --pr
...
+```
+
+### Deploying to Vercel without any checks 
+To ignore all types / error checks while deploying, use : 
+```shell
+yarn vercel:yolo
+```
+
+### Disabling Github Workflow
+Delete `.github` directory
```
carletex, line 115: Same, a line about github workflow.

carletex (commented):
This is great @technophile-04 !!

I left some comments (I'm not the best writer... note the "blabla" haha) but I think we could include some of that stuff.

Feel free to update it, or we can also ask @edakturk14 for help (I already saw that you tagged her as a reviewer ;))

carletex:
I think this is perfect @technophile-04 Thank you!!

We can iterate later with Eda

PR #209, opened 2023-02-23
Add UI for displaying multiple contracts

### LINKED ISSUE #186: Multiple deployed contracts UI
I didn't find any discussion about handling multiple contracts. I think it could be nice to add some UI for it (that would only render when there actually are multiple contracts).

Should this be explored further and build something like this:
image

Or is it 'good enough' to let a developer figure it out on his own, and just add additional ContractUI components underneath like so:
```
<div className="flex justify-center">
  <div className="grid grid-cols-1 lg:grid-cols-6 py-12 px-10 lg:gap-12 w-full max-w-7xl my-0">
    <ContractUI contractName="YourContract" />
    <ContractUI contractName="OtherContract" />
  </div>
</div>
```

Which looks like this:
image

sverps:
Fixes #186

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
@@ -77,7 +77,7 @@ const ContractUI = ({ contractName, className = "" }: TContractUIProps) => {
           <div className="bg-base-100 border-base-300 border shadow-md shadow-secondary rounded-3xl px-6 lg:px-8 mb-6 space-y-1 py-4">
             <div className="flex">
               <div className="flex flex-col gap-1">
-                {contractName}
+                <span className="font-bold">{contractName}</span>
                 <Address address={contractAddress} />
                 <div className="flex gap-1 items-center">
                   <span className="font-bold text-sm">Balance:</span>
--- a/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
@@ -20,7 +20,7 @@ type TContractUIProps = {
 /**
  * UI component to interface with deployed contracts.
  **/
-const ContractUI = ({ contractName, className }: TContractUIProps) => {
+const ContractUI = ({ contractName, className = "" }: TContractUIProps) => {
   const configuredChain = getTargetNetwork();
   const provider = useProvider();
   const [refreshDisplayVariables, setRefreshDisplayVariables] = useState(false);
--- a/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
@@ -14,12 +14,13 @@ import Spinner from "~~/components/Spinner";
 
 type TContractUIProps = {
   contractName: string;
+  className?: string;
 };
 
 /**
  * UI component to interface with deployed contracts.
  **/
-const ContractUI = ({ contractName }: TContractUIProps) => {
+const ContractUI = ({ contractName, className }: TContractUIProps) => {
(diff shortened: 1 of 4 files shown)


packages/nextjs/tailwind.config.js
```diff
@@ -1,7 +1,29 @@
...
+          "scrollbar-width": "none",
+
+          /* Safari and Chrome */
+          "&::-webkit-scrollbar": {
+            display: "none",
+          },
+        },
+      });
+    }),
+  ],
```
carletex, line 26: In order to keep things minimal (something that we always try do on SE2), I'd remove this, since it add some noise to the file for lil gain :D

Debugging multiple contracts on mobile doesn't seem the the normal use case :D

We can just wrap the tabs

packages/nextjs/hooks/scaffold-eth/useDeployedContractNames.ts
```diff
@@ -0,0 +1,33 @@
+import { useEffect, useState } from "react";
+import { useProvider } from "wagmi";
+import { getTargetNetwork } from "~~/utils/scaffold-eth";
+
+import contracts from "~~/generated/hardhat_contracts.json";
```
carletex, line 5: Check `useDeployedContractInfo.ts` on how we are handling the import of this file.

I think we should avoid the NextJS "full-modal" error when people forget to `yarn deploy`. It happens on every page, not only on debug.

image

In some cases, people would also use SE2 and won't need a contract (like doing wallet signing to a backend).

It'll also avoid the github action error (although we might merge this at some point https://github.com/scaffold-eth/se-2/pull/191)

carletex (changes requested):
Looking good @sverps Thanks!

Did a quick review and left a couple of comments.

packages/nextjs/hooks/scaffold-eth/useDeployedContractNames.ts
sverps: Ah yes, good point :+1:

packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
```diff
@@ -66,12 +67,15 @@ const ContractUI = ({ contractName }: TContractUIProps) => {
   }
 
   return (
-    <div className="grid grid-cols-1 lg:grid-cols-6 py-8 lg:py-12 px-6 lg:px-10 lg:gap-12 w-full max-w-7xl my-0">
+    <div
+      className={`grid grid-cols-1 lg:grid-cols-6 py-8 lg:py-12 px-6 lg:px-10 lg:gap-12 w-full max-w-7xl my-0 ${className}`}
```
rin-st, line 71: `className` will be `undefined` without default value

packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
```diff
@@ -62,16 +63,21 @@ const ContractUI = ({ contractName }: TContractUIProps) => {
   }
 
   if (!contractAddress) {
-    return <p className="text-3xl mt-14">No Contract found!</p>;
+    return (
+      <p className="text-3xl mt-14">
+        {`No contract found by the name of "${contractName}" on chain "${configuredChain}"!`}
```
carletex: image

Need `${configuredChain.name}`

carletex (approved):
This looks great @sverps 

I pushed a fix for an error and a couple of style tweaks.

image

Thanks!

packages/nextjs/components/scaffold-eth/Contract/ContractUI.tsx
carletex: Fixed on https://github.com/scaffold-eth/se-2/pull/209/commits/846b727b631235638ceb314785d621cce19aa56d

PR #211, opened 2023-02-23
Argument "args" on scaffoldContractRead

carletex:
This is kind of a hotfix (until #147 #207 or #200 is merged). That's why I went with `any[]` :D 

On yesterday's hacking session, we realized we couldn't read contract data that require params.

e.g.

```
  const { data } = useScaffoldContractRead("YourContract", "userGreetingCounter", [
    "0xWHATEVER_ADDRESS",
  ]);
```

Technically you could pass it on `readConfig`, but I think this is more clear. Should we also add `writeConfig` on `scaffoldContractWrite`? (we can leave it for other PR)

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
@@ -8,11 +8,13 @@ import { BigNumber } from "ethers";
  * @dev wrapper for wagmi's useContractRead hook which loads in deployed contract contract abi, address automatically
  * @param contractName - deployed contract name
  * @param functionName - name of the function to be called
- * @param readConfig   - wagmi configurations
+ * @param args - args to be passed to the function call
+ * @param readConfig - extra wagmi configuration
  */
 export const useScaffoldContractRead = <TReturn extends BigNumber | string | boolean = any>(
   contractName: string,
   functionName: string,
+  args?: any[],
   readConfig?: Parameters<typeof useContractRead>[0],
 ) => {
   const configuredChain = getTargetNetwork();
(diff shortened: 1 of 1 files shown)


rin-st:
Have you tried 
```
const { data } = useScaffoldContractRead("YourContract", "userGreetingCounter", {
    args: ["0xWHATEVER_ADDRESS"]
});
```
?

carletex:
@rin-st Yep, that worked. See my PR comment:

> Technically you could pass it on readConfig, but I think this is more clear.

I was trying to be more explicit (and using the same signature as in useScaffoldContractWrite).

But if we think using the `readConfig` to override the call is clear enough, happy to close this!

technophile-04:
> Technically you could pass it on readConfig, but I think this is more clear. Should we also add writeConfig on scaffoldContractWrite? (we can leave it for other PR)

Lol yup love this pattern makes more intutive, +1 for `write` too !

rin-st:
> Yep, that worked. See my PR comment:

oops, sorry :)

I think if you're adding args here, you should remove it from config, so user can't add args here and in config

carletex:
> I think if you're adding args here, you should remove it from config, so user can't add args here and in config

Not sure what you mean. `readConfig` is meant to pass our hook some extra wagmi config props (to override the ones we've set or to add props that we are not using).

I don't think people do this:
```js
const { data } = useScaffoldContractRead("YourContract", "userGreetingCounter", ["0xWHATEVER_ADDRESS"], {
    args: ["some_other_stuff??"]
});
```

And if they do.. :unamused: haha

I guess you mean to remove `args` from `readConfig`?

rin-st:
> I guess you mean to remove args from readConfig?

yep

> And if they do.. 😒 haha

if we'll give them a possibility to do it, they'll do 😄 . Not critical, we can do it later. And they also can pass `functionName` to `readConfig` too

carletex:
> if we'll give them a possibility to do it, they'll do smile . Not critical, we can do it later. And they also can pass functionName to readConfig too

Yep, I see your point. As you said, not critical, so merging it for now.

Thanks!

PR #213, opened 2023-02-25
fix: allow only numerical input for uint256 field

### LINKED ISSUE #212: uint256 input should not throw errors with string input
Entering a text value and trying to parse it leads to following error:

image

~I propose to only allow numerical input when type is uint256~

sverps:
Fixes #212

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/InputUI.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/InputUI.tsx
@@ -47,7 +47,7 @@ const InputUI = ({ setForm, form, stateObjectKey, paramType }: TInputUIProps) =>
           }}
         />
       ) : (
-        <div className="flex items-end border-2 border-base-300 bg-base-200 rounded-full text-accent justify-between items-center">
+        <div className="flex items-center justify-between border-2 border-base-300 bg-base-200 rounded-full text-accent">
           <input
             placeholder={paramType.name ? paramType.type + " " + paramType.name : paramType.type}
             autoComplete="off"
--- a/packages/nextjs/components/scaffold-eth/Contract/InputUI.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/InputUI.tsx
@@ -1,7 +1,8 @@
 import { utils } from "ethers";
 import { FunctionFragment } from "ethers/lib/utils";
-import React, { Dispatch, ReactElement, SetStateAction } from "react";
+import React, { Dispatch, ReactElement, SetStateAction, useState } from "react";
 import { AddressInput } from "~~/components/scaffold-eth";
+import { NUMBER_REGEX } from "./utilsComponents";
 
 import { StringToBytesConverter, StringToBytes32Converter, UintToEtherConverter } from "./utilsDisplay";
 
@@ -17,6 +18,8 @@ type TInputUIProps = {
  * Generic Input component to handle input's based on their function param type
  */
 const InputUI = ({ setForm, form, stateObjectKey, paramType }: TInputUIProps) => {
+  const [inputError, setInputError] = useState(false);
+
   let inputSuffix: ReactElement = <></>;
 
   switch (paramType.type) {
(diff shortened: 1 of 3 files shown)


packages/nextjs/components/scaffold-eth/Contract/InputUI.tsx
```diff
@@ -60,6 +60,7 @@ const InputUI = ({ setForm, form, stateObjectKey, paramType }: TInputUIProps) =>
               formUpdate[event.target.name] = contractFunctionArgument;
               setForm(formUpdate);
             }}
+            type={paramType.type === "uint256" ? "number" : undefined}
```
rin-st, line 63: do we need that arrows?

Screenshot 2023-02-25 at 19 14 40

technophile-04: ++ vote to remove it

technophile-04:
Also a small nitpick, I think it was introduced in #206 

The `*` is not centered correctly aligned: 
Screenshot 2023-02-25 at 10 27 04 PM

sverps:
> The `*` is not centered correctly aligned:

I noticed the method to align the asterisk was to place it at flex-end, to sort of 'simulate' being centered. This means the fields with `#` were a bit off-center. I fixed it by using a vertically centered alternative to the asterisk.

Additionally, rather than forcing numerical input, I aligned the behavior with that of `TxValueInput` and when the value is invalid (e.g. string input for uint256 field), the form now shows a red border indicating an issue.

technophile-04:
Tysm @sverps love the new approach and it works great !!! lol sorry for being PITA

packages/nextjs/components/scaffold-eth/Contract/utilsComponents.tsx
```diff
@@ -83,14 +83,14 @@ const TxValueInput = ({ setTxValue, txValue }: { setTxValue: Dispatch<SetStateAc
           data-tip="Multiply by 10^18 (wei)"
         >
           <button
-            className="cursor-pointer text-xl font-semibold pt-1 px-4 text-accent"
+            className="cursor-pointer text-l font-semibold px-4 text-accent"
```
technophile-04, line 86: just small changes, we can change `text-l` -> `text-xl`.....I think it was added by mistake since `text-l` doesn't exist in tailwind

Also I think we should keep it `text-xl` instead of removing it I like big symbols

packages/nextjs/components/scaffold-eth/Contract/utilsDisplay.tsx
```diff
@@ -98,8 +98,8 @@ export const UintToEtherConverter = ({ form, setForm, stateObjectKey }: IUtility
   }, [form, setForm, stateObjectKey]);
 
   return (
-    <div className="cursor-pointer text-xl font-semibold px-4 text-accent" onClick={convertEtherToUint}>
-      *
+    <div className="cursor-pointer text-l font-semibold px-4 text-accent" onClick={convertEtherToUint}>
```
technophile-04, line 101: same `text-l` -> `text-xl`

packages/nextjs/components/scaffold-eth/Contract/utilsComponents.tsx
sverps: Oops, wanted to put `text-lg`. I modified it to keep the sizing of the asterisk about the same, but if everyone is fine with a larger character, we can keep it on text-xl as well. Either works for me.

packages/nextjs/components/scaffold-eth/Contract/utilsDisplay.tsx
sverps: fixed

sverps:
Here's what the asterisk will look like after this PR:
image

packages/nextjs/components/scaffold-eth/Contract/utilsComponents.tsx
technophile-04: cc @carletex for `xl` and `lg`... both look great to me, maybe he can change it while merging

technophile-04 (approved):

Issue #118, opened 2023-01-30
Update folder structure


rin-st:
- Update components to
```
Components
+-- Ui
|    +-- {simple universal components like button}
+-- Page1
|    +-- {components for Page1 only}
+-- Page2
|    +-- {components for Page2 only}
+-- Shared
|    +-- {shared complex components}
```
- Remove `scaffold-eth` folders from components, hooks, and utils.
- I believe we should have one file for one react component. So I think it will be better to remove `utilsSomething.tsx` files and replace them with files/folders

@carletex @technophile-04

carletex:
I appreciate this @rin-st We should have this kind of discussions to keep iterating & improving.

> Remove scaffold-eth folders from components, hooks, and utils.

The reason for it: One day, **scaffold-eth components & hooks** (which we can call "the core" of SE-2) might be a separate package (or packages) so people can use it in their dapps (e.g. `@scaffold-eth/burner-wallet`, `@scaffold-eth/contract-debug`, `@scaffold-eth/eth-address-input`). So want to keep it separate in the app.

We are building a tool for building dapps, it's not a final app, so people will be creating their own components / hooks as they need....and use ours (that for now live inside the project), need to be separated from theirs-

ExampleUI is a good example (:D) on how people would use SE-2, that's why I really like how you organized the code / folder structure on https://github.com/scaffold-eth/se-2/pull/115

> I believe we should have one file for one react component. So I think it will be better to remove utilsSomething.tsx files and replace them with files/folders

I've seen both and I don't feel strong about any of the 2:

**Option 1**:
+-- utils
++|----  network.ts (This file export a bunch of utils related to networks (getNetworkDetailsByChainId, getBlockExplorerTxLink, etc)

**Option 2**:
+-- utils
++|---- network
+++|------ getNetworkDetailsByChainId.ts
+++|------ getBlockExplorerTxLink

I personally don't see a huge benefit in any of the two. For me it's just a matter of preference / style.

rin-st:
> The reason for it: One day, scaffold-eth components & hooks (which we can call "the core" of SE-2) might be a separate package (or packages) 

Good point. So it's everything except the example UI page/components?

> I've seen both and I don't feel strong about any of the 2:

[...]

PR #220, opened 2023-02-28
refactor: input components separation of concerns

### LINKED ISSUE #219: Code improvement: input fields
## Current situation

The current input fields are a bit spread out:
- `AddressInput.tsx`
- `EtherInput.tsx`
- `TxValueInput` (inside `utilsComponents.tsx`)
- `InputUI.tsx`

A lot of the styling is duplicated across these components.

The `TxValueInput` is essentially the same as the `InputUI` with type of `uint256`, their asterisk behavior is essentially the same (multiply by 10^18).

## Desired situation

Styles should be extracted into a common component (i.e. `InputBase`).

Functionally, the components should then implement formatting and utils (i.e. ENS resolver for Addresses, Eth <-> USD switcher for Ether input, asterisk/hash toggles for various inputs) in separate components based on their input type, but agnostic to how/where they would be used.

Then the contract ui can leverage these generic components to build its ui and update its form values by passing the proper params.

This creates a better separation of concerns and uniformity across different input components, and allows a more flexible usage of underlying components.

sverps:
Fixes #219


sverps:
> Noticed one thing - uint input doesn't work after converting it back. And should it convert that way?

Looks like issue already exists on `main` branch. It's a good question, though. We should rethink the asterisk behavior and probably add a tooltip on it that explains what it does.

The multiplication by 10^18 is useful for payable value, since that is an ether value and it's handy to not have to make this conversion yourself.

For an actual uint256 argument, it's rather arbitrary to have it multiply. It's only assumed it's some token amount or balance since the type requires 256 bits.

technophile-04:
@sverps this looks great! a lot cleaner and better !! 

A bug which I noticed : 

https://user-images.githubusercontent.com/80153681/221964629-3dd9e0a2-6b90-44b3-bf2c-1208494f579f.mov

I see there are two utility which are handling this conversion
1. https://github.com/scaffold-eth/se-2/blob/5bc5a6b7363cc491618c09cefec2446615c4158d/packages/nextjs/components/scaffold-eth/Contract/utilsComponents.tsx#L49
2. https://github.com/scaffold-eth/se-2/blob/5bc5a6b7363cc491618c09cefec2446615c4158d/packages/nextjs/components/scaffold-eth/Contract/utilsDisplay.tsx#L86

I like the behavior of `TxnValueInput` which is used by `WriteOnlyFunctionFrom` this is `main` branch behavior:
 

https://user-images.githubusercontent.com/80153681/221965613-942d2215-9479-41e7-990e-8b5450c53ba0.mov

So It would be great if we keep the behavior of both `Read` and `Write` similar to `TxnValueInput` i.e it has a `tooltip` and keeps on multiplying number by 10^18

packages/nextjs/components/scaffold-eth/Contract/ContractInput.tsx
```diff
@@ -0,0 +1,46 @@
...
+      name: stateObjectKey,
+      value: form[stateObjectKey],
+      placeholder: paramType.name ? `${paramType.type} ${paramType.name}` : paramType.type,
+      onChange: (value: any) => {
+        const formUpdate = { ...form };
+        formUpdate[stateObjectKey] = value;
+        setForm(formUpdate);
+      },
+    }),
+    [form, paramType.name, paramType.type, setForm, stateObjectKey],
```
technophile-04, line 28: NITPICK: Since `form` here is an object, I think `useMemo` might be not usefull here since changes the dependencies are compared using [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) in react.

packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
```diff
@@ -1,13 +1,14 @@
...
+import { useEffect, useState } from "react";
 import Blockies from "react-blockies";
 import { useEnsAddress, useEnsAvatar, useEnsName } from "wagmi";
 import { isAddress } from "ethers/lib/utils";
+import { InputBase } from "./InputBase";
 
-type TAddressInputProps = {
-  onChange?: (arg: string) => void;
-  placeholder?: string;
+type AddressInputProps = {
```
technophile-04, line 7: I was wondering if we should make a common interface as : 
```js 
Interface IInput {
  value: string;
  onChange: (newValue: string) => void;
  name?: string;
  placeholder?: string;
}
```
and extend other Inputs like `AddressInput`, `Bytes32Input` etc from it because most of them have same properties : `name`, `placeholder` etc

sverps:
I re-added the tooltip (seems I overlooked this behavior, oops :sweat_smile: ).
Additionally, when the current valuee is already `BigNumbeer`, I don't think it makes sense to multiply it again by 10^18, so maybe better to not allow this option in that case.
I also improved the error state for values that exceed the allowed amount of bits.


https://user-images.githubusercontent.com/15879327/222087921-7d235cf6-3e01-4081-bca1-6863277ffafa.mp4

packages/nextjs/components/scaffold-eth/Contract/ContractInput.tsx
sverps: But `form` is the output of a `useState` hook in the parent component, and we don't modify it except by using `setForm` so it's reference should be stable across renders, I believe.

packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
sverps: Good point! Made a common input props interface.

(As a sidenote, I think we shouldn't prefix interfaces or types with I or T. Single letter prefixes are, in my experience, usually reserved for generic types)

packages/nextjs/components/scaffold-eth/Contract/ContractInput.tsx
rin-st: as I understand to use `useMemo` here, you need to memorize inputs on `ReadOnlyFunctionForm` and `WriteOnlyFunctionForm` since at least `setForm` will change on every render. And also form will change on every other input change too. Using global state for forms is pain, so you probably don't need useMemo here at all

packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
```diff
@@ -1,22 +1,17 @@
...
+import { CommonInputProps } from "./utils";
 
 // ToDo:  move this function to an utility file
 const isENS = (address = "") => address.endsWith(".eth") || address.endsWith(".xyz");
 
 /**
  * Address input with ENS name resolution
  */
-const AddressInput = ({ value, name, placeholder, onChange }: TAddressInputProps) => {
+export const AddressInput = ({ value, name, placeholder, onChange }: CommonInputProps) => {
```
rin-st, line 14: I don't remember when but as I remember, @carletex mentioned that he prefers `export default` for components since we can use lazy loading in future

packages/nextjs/components/scaffold-eth/Input/InputBase.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  const handleChange = useCallback(
+    (e: ChangeEvent<HTMLInputElement>) => {
+      onChange(e.target.value as unknown as T);
+    },
+    [onChange],
+  );
+
+  return (
+    <div className={`flex border-2 border-base-300 bg-base-200 rounded-full text-accent ${modifiers.join(" ")}`}>
```
rin-st, line 36: `${modifiers[0]}` ? Or it can be just a string

packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
sverps: Yeah there's an issue open about it, I still need to check out the details and update my PR accordingly :+1:

packages/nextjs/components/scaffold-eth/Input/UintInput.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { CommonInputProps, NUMBER_REGEX, UintVariant } from "./utils";
+
+type UintInputProps = CommonInputProps<string | BigNumber> & {
+  variant?: UintVariant;
+};
+
+export const UintInput = ({ value, onChange, name, placeholder, variant = UintVariant.UINT256 }: UintInputProps) => {
+  const [inputError, setInputError] = useState(false);
+  const convertEtherToUint = useCallback(() => {
+    if (!value || value instanceof BigNumber) {
```
rin-st, line 13: But what if the user wants to pass some small `value` as `BigNumber` and he wants it to be a wei? For example for prefilling input. Or we don't want to consider this?

packages/nextjs/components/scaffold-eth/Input/InputBase.tsx
sverps: The idea was to allow concurrent modifiers in case we want to expand the input component states (i.e. when `error` and `disabled` are active at the same time). Since it isn't the case yet, I'll simplify it a bit. :+1:

packages/nextjs/components/scaffold-eth/Input/UintInput.tsx
sverps: Hmmm, fair point. Maybe it's less confusing for the user to not assume anything and let him multiply as many times as he wishes... :smiley:

packages/nextjs/components/scaffold-eth/Contract/ContractInput.tsx
sverps: I removed it, looks like you guys were right. :+1: 
I generally manage more complex state with reducers or in the case of forms with `formik`.

packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
sverps: I found the discussion (#118). The issue is with barrel files.
I removed all the barrel files and updated imports. I think we should keep named exports, as it will be (slightly) easier to add barrel files for a potential future scaffold-eth package if ever the nextjs issue is resolved.

packages/nextjs/components/scaffold-eth/Input/IntInput.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const convertEtherToUint = useCallback(() => {
+    if (!value) {
+      return;
+    }
+    onChange(ethers.utils.parseEther(value.toString()));
+  }, [onChange, value]);
+
+  useEffect(() => {
+    if (
+      (value && typeof value === "string" && !SIGNED_NUMBER_REGEX.test(value) && value !== "-") ||
```
rin-st, line 21: Is it real to combine `IntInput` and `UintInput`? It seems the only difference is this `if` condition and a default value of `variant` prop

technophile-04:
Did some testing and everything looks really great !! Love the refactor!

Noticed this while testing the array : 
Screenshot 2023-03-02 at 9 26 10 PM

We can refer to #156 if needed because we have them in `main` branch, but was wondering if we should handle  in this PR or separate PR 

Small nitpick(can be ignored completely) : 

Since we have `enums` of variants of `uint` and `int` may we can only show `*` if its `  >= uint64` and `>= int64`  or only for `256` like currently on `main`.....but nevertheless this is a kind of overhead and we can completely ignore it

packages/nextjs/components/scaffold-eth/Input/IntInput.tsx
sverps: Done

sverps:
> We can refer to #156 if needed because we have them in `main` branch, but was wondering if we should handle in this PR or separate PR

Ah if this was working in main branch, I'll add it in my refactor as well. Wouldn't want to introduce a regression. Thanks for pointing it out :+1:

sverps:
I removed the `MissingTypesInput` and now param types that have no specific input field for it fallback to the input base. Then we're sure things keep working as before.

In future issues, we could expand the set of inputs to create better validation for other inputs (such as different array types) etc. But at least they'll keep working... :sweat_smile:

rin-st:
tried to open example-ui page from your branch, but it's broken

sverps:
> tried to open example-ui page from your branch, but it's broken
> 
> 

Weird, I don't get that error, can you share a more detailed stacktrace?

technophile-04:
Noticed some minor bugs while testing : 
1. You can't seems to send the transaction without setting `value` if the function is `payable`. 
2. The input red-highlighted, if you try to pass an array if its of `uint256[]`, although if you send transaction it goes correctly. 

Demo : 

https://user-images.githubusercontent.com/80153681/222731134-6fc5e9f6-dbbe-48ee-ae80-e76a235154c3.mov

I used this contract while testing -> https://github.com/scaffold-eth/se-2/pull/156/commits/1f22545814be26d81897a42fc99c7e30863d4ab2 . 

> tried to open example-ui page from your branch, but it's broken

It works for me too, maybe you changed some stuff in `YourContract.sol` for testing and this lead `example-ui` erroring out because its not inline with changes in `YourContract.sol`...This happened with me while testing

rin-st:
> It works for me too, maybe you changed some stuff in YourContract.sol for testing and this lead example-ui erroring out because its not inline with changes in YourContract.sol...This happened with me while testing

Strange bug. I changed contract before, but redeploying the contract, `hardhat clean`, manually removing old hh generated data, and removing browser cache didn't help for some reason - useContractRead returned value from the old contract on `example-ui` page. Added `toString()` to bugged value in the code, and it worked. Removed `toString()`, and it still works. 🤷‍♂️

sverps:
> Noticed some minor bugs while testing :
> 
> 1. You can't seems to send the transaction without setting `value` if the function is `payable`.
> 2. The input red-highlighted, if you try to pass an array if its of `uint256[]`, although if you send transaction it goes correctly.

Thanks a lot for all the testing. Fixed both regressions. :+1: 
I have the feeling it's starting to look solid :rocket:

carletex:
We can apply what we have learned in #227 (and fix conflicts :D)

sverps:
> We can apply what we have learned in #227 (and fix conflicts :D)

I'll merge main and align all inputs :+1:

sverps:
Went through the whole thing and did some minor cleanup and fixes. Should be good now. Ready for last reviews.

rin-st:
looks good to me 👍

rin-st (approved):

carletex (commented):
This is great @sverps, everything looks cleaner and organized! And thanks all for the review & feedback.

I left a few review comments.

packages/nextjs/utils/scaffold-eth/index.ts
```diff
@@ -2,4 +2,3 @@ export { default as getLocalProvider } from "./getLocalProvider";
 export * from "./networks";
 export { default as notification } from "./notification";
 export { default as fetchPriceFromUniswap } from "./fetchPriceFromUniswap";
-export { default as parseTxnValue } from "./parseTxnValue";
```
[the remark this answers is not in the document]
sverps: I think I removed things from barrel files where I touched them in relation to #118, but it's probably better to keep them in barrel files and resolve the other issue as part of another PR (if needed)

packages/nextjs/components/scaffold-eth/index.tsx
```diff
@@ -2,7 +2,5 @@ export { default as FaucetButton } from "./FaucetButton";
 export { default as Faucet } from "./Faucet";
 export { default as Address } from "./Address";
 export { default as Balance } from "./Balance";
-export { default as AddressInput } from "./AddressInput";
-export { default as EtherInput } from "./EtherInput";
```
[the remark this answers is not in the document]
sverps: Fixed

packages/nextjs/components/scaffold-eth/Input/utils.ts
```diff
@@ -0,0 +1,108 @@
...
+  INT232 = "int232",
+  INT240 = "int240",
+  INT248 = "int248",
+  INT256 = "int256",
+}
+
+export const SIGNED_NUMBER_REGEX = /^-?\d+\.?\d*$/;
+export const UNSIGNED_NUMBER_REGEX = /^\.?\d+\.?\d*$/;
+
+export const isValid = (dataType: IntegerVariant, value: BigNumberish, strict = true) => {
```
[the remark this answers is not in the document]
sverps: Changed it to `isValidInteger`

packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
```diff
@@ -0,0 +1,70 @@
+import { useEffect } from "react";
```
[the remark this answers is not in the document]
sverps: I think the original behavior is better and more intuitive for an end user. A user that enters `atg.eth` doesn't expect is to suddenly show a different ens, even more so in cases where the ens don't look related at all to each other.

carletex:
Amazing stuff @sverps 

Thanks <3!!

PR #221, opened 2023-02-28
fix: number input errors

rin-st:
added handling input errors for all the uints/ints

before:

https://user-images.githubusercontent.com/25638585/221921745-fada2832-0f95-40af-9e61-24f791fec712.mov

after:

https://user-images.githubusercontent.com/25638585/221921848-7015c456-1ed5-43d4-99b5-1a7c2bc05bd5.mov

### DIFF
--- a/packages/hardhat/contracts/YourContract.sol
+++ b/packages/hardhat/contracts/YourContract.sol
@@ -18,6 +18,8 @@ contract YourContract {
     string public greeting = "Building Unstoppable Apps!!!";
     bool public premium = false;
     uint256 public totalCounter = 0;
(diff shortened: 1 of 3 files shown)


sverps:
Changes look nice :+1: 

> converted to draft because of #220 . Will continue if needed after merging it

I think we can incorporate your approach neatly into the refactor (or add it after). I think rather than having a `Uint256Input`, we could capture all ints and uints in a `UintInput` and a IntInput` component instead.

rin-st:
fixed in https://github.com/scaffold-eth/se-2/pull/220

PR #227, opened 2023-03-06
fix: EtherInput now also works as a controlled input

### LINKED ISSUE #226: EtherInput as controlled input is broken
It doesn't take the `value` prop into account, doesn't really do anything with it. This results in the component only being usable as an uncontrolled input. A change of `value` outside of the component is not propagated to the actual input.

# Reproduce

Go on the home page for example and add the following state:
```javascript
const [amount, setAmount] = useState("1");
```

Then also add this component anywhere:
```javascript
<EtherInput value={amount} name="amount" onChange={newAmount => setAmount(newAmount)} />
```

The input stays blank (instead of being initialized to "1").

Triggering `setAmount` from anywhere outside of the `EtherInput` component equally doesn't reflect in a change of the input value.

sverps:
Fixes #226


packages/nextjs/components/scaffold-eth/EtherInput.tsx
```diff
@@ -2,7 +2,35 @@ import React, { ChangeEvent, useEffect, useState } from "react";
 import { useAppStore } from "~~/services/store/store";
 import { ArrowsRightLeftIcon } from "@heroicons/react/24/outline";
 
-type TEtherInputProps = {
+function etherValueToDisplayValue(usdMode: boolean, etherValue: string, ethPrice: number) {
```
rin-st, line 5: since functions almost the same I'd propose something like this

```
function multiplyValue(applyMultiplying: boolean, value: string, multiplier: number) {
  if (applyMultiplying && multiplier) {
    const parsedEthValue = parseFloat(value);
    if (!Number.isNaN(parsedEthValue)) {
      return (parsedEthValue * multiplier).toString();
    }
  }

  return value;
}

function etherValueToDisplayValue(usdMode: boolean, etherValue: string, ethPrice: number) {
  return multiplyValue(usdMode, etherValue, ethPrice);
}

function displayValueToEtherValue(usdMode: boolean, displayValue: string, ethPrice: number) {
  return multiplyValue(usdMode, displayValue, 1 / ethPrice);
}
```

rin-st:
Found another bug, it is in `main` already, you can fix it here or create a new issue.

$ mode is buggy sometimes (because of js multiplying?). Probably `toFixed(2)` can help.

Adding 99 here and trying to delete symbols
https://user-images.githubusercontent.com/25638585/223387593-3c5c56c3-ffc3-4429-a039-1ec16ed6869d.mov

packages/nextjs/components/scaffold-eth/EtherInput.tsx
sverps: Added a `toFixed(2)` for the usd display value, so not sure if we can still apply this suggestion in some way. Let me know if you have an idea.

sverps:
> $ mode is buggy sometimes (because of js multiplying?). Probably `toFixed(2)` can help.

The issue was a bit tricky. It's caused by the `displayValue` not being the source of truth in `usdMode`, basically. If you edit it, it updates the `value`/`ethValue`, which can then trigger the `displayValue` to be updated again to a different value than initial due to floating point rounding errors.

The solution I found (maybe there's another way) is to internally keep track of the last set ethValue with a reference, and only update the `displayValue` if the outside `value` differs from this last set value.

It's the downside of allowing the same input component to be used as controller or uncontrolled input.

rin-st:
It doesn't work for me when uncontrolled 🤷‍♂️

rin-st:
Tried to play with code and simplify things. What do you think about it? There is no error handling, but except that it works too.

```
import React, { ChangeEvent, useState } from "react";
import { useAppStore } from "~~/services/store/store";
import { ArrowsRightLeftIcon } from "@heroicons/react/24/outline";

function etherToUsd(etherValue: string, ethPrice: number) {
  if (ethPrice) {
    const parsedEthValue = parseFloat(etherValue);
    if (Number.isNaN(parsedEthValue)) {
      return etherValue;
    } else {
      return (parsedEthValue * ethPrice).toFixed(2);
    }
  } else {
    return etherValue;
  }
}

function usdToEther(displayValue: string, ethPrice: number) {
  if (ethPrice) {
    const parsedDisplayValue = parseFloat(displayValue);
    if (Number.isNaN(parsedDisplayValue)) {
      // Invalid number.
      return displayValue;
    } else {
      // Compute the ETH value if a valid number.
      return (parsedDisplayValue / ethPrice).toString();
    }
  } else {
    return displayValue;
  }
}

type EtherInputProps = {
  onChange?: (arg: string) => void;
  placeholder?: string;
  name?: string;
  value?: string;
};

/**
 * Input for ETH amount with USD conversion.
 *
 * onChange will always be called with the value in ETH
 */
export default function EtherInput({ value = "", name, placeholder, onChange }: EtherInputProps) {
  const [innerValue, setInnerValue] = useState(value);
  const isControlled = !!onChange;
  const ethValue = isControlled ? value : innerValue;
  const onEthValueChange = isControlled ? onChange : setInnerValue;

  const ethPrice = useAppStore(state => state.ethPrice);
  const [usdMode, setUsdMode] = useState(false);
  const [usdValue, setUsdValue] = useState(etherToUsd(ethValue, ethPrice));

  const handleChangeNumber = (event: ChangeEvent<HTMLInputElement>) => {
    const newValue = event.target.value;
    setUsdValue(usdMode ? newValue : etherToUsd(newValue, ethPrice));
    onEthValueChange(usdMode ? usdToEther(newValue, ethPrice) : newValue);
  };

  const toggleMode = () => {
    setUsdMode(!usdMode);
  };

  return (
    <>
      <div className="rounded-full border-2 border-base-300">
        <div className="form-control grow">
          <div className="flex w-full items-center">
            <span className="pl-4 text-primary">{usdMode ? "$" : "Ξ"}</span>
            <input
              name={name}
              type="text"
              placeholder={placeholder}
              className="input input-ghost pl-1 focus:outline-none focus:bg-transparent focus:text-gray-400 h-[2.2rem] min-h-[2.2rem] border w-full font-medium placeholder:text-accent/50 text-gray-400 grow"
              value={usdMode ? usdValue : ethValue}
              onChange={handleChangeNumber}
            />
            <button
              className="btn btn-primary h-[2.2rem] min-h-[2.2rem]"
              onClick={toggleMode}
              disabled={!usdMode && !ethPrice}
            >
              <ArrowsRightLeftIcon className="h-3 w-3 cursor-pointer" aria-hidden="true" />
            </button>
          </div>
        </div>
      </div>
    </>
  );
}
```

sverps:
I think we're over-complicating things. Since the component is currencly only used in the `Faucet`, and it receives a `value` and an `onChange` prop there, it is used as a controlled input.

I see little value in making the component more complex than necessary for little gain, so I think we should just align it further with the other inputs and make it a regular controlled input.

sverps:
> It doesn't work for me when uncontrolled man_shrugging

Can you explain in more detail what is not working? For me it works fine with a state managed outside the component or using the internal state. Like so:

```javascript
<EtherInput value={amount} name="amount" onChange={newAmount => setAmount(newAmount)} />
```

or

```javascript
<EtherInput name="amount" onChange={newAmount => setAmount(newAmount)} />
```

rin-st:
Thanks for your comments!

> In your current approach, there's no way to actually use the value of the input without providing an onChange prop. So not much use for the component used like that.

We can use form. See the first input in [Example](https://beta.reactjs.org/reference/react-dom/components/input#reading-the-input-values-when-submitting-a-form). But the value used on submit can be eth or usd. Or we can create hidden input with eth value and use it.

So, `onChange` in uncontrolled input is not a must. And in your case, `EtherInput` doesn't work without `onChange`: `<EtherInput name="amount" />`

https://user-images.githubusercontent.com/25638585/223648146-b6666110-683b-4a3d-b934-26a459dec1f4.mov

---
> I don't think looking at the onChange prop to determine if it's controlled or not is right. It's about whether or not the actual state is managed inside the component or outside

As I understand, in our case controlled = managed outside, uncontrolled = managed inside. But yes, `onChange` prop is not the better way to figure it out, probably I need to check `value` prop.

rin-st:
Please correct me if I'm wrong, @technophile-04, @carletex, as I understand, there are no uncontrolled inputs in our app.

Why do we need to maintain uncontrolled behavior? For users who will decide to use forms?

technophile-04:
Did some testing and everything works nicely 🙌 

> controlled = managed outside, uncontrolled = managed inside 

> we're wasting time trying to make it work for uncontrolled

Yup I see most of the people using controlled only while using our components

technophile-04 (approved):

carletex:
> Why do we need to maintain uncontrolled behavior? For users who will decide to use forms?

It started with the `AddressInput` component. Adding the uncontrolled behav was basically adding this line (+ some checks):
```ts
  const currentValue = value !== undefined ? value : address || "";
```

Besides using forms (on regular HTML inputs, you have the option of use them as controlled VS uncontrolled, even if the controlled way is the recommended way)... I liked the idea of adding `<AddressInput />` into the page with no props and you could see the component doing its work (ens resolution, ens avatar, blockie, etc). Maybe a silly thing, but adding it was trivial.

I though it was a good idea to for the `EtherInput` component too. But it's more complex by nature, and yep, the controlled input was broken (in a weird way) but also working in "controlled" & "uncontrolled" mode (in a weird way :D)

I like the component refactor in this PR (way more clean now), but I don't think it's working as I understand it should:

**Uncontrolled:**
- `<EtherInput />` This doesn't work. It works on main.

**Controlled**
- `<EtherInput value="12" />`. This shows the correct value of 12 (and its USD value), but it let you change the ETH value on the input (it shouldn't in a controlled component). Same error in main (+ doesn't show the init value). It doesn't make sense to use the component like this tho, but, it tell us that we are not handling the "controlled" input the right way.
- `<EtherInput value={amount} onChange={v => setAmount(v)} />`. This works perfectly, better than the exiting main, since it takes in consideration the init value.


> upd. It looks like you changed all the inputs behavior to controlled in https://github.com/scaffold-eth/se-2/pull/220 . Probably we're wasting time trying to make it work for uncontrolled

It all boils to this. What do you all think? My opinion:

- Make uncontrolled work if it doesn't take too much effort / complexity. 
  - If it does, nix it (I think #220 does it by making onChange mandatory)
- Make the controlled version work as it should. (#220 has the same problem, where you can still change the input on a hardcoded value and not update).

In any case, we are not wasting time! :P These tinkering / discussions are good for the project.

Thank you all <3

sverps:
> What do you all think? My opinion

I think we should always force all inputs to be controlled (i.e. required `value` and `onChange`). Doing this consistently has the following advantages:

- Makes it's usage more intuitive (i.e. you see one component used in a certain way and you automatically know how to use another input). No edge-cases or differences between different components leads to less confusion and user error.
- Makes the code cleaner and simpler, and less prone to bugs. Users that want to dive in and look and play with inner workings will understand it better. Maybe someone wants to make an exotic input that handles some specific contract interaction.

carletex:
I lean into having both, as any HTML input that you might use in React.

But after reading your comments and following the simple & minimal motto of SE-2....let's go with **controlled only**. We can always change our mind in the future.

We need to tweak this PR to make it totally controlled. Right now `<EtherInput value="1" />` let's you change the value. It should have the same behaviour as `<input type="text" value="hola" />`

Thanks

sverps:
> We need to tweak this PR to make it totally controlled. Right now `<EtherInput value="1" />` let's you change the value. It should have the same behaviour as `<input type="text" value="hola" />`

Ok, I'm on it :+1:

rin-st:
I'm feeling guilty for adding new comments to this pr 😄 , but
should it take the old value when you're changing the mode from USD to Eth and value ends with `.`, or should it take value without a dot? For example, you have `$ 1.5`, change to `$ 1.`, convert to Eth -> it converts 1.5, not 1.

https://user-images.githubusercontent.com/25638585/224101674-095c1b9d-5b44-4aaa-a034-e552a8567ef9.mov

sverps:
> I'm feeling guilty for adding new comments to this pr 

I always welcome all feedback. No worries :sweat_smile: 

> should it take the old value when you're changing the mode from USD to Eth and value ends with `.`, or should it take value without a dot? For example, you have `$ 1.5`, change to `$ 1.`, convert to Eth -> it converts 1.5, not 1.

I've considered this scenario, and there's no fix without trade-offs. If we make it take the new value already when there is a trailing period, it will result in the period being removed automatically as soon as the last decimal digit is removed, which felt a bit jumpy and weird.

I think it's more likely that a user would want to change (for example) $ 1.5 into $ 1.6 by removing the `5` and typing a `6`, than that he would convert usd to eth when there is a trailing period. For that reason, I opted for the current behavior.

rin-st:
what if change `toggleMode` to 
```
const toggleMode = () => {
    if (transitoryDisplayValue) {
      const newEthValue = displayValueToEtherValue(usdMode, transitoryDisplayValue.split(".")[0] || "0", ethPrice);
      onChange(newEthValue);
    }
    setUSDMode(!usdMode);
  };
```
it seems it fixes the issue

packages/nextjs/components/scaffold-eth/EtherInput.tsx
```diff
@@ -14,69 +46,45 @@ type TEtherInputProps = {
...
-      ethNewValue = newValue;
     }
 
-    setEthValue(ethNewValue);
-    setDisplayValue(newValue);
-    if (onChange) {
-      onChange(ethNewValue);
+    // Since the display value is a derived state (calculated from the ether value), usdMode would not allow introducing a decimal point.
+    // This condition handles a transitory state for a display value with a trailing decimal sign
+    if (newValue.endsWith(".")) {
```
rin-st, line 76: ` || newValue.endsWith(".0")` or you can't write for example $ 1.02

sverps: Ah good catch :+1:

sverps:
> it seems it fixes the issue

Found an alternative solution that fixes more things.

Deeper down, the issue was that the eth value wasn't updated yet when a transitory state was set. For example, when the value is `$ 1.5` and you remove the `5` to leave `$ 1.`, the controlling state will not be updated yet.

This last fix ensures the change handler is always triggered when the eth value changes.

rin-st (approved):

rin-st:
Could you please also add
```
if (newValue && !NUMBER_REGEX.test(newValue)) {
  return;
}
```
on line 79? 
Without it, you can write whatever you want in eth mode, and things like 0.0.0....0.. or again everything if you start by letter in usd mode. If I miss something and it doesn't solve the issue we can do it in another pr.

sverps:
> Could you please also add

Done. Thanks a lot for your thorough review and testing!

carletex:
DX perspective:

When I use the component, it's weird that only `onChange` is required. In my mind, "value" and "onChange" should be required.

image
My IDE autocompletion

With value required:
image

I see the use case the other way around: onChange not required (for disabled / read-only inputs) if we want to make that happen. But I don't see why you would want to NOT provide a value.

This is going to be the 99.9% use case for controlled components, right?:

```tsx
const [sendValue, setSendValue] = useState("");

<EtherInput placeholder="Amount to send" value={sendValue} onChange={value => setSendValue(value)} />
```

Am I missing something? Thanks!

sverps:
> When I use the component, it's weird that only onChange is required. In my mind, "value" and "onChange" should be required.

Fixed :+1: 

Turns out typescript handles `value?: string` differently from `value: string | undefined`. It's the second one we need. That will force the user to provide the prop, but doesn't necessarily force it to be defined. The following still works fine:

```javascript
const [sendValue, setSendValue] = useState<string>();

<EtherInput placeholder="Amount to send" value={sendValue} onChange={value => setSendValue(value)} />
```

carletex:
> Turns out typescript handles value?: string differently from value: string | undefined.

Yep! That might be it, but I'll have to use it more to see.

`value: string` could be another option, since I always do `useState("")`, instead of `useState<string>();`, where the type is inferred. But the undefined pattern can be useful in some use cases, and most of wagmi hooks return "whateverType | undefined", so this approach plays nicely with it.

We might be splitting hairs here haha

Thanks you all! <3

PR #228, opened 2023-03-07
remove restriction for useScaffoldContractRead generic

technophile-04:
Noticed it while hacking se-2 : 

1. You are not able to pass other types as generic expect `BigNumber | boolean | string`, I think we should not restrict just to `BigNumber | boolean | string` and allow developers to pass in what they want, since sometime your function may return `struct` or other types. 

2. Added `polygonMumbai` to `.env's`, lots of people use `mumbai` for hacking and it would great if we add reference in comments itself so that developers don't have to go to wagmi's doc (we can completely ignore this though)

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
@@ -2,7 +2,6 @@ import { useContractRead } from "wagmi";
 import type { Abi } from "abitype";
 import { useDeployedContractInfo } from "./useDeployedContractInfo";
 import { getTargetNetwork } from "~~/utils/scaffold-eth";
-import { BigNumber } from "ethers";
 
 /**
  * @dev wrapper for wagmi's useContractRead hook which loads in deployed contract contract abi, address automatically
@@ -11,7 +10,7 @@ import { BigNumber } from "ethers";
  * @param args - args to be passed to the function call
  * @param readConfig - extra wagmi configuration
  */
-export const useScaffoldContractRead = <TReturn extends BigNumber | string | boolean = any>(
+export const useScaffoldContractRead = <TReturn = any>(
   contractName: string,
   functionName: string,
   args?: any[],
--- a/packages/nextjs/.env.development
+++ b/packages/nextjs/.env.development
@@ -1,5 +1,5 @@
 # The network where your DApp lives in.
-# mainnet / hardhat / sepolia / polygon / optimism / arbitrum
+# mainnet / hardhat / sepolia / polygon / optimism / arbitrum / polygonMumbai
 # Check supported chains - https://wagmi.sh/core/chains#supported-chains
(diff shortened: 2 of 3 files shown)


packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
```diff
@@ -11,7 +10,7 @@ import { BigNumber } from "ethers";
  * @param args - args to be passed to the function call
  * @param readConfig - extra wagmi configuration
  */
-export const useScaffoldContractRead = <TReturn extends BigNumber | string | boolean = any>(
+export const useScaffoldContractRead = <TReturn = any>(
```
sverps, line 13: Not a fan of using `any`. 
We should improve the restrictions we put to allow all possible return types. This way users will better understand what it can return by looking at the allowed types.

I suggest to add a `Struct` type:

```javascript
interface Struct {
  [key: string | number]: BigNumber | string | boolean | Struct;
}
```

Then we can upate the generic type like so:

```javascript
export const useScaffoldContractRead = <TReturn extends BigNumber | string | boolean | Struct = any>
```

Then you can use the hook like so:
```javascript
const { data } = useScaffoldContractRead<{ keyName: boolean }>( ...
```

or

```javascript
const { data } = useScaffoldContractRead<[ boolean ]>( ...
```

sverps (changes requested):

packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
technophile-04: > Not a fan of using any.

+++ , Really love your suggestions but again user may have different return types eg. [BigNumber, string] or even [boolean] is not working and by restricting they won't be able to pass in what they think is the correct type. Also, there might be different combinations that we could miss that's why kept it loose with `any`.

I 100% agree and really don't want `any` to get into the codebase but this was a quick patch/hack until #228 is ready. My thought was if the user is passing `generic` param to `useScaffoldContractRead` then he is a bit mature and know's what's are his `return types` are.

sverps: After exploring it a bit further and trying to expand the type to allow the various different return types (including nested structs), the type started to become so broad that it was practically `any`... So little benefit...

> My thought was if the user is passing generic param to useScaffoldContractRead then he is a bit mature and know's what's are his return types are.

Fair point :+1:

sverps (approved):

rin-st (approved):

packages/nextjs/hooks/scaffold-eth/useScaffoldContractRead.ts
carletex: Love this discussion. Thanks.

carletex:
This makes sense. Thanks for the discussion!

PR #231, opened 2023-03-09
Use config file for network and other configuration

### LINKED ISSUE #230: Migrate from env's to config fiile
I think we should use `*.config.ts`  file in our frontend for `targeetNetwork`, `pollingInterval` etc instead of relying on `env's`. 

### Reasons / Advantages : 
1. The main reason for putting this discussion is #202, using a `.ts` will help us infer literal types of `targetNetwork` which would be a great help for issue 202, as currently I can the showstopper for issue 202 is figuring out proper `targetNetwork` at compile time itself which will lead to type autocompletion as mentioned in this comment -> https://github.com/scaffold-eth/se-2/pull/202#issuecomment-1441963792
2. We will be able to force people to use `supported` wagmi networks only and it will also eleminate the chances of spelling mistakes and people refering / finding chains on wagmi docs. checkout out : 

https://user-images.githubusercontent.com/80153681/224123396-cefcf9de-d763-4c93-bee8-60e440562c4a.mov

 3. Also our single source of truth will be hand in hand with `wagmi` referring to -> #188 
 4. Currently if people use chains like `polygonMumbai` / others and make transactions, you don't get a link to their block explorer, since we are currently handling only `etherscan` checkout -> https://github.com/scaffold-eth/se-2/blob/5fb5e5f90e655d6d2b0df04dafe19c85a13f5632/packages/nextjs/utils/scaffold-eth/networks.ts#L70
 
You don't get any link to `blockexplorer` if you make transaction on Mumbai testnet or similar : 
Screenshot 2023-03-10 at 12 22 07 AM

while using wagmi's chain we can already leverage it(we can handle this in the current main branch too though). 

---

I think the reason we were not using `*config.ts` and were using `env's` initially was that it made more sense at that time since we were not that much relying on env's and them being typed also its easy to leverage env's in Next. 

some discussion on this topic was also found here -> https://github.com/scaffold-eth/se-2/issues/1#issuecomment-1259236623

--- 

Would love to hear the thoughts of other on this 🙌, specifically  : 
1. Do we really need `*.config.ts`, will it be helpful that much and improve DX ? 
2. If at all we plan to create `*.config.ts` should make it global like one `*.config.ts` file for both `hardhat` and `nextjs` ? 
3. ....

cc @carletex @rin-st @sverps

technophile-04:
Fixes #230


packages/nextjs/scaffold.config.ts
```diff
@@ -0,0 +1,13 @@
+import * as chain from "wagmi/chains";
+
+type TScaffoldConfig = {
+  targetNetwork: chain.Chain;
+  pollingInterval: number;
+};
+
+const ScaffoldConfig: TScaffoldConfig = {
```
sverps, line 8: I think we should settle on a styleguide and apply it on the whole project (I like [google's styleguide](https://google.github.io/styleguide/tsguide.html) as a good starting point).

That means here we'd follow default naming convention and make the const `lowerCamelCase`.

As for the type, but I think the convention should be to not prefix or suffix it and use regular `UpperCamelCase`, since typescript itself already contains the info about what kind of type it is (https://google.github.io/styleguide/tsguide.html#naming-style). 
Prefixes should, in my opinion, be reserved for generic types, to mark a clear distinction between defined types and generics.

@rin-st @carletex any thoughts?

rin-st: I agree 👍

carletex: I'm up for a having a styleguide.

Besides applying it here, we should create an issue an for it, discuss what guidelines we want to follow..... and make a PR refactoring all the names / types / etc.

sverps: Opened an issue about it :+1:

packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -56,26 +56,32 @@ export const NETWORKS_EXTRA_DATA: Record<string, TChainAttributes> = {
...
 
-  if (chainId === 31337 || chainId === 1337) {
-    // If its localChain then return empty sting
+  const chainNames = Object.keys(chains);
+
+  const targetChainArr = chainNames.filter(chainName => {
+    const wagmiChain = chains[chainName as keyof typeof chains];
+    return wagmiChain.id === chainId;
+  });
+
```
carletex, line 68: An idea (without testing it :angel: ).

What if we make the argument `network` to be wagmi's `Chain`? So we don't need to do this conversion.

In the transactor, we could get the connected network with wagmi's `useNetwork()` (and maybe throw an error / return if the signer is on a different network)

```tsx
// useTransactor
const { chain } = useNetwork();
// check that network.chainId === chain.id => if not, error / return
getBlockExplorerTxLink(chain, ...)
```

Since we are using wagmi's Chains as source of truth, maybe we should rely on them whenever we can? Instead of using internal ethers/providers.

technophile-04: Yup makes sense, lol even I was a bit uncomfortable while writing the conversion, 

> What if we make the argument network to be wagmi's Chain? So we don't need to do this conversion.

This is a great idea but the problem is `useTransactor` accepts `custom signer` too and the reason we are relying on the signers `network` is that : 
NITPICK: There might a case where people pass in the custom signer without connecting their wallet and in that case `useNetwork` and signer `network` will be different and might cause some inconsistency, we are using similar pattern in `Faucet.jsx`. 

Again the above NITPICK is a very rare case. 

Will try exploring some more alternatives 🙌

carletex: Hey Shiv, good point!!

After pulling the threads on this, some notes:
- The only instance where we are passing a custom signer, is for the hardhat faucet (where we want to send funds from a hardhat account with funds). Can we think about other use case? Since we are in the front-end, 99.99% of the times we'll want to use the connected signer (besides the faucet, and maybe a exotic build where we do deterministic private key derivation or something)
- `getBlockExplorerTxLink` doesn't return anything for the local chain (the only instance where we are using a custom signer for the transactor :thinking: )
- `getLocalProvider` is only used in the Faucet stuff. Could be an unneeded early abstraction.
- Maybe an util like `getChainFromChainId` that returns a wagmi Chain for a given Id?

In any case, no need to make any updates rn!! just thinking out loud and pointing out some possible inconsistencies. Maybe we made some past decisions that could use some refactors for the sake of consistency / simplicity / better DX.

Thanks!

technophile-04: Tysm everything makes sense!!

> Maybe we made some past decisions that could use some refactors for the sake of consistency / simplicity / better DX

+++

technophile-04:
### Small update 

Moved `scaffold.config` to root and same is being utilised by both 🙌 

Will be testing more and try to make things more consistent

sverps:
> btw I came across some patterns where people have created `config` directory similar to `nextjs` and `harhdat` where all the base configurations live like `.eslintrc`, `.tsconfig` etc, just wanted to put here since we are in "refactoring phase" :) and also we can put `scaffold.config` inside it. I like the way currently it is now but would love to know others thought

I'm used to those types of config files being at the root of their package, that's where I would generally look for them, so it doesn't bother me personally.

Additionally, do you know if all config files will play nice if they're not at the root of a package (e.g. `tailwind.config.js`, `postcss.config.js`, `next.config.js`)? I'd only consider it if all configs can be grouped, otherwise there will be 2 locations for config files, which would be a bit confusing.

scaffold.config.ts
```diff
@@ -0,0 +1,13 @@
+import * as chain from "@wagmi/chains";
+
+type ScaffoldConfig = {
+	targetNetwork: chain.Chain;
+	pollingInterval: number;
```
carletex, line 5: I think this are tabs

packages/nextjs/tsconfig.json
```diff
@@ -15,7 +15,8 @@
     "jsx": "preserve",
     "incremental": true,
     "paths": {
-      "~~/*": ["./*"]
+      "~~/*": ["./*"],
+      "@root/*": ["../../*"]
```
carletex, line 19: I still get this on my nextjs console:

image

I ran `yarn install` and restarted nextjs. Am I missing something?

carletex:
> Moved scaffold.config to root and same is being utilised by both next and hardhat

This is great! I think having an unique config file it's more dx friendly. A few thoughts/questions: 

- Could/Should we add a single source of truth here about networks? We have the `networks.ts` file on next, and we use the networks also in `hardhat.config.ts`. I'm just wondering if we could define a network object with all the data we need in scaffold.config, for hardhat & next (wagmi) to use. "Networks" feel like a key & transversal piece for SE-2 (or any dApp)
- Should we also support env vars in the config file? Specifically for secrets, since people are going to commit the config file. Things like "DEPLOYER_PRIVATE_KEY" or even alchemy/infura keys that could live in config.
```ts
const scaffoldConfig: ScaffoldConfig = {
	targetNetwork: chain.hardhat,
	pollingInterval: 30000,
        deployerdPrivateKey: process.env.DEPLOYER_PRIVATE_KEY
} as const;
```
I think we'll just need the dotenv package on the root. Handling all the config & env vars in the root seems more friendly. I think it will work on Vercel when manually defining the env vars in their UI.

This is just me thinking out loud, would love to hear your view on it.

> btw I came across some patterns where people have created config directory similar to nextjs and harhdat where all the base configurations live like .eslintrc, .tsconfig etc, just wanted to put here since we are in "refactoring phase" :) and also we can put scaffold.config inside it. I like the way currently it is now but would love to know others thought
> I'm used to those types of config files being at the root of their package, that's where I would generally look for them, so it doesn't bother me personally.

Since this is a monorepo with two packages, it also doesn't bother me having separate config files. But I've also seen monorepos with a **shared** (or config) directory.. I guess it's an alternative to having everything shared in the root (when is too bloated). It's definitely something that we should explore, but maybe in the future / another PR? This PR is already a big refactor.

packages/nextjs/tsconfig.json
technophile-04: Yes !! encountered the same problem for both, don't know for some reason it was working but I am sure i am missing some configuration, working on it

technophile-04:
### Update 

Now `scaffold.config.ts` at root should work but there are some catches :  

The way the front end was deployed previously is changed. 

Since we are utilizing `scaffold.config.ts` which is at root, outside the `nextjs` dir when you deploy to vercel it gives the below error : 
Screenshot 2023-03-15 at 9 15 08 PM

Previously we were uploading the whole source code to `vercel` and then running `build` there. 

Now, the flow is: first you need to build locally and then publish only `build` dir to vercel similar to surge.sh in SE

I have create script alias which will do it : 
`yarn vercel:prod` - will build and deploy to vercel on production
`yarn vercel:dev` - will build and deploy as preview
`yarn vercel:yolo` - ignore errors and deploy to production



### Approaches explored : 
Although using a `.ts` file at monorepo root seems like very simple / famous case but was not able to find good examples/references to deal with it having nextjs. 

1. Using `externalDir` which is `experimental` feature by nextjs, it seems to do the work but it prints the warning when you start the app : 

Screenshot 2023-03-15 at 9 44 32 PM

checkout this disccusion / comment => https://github.com/vercel/next.js/discussions/32192#discussioncomment-1761791

And this issue: https://github.com/vercel/next.js/issues/20374 why there are problems. 


2. Currently using nextjs built in [`module-transpilation`](https://nextjs.org/blog/next-13-1#built-in-module-transpilation-stable) which was built in nextjs by author of [`next-transpile-modules`](https://www.npmjs.com/package/next-transpile-modules), I saw people using [`next-transpile-modules`](https://github.com/martpie/next-transpile-modules#what-problems-does-it-solve) as a workaround for the problem which we faced and now its inbuild so utilzing it. 

I tried comparing build size of both approaches and they came to be same : 
| using `module-transpilation` | using `experimental` | 
| -----------------------------| ----------------------| 
| Screenshot 2023-03-15 at 10 16 50 PM |  Screenshot 2023-03-15 at 10 17 19 PM | 



3. Others, there some approaches where people used reconfigured `webpack` in `next.config.js` but found them really hacky and not future proof 

4. Maybe we can just put back `scaffold.config.ts` inside `nextjs`  

### Some thoughts : 
1. Yes I think new deployment flow is a bit off but I think if we want `scaffold.config.ts` to stay out of `nextjs` dir then we need to go with current approach, btw the current approach build time is really fast.  Also would love to know others thoughts / approaches to conquer this
2. This  was just a draft to see how it will look if we put `scaffold.config.ts` in `root`, happy to revamp this whole thing 🙌


--- 

Regarding putting `.envs` in `scaffold.config.ts` love the idea even tried exploring it but again there seems problem with `nextjs` using external env's 

People seem to use `dotenv` inside `next.config.js` to read from external `.env` and pass in  env's  inside [nextConfig](https://nextjs.org/docs/api-reference/next.config.js/environment-variables) the problem with this approach is that the env's you pass in are inlined in javascript while bundling i.e only allowed to pass in `NEXT_PUBLIC` and this will lead to exposing unnecessary keys if you are not mindful + found this approach a bit hacky 

cehkcout -> https://stackoverflow.com/questions/69259896/set-environment-variables-outside-of-pages-dir-in-nextjs
and https://stackoverflow.com/questions/59462614/how-to-use-different-env-files-with-nextjs/69619744#69619744

PS : the above stackoverflow links were picked from differnt next js issues 

I think maybe we can keep our api key's as hardcoded in `scaffold.config.ts` so that it can be used in both `hardhat.config.ts` and `wagmiClient` and let `nextjs` handle `env's` since `nextjs` aslo has `env.production` and etc we would manually need to handle its a bit messy chekcout -> https://stackoverflow.com/a/74154252/16480510

Also I saw one pattern where env's were managed in `.yml` file checkout -> https://stackoverflow.com/a/69619744/16480510



--- 

cc @carletex @rin-st @sverps please feel free to suggest other approaches/suggestions happy to give it another try, I am sure I might have missed something 🙌

carletex:
Hey @technophile-04 Thank you for this and your detailed explanation. Super helpful!

I feel that this got out of hand: my fault for suggesting the root config file haha. But we had to explore it and see what the issues were. I still feel that we want to aim for it, but let's do it in a more iterative / incremental way.

**Proposal**

Since the most important reason for this was to enable #202, I think we should go with the simple approach for now, and keep exploring the root / shared stuff for the monorepo.

Let's create a config file only for Next (replacing the .env.* files). This way we won't have any issues with Vercel are the changes will be minimal.

What do you think?

Thank you Shiv!

technophile-04:
In the latest commit [4c4e4a6](https://github.com/scaffold-eth/se-2/pull/231/commits/4c4e4a6a81349939faf801649a0595c10549f31a) now deployments are back to normal and no need to prebuilt stuff as mentioned here https://github.com/scaffold-eth/se-2/pull/231#issuecomment-1470438008 but just a very small catch 🙃 : 

Although this is one time config, you need to mention the path to `nexjts` dir : 
Screenshot 2023-03-16 at 2 40 32 PM

###  Reason : 
Since we using monorepo, vercel offically suggests to invoke `vercel cli` from root of monorepo and then pass in source code location to solve the problem mentioned here [4c4e4a6](https://github.com/scaffold-eth/se-2/pull/231/commits/4c4e4a6a81349939faf801649a0595c10549f31a)

Screenshot 2023-03-16 at 2 43 35 PM

Screenshot 2023-03-16 at 2 44 15 PM

Checkout Vercel docs monorepo FAQ -> https://vercel.com/docs/concepts/monorepos
 
 
 ### Now the flow for the target network is : 
 `scaffold.config.ts` is a single source of truth for both `nextjs` and `hardhat`. 
 
Just doing `yarn deploy` will deploy to network configured  in `scaffold.config.ts` and `yarn start` will use it as target network in frontend. 


### Flow for configuring env's : 
How env's works remain the same as the `main` branch. 

What's changed? 
I have hardcoded our API keys inside `scaffoldConfig` so that they can be used by both hardhat and `nextjs` for example provider key also removed .env.* from frontend since all handled by `scaffold.config` now 

reason: It's now saying that `Scaffold-eth` is "configured" with some handy api keys which you can access through `scaffold.config`. 

---

TODO : 
- [ ] Handle this properly https://github.com/scaffold-eth/se-2/pull/231#discussion_r1132625398
- [ ] If we plan to have global env's then implement them, checkout out the last section for resistance faced https://github.com/scaffold-eth/se-2/pull/231#issuecomment-1470438008

Happy to handle them both in different PRs, since this is already big and don't want us blocking on #202 🙌

package.json
```diff
@@ -25,12 +25,16 @@
     "next:check-types": "yarn workspace @se-2/nextjs check-types",
     "postinstall": "husky install",
     "precommit": "lint-staged",
-    "vercel": "yarn workspace @se-2/nextjs vercel",
-    "vercel:yolo": "yarn workspace @se-2/nextjs vercel:yolo"
+    "vercel:yolo": "vercel --build-env NEXT_PUBLIC_IGNORE_BUILD_ERROR=true"
   },
   "packageManager": "yarn@3.2.3",
   "devDependencies": {
+    "@types/node": "^18.15.3",
```
rin-st, line 32: Are we really need it? It seems everything works without it

technophile-04: Tysm !! yes yes we can remove it, it was added here [8f19b21](https://github.com/scaffold-eth/se-2/pull/231/commits/8f19b21cd837aaa9bc48927d74942dcece9940da) and since now we are not using `.env` at root we can completely remove it

packages/nextjs/next.config.js
```diff
@@ -9,6 +9,7 @@ const nextConfig = {
   eslint: {
     ignoreDuringBuilds: process.env.NEXT_PUBLIC_IGNORE_BUILD_ERROR === "true",
   },
+  transpilePackages: ["se-2"],
```
rin-st, line 12: could you explain please, why we need it?

technophile-04: if you look at https://github.com/scaffold-eth/se-2/pull/231#issuecomment-1470438008 in the section 

> Approaches explored :
> 2. 

TLDR;
 Since we have `scaffold.config.ts` outside the `nextjs` directory if you don't pass in this it gives this error -> https://github.com/scaffold-eth/se-2/pull/231#discussion_r1135317725
 
 The reason I have put `se-2` there is because we had `scaffold.config.ts` at `se-2` workspace and needed to transpile it during `next` build  : 
 
Screenshot 2023-03-16 at 9 16 32 PM

Although would highly recommend giving a quick read of the section  `Approaches explored ` especially the links  mentioned in that section 🙌

Also would love to hear if there are any alternatives / suggestions 🙌

carletex:
Great job @technophile-04 

I think we are not far, but still doesn't feel 100% right to me.

### 1. Vercel root directory

Screenshot 2023-03-16 at 2 40 32 PM

It might be like a little thing, but for me, it decreases the deploy experience by a bunch. I tried to search for options for the CLI / vercel.json config file, but it doesn't seem that you can provide a default `root directory` option. The `--cwd` option doesn't work either.

Not sure what we could try.

### 2. .env vars

> If we plan to have global env's then implement them, checkout out the last section for resistance faced https://github.com/scaffold-eth/se-2/pull/231#issuecomment-1470438008

I think we need .env vars for secrets. 

**Answer 1**
Right now on `scaffold.config.ts`:
```ts
const scaffoldConfig: ScaffoldConfig = {
  // ....
  deployerPrivateKey:
    "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
 // ..
} as const;
```

and then `const deployerPrivateKey = process.env.DEPLOYER_PRIVATE_KEY ?? scaffoldConfig.deployerPrivateKey;` on `hardhat.config.ts`. It's confusing, since we have the `deployerPrivateKey` on the config file and it's not clear that people need to create a `.env` file inside the hardhat folder

I think it should be:
```ts
const scaffoldConfig: ScaffoldConfig = {
  // ....
  deployerPrivateKey: process.env.deployerPrivateKey || 
    "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
 // ..
} as const;
```
So people can create a `.env` file in the root folder (we might have a `.env.example`) that it's meant for secrets. And secrets consumers just read `scaffoldConfig.deployerPrivateKey`. Same for `alchemyApiKey` and `etherscanApiKey`.

And make sure people understand that they should never use `scaffoldConfig` to store secrets directly (even if we are doing so :sweat_smile:)

Maybe we could pass the secrets/env vars like this to Vercel:

image

If people create/update an env var from the Vercel UI, they will take effect if they rebuild (they are being used on the scaffold.config.file)?

**Answer 2**

Use `scaffold.config.ts` only for public variables (targetNetwork && pollingInterval), and handle the secrets env vars in its own package:
- **Hardhat.**. Same as in main (.env file)
- **NextJS**.  We'd replace `.env.*` with the config file, but people could still use `.env` / `.env.local` / Vercel UI to set up secrets and use them in their API routes or whatever.

This way we don't mix config & secret env vars... and things might be simpler.

> Handle this properly https://github.com/scaffold-eth/se-2/pull/231#discussion_r1132625398

This is not a big deal for now. We could create an issue for that refactor.

---

Definitely having a root config file makes things a bit more complicated. I'll be up for it if we find a nice balance. But https://github.com/scaffold-eth/se-2/pull/231#issuecomment-1471517364 is still on my mind.

Thoughts?

technophile-04:
Tysm all for the review !! 🙌

> I think we are not far, but still doesn't feel 100% right to me.

Lol yes !, even I have this feeling it doesn't feel 100% right

I have made #246 which has config in `frontend` only as suggested in https://github.com/scaffold-eth/se-2/pull/231#issuecomment-1471517364 

I plan to explore more on putting `.config` at the root, Planning to also open some issues/issues on `vercel` github to find/suggest better solutions. Example: configuring project root as Carlos mention for `verceli-cli` even I tried finding hard and there were no options to configure it 


Marking this as a draft, Thanks 🙌

technophile-04:
I have created a discussion on Vercel for Pt.1 of https://github.com/scaffold-eth/scaffold-eth-2/pull/231#issuecomment-1472265152 -> https://github.com/orgs/vercel/discussions/2145

Closing this for now, since not under active development and until SE-2 cli is up

PR #239, opened 2023-03-13
fix: removes arrows for number inputs

rin-st:
Removes arrows for number inputs
image

https://stackoverflow.com/a/74043438

### DIFF
--- a/packages/nextjs/styles/globals.css
+++ b/packages/nextjs/styles/globals.css
@@ -7,6 +7,16 @@
   background: hsl(var(--b2));
 }
 
+@layer base {
+  input[type="number"]::-webkit-outer-spin-button,
+  input[type="number"]::-webkit-inner-spin-button,
+  input[type="number"] {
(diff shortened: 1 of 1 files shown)


carletex:
Hey Rinat!

we are not using any `type="number"` input in our code, right? I think you tweaked the exampleUI input for testing?

In any case, I think this is too opinionated, even for us haha.

rin-st:
Hey @carletex ! 👋 

Yep, we don't have input `type="number"` for now. But it's much easier to use it sometimes to work with numbers and other devs most likely will use it. I never met the app where that arrows weren't hidden :D so I thought it's a common thing.

Anyway, I had to create the issue first. Closing this?

carletex:
I agree @rin-st!

But we have to draw the line somewhere, and since we are also trying to be simple & minimal (and we are not using any input numbers for now) I think we can omit this for now.

thanks!

PR #242, opened 2023-03-14
Conventions on barrel files, import/exports, etc

carletex:
This is the first proposal for the things we discussed on https://github.com/scaffold-eth/se-2/issues/118

We were using a bunch of different conventions, and doing this feels really good :D But happy to re-thinking any of the conventions. 

I think being consistent is the key here (we could also write this kind of stuff in the GitHub Wiki / Contributing file or something).

What I did:
- **Barrel files** for scaffold-eth components/hooks/utils (no barrels for _app_ components)
  - I wasn't sure about deeper nested files (like components/scaffold-eth/Inputs/). I'm re-exporting them in the root scaffold-eth barrel file, so you'll import AddressInput as `import { AddressInput } from "~~/components/scaffold-eth";` instead of `import { AddressInput } from "~~/components/scaffold-eth/Input";`
- No more default exports (also for app components) => Switch to **named exports**.
- No "functions" => Converted to **arrow functions**
- No relative imports ("./" or "../") => use **~~** **imports**
- **No React** import when not needed
- **Sort imports** (using _eslint-plugin-simple-import-sort_). It'll give just a warning, but if you have eslint --fix on save in your IDE it'll sort them for you

Thoughts? 

Fixes https://github.com/scaffold-eth/se-2/issues/118


rin-st:
Awesome!

> I wasn't sure about deeper nested files ... so you'll import AddressInput as `import { AddressInput } from "~~/components/scaffold-eth";` instead of `import { AddressInput } from "~~/components/scaffold-eth/Input"`

I'm not sure too. I'd prefer second because it's
1) grouped
2) more explicit
3) from import you understand what you're exporting. I'll try to explain. Let's say we'll have Select component and Option subcomponent. Import will be `import { Option } from "~~/components/scaffold-eth"`.  When new dev will read the code he is unlikely understands what `Option` is. But `import { Option } from "~~/components/scaffold-eth/select"` is more clear. Yes, an example is not the best :D but I hope you understand me. Yes, we need to name components better, probably `SelectOption` or `Select.Option` etc, but when the codebase grows we will see that things.
---
> No "functions" => Converted to arrow functions

Google [don't like it](https://google.github.io/styleguide/tsguide.html#function-expressions), but since we don't use `this` or `arguments` and don't often need hoisting, arrow functions are good too (and easier to write).

rin-st (approved):

technophile-04:
Love the changes !!! 

> I wasn't sure about deeper nested files ... so you'll import AddressInput as `import { AddressInput } from "~~/components/scaffold-eth";` instead of `import { AddressInput } from "~~/components/scaffold-eth/Input"` 

I tend to like the first, reason(seeing it from using se-2 instead of contributing/maintaining perspective) : 

1. Its more declarative, also won't clutter the file which is importing it. Example : 
    ```ts
     import { AddressInput, Xyz } from "~~/components/scaffold-eth";

      ----

     import { AddressInput  } from "~~/components/scaffold-eth";
     import { Xyz } from "~~/components/scaffold-eth/xyz"

     ```
     As we grow in the future we will be adding components and this will clutter the file of the developer who is utilizing them. 

2. No doubt second is more explicit and easy to understand as Rinat explained but it has a very small downside that if we plan to move components internally eg : Moving `AddressInput` from `scaffold-eth/Input` -> `scaffold-eth/Address` then we have to change all the file utilizing it (sorry for bad example but I hope you all got the feelings) 

But yeah these are personal preferences and both ways make sense. 

> Sort imports (using eslint-plugin-simple-import-sort). It'll give just a warning, but if you have eslint --fix on save in your IDE it'll sort them for you

Love it !! Although had come problem configuring `eslint --fix on save` in my `Vscode` due to `prettier` and `eslint` extensions conflicting but it might be due to my previous setup. 

I was thinking maybe instead of sorting it on the `lint`  level maybe we sort it using `prettier` using https://github.com/trivago/prettier-plugin-sort-imports  ? So that you don't get a yellow squiggly line if you don't have your IDE configured properly and do it manually? Just thinking out loud and we can for sure ignore this suggestion !!

technophile-04 (approved):

PR #243, opened 2023-03-15
feat: sepolia alchemy rpc url

rin-st:
Alchemy now supports sepolia
https://www.alchemy.com/overviews/sepolia-testnet


carletex:
Thanks @rin-st 

I also updated hardhat-deploy, where I made a PR a few weeks ago to support sepolia (so we can verify contracts) https://github.com/wighawag/hardhat-deploy/pull/418.. but forgot to update it here :D

PR #248, opened 2023-03-19
Fix: Infer types literal for scaffold.config

technophile-04:
### Description : 
This is a small patch for #246, it seems that in that PR when we try to access `targetNetwork` types widening was happens 
Screenshot 2023-03-20 at 12 16 01 AM

Now : 
Screenshot 2023-03-20 at 12 17 01 AM

Had to upgrade prettier since support typescript `satisfies` was added in 2.8 checkout -> https://prettier.io/blog/2022/11/23/2.8.0.html

### DIFF
--- a/packages/nextjs/components/Footer.tsx
+++ b/packages/nextjs/components/Footer.tsx
@@ -23,7 +23,7 @@ export const Footer = () => {
                 <span>{ethPrice}</span>
               </div>
             )}
-            {Number(scaffoldConfig.targetNetwork.id) === hardhat.id && <Faucet />}
+            {(scaffoldConfig.targetNetwork.id as number) === hardhat.id && <Faucet />}
           </div>
           <SwitchTheme className="pointer-events-auto" />
         </div>
--- a/packages/nextjs/components/Footer.tsx
+++ b/packages/nextjs/components/Footer.tsx
@@ -23,7 +23,7 @@ export const Footer = () => {
                 <span>{ethPrice}</span>
               </div>
             )}
-            {scaffoldConfig.targetNetwork.id === hardhat.id && <Faucet />}
+            {Number(scaffoldConfig.targetNetwork.id) === hardhat.id && <Faucet />}
           </div>
           <SwitchTheme className="pointer-events-auto" />
         </div>
--- a/packages/nextjs/components/Footer.tsx
+++ b/packages/nextjs/components/Footer.tsx
@@ -3,14 +3,15 @@ import { CurrencyDollarIcon } from "@heroicons/react/24/outline";
 import { HeartIcon } from "@heroicons/react/24/outline";
 import { SwitchTheme } from "~~/components/SwitchTheme";
 import { Faucet } from "~~/components/scaffold-eth";
-import scaffoldConfig from "~~/scaffold.config";
 import { useAppStore } from "~~/services/store/store";
+import { getTargetNetwork } from "~~/utils/scaffold-eth";
 
 /**
  * Site footer
  */
 export const Footer = () => {
   const ethPrice = useAppStore(state => state.ethPrice);
+  const configuredNetwork = getTargetNetwork();
 
   return (
     <div className="min-h-0 p-5 mb-11 lg:mb-0">
@@ -23,7 +24,7 @@ export const Footer = () => {
                 <span>{ethPrice}</span>
               </div>
             )}
-            {(scaffoldConfig.targetNetwork.id as number) === hardhat.id && <Faucet />}
+            {configuredNetwork.id === hardhat.id && <Faucet />}
           </div>
           <SwitchTheme className="pointer-events-auto" />
         </div>
--- a/packages/nextjs/components/Footer.tsx
+++ b/packages/nextjs/components/Footer.tsx
@@ -11,7 +11,6 @@ import { getTargetNetwork } from "~~/utils/scaffold-eth";
  */
 export const Footer = () => {
   const ethPrice = useAppStore(state => state.ethPrice);
-  const configuredNetwork = getTargetNetwork();
 
   return (
     <div className="min-h-0 p-5 mb-11 lg:mb-0">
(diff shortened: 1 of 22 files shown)


packages/nextjs/components/Footer.tsx
```diff
@@ -23,7 +23,7 @@ export const Footer = () => {
                 <span>{ethPrice}</span>
               </div>
             )}
-            {scaffoldConfig.targetNetwork.id === hardhat.id && <Faucet />}
+            {Number(scaffoldConfig.targetNetwork.id) === hardhat.id && <Faucet />}
```
technophile-04, line 26: Had to use this hack because it was giving below error : 
Screenshot 2023-03-20 at 12 44 12 AM

Other option will be using `// @ts-ignore`

sverps: I think rather than solving it with javascript, you can solve it by simply casting it to `number`, this should widen its type:

```typescript
(scaffoldConfig.targetNetwork.id as number) === hardhat.id
```

packages/nextjs/scaffold.config.ts
```diff
@@ -19,6 +19,6 @@ const scaffoldConfig: ScaffoldConfig = {
   // It's recommended to store it in an env variable:
   // .env.local for local testing, and in the Vercel/system env config for live apps.
   alchemyApiKey: process.env.NEXT_PUBLIC_ALCHEMY_API_KEY || "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF",
-} as const;
+} satisfies ScaffoldConfig;
```
sverps, line 22: Nice, first time I see an actual use-case for `satisfies` :+1:

sverps (changes requested):
See comment about type casting the id to `number` to make it less hacky

sverps (approved):

packages/nextjs/components/Footer.tsx
rin-st: it works without `Number` or `as`
`scaffoldConfig.targetNetwork.id === hardhat.id && <Faucet />`

packages/nextjs/scaffold.config.ts
```diff
@@ -6,7 +6,7 @@ type ScaffoldConfig = {
   alchemyApiKey: string;
 };
 
-const scaffoldConfig: ScaffoldConfig = {
+const scaffoldConfig = {
   // The network where your DApp lives in
   targetNetwork: chains.hardhat,
```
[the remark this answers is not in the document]
sverps: That would probably widen the type here, which is what we want to avoid. Otherwise the pr that I'm working on wouldn't work. We want the type to be exact, not `number`

rin-st (approved):

carletex:
Ok, a recap so I understand.

We need to get the actual targetNetwork.id value (and not the widened type) for #202 to work. We solved that using the `satisfies` operator.

The issue then is that TS complains when we compare `scaffoldConfig.targetNetwork.id` (resolved to a specific number) with another specific number. "Types have no overlap". We need to assert the type (as number) to remove the error.

I don't love the casting here, seems counterintuitive, since you are working with the wagmi Chain type, but it's not really resolving as such. But I could live with it.

Just throwing another idea, and see if you all think it's worth implementing it:

As we had before https://github.com/scaffold-eth/se-2/pull/246, have a `getTargetNetwork` util than returns the selected network in config:

```ts
const getTargetNetwork = (): Chain => { }
```

So benefits I see:
- It solves this error.
- We can directly access the config when need the specific value (for #202)
- We could extend the targetNetwork value there (e.g. adding the colors)
- It's a wrapper, so if we ever change the way of defining the targetNetwork, we'd just need to change it here (e.g. we move the scaffold.config to the root).

Shiv already mentioned why he decided to remove this extra function in https://github.com/scaffold-eth/se-2/pull/246#discussion_r1139234630. But with this in mind, maybe our reasoning is different.

Anyway, not feeling super strong about this, just wanted to put it out there and see what you all think.

Happy to merge as it's.

sverps:
I like your idea of keeping a util function like this. It would still remain clear what this value gets, and it makes it more flexible and also easier for someone to tinker and make an interface that could allow switching network. :+1:

technophile-04:
Love the suggestion Carlos !!  100% agree with the benefits mentioned  

I also see people extending `NETWORKS_EXTRA_DATA` : 
```ts
const NETWORKS_EXTRA_DATA = {
  [chains.goerli.id]: {
    color: "#b8af0c",
    daiContractAddress : "0xabc..."
  },
  [chains.mainnet.id]: {
    color: "#ff8b9e",
    daiContractAddress : "0xdef..."
  },
 } 
```

and then using : 
```ts
const daiAddress = getTargetNetwork().daiContractAddress; 
```

packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -78,3 +79,16 @@ export function getBlockExplorerTxLink(network: Network, txnHash: string) {
 
   return `${blockExplorerTxURL}/tx/${txnHash}`;
 }
+
+/**
+ * @returns targetNetwork object consisting targetNetwork from scaffold.config and extra network meta data
+ */
+
+export function getTargetNetwork(): chains.Chain & Partial<TChainAttributes> {
```
technophile-04, line 87: using `Partial<TChainAttributes>` here since there might be the cases where : 
```js
return {
    ...configuredNetwork,
    ...NETWORKS_EXTRA_DATA[configuredNetwork.id] -> undefined,
  };
```

And typescript will suggest the developer to first check if propery mentioned in `TChainAttributes` exists and then use it. 

Suggestion are open 🙌

sverps (approved):

rin-st (approved):

carletex:
Great @technophile-04 , thanks!

Not a big deal, but since I've been obsessing with consistency lately, I removed the uses of `scaffoldConfig.targetNetwork` and now we are only using `getTargetNetwork()`

1. `const configuredNetwork = getTargetNetwork();` when multiple uses
2. `getTargetNetwork().id` when only 1 use.

This way we signal builders the way we prefer them to access the targetNetwork.

PR #255, opened 2023-03-23
feat: make contract and hook types work without a generated output

### LINKED ISSUE #254: Vercel builds fails
I connected this https://github.com/BuidlGuidl/hackathon-se2 to Vercel in the Vercel UI, so it deploys when I push to `main`.

It fails when I try to deploy (The initial deploy from the cli worked). It gets these errors:

image

I guess this is happening since #202 and we talked about this before.

- I just added the generated `hardhat_contract.ts` file to the repo (even tho my project won't use any contract). Not the best solution because:
 
    1. Doesn't make sense when your project doesn't use any contract (a lot of times you build dapps that don't use a custom contract)
    2. If you use a contract, the file will change all the time when you are developing locally (the address of the local contract will change on each deployment).

What solutions do we have?

sverps:
Fixes #254

A dev that doesn't need any custom contract can simply remove the `hardhat_contracts.ts` import in the scaffold.config, and set `contracts: null`.

Typescript won't complain due to types resolving differently when `contracts` is set to `null` in the config.


packages/nextjs/scaffold.config.ts
```diff
@@ -1,9 +1,23 @@
+import { Abi } from "abitype";
 import * as chains from "wagmi/chains";
+import contracts from "~~/generated/hardhat_contracts";
```
carletex, line 3: This still makes the app crash if you do `yarn start` and the file is not there.

How about:

```ts
let contracts;
try {
  contracts = require("~~/generated/hardhat_contracts");
} catch (error) {
  // Silently fail. The builder might not want to use a contract deployed from SE2
}
```

sverps: I'm guessing that will break the narrow typing. I don't think there's a way to get around that. Next best thing we can do is explain very well in the setup that if the user wants a contract, he has to deploy it first, and if he doesn't want it, he should modify this in the config.

packages/nextjs/scaffold.config.ts
carletex, line 1: Not introduced in this PR, but since we are dealing with the `hardhat_contract` file maybe we can take a look (happy to create another issue)

image

"Sometimes" (Haven't been able to reproduce consistently, but happen on app start up.. sometimes) this show.

It's weird because it's referencing the **JSON** file. Maybe because it gets created and the removed? Or because it has the same name as the **TS** one so the imports get confused?

packages/nextjs/scaffold.config.ts
carletex: > The aim of my PR is to make the changes a dev needs to make to use se-2 without contract as minimal as possible.

I have the same aim :)

What about?

```ts
// If you don't use custom contract for this SE-2 app:
// - Remove the import line below
// - Se "contracts: null" in the scaffoldConfig below
import contracts from "~~/generated/hardhat_contracts";

```

packages/nextjs/scaffold.config.ts
sverps: Hmmm, interesting. I think your guess is correct. When you deploy an update to your contract (and next is already running), then there's a short time where `hardhad_contacts.json` exists, and it tries importing that, but then it gets removed... etc.

I think we should fix it by not putting the `json` output in the next folder.
`"deploy": "hardhat deploy --export-all ../nextjs/generated/hardhat_contracts.json && yarn generateTsAbis",`

We could export the `json` file to `/packages/hardhat/generated` or something, and then use that as the source for the `generateTsAbis`. Makes a bit more sense that way.

carletex: > We could export the json file to /packages/hardhat/generated or something, and then use that as the source for the generateTsAbis. Makes a bit more sense that way

Yep! This would work :)

Thanks

packages/nextjs/scaffold.config.ts
```diff
@@ -19,6 +33,9 @@ const scaffoldConfig = {
   // It's recommended to store it in an env variable:
   // .env.local for local testing, and in the Vercel/system env config for live apps.
   alchemyApiKey: process.env.NEXT_PUBLIC_ALCHEMY_API_KEY || "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF",
+
+  // This is the contract data that was generated by "yarn deploy", or null if no local contracts are used
+  contracts,
```
carletex, line 38: When setting this to `null`, I get the following type error in the project:

image

packages/nextjs/scaffold.config.ts
sverps: Ah yes :+1: 

Would probably also be a good idea to update the readme a bit and explain how to use the scaffold.config

packages/nextjs/hooks/scaffold-eth/useAutoConnect.ts
carletex, line 1: @sverps Added this check (for the burnerwallet autoconnect), and another one in the Faucet, to avoid UI & console errors when no contracts.

packages/nextjs/scaffold.config.ts
carletex: db5d16bd28bdde35176faa7eb9da6dabef3549be

carletex: I leave updating the README for another PR. We also need to include some other stuff.

packages/nextjs/scaffold.config.ts
sverps: Hmmm, strange that this didn't fail before. I do see it now, though. Fixed it :+1:

packages/hardhat/scripts/generateTsAbis.ts
```diff
@@ -1,21 +1,22 @@
...
   }, "");
 
+  if (!fs.existsSync(TARGET_DIR)) {
+    fs.mkdirSync(TARGET_DIR);
+  }
   fs.writeFileSync(
-    `${GENERATED_PATH}.ts`,
+    `${TARGET_DIR}hardhat_contracts.ts`,
     prettier.format(`export default {${fileContent}} as const;`, { parser: "typescript" }),
   );
```
carletex, line 19: Do we want to keep the `json` file?

It might be confusing: sometimes I search for "hardhat_contract" in my IDE to check that file and now it shows both.

If we remove it, we wouldn't need to create a folder, It could be a tmp file inside the scripts folder that gets removed.

Or am I missing something?

sverps: Fair point. I figured there was no harm in keeping it and being able to see the actual output, but your point makes sense. I'll remove it. Might even make more sense to just write it in an actual temp dir. I'll fix it :+1:

sverps:
Found another confusing thing: the `generateTsAbi` script just spits out an error when called directly (which is logical since the json doesn't exist). This can be rather confusing for a user that is exploring which scripts exist in the `package.json`. He might not understand that this script is only used as part of the deploy and is not to be used directly (even though the name suggests it could).

So I removed the separate script for it and just added it to the deploy script instead.

carletex:
> So I removed the separate script for it and just added it to the deploy script instead.

Yep, nice one. Make sense!

This looks great. Merging it. 

Thank you @sverps

PR #257, opened 2023-03-24
Allow burner for configured chain

### LINKED ISSUE #235: Allow Burner Wallets on every network
Right you can only use burner wallet on the local chain.

We do a bunch of talks / onboardings and we use burner wallets to let people participate without having a wallet.

e.g. https://twitter.com/austingriffith/status/1633942545809166336

**Solution**

I propose having an _internal_ flag in `RainbowKitCustomConnectButton.tsx` on `tempAutoConnectConfig` like `burnerLocalOnly` (defaults to true) that when set to `false` will enable the burner for any chain that it's configured.

I don't want to include it on an `.env` file (or scaffold config file https://github.com/scaffold-eth/se-2/issues/230) yet, since it seems very niche for us and don't want to pollute the config. In any case, it'll be easy to do it in the future.

The files that we needed to tweak to make it work (in a dirty way :)) were:

- `burnerWalletConfig.ts`: The `hidden: ()`  prop
- `wagmiConnectors.tsx`: The available chains `burnerChains = configureChains`
- `BunerConnector.tsx`: `defaultBurnerChainId` and `getAccount`

technophile-04:
Fixes #235 

### Flow : 
If `enableBurnerWalletOnLoad` is set to `true` in `RainbowKitCustomConnectButton.tsx` then whenever you load dApp it will get connected to the burner account unless you have connected to your own wallet. 

Currently, we are always showing the option for a burner wallet when connecting : 
Screenshot 2023-03-24 at 2 16 10 PM

I think it will be great if we keep it since its a charm of scaffold-eth 😉 but if the developer want to forcefully don't show it then it he might do `hidden () => true` in `burnerWalletConfig` 

Should we move it to `scaffold.config` ? Didn't want to clutter the config as mentioned in #235 

https://s-taupe.vercel.app

Added comments below for the reason for the approach taken


packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
```diff
@@ -6,7 +6,7 @@ import { getTargetNetwork } from "~~/utils/scaffold-eth";
 
 // todo: move this later scaffold config.  See TAutoConnect for comments on each prop
 const tempAutoConnectConfig: TAutoConnect = {
-  enableBurnerWallet: true,
+  enableBurnerWalletOnLoad: true,
```
technophile-04, line 9: Should we move these options to `scaffold.config` ? 

It s only being used by `useAutoConnect` so maybe inline them at the top of  hook ? 

Confused a bit since I think people would like toggle this option

packages/nextjs/services/web3/wagmi-burner/BurnerConnector.ts
```diff
@@ -42,22 +42,19 @@ export class BurnerConnector extends Connector<StaticJsonRpcProvider, BurnerConn
 
     this.provider = new StaticJsonRpcProvider(chain.rpcUrls.default.http[0]);
     const account = await this.getAccount();
-    const chainId = await this.getChainId();
```
technophile-04, line 45: Remove this call and below check. 

Reason : 
Since this function was making an RPC call `eth_getChainId` it made the exprirence of connecting to burner wallet very laggy, 

Also if you look at the function `getChainId`
```js
async getChainId(): Promise<number> {
    const network = await this.provider?.getNetwork();
    const chainId = network?.chainId ?? this.options.defaultChainId;
    if (chainId == null) {
      throw new BurnerConnectorError(BurnerConnectorErrorList.chainIdNotResolved);
    }

    return Promise.resolve(chainId);
  }
```

It's falling back to `this.options.defaultChainId` which will be our `configuredNetwork`, which is single source of truth, hence its redundant to make an RPC call with provider network to get the chainId. 


https://user-images.githubusercontent.com/80153681/227477053-be2b9688-58a3-460f-87e2-b5b594bb6985.mov

carletex:
hey @technophile-04 thanks for this!

> Should we move these options to scaffold.config ?

Yes, let's move it to scaffold.config!

---

Not sure if I understand your approach: enableBurnerWallet => enableBurnerWalletOnLoad? And we also need the flag for local only, right?

The way I see it: this should be the config options (in `scaffold.config.ts`):

```ts
const scaffoldConfig = {
  // ....
  burnerWallet: {
    // If not enabled => it won't show anywhere
    enabled: true,
    // If enabled + autoconnect => its connect on page load. If false => you need to click in the modal to connect to it
    autoconnect: true,
    // If true => only works on localhost. If false => works on any network
    onlyLocal: true,
  },
} satisfies ScaffoldConfig;
```

Make sense?

packages/nextjs/services/web3/wagmiConnectors.tsx
```diff
@@ -70,7 +57,7 @@ export const wagmiConnectors = connectorsForWallets([
       braveWallet({ chains: appChains.chains }),
       coinbaseWallet({ appName: "scaffold-eth", chains: appChains.chains }),
       rainbowWallet({ chains: appChains.chains }),
-      burnerWalletConfig({ chains: burnerChains.chains }),
+      burnerWalletConfig({ chains: [appChains.chains[0]] }),
```
technophile-04, line 60: Removed `burnerChains` and inlined it with `appChains.chains[0]`  because `configuredNetwork` will be 0th network and also other stuff from `burnerChains` are not used like `burnerChains.providers` .... etc

packages/nextjs/scaffold.config.ts
```diff
@@ -19,6 +24,21 @@ const scaffoldConfig = {
...
+    // Only show the Burner Wallet when running on hardhat network
+    onlyLocal: false,
+  },
+
+  /**
+   * Auto connect:
+   * 1. If the user was connected into a wallet before, on page reload reconnect automatically
+   * 2. If user is not connected to any wallet:  On reload, connect to burner wallet if burnerWallet.enabled is true && burnerWallet.onlyLocal is false
+   */
+  walletAutoConnect: true,
```
carletex, line 41: Good catch!!! This is not only burner wallet, so it should live here, not inside the burner wallet config ;)

packages/nextjs/scaffold.config.ts
```diff
@@ -4,11 +4,16 @@ type ScaffoldConfig = {
...
+    enabled: boolean;
+    onlyLocal: boolean;
+  };
+  walletAutoConnect: boolean;
 };
 
 const scaffoldConfig = {
   // The network where your DApp lives in
-  targetNetwork: chains.hardhat,
+  targetNetwork: chains.goerli,
```
sverps: Don't forget to put it back to hardhat when PR is finalized :)

packages/nextjs/services/web3/wagmi-burner/burnerWalletConfig.ts
```diff
@@ -20,8 +25,17 @@ export const burnerWalletConfig = ({ chains }: BurnerWalletOptions): Wallet => (
   name: burnerWalletName,
   iconUrl: "https://avatars.githubusercontent.com/u/56928858?s=200&v=4",
   iconBackground: "#0c2f78",
-  //todo add conditions to hide burner wallet
-  // hidden: () => true,
+  hidden: () => {
+    if (!burnerConfig.enabled) {
```
sverps, line 29: What should the default behavior be (i.e. when enabled is not defined)?

If we want to enable it by default, maybe we should put

```typescript
if (burnerConfig.enabled === false) {
  return true;
}
```

Then an undefined setting will fall through to the rest of the check and eventually default to not be hidden.

packages/nextjs/scaffold.config.ts
technophile-04: Tysm 🙌 Fixed at [8c7fa2c](https://github.com/scaffold-eth/se-2/pull/257/commits/8c7fa2c00c60ff1b266c5b228d4b6b38580b5a68)

packages/nextjs/hooks/scaffold-eth/useAutoConnect.ts
```diff
@@ -61,7 +56,7 @@ const getInitialConnector = (
  * Automatically connect to a wallet/connector based on config and prior wallet
  * @param config
```
carletex, line 57: Remove?

packages/nextjs/hooks/scaffold-eth/useAutoConnect.ts
```diff
@@ -10,42 +10,37 @@ export type TAutoConnect = {
...
    */
-  enableBurnerWallet: boolean;
-  /**
-   * Auto connect:
-   * 1. If the user was connected into a wallet before, on page reload reconnect automatically
-   * 2. If user is not connected to any wallet:  On reload, connect to burner wallet
-   */
-  autoConnect: boolean;
+  burnerEnabled: boolean;
+  walletAutoConnect: boolean;
```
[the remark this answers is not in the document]
technophile-04: yes yes it should be removed

packages/nextjs/hooks/scaffold-eth/useAutoConnect.ts
technophile-04: yup thanks, fixed at [6ccdf5a](https://github.com/scaffold-eth/se-2/pull/257/commits/6ccdf5a85c83dc58f668fe66b459134fb69261f4)

packages/nextjs/scaffold.config.ts
```diff
@@ -37,6 +42,20 @@ const scaffoldConfig = {
   // .env.local for local testing, and in the Vercel/system env config for live apps.
   alchemyApiKey: process.env.NEXT_PUBLIC_ALCHEMY_API_KEY || "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF",
 
+  // Burner Wallet configuration
+  burnerWallet: {
+    // Set it to false to completely remove burner wallet from all networks
+    enabled: true,
+    // Only show the Burner Wallet when running on hardhat network
+    onlyLocal: false,
```
carletex, line 50: I think this should default to "true"? What do you think @technophile-04 ?

packages/nextjs/services/web3/wagmi-burner/burnerWalletConfig.ts
sverps: True. There's some workarounds (such as casting to `ScaffoldConfig` in places where we don't need the narrow typing), but probably not worth it while our config is still relatively simple.

packages/nextjs/scaffold.config.ts
technophile-04: Umm actually se-1 has it by default but yeah I feel since se-2 is bit more on side of production  build it makes sense to make it true, updated it at [3a96074](https://github.com/scaffold-eth/se-2/pull/257/commits/3a96074eaa9ad9c11bba0ed033a7c6dbe1aaf8be) we should probably mention it in the README.md

carletex:
Really nice job @technophile-04 

Thank you

packages/nextjs/scaffold.config.ts
technophile-04: Just wondering should I update the README.md in this PR itself ? I feel we should have a small dedicated section to `scaffold.config` in README.md a minimal section

PR #258, opened 2023-03-24
Сlickable logo

Bogarne:
The picture and text make up the logo, and it would be correct to make both clickable.


technophile-04:
Tysm @Bogarne  🚀 !!

technophile-04 (approved):

PR #260, opened 2023-03-24
Generate empty hardhat_contract.ts file on postinstall

carletex:
Just tinkering around. This PR might end on the trash :D

Since we merged #255 I was thinking about how we could improve it. I was thinking about generating an «empty» "hardhat_contracts.ts" file on `yarn install`.

This is a quick & dirty attempt to do so.

Does this make sense? I have some issues with `contract.types.ts` and am not sure if there is a workaround (is it breaking the narrow typing?)

cc @sverps


sverps:
Your approach of generating an empty record as sort of 'default' contracts file gave me some more ideas to try out. I pushed a commit to your branch, it seems to be working nicely now, and I believe this is exactly what we wanted to have (the project being in a compilable state after just running `yarn` without the need to mess in the config to build a dapp without contract) 

I even managed to remove the contracts import from the scaffold.config, this pass-through is obsolete with this solution.

The only thing that I didn't get to work was the `postinstall` script. For some reason, it doesn't trigger when I run `yarn`. I have to call it separately. So not sure what's going on there.

technophile-04:
> The only thing that I didn't get to work was the postinstall script. For some reason, it doesn't trigger when I run yarn. I have to call it separately. So not sure what's going on there.

Did some digging and I think Yarn has some issues, it only runs `postinstall` scripts  when there are `dependencies` changes. checkout this -> https://github.com/yarnpkg/yarn/issues/5476

---

I tried doing a fresh clone of the repo and ran `yarn` it generated the empty `hardhat_contracts.ts` file 🙌

carletex:
This looks great @sverps Thank you!

There is only one thing left

image

The tweak I made for the Faucet & the autoconnect don't work, since `contracts = {}` is not a falsy. I guess we could:

a) `Object.keys(obj).length` check in the condition.
b) Default file being something like: `export default false as const;` (not sure how easy would be to fix the types)

I did **a** in 9913f5b21c7a8b515ca45d7018d81c90a3162fab, but if the types **b** are easily fixable, happy to switch. 

> The only thing that I didn't get to work was the postinstall script. For some reason, it doesn't trigger when I run yarn. I have to call it separately. So not sure what's going on there.
> Did some digging and I think Yarn has some issues, it only runs postinstall scripts when there are dependencies changes. checkout this -> https://github.com/yarnpkg/yarn/issues/5476
> I tried doing a fresh clone of the repo and ran yarn it generated the empty hardhat_contracts.ts file raised_hands

Yep, also noticed this. In our case, I think we don't mind... since it'll be generated on a fresh clone.

carletex:
Merging this since we are about to announce the SE-2 hackathon, and want this to be in :)

We can rethink some stuff later.

PR #264, opened 2023-03-27
refactor: small improvement for empty contracts

sverps:
@carletex I cleaned it up a little and the empty contracts is now `null`, that way we no longer need to check `Object.keys().length`, which is a bit cleaner.


technophile-04:
Looks clean !! Tested out and it works nicely 🙌

technophile-04 (approved):

carletex:
Amazing, thanks!!!

PR #266, opened 2023-03-29
use targetNetwork blockExplorer link for AddressComponent

### LINKED ISSUE #265: Block explorer on Address Component
On the Address component, we have

```ts
const blockExplorerLink = (address: string, blockExplorer?: string) =>
  `${blockExplorer || "https://etherscan.io/"}address/${address}`;
```

and it's always called like:

```ts
const explorerLink = blockExplorerLink(address, blockExplorer);
```

Leveraging on the block explorer that we pass as a prop to the Address component.

I think we should use the block explorer (defined on wagmi) for the current network (I don't see the value of pointing to a different network, except in local testing)

I did this quick & dirty fix in a project: 
https://github.com/BuidlGuidl/hacker-houses-streams/commit/f15ba44b39f9c33780c76477562cf2d382a7904b

*** mmight be exactly what we need... or not :D), but you can take it as a reference

technophile-04:
Fixes #265 

- Liked the approach mentioned by Carlos in https://github.com/scaffold-eth/se-2/issues/265#issue-1645509779 so used it. 

- Moved this function to `network.ts` below `getBlockExplorerTxLink` as `getBlockExplorerAddressLink`, don't know if it was worth it.....Should we move it back to the Address component?  Its seems little consistent in `network.ts`

Also was wondering if should we handle https://github.com/scaffold-eth/se-2/pull/231#discussion_r1132625398 in this PR itself?

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Address.tsx
+++ b/packages/nextjs/components/scaffold-eth/Address.tsx
@@ -5,13 +5,10 @@ import Blockies from "react-blockies";
 import { CopyToClipboard } from "react-copy-to-clipboard";
 import { useEnsAvatar, useEnsName } from "wagmi";
 import { CheckCircleIcon, DocumentDuplicateIcon } from "@heroicons/react/24/outline";
-
-const blockExplorerLink = (address: string, blockExplorer?: string) =>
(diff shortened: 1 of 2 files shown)


rin-st (approved):

carletex:
Looking great! Thanks <3

PR #267, opened 2023-03-29
Better msg for tx

austintgriffith:
instead of saying "mining"


austintgriffith:
lol wtf 
[the ol' wraparound distance]

PR #270, opened 2023-03-30
Use cacheTime with usePrepare

technophile-04:
This is solution 2 to #269 

### Context : 
We are using a contract from https://github.com/scaffold-eth/se-2/pull/215#issue-1601764822 check out its description on what the contract does. 

### Problem with `usePrepareContract` : 

https://user-images.githubusercontent.com/80153681/228768624-c5cf09f6-19d9-49d8-85c3-cb2bcc6b5721.mov

I have you try to `checkout` immediately after `checkIn` its says not checked In although we have checked in. 

### Reason for the problem : 

https://user-images.githubusercontent.com/80153681/228769555-ed2b85c8-8373-435a-b252-ecf75c1b2640.mov

Since we are using `usePrepareContract`, you see in above video when the app loads it seems that `usePrepareContract` is making an `eth_estimateGas` call which returns error `Not check in` initially which is true since we have not checked in but after you `checkIn` and try to `checkout` immediately it seems that `usePrepare` uses previous result as cached and saying `Not checkin`....but if you refresh the app and try `checkout`it work since now on app load `eth_estimateGas` will pass without error. 

### Solution : 
Disabling cache by setting [`cacheTime : 0`]((https://wagmi.sh/react/prepare-hooks/usePrepareContractWrite#cachetime-optional)) for wagmi's prepared config. Ideally this should disable `cache` for prepared config and `usePreapreContract` should make a new request  `eth_estimateGas` every time you click `checkout` without relying on any previous result. 

For some reason this seems not to be working :(, Am I missing something?

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn lint-staged --verbose
+# yarn lint-staged --verbose
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-# yarn lint-staged --verbose
+yarn lint-staged --verbose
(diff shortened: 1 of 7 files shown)


technophile-04:
### Reason why this is not working : 

The reason for it to be not working is because `config` for `usePrepareContractWrite` is not changing because of this it not making any new requests.... Thanks to Carlos 🙌

PR #272, opened 2023-03-30
Add useScaffoldEventHistory hook

### LINKED ISSUE #142: useScaffoldEventsRead hook

damianmarti:
useScaffoldEventHistory hook allows to get the events from a contract.

Only contractName and eventName are mandatory.

You can choose if you want to get the block, transaction or receipt data with the blockData, transactionData and receiptData flags. Default to false to avoid RPC calls.

 ~~fromBlock default value is set to 1000 blocks before the current block.~~ Edited: fromBlock is mandatory now.

If you want to get all the events from the contract, set it to the block number when the contract was deployed. 

Use filters to filters the events by the indexed parameters.

Example usage:

```
  const { data: events, isLoading: isLoadingEvents } = useScaffoldEventHistory({
    contractName: "YourContract",
    eventName: "Withdraw",
    fromBlock: 0,
    blockData: true,
    transactionData: true,
    receiptData: true,
    filters: {
    	to: "0x5dCb5f4F39Caa6Ca25380cfc42280330b49d3c93",
    }
  });
```

The events are not re-fetched automatically. If you want to keep the events data updated, you have to call the hook again.

Should we add support for automatic re-fetching on the hook?

closes #142


packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
```diff
@@ -0,0 +1,124 @@
...
+        console.log("contract not found");
+        return;
+      }
+      try {
+        if (!fromBlock) {
+          fromBlock = await provider.getBlockNumber() - 1000;
+        }
+
+        const fragment = contract.interface.getEvent(eventName);
+        if (!fragment) {
```
sverps, line 61: Since our `eventName` is strongly typed to only allow valid event names, the fragment should always be defined, right? Maybe this condition can be improved a bit (or omitted by asserting its type?).

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
```diff
@@ -0,0 +1,124 @@
...
+  contractName: TContractName;
+  eventName: TEventName;
+  fromBlock?: number;
+  filters?: any;
+  blockData?: boolean;
+  transactionData?: boolean;
+  receiptData?: boolean;
+}) => {
+  const [events, setEvents] = useState(<any>[]);
+  const { data: deployedContractData } = useDeployedContractInfo(contractName);
```
sverps, line 40: grab the loading state from here and pass it outwards

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
```diff
@@ -0,0 +1,124 @@
...
+          }
+          newEvents.push(log);
+        }
+        setEvents(newEvents.reverse());
+      }
+      catch (e) {
+          console.log(e);
+      }
+    }
+    readEvents();
```
sverps, line 120: Only call `readEvents` if `useDeployedContractInfo` is not loading

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
```diff
@@ -0,0 +1,124 @@
...
+ * @param config - The config settings
+ * @param config.contractName - deployed contract name
+ * @param config.eventName - name of the event to listen for
+ * @param config.fromBlock - the block number to start reading events from (default: 1000 blocks before current block)
+ * @param config.filters - filters to be applied to the event (parameterName: value)
+ * @param config.blockData - if set to true it will return the block data for each event (default: false)
+ * @param config.transactionData - if set to true it will return the transaction data for each event (default: false)
+ * @param config.receiptData - if set to true it will return the receipt data for each event (default: false)
+ */
+export const useScaffoldEventRead = <
```
sverps, line 19: Would `useScaffoldEventHistory` make more sense here? Not entirely sure, but that might make the distinction between this hook and the `useScaffoldEventSubscriber` clearer I think.

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
sverps, line 1: I think there  might be an issue with your autoformatter. Prettier should automatically format this according to the `.prettierrc.json` config file. Can you check if you can have you IDE format your files on save?

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
```diff
@@ -0,0 +1,124 @@
...
+        setEvents(newEvents.reverse());
+      }
+      catch (e) {
+          console.log(e);
+      }
+    }
+    readEvents();
+  }, [provider, fromBlock, contractName, eventName, deployedContractData?.address, contract]);
+
+  return events;
```
sverps, line 123: Would be good to align the return value here with the other hooks.

Something like this maybe (then we can add some additional ways to interact with the hook as well and have it update, etc.):

```typescript
return {
  data: events,
  isLoading: ...,
  refetch: ...,
}
```

sverps:
Hey @damianmarti , thanks for the PR. This hook looks really useful! Made some small comments/suggestions for additional improvements, but it's already looking good so far! :+1:

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  useEffect(() => {
+    async function readEvents() {
+      if (!contract) {
+        console.log("contract not found");
+        return;
+      }
+      try {
+        if (!fromBlock) {
+          fromBlock = await provider.getBlockNumber() - 1000;
```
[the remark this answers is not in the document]
technophile-04: won't `fromBlock = 0` be a bit overkill default value 😅 ? Also, I think it will increase fetching time on chains other than a hardhat

Maybe We mention it in JSDoc comments that by default it starts filtering from the last 1000 blocks

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
technophile-04: > Would useScaffoldEventHistory make more sense here? Not entirely sure, but that might make the distinction between this hook and the useScaffoldEventSubscriber clearer I think.

++, `useScaffoldEventHistory` love this name !!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
technophile-04: +++ I think isLoading will be must

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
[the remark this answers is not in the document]
sverps: Yeah, true. Not sure what the best choice is. Maybe no default and make it explicit? I'm just not a fan of magic numbers in code, and 1000 feels a bit like it, could just as well be 50 or 100 or 500... So what about making the param mandatory? (just thinking out loud here :sweat_smile: )

technophile-04:
> Hey @damianmarti , thanks for the PR. This hook looks really useful! Made some small comments/suggestions for additional improvements, but it's already looking good so far! 👍

+++ Tested it and works really nicely, Damu always nails it !!

--- 

### Some thoughts : 

Don't know how easy/hard would it be to add `watch : true` / `pollTime` which which fetch `onBlock` or on the interval mentioned similar to [`useEventListnerHook`](https://github.com/scaffold-eth/eth-ui/blob/next/packages/eth-hooks/src/hooks/useEventListener.ts) in se-1. I see there they are using `useQuery` which gives the power of `pollTime` by default which we might to think a bit and do it with raw setInterval (I don't have any idea just thinking out loud ). 


Nevertheless I even really like current version of this hook and solves a lot of things already.

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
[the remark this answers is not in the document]
damianmarti: As @technophile-04 said, I didn't want to use fromBlock = 0 as default to avoid a long fetching time if you are using it on a test or mainnet.

Maybe we can set the fromBlock param mandatory, so the user has to think about what block number he needs to set and suggest that if you want all the emitted events, you should set the fromBlock equal to the block from the contract deployment.

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
damianmarti: Yes! You are right! What do you mean by "or omitted by asserting its type?"?

damianmarti:
> > Hey @damianmarti , thanks for the PR. This hook looks really useful! Made some small comments/suggestions for additional improvements, but it's already looking good so far! +1
> 
> +++ Tested it and works really nicely, Damu always nails it !!
> 
> ### Some thoughts :
> Don't know how easy/hard would it be to add `watch : true` / `pollTime` which which fetch `onBlock` or on the interval mentioned similar to [`useEventListnerHook`](https://github.com/scaffold-eth/eth-ui/blob/next/packages/eth-hooks/src/hooks/useEventListener.ts) in se-1. I see there they are using `useQuery` which gives the power of `pollTime` by default which we might to think a bit and do it with raw setInterval (I don't have any idea just thinking out loud ).
> 
> Nevertheless I even really like current version of this hook and solves a lot of things already.

I think that maybe I can finish the hook without the watch feature, merge it, and then I can try to add the watch feature if we want. What do you think?

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
damianmarti: Should this bad formatting be checked by the linter at the pre-commit hook?

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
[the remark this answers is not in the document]
damianmarti: fromBlock is mandatory now!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
damianmarti: Removed! It was throwing an exception anyway, so the catch block can catch it if it's needed.

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
damianmarti: Added!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
damianmarti: Added!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
damianmarti: Changed!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
damianmarti: Fixed!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
damianmarti: Added isLoading!

damianmarti:
Thanks for all the feedback!! 

I already made the changes, please let me know if there is something else to improve! Thanks!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -0,0 +1,122 @@
...
+          topics.push(...indexedTopics);
+        }
+
+        const logs = await provider.getLogs({
+          address: deployedContractData?.address,
+          topics: topics,
+          fromBlock: fromBlock,
+        });
+        const newEvents = [];
+        for (let i = 0; i < logs.length; i++) {
```
sverps, line 84: If we start from the end and count down, then we won't need to do `newEvents.reverse()` at the end. Small optimization.

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -0,0 +1,122 @@
...
+    abi: deployedContractData?.abi,
+    signerOrProvider: provider,
+  });
+
+  useEffect(() => {
+    async function readEvents() {
+      if (!contract) {
+        console.log("contract not found");
+        return;
+      }
```
sverps, line 55: I think we should probably set `isLoading` to false here before returning. This condition will only be met if we finished loading the `deployedContractData` and this data doesn't result in a valid contract. All that we were waiting for is settled here. Maybe we can add an `error` state that we set and return from the hook, rather than the console.log.

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
sverps: > What do you mean by "or omitted by asserting its type?

Hadn't opened the PR in vscode yet, so wasn't sure if typescript would complain about fragment possibly being undefined or not. Guess it wasn't an issue, based on your most recent changes, nice :+1:

packages/nextjs/hooks/scaffold-eth/useScaffoldEventRead.ts
sverps: Yeah the linter should automatically run and format touched files in the pre-commit hook (`yarn next:lint --fix`)

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
technophile-04: Yup the `error` state will be really great, also I think instead we should use `console.error` cause thats what are preferring while logging error 🙌

technophile-04:
> I think that maybe I can finish the hook without the watch feature, merge it, and then I can try to add the watch feature if we want. What do you think? 

Yes yes, even the current state of hooks is really helpful and great

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
damianmarti: Added the error state, set isLoading to false in error cases, and changed to use console.error.

I had to change the useDeployedContractInfo hook, to set the isLoading flag to false if the contract code is not found. https://github.com/scaffold-eth/se-2/pull/272/commits/fa6099cb5d5e07b2ccb32127c712cb8aba1dcc7f

technophile-04 (approved):
Looks really nice ! Tested and everything works like a charm! 🙌

Added few nitpicks which I thnk can be ignored

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -38,6 +38,7 @@ export const useScaffoldEventHistory = <
 }) => {
   const [events, setEvents] = useState(<any>[]);
   const [isLoading, setIsLoading] = useState(true);
+  const [error, setError] = useState(<string>"");
```
sverps, line 41: No need to use assertion here, since `""` is already of type `string`.

Personally, I prefer to let states be `undefined` rather than "empty", otherwise there are cases where it's difficult to distinguish between an actual empty array/string and a state that is waiting for an async proceess to complete.

So I'd go for this:
```typescript
const [error, setError] = useState<string>();
```

That tells the type it will be `string | undefined`.

Same comment applies to line 39, where I'd go for

```typescript
const [events, setEvents] = useState<any[]>();
```

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
```diff
@@ -30,16 +30,22 @@ export const useDeployedContractInfo = <TContractName extends ContractName>(cont
...
+
+  if (status === ContractCodeStatus.NOT_FOUND) {
+    isLoading = false;
+  }
+
   return {
     data: status === ContractCodeStatus.DEPLOYED ? deployedContract : undefined,
-    isLoading: status === ContractCodeStatus.LOADING,
+    isLoading: isLoading,
   };
```
sverps, line 50: This change can be reverted, I believe. If status equals `NOT_FOUND`, then `isLoading` would already be `false`, so no need for this additional state check. Setting `isLoading` in the returned object directly is fine.

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -49,74 +50,82 @@ export const useScaffoldEventHistory = <
...
-        return;
-      }
       try {
-        const fragment = contract.interface.getEvent(eventName);
-        const emptyIface = new ethers.utils.Interface([])
-        const topicHash = emptyIface.getEventTopic(fragment)
-        const topics = <any>[topicHash];
+        if (!deployedContractData || !contract) {
+          console.error("Contract not found");
+          setEvents([]);
```
sverps, line 56: In relation to my other comment, here we can then do `setEvents(undefined)` to make the difference between "no info could be determined" and "we found the results, but it's empty"

sverps:
Added my last comments. I think after that it should be good for me :+1: 
Nice work, @damianmarti 

@carletex would be good if you had a look at this as well. Nice hook to add to the project, I think

carletex:
> @carletex would be good if you had a look at this as well. Nice hook to add to the project, I think

@sverps Oh yes, sorry I didn't say anything before. Damu implemented this for https://github.com/BuidlGuidl/hacker-houses-streams, a project where we needed to list historical events...and told him to make a PR here!

So @sverps || @technophile-04 feel free to merge when ready!

carletex:
And great work @damianmarti <333

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -0,0 +1,131 @@
...
+  const contract = useContract({
+    address: deployedContractData?.address,
+    abi: deployedContractData?.abi,
+    signerOrProvider: provider,
+  });
+
+  useEffect(() => {
+    async function readEvents() {
+      try {
+        if (!deployedContractData || !contract) {
```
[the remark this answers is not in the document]
damianmarti: Changed!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
damianmarti: Changed! Thanks!

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
damianmarti: Reverted!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
damianmarti: Changed!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -0,0 +1,128 @@
...
+            log: logs[i],
+            args: contract.interface.parseLog(logs[i]).args,
+            block: block,
+            transaction: transaction,
+            receipt: receipt,
+          }
+          newEvents.push(log);
+        }
+        setEvents(newEvents);
+        setError("");
```
sverps, line 108: Should be `setError(undefined)`

damianmarti: No, I really appreciate all your comments, so in the end, we have an excellent-quality code!!! 

Changed!

packages/nextjs/components/example-ui/ContractData.tsx
```diff
@@ -30,6 +35,19 @@ export const ContractData = () => {
...
   });
 
+  const {
+    data: events,
+    isLoading: isLoadingEvents,
+    error: errorReadingEvents,
+  } = useScaffoldEventHistory({
+    contractName: "YourContract",
+    eventName: "GreetingChange",
+    fromBlock: Number(process.env.NEXT_PUBLIC_DEPLOY_BLOCK) || 0,
```
technophile-04, line 45: Maybe we can remove the use of `process.env` and put a straight 0 to keep things consistent?

Reason : 
We currently don't have  any `.env.development` or `.env.production`  

Reason to have them : 
Maybe are we hinting to the user to store the deploy block in `env`  so that they can use same it in another place

Just confused a bit, maybe we keep it as it is. 

cc @sverps , @damianmarti please feel free to merge I think the hook is really looking solid 🔥

technophile-04 (approved):

packages/nextjs/components/example-ui/ContractData.tsx
damianmarti: It was added by @carletex at the hacker houses build.

I think that it's a nice hint to the user to use the deploy block in the env file.

technophile-04: Got it makes sense 🙌

technophile-04:
> So @sverps || @technophile-04 feel free to merge when ready!

Merging this, look really awesome 🚀 !!

PR #282, opened 2023-04-04
Un-gitignore hardhat_contracts and rename it to deployed_contracts

### LINKED ISSUE #278: un-gitignore hardhat_contracts.ts
- We can remove the postinstall script.
- When you deploy your contracts for the your dApp, you need the file to be committed anyway (for Vercel auto-deployements at least)
- Maybe we can also rename it to `deployed_contracts.ts`? Already thinking about foundry....
- It will fix https://github.com/scaffold-eth/se-2/issues/275

The only CONS that I see is that you will always see change on that file when tinkering locally with the smart contracts. But not a big deal.

carletex:
Fixes #278 

- hardhat_contracts => deployed_contracts
- add it to the repo. I'm starting to like this, because it makes people more conscious about where the contract info is.
- remove postinstall script


packages/hardhat/scripts/generateTsAbis.ts
```diff
@@ -14,7 +14,7 @@ function main() {
     fs.mkdirSync(TARGET_DIR);
   }
   fs.writeFileSync(
-    `${TARGET_DIR}hardhat_contracts.ts`,
+    `${TARGET_DIR}deployed_contracts.ts`,
     prettier.format(`export default {${fileContent}} as const;`, { parser: "typescript" }),
```
technophile-04, line 18: `NITPICK:`

Maybe we do : 
```ts
 fs.writeFileSync(
    `${TARGET_DIR}deployed_contracts.ts`,
    prettier.format(`const contracts = {${fileContent}} as const; \n\n export default contracts`, {
      parser: "typescript",
    }),
  );
```

#### Reason : 

The initial state of `deployed_contracts.ts` is :
```ts
const contracts = null;
export default contracts;
```

Final state of `deplyoyed_contracts.ts` after running `yarn deploy` : 
```ts
export default {
  //...
} as const
```

It would be great if keep final state similar to inital, just to keep it consistent 

Final  state : 
```ts
const contracts = {
// ...
} as const 

export default contracts;
```

We also get the added benefit of auto import which might be helpful if people want to get abi : 
Screenshot 2023-04-04 at 4 13 02 PM


Maybe we can change name from `contracts` -> `contractsAbis` ?

technophile-04 (approved):
> add it to the repo. I'm starting to like this, because it makes people more conscious about where the contract info is.

+++, love this approach

packages/hardhat/scripts/generateTsAbis.ts
sverps: The auto-import would be very helpful indeed :+1: 

No strong opinion on the name, but kinda make sense, not sure what users would find the most intuitive.

sverps (approved):

packages/hardhat/scripts/generateTsAbis.ts
carletex: > It would be great if keep final state similar to inital, just to keep it consistent

I buy that! Fixed on 819b4df1e4dca9ac0535b94164a536328633f240

> We also get the added benefit of auto import which might be helpful if people want to get abi :
> The auto-import would be very helpful indeed

I think we should recommend people to use the `useDeployedContractInfo` hook.

> Maybe we can change name from contracts -> contractsAbis ?

I could see `deployedContracts`. But `contractsAbis` doesn't seem right, since the file stores more than that (deployed network, deployed address)

technophile-04: > I think we should recommend people to use the useDeployedContractInfo hook.

yes yes 💯

> I could see deployedContracts. But contractsAbis doesn't seem right, since the file stores more than that (deployed network, deployed address)

Agreed, I was just suggesting a different name because it has conflicts with `~/utils/scaffold-eth/contracts`,  BTW I love `deployedContracts` name

carletex: Changed the var name and also the file name.

`deployedContracts.ts` is more aligned to our naming conventions, right?

technophile-04: Lol don't know why `deployedContracts.ts` makes me feel uneasy 😂 maybe I am habitat to seeing a file with `_` in the generated directory. 

But yeah `deployedContracts.ts` seems more aligned 🙌

PR #283, opened 2023-04-05
Add useScaffoldContract hook

damianmarti:
Wrapper around wagmi useContract to load the contract from the contract name.

Mandatory parameter: contractName

Optional parameter: signerOrProvider, if not set, it will use the default provider

Example usage:

`const { data: yourContract } = useScaffoldContract({ contractName: "YourContract" });`


sverps (approved):
Looks nice :+1: 
No comments this time :sweat_smile:

packages/nextjs/hooks/scaffold-eth/useScaffoldContract.ts
```diff
@@ -0,0 +1,32 @@
+import { ethers } from "ethers";
+import { useContract, useProvider } from "wagmi";
+import { useDeployedContractInfo } from "~~/hooks/scaffold-eth";
+import { ContractName } from "~~/utils/scaffold-eth/contract";
+
+/**
+ * @dev Gets a deployed contract from contractName and returns a contract instance
```
technophile-04, line 7: This makes sense but  maybe we write : 
`Gets a deployed ethers.Contract instance from contractName`? 

Reason : 
it has small ambiguity with `useDeployedContractInfo` : 
https://github.com/scaffold-eth/se-2/blob/dedd249fed95a858845164424529564b6b51949b/packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts#L8

technophile-04:
added a small `nitpick` comment 😅

It looks really good to me 🔥

packages/nextjs/hooks/scaffold-eth/useScaffoldContract.ts
sverps: I think the comment for the new hook is fine, but can maybe be slightly reworded like so:
`Gets a deployed contract by contract name and returns a contract instance`

The comment for the `useDeployedContractInfo` is probably the one that we should update, because it doesn't return the deployed contract, but rather the contract's definition/info. So there we could put:
`Gets the matching contract definition from the contracts file generated by \`yarn deploy\` .`

technophile-04: Thanks !! Yup its makes more sense to update `useDeployedContractInfo` comment 

updated it at [b63a072](https://github.com/scaffold-eth/se-2/pull/283/commits/b63a072aa2aa649087b6b7d56e4488927c8412ef)

damianmarti:
Thanks @sverps @technophile-04 !!!

Issue #232, opened 2023-03-10
TypeScript styleguide


sverps:
I propose choosing and following a style guide to get more consistency throughout the code.

My choice would be to follow [google's style guide](https://google.github.io/styleguide/tsguide.html).

Some possible exceptions:

- To avoid having to rename practically all files, I think we could opt to disregard the filename rule.
- [anything else?]

rin-st:
Agree with everything @carletex wrote.
>Interface vs Types

I prefer to use Types
---
Again, regarding [default export](https://google.github.io/styleguide/tsguide.html#exports)

Next.js supports lazy loading for named exports using `next/dynamic` [docs](https://nextjs.org/docs/advanced-features/dynamic-import#with-named-exports).

So, if you're ok to write one additional `then` for potential lazy loading, we probably don't need default exports anymore

ByteAtATime:
My personal opinion, I think we can leave default exports as-is, since it's very common in React.

PR #292, opened 2023-04-07
Fix autoConnect when contract not deployed

technophile-04:
### Issue : 
It seems that due to the below check autoConnect does not work if you use SE-2 that doesn't use deployed contract : 
https://github.com/scaffold-eth/se-2/blob/ce83d166df62083a3aac86a6dda1802737a02fa8/packages/nextjs/hooks/scaffold-eth/useAutoConnect.ts#L68

### Demo not working : 
Selected Mainnet as `targetNetwork`

https://user-images.githubusercontent.com/80153681/230662110-f12dac45-13b0-444b-bcb2-0c552a10504a.mov

--- 

Fixed by removing the `if(contract)` from both `useAutoConnect` & `Faucet`, it seems that check was added due console getting bloated with errors and warning but I think it was fixed in other commits 

checkout OG comment : https://github.com/scaffold-eth/se-2/pull/255/files/43f82e58bab2f624b60f520e68cfff749fe541e7#r1147414048

#### Demo with only `yarn chain` runnig and deployed contract is null : 

https://user-images.githubusercontent.com/80153681/230661918-c66089cb-5b29-4914-a59d-7588a0818ebe.mov

#### Demo with `yarn chain` not running : 

https://user-images.githubusercontent.com/80153681/230662002-1fd3e69f-b790-4e40-b534-db8f6e537a1f.mov

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,6 @@ If you want to redeploy to the same production URL you can run `yarn vercel --pr
 
 **Hint**: We recommend connecting the project GitHub repo to Vercel so you the gets automatically deployed when pushing to `main`
 
-**Hint**: If you don't have any contract deployed through SE-2 and plan to just deploy NextJS app maybe only with "connect wallet" functionality you can do so by just setting `targetNetwork : chains.mainnet` or your preferred network in `packages/nextjs/scaffold.config.ts`, rest all instruction remains same as [above](#deploying-your-nextjs-app)
-
 ## Hook Example
 
 - [useScaffoldContractRead](#usescaffoldcontractread)
--- a/README.md
+++ b/README.md
@@ -123,6 +123,8 @@ If you want to redeploy to the same production URL you can run `yarn vercel --pr
 
 **Hint**: We recommend connecting the project GitHub repo to Vercel so you the gets automatically deployed when pushing to `main`
 
+**Hint**: If you don't have any contract deployed through SE-2 and plan to just deploy NextJS app maybe only with "connect wallet" functionality you can do so by just setting `targetNetwork : chains.mainnet` or your preferred network in `packages/nextjs/scaffold.config.ts`, rest all instruction remains same as [above](#deploying-your-nextjs-app)
+
 ## Hook Example
 
 - [useScaffoldContractRead](#usescaffoldcontractread)
--- a/packages/nextjs/components/scaffold-eth/Faucet.tsx
+++ b/packages/nextjs/components/scaffold-eth/Faucet.tsx
@@ -25,23 +25,18 @@ export const Faucet = () => {
   const faucetTxn = useTransactor(signer);
 
   useEffect(() => {
-    const getFaucetAddress = async () => {
-      try {
-        if (provider) {
-          const accounts = await provider.listAccounts();
-          setFaucetAddress(accounts[FAUCET_ACCOUNT_INDEX]);
-        }
-      } catch (error) {
+    provider
+      ?.listAccounts()
+      .then(accounts => setFaucetAddress(accounts[FAUCET_ACCOUNT_INDEX]))
+      .catch(error => {
         notification.error(
           <>
             <p className="font-bold mt-0 mb-1">Cannot connect to local provider</p>
             <p className="m-0">Did you forget to run `yarn chain`?</p>
           </>,
         );
-        console.error("⚡️ ~ file: Faucet.tsx:getFaucetAddress ~ error", error);
-      }
-    };
-    getFaucetAddress();
+        console.error("⚡️ ~ file: Faucet.tsx:useEffect ~ error", error);
+      });
   }, [provider]);
 
   const sendETH = async () => {
--- a/packages/nextjs/components/scaffold-eth/Faucet.tsx
+++ b/packages/nextjs/components/scaffold-eth/Faucet.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
 import { ethers } from "ethers";
 import { useNetwork } from "wagmi";
 import { hardhat, localhost } from "wagmi/chains";
@@ -20,7 +20,7 @@ export const Faucet = () => {
   const [sendValue, setSendValue] = useState("");
 
   const { chain: ConnectedChain } = useNetwork();
-  const provider = getLocalProvider(localhost);
+  const provider = useMemo(() => getLocalProvider(localhost), []);
   const signer = provider?.getSigner(FAUCET_ACCOUNT_INDEX);
   const faucetTxn = useTransactor(signer);
 
@@ -34,7 +34,7 @@ export const Faucet = () => {
       } catch (error) {
         notification.error(
           <>
-            <p className="font-bold mt-0 mb-1gi">Cannot connect to local provider</p>
+            <p className="font-bold mt-0 mb-1">Cannot connect to local provider</p>
             <p className="m-0">Did you forget to run `yarn chain`?</p>
           </>,
         );
@@ -42,8 +42,7 @@ export const Faucet = () => {
       }
     };
     getFaucetAddress();
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [JSON.stringify(provider)]);
+  }, [provider]);
 
   const sendETH = async () => {
     try {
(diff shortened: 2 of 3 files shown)


carletex:
Thanks for the videos and the explanation, Shiv.

> Demo with only yarn chain runnig and deployed contract is null :

This one looks good!

> Demo with yarn chain not running

I added this one to avoid those errors on projects that don't use contracts. E.g: https://hackathon.buidlguidl.com/

What would be the alternative? Switching targetNetwork to something different than hardhat?

damianmarti:
> Thanks for the videos and the explanation, Shiv.
Really nice explanations and videos to save PR review time!!

> 
> > Demo with only yarn chain runnig and deployed contract is null :
> 
> This one looks good!
> 
> > Demo with yarn chain not running
> 
> I added this one to avoid those errors on projects that don't use contracts. E.g: https://hackathon.buidlguidl.com/
> 
> What would be the alternative? Switching targetNetwork to something different than hardhat?
Maybe something like targetNetwork: false?

sverps:
> Maybe something like targetNetwork: false?

Not a fan of assigning values that don't really match the variable. Would also break some types where `targetNetwork` is used and expected to be a network.

We could, however, add another config in `scaffold.config`, something like `autoConnect.enabled`?

damianmarti:
> > Maybe something like targetNetwork: false?
> 
> Not a fan of assigning values that don't really match the variable. Would also break some types where `targetNetwork` is used and expected to be a network.
> 
> We could, however, add another config in `scaffold.config`, something like `autoConnect.enabled`?

You are right! 

And what about targetNetwork: null? I mean, if you are not using a chain at all, sounds right not to be needed to configure any network. But it's true that if we set it to null, we should take care that the targetNetwork can be null...

Maybe we can just add another config param and that's it!

technophile-04:
Just summarising the discussion so far:- 

#### Reason for discussion : 
We don't want to show a "Cannot connect to the local provider" and other errors if the user wants to build only the frontend app. 

#### Solution discussed : 
Giving users a single config option to do so 

#### Question : 
Just wanted to know what this option should actually do meaning should it remove "Connecting with wallet" and any other stuff?  

The reason I am saying this is because people using SE-2 will surely at least want "Connecting with wallet" functionality be it signing message or other, If they want this functionality it will lead to the user configuring `targetNetwork` to some `network` because it will also handle "Switch to right network" and other things

So I see telling people to configure the network to "mainnet" or another network in the README as the best option to get started building only frontend since requires 0 to minimal changes and aslo won't confuse with more options. 

I really appreciate the solutions suggested and am happy to explore but had little doubt about what things should we should "turn off" with this configuration option ?   

Tysm all

sverps:
> So I see telling people to configure the network to "mainnet" or another network in the README as the best option to get started building only frontend since requires 0 to minimal changes and aslo won't confuse with more options.

Yeah, lets keep things simple and just explain it in the readme, since the existing config already allows users to resolve the issue. :+1:

damianmarti:
> > So I see telling people to configure the network to "mainnet" or another network in the README as the best option to get started building only frontend since requires 0 to minimal changes and aslo won't confuse with more options.
> 
> Yeah, lets keep things simple and just explain it in the readme, since the existing config already allows users to resolve the issue. +1

I agree with that too! 

And maybe @carletex can check if setting the network to mainnet fix the issue that he had at https://hackathon.buidlguidl.com/

technophile-04:
> And maybe @carletex can check if setting the network to mainnet fix the issue that he had at https://hackathon.buidlguidl.com/ 

cc @carletex  for review and merge / thoughts

README.md
```diff
@@ -123,6 +123,8 @@ If you want to redeploy to the same production URL you can run `yarn vercel --pr
 
 **Hint**: We recommend connecting the project GitHub repo to Vercel so you the gets automatically deployed when pushing to `main`
 
+**Hint**: If you don't have any contract deployed through SE-2 and plan to just deploy NextJS app maybe only with "connect wallet" functionality you can do so by just setting `targetNetwork : chains.mainnet` or your preferred network in `packages/nextjs/scaffold.config.ts`, rest all instruction remains same as [above](#deploying-your-nextjs-app)
```
technophile-04, line 126: Added a small `hint` it can be improved a lot please feel free to suggest idea, aslo was wondering should we create a separate section ? I feel like its not worth it because README.md has really grown alot

packages/nextjs/components/scaffold-eth/Faucet.tsx
```diff
@@ -28,7 +27,7 @@ export const Faucet = () => {
   useEffect(() => {
```
sverps, line 27: This whole `useEffect` feels a bit off.

Maybe this can be cleaned up like so:
```typescript
useEffect(() => {
  provider
    ?.listAccounts()
    .then(accounts => setFaucetAddress(accounts[FAUCET_ACCOUNT_INDEX]))
    .catch(error => {
      notification.error(
        <>
          <p className="font-bold mt-0 mb-1gi">Cannot connect to local provider</p>
          <p className="m-0">Did you forget to run `yarn chain`?</p>
        </>,
      );
      console.error("⚡️ ~ file: Faucet.tsx:getFaucetAddress ~ error", error);
    });
}, [provider]);
```

I also don't understand why the JSON stringify is needed. Feels wrong, and there's probably a cleaner way to fix the issue that it might cause.

technophile-04: > I also don't understand why the JSON stringify is needed. Feels wrong, and there's probably a cleaner way to fix the issue that it might cause.

Yeah I agree but since the provider is an `object` as a dependency and react uses referential diff for deps array which causes `useEffect` to run multiple time : 

https://user-images.githubusercontent.com/80153681/233550082-aee7cfa0-be73-414d-9e09-b7a21fb344eb.mov

I think `JSON.stringigy` is one of the used options to do deep equals

 Nevertheless, I memoized the provider since `localhost` is static by wrapping  `getLocalProvider` with `useMemo` and it does the job  [eb3a0de](https://github.com/scaffold-eth/scaffold-eth-2/pull/292/commits/eb3a0dec3518affa4036953cd5da1b714598872e) thanks 🙌

technophile-04: > Maybe this can be cleaned up like so:

Not strong on this, but looks minimal and nice, I think it's worth mentioning in #232 which pattern are we following `then/catch` vs `async/await` 🙌

carletex: I agree that it looks a bit better, but I personally prefer to follow the async/await pattern like in the rest of the project.

carletex (commented):
Left some minor comments @technophile-04 thanks! ;)

packages/nextjs/components/scaffold-eth/Faucet.tsx
sverps: Sure, no issues with keeping things consistent :+1:

README.md
technophile-04: Update the hint text a bit to fit in properly, hope it makes sense : 
Screenshot 2023-04-21 at 7 45 10 PM

Before : 
Screenshot 2023-04-21 at 7 26 36 PM

The empty space looked weird to me but happy to revert 🙌

packages/nextjs/hooks/scaffold-eth/useAutoConnect.ts
```diff
@@ -65,10 +64,6 @@ export const useAutoConnect = (): void => {
   }, [accountState.isConnected, accountState.connector?.name]);
 
   useEffectOnce(() => {
-    if (!contracts) {
-      return;
-    }
-
```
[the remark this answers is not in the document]
technophile-04: Yes :( , But console looks clean if you change  the `targetNetwork` 🙌

packages/nextjs/components/scaffold-eth/Faucet.tsx
```diff
@@ -21,22 +20,28 @@ export const Faucet = () => {
   const [sendValue, setSendValue] = useState("");
 
   const { chain: ConnectedChain } = useNetwork();
-  const provider = getLocalProvider(localhost);
+  const provider = useMemo(() => getLocalProvider(localhost), []);
```
carletex, line 23: What if we move this outside/above of the component?

```
const provider = getLocalProvider(localhost);

export const Faucet = () => { 
//... 
}
```

Would it have the same effect?

PR #294, opened 2023-04-08
added screens in the tailwind config file

IbrahimIjai:
I noticed that "screens" were missing in the tailwind config file, I added various screen sizes that will help to easily create responsive UI easier and faster.

screens: {
        sm: "640px", // => @media (min-width: 640px) { ... }
        md: "768px", // => @media (min-width: 768px) { ... }
        lg: "1024px", // => @media (min-width: 1024px) { ... }
        xl: "1280px", // => @media (min-width: 1280px) { ... }
        "2xl": "1536px", // => @media (min-width: 1536px) { ... }
      },

### DIFF
--- a/packages/nextjs/tailwind.config.js
+++ b/packages/nextjs/tailwind.config.js
@@ -112,6 +112,13 @@ module.exports = {
         "pulse-fast": "pulse 1s cubic-bezier(0.4, 0, 0.6, 1) infinite",
         zoom: "zoom 1s ease infinite",
       },
+      screens: {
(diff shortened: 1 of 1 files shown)


sverps:
Thanks for the suggestion, but tailwind comes built-in with default media queries, and they match with what you added in the PR. So no need to set these manually in the config, we can use them our of the box.

IbrahimIjai:
Wow, I have always configure those. Will go through the documentation again. Thanks

Issue #286, opened 2023-04-06
Notifications don't follow the website theme


kevinjoshi46b:
Why not expand themes to notifications as well?

Something like this:

![a6131032-36ad-41b8-9c8f-6fdce6b8146a](https://user-images.githubusercontent.com/62877695/230453037-10c8181e-2e18-4b97-bd55-13435f46e032.jpg)

![f31864b8-5184-4807-8314-1afcd39c88ae](https://user-images.githubusercontent.com/62877695/230453115-7da4cba7-1db8-4304-a265-0249359bbfc3.jpg)

sverps:
Sounds good. Want to tackle this and create a PR? Or do you prefer we look into it when we have some time?

kevinjoshi46b:
@sverps I would love to upgrade the notifications myself, I will open a PR for it in some time. You can assign the issue to me.

kevinjoshi46b:
I also had one more suggestion.

For making the notifications UI/UX much better why don't we just display some message like "Oops something went wrong! Check console for more details" something on these lines and console log the error.

Since bombarding the huge error message in the notification itself could be overwhelming for normal web3 users.

sverps:
Maybe we can keep that for another issue/pr. There's some pros and cons to that.

KcPele:
> Maybe we can keep that for another issue/pr. There's some pros and cons to that.

what he said here, this is the issue #301 What this issue is handling has to do with the theme

kevinjoshi46b:
Solved by #298

PR #298, opened 2023-04-10
feat: notifications now follow the theme

kevinjoshi46b:
Solves #286 

Notifications will look like this now:

Screenshot 2023-04-10 113745

Screenshot 2023-04-10 113736

Screenshot 2023-04-10 113801

Screenshot 2023-04-10 113809

### DIFF
--- a/packages/nextjs/utils/scaffold-eth/notification.tsx
+++ b/packages/nextjs/utils/scaffold-eth/notification.tsx
@@ -49,14 +49,14 @@ const Notification = ({
   return toast.custom(
     t => (
       <div
-        className={`flex flex-row items-center justify-between max-w-sm bg-white p-4 text-black shadow-2xl hover:shadow-none transform-gpu translate-y-0 hover:translate-y-1 rounded-md relative transition-all duration-500 ease-in-out space-x-2 ${
+        className={`flex flex-row items-start justify-between max-w-sm bg-base-200 p-4 shadow-2xl hover:shadow-none transform-gpu translate-y-0 hover:translate-y-1 rounded-md relative transition-all duration-500 ease-in-out space-x-2 ${
           t.visible ? "top-0" : "-top-96"
         }`}
       >
-        <div className="text-2xl self-start mb-0 mt-3">{icon ? icon : ENUM_STATUSES[status]}</div>
-        <div className="mt-3 break-all">{content}</div>
+        <div className="text-2xl self-start">{icon ? icon : ENUM_STATUSES[status]}</div>
+        <div className={`break-all ${icon ? "mt-1" : ""}`}>{content}</div>
 
-        <div className="absolute top-1 right-1 cursor-pointer text-lg" onClick={() => toast.dismiss(t.id)}>
+        <div className={`cursor-pointer text-lg ${icon ? "mt-1" : ""}`} onClick={() => toast.dismiss(t.id)}>
           <XMarkIcon className="w-6 cursor-pointer" onClick={() => toast.remove(t.id)} />
         </div>
       </div>
--- a/packages/nextjs/utils/scaffold-eth/notification.tsx
+++ b/packages/nextjs/utils/scaffold-eth/notification.tsx
@@ -49,8 +49,11 @@ const Notification = ({
   return toast.custom(
     t => (
       <div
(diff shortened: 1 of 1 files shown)


packages/nextjs/utils/scaffold-eth/notification.tsx
```diff
@@ -49,14 +49,14 @@ const Notification = ({
   return toast.custom(
     t => (
       <div
-        className={`flex flex-row items-center justify-between max-w-sm bg-white p-4 text-black shadow-2xl hover:shadow-none transform-gpu translate-y-0 hover:translate-y-1 rounded-md relative transition-all duration-500 ease-in-out space-x-2 ${
+        className={`flex flex-row items-start justify-between max-w-sm bg-base-200 p-4 shadow-2xl hover:shadow-none transform-gpu translate-y-0 hover:translate-y-1 rounded-md relative transition-all duration-500 ease-in-out space-x-2 ${
           t.visible ? "top-0" : "-top-96"
```
damianmarti, line 53: Why do you need the -top-96 when not visible?

kevinjoshi46b: I didn't notice it till you pointed it out.
I just checked, it seems that "-top-96" position helps create the slide up out of the screen animation, without it the notification just disappears in place but this brings up two points what about notifications with position "bottom-xxx"?
- In the case where notification position is bottom, when the notification is about to disappear it moves up to somewhere in the middle of the screen and seems to look like a glitch (the screenshot bellow explains it better) (will fix this in my next commit)

From here:
image

To here:
image

- When hovering on the notification the shadow is set to none and the notification moves down by one ("translate-y-1") this animation looks weird when notification position is bottom, I think it should have a mirror effect and the notification should move up by 1 ("translate-x-1"). Would love to have your opinion on this before I change it.

damianmarti: Maybe we can check how it feels if the notification is at the top.

kevinjoshi46b: I didn't get you there

damianmarti: Sorry. 

About "When hovering on the notification the shadow is set to none and the notification moves down by one ("translate-y-1") this animation looks weird when the notification position is bottom, I think it should have a mirror effect and the notification should move up by 1 ("translate-x-1"). Would love to have your opinion on this before I change it."

Maybe the animation looks good when the notification is at the top, so if you change it to "translate-x-1", now the notification at the top will look weird.

Please, let me know if I'm not clear enough.

kevinjoshi46b: Done

technophile-04:
Thanks @kevinjoshi46b for the PR and @damianmarti for the review !!

> In the case where notification position is bottom, when the notification is about to disappear it moves up to somewhere in the middle of the screen and seems to look like a glitch (the screenshot bellow explains it better) (will fix this in my next commit)

Yeah this seemed to an issue, thanks for tackling it 

---


One minor change which I noticed was the position of the "X" icon : 

`main` branch : 
Screenshot 2023-04-12 at 8 15 41 PM

`this PR` : 
Screenshot 2023-04-12 at 8 13 47 PM

Was this intentional @kevinjoshi46b ? Both look fine to me but `main` branch seems a bit standardized with "X" at the top right not taking up space

--- 

Rest everything looks good to me, thanks @kevinjoshi46b 🙌

technophile-04 (approved):

kevinjoshi46b:
> Was this intentional @kevinjoshi46b ? Both look fine to me but `main` branch seems a bit standardized with "X" at the top right not taking up space

I agree that position of "X" in the top right being standardized but I feel it looks like a standard only for models that have a lot of content in them.
What I have observed is for notifications that are meant to be just 1-2 lines having the "X" inline with the content looks more natural. That is also how [MUI](https://mui.com/material-ui/react-snackbar/) and many other libraries have implemented it.

But ya this is not that big of a deal if we want we can revert to the original positioning as well.

kevinjoshi46b:
> 💯 agree, I just mentioned this because I noticed it while testing for long error messages because lol sometimes we get really big error messages check #301 but current looks good to me

Ya I am looking forward to that issue being resolved!
Those json notifications might get too overwhelming for non-dev users 🫠

Issue #284, opened 2023-04-06
Rainbowkit dark theme


kevinjoshi46b:
Why isn't [rainbowkit's dark theme](https://www.rainbowkit.com/docs/theming) not used? Is it because its dark theme colors do not match scaffoldEthDark theme's color scheme?

kevinjoshi46b:
For reference this is how the dark theme model looks:

![image](https://user-images.githubusercontent.com/62877695/230529605-1f276627-08a5-41a0-bb31-6c0f333b7980.png)

And to what I could understand we can only change the buttons colors.

sverps:
Would indeed be nice to align this with the dark/light theme :+1: 
Shall I assign you for this one as well?

kevinjoshi46b:
@sverps Ya sure go ahead

damianmarti:
Done at https://github.com/scaffold-eth/se-2/pull/299

PR #299, opened 2023-04-10
feat: Added rainbowkit dark theme

kevinjoshi46b:
Solves #284

### DIFF
--- a/packages/nextjs/pages/_app.tsx
+++ b/packages/nextjs/pages/_app.tsx
@@ -1,9 +1,10 @@
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
 import type { AppProps } from "next/app";
-import { RainbowKitProvider } from "@rainbow-me/rainbowkit";
+import { RainbowKitProvider, darkTheme, lightTheme } from "@rainbow-me/rainbowkit";
 import "@rainbow-me/rainbowkit/styles.css";
 import NextNProgress from "nextjs-progressbar";
 import { Toaster } from "react-hot-toast";
+import { useDarkMode } from "usehooks-ts";
 import { WagmiConfig } from "wagmi";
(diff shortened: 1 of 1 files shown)


packages/nextjs/pages/_app.tsx
```diff
@@ -17,17 +18,27 @@ import "~~/styles/globals.css";
...
+  const { isDarkMode } = useDarkMode();
 
   useEffect(() => {
     if (price > 0) {
       setEthPrice(price);
     }
   }, [setEthPrice, price]);
 
+  useEffect(() => {
+    setIsDarkTheme(isDarkMode);
```
damianmarti, line 31: Why do you need the isDarkTheme state var? Can you use isDarkMode directly?

kevinjoshi46b: We can't.
Initially even I approached it the same way by directly using the isDarkMode variable but seems that the value for isDarkMode variable defaults to false during server side rendering and it does get updated when it reaches the client side but the change is not seen unless useEffect and another variable is used to then re render on the client side on the initial load. Obviously everything works fine after the website is mounted and the theme is changed.

You can re produce this and check for yourself by, setting your OS theme to be darkmode and refresh the website. The website will load in dark theme but the rainbowkit model will still be in light mode unless you switch the theme to light and then dark mode again.

damianmarti: Ok! Please add a small comment in the code explaining it, so we know why it's needed.

kevinjoshi46b: Done

sverps (approved):

packages/nextjs/pages/_app.tsx
damianmarti: Thanks!!

PR #302, opened 2023-04-10
Hook example

KcPele:


damianmarti:
Thanks, @KcPele !! Maybe you can add an example from useScaffoldContract too.

README.md
```diff
@@ -110,6 +115,78 @@ You can verify your smart contract on Etherscan by running:
...
+- useDeployedContractInfo
+
+  **useScaffoldContractRead**:
+  Use this hook to read a value from your deployed contracts.
+
+  ```ts
+  const { data: totalCounter } = useScaffoldContractRead({
+    contractName: "YourContract",
+    functionName: "getGreeting",
+    // if it does not accept any args it should be args: []
```
sverps, line 133: The args param can bee omitted entirely when no args needed for the function call, no need to pass empty array.

KcPele: ok,

README.md
```diff
@@ -110,6 +115,94 @@ You can verify your smart contract on Etherscan by running:
...
+
+  ```ts
+  const { data: yourContract } = useScaffoldContract({
+    contractName: "YourContract",
+  });
+  // will return the greeting and can be call in any function unlike useScaffoldContractRead
+  await yourContract.greeting();
+
+  //will be use to write to a contract and can be called in any function
+  await yourContract.setGreeting("Hello");
```
damianmarti, line 139: Please check if this works. The hook uses the provider as default, so I think that you should pass the signer to the hook if you want to call write methods.

damianmarti (changes requested):

README.md
```diff
@@ -110,6 +115,94 @@ You can verify your smart contract on Etherscan by running:
...
+  ```ts
+  const {
+    data: events,
+    isLoading: isLoadingEvents,
+    error: errorReadingEvents,
+    } = useScaffoldEventHistory({
+    contractName: "YourContract",
+    eventName: "GreetingChange",
+    fromBlock: //block number here,
+    blockData: true,
```
damianmarti, line 194: You can add the other optional parameters here and let the user know what are the mandatory parameters.

KcPele: I think Typescript will assist with that. the main aim of the doc is to show how to use it with its required parameters

technophile-04:
Tysm @KcPele 🙌

### Adding some notes : 

#### 1. Finding the right place for `Hook's Example` : 
I think the current placement after `Deploying your Smart Contracts to a Live Network` breaks the flow of "Shipping your DApp" maybe we put it after `Deploying your NextJS App` or at the end before `Contributing to Scaffold-Eth 2` 

#### 2. Maybe keep the order in order of their usability : 
1)useScaffoldContractRead 2) useScaffoldContractWrite 3) useScaffoldEventHistory 4) useScaffoldEventSubscriber 5) useDeployedContractInfo 6) useScaffoldContract because having `useScaffoldContract ` first might give the developer the wrong recommendation of using it for `read` / `write`  

#### 3. Mentioning `YourContract.sol` as an example used. 
I like how wagmi explains hooks by mentioning the contract used checkout  [useScaffoldContactRead](https://wagmi.sh/react/hooks/useContractRead#usage). 
Similar to them we can mention : 
The following examples use the [YourContract.sol](https://github.com/scaffold-eth/se-2/blob/main/packages/hardhat/contracts/YourContract.sol) which comes with SE-2. 

This will avoid us from doing : 
```js
// if it does not accept any args it can be optional
  args: ["ARGUMENTS IF THE FUNCTION ACCEPTS ANY"],
```

Please don't change anything yet, I feel Eda will be really a great person to guide us on this

edakturk14 (approved):
Looks great! Thanks for making these changes!

damianmarti:
> Tysm @KcPele raised_hands
> 
> ### Adding some notes :
> #### 1. Finding the right place for `Hook's Example` :
> I think the current placement after `Deploying your Smart Contracts to a Live Network` breaks the flow of "Shipping your DApp" maybe we put it after `Deploying your NextJS App` or at the end before `Contributing to Scaffold-Eth 2`

I think it's better to add the Hooks Examples before the "Deploying your Smart Contracts to a Live Network" section.

> 
> #### 2. Maybe keep the order in order of their usability :
> 1)useScaffoldContractRead 2) useScaffoldContractWrite 3) useScaffoldEventHistory 4) useScaffoldEventSubscriber 5) useDeployedContractInfo 6) useScaffoldContract because having `useScaffoldContract ` first might give the developer the wrong recommendation of using it for `read` / `write`

I like this order.

> 
> #### 3. Mentioning `YourContract.sol` as an example used.
> I like how wagmi explains hooks by mentioning the contract used checkout [useScaffoldContactRead](https://wagmi.sh/react/hooks/useContractRead#usage). Similar to them we can mention : The following examples use the [YourContract.sol](https://github.com/scaffold-eth/se-2/blob/main/packages/hardhat/contracts/YourContract.sol) which comes with SE-2.
> 
> This will avoid us from doing :
> 
> ```js
> // if it does not accept any args it can be optional
>   args: ["ARGUMENTS IF THE FUNCTION ACCEPTS ANY"],
> ```
> 
> Please don't change anything yet, I feel Eda will be really a great person to guide us on this

KcPele:
> I think we can make these changes suggested by @technophile-04 and then merge the PR. What do you think @KcPele ?

OK

KcPele:
> > Tysm @KcPele raised_hands
> > ### Adding some notes :
> > #### 1. Finding the right place for `Hook's Example` :
> > I think the current placement after `Deploying your Smart Contracts to a Live Network` breaks the flow of "Shipping your DApp" maybe we put it after `Deploying your NextJS App` or at the end before `Contributing to Scaffold-Eth 2`
> 
> I think it's better to add the Hooks Examples before the "Deploying your Smart Contracts to a Live Network" section.
> 
> > #### 2. Maybe keep the order in order of their usability :
> > 1)useScaffoldContractRead 2) useScaffoldContractWrite 3) useScaffoldEventHistory 4) useScaffoldEventSubscriber 5) useDeployedContractInfo 6) useScaffoldContract because having `useScaffoldContract ` first might give the developer the wrong recommendation of using it for `read` / `write`
> 
> I like this order.
> 
> > #### 3. Mentioning `YourContract.sol` as an example used.
> > I like how wagmi explains hooks by mentioning the contract used checkout [useScaffoldContactRead](https://wagmi.sh/react/hooks/useContractRead#usage). Similar to them we can mention : The following examples use the [YourContract.sol](https://github.com/scaffold-eth/se-2/blob/main/packages/hardhat/contracts/YourContract.sol) which comes with SE-2.
> > This will avoid us from doing :
> > ```js
> > // if it does not accept any args it can be optional
> >   args: ["ARGUMENTS IF THE FUNCTION ACCEPTS ANY"],
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Please don't change anything yet, I feel Eda will be really a great person to guide us on this

I had to make use of YourContract because it gives clarity to users and they can easily know where it is coming from.

damianmarti:
Sorry @KcPele . One more small comment. I think that it will be useful if each hook name is a subtitle, using ###, so in the generated README you have an anchor to each hook to send the link if anyone needs it. What do you think?

KcPele:
> 

Done Ser

PR #303, opened 2023-04-10
Add a command to deploy on IPFS with Spheron CLI

rekpero:
I've added a command `npm run spheron` or `yarn spheron` to build directly, export it to static files and then upload to ipfs using spheron CLI.
You can even change it to Filecoin or Arweave by changing the `spheron.json` inside nextjs to `filecoin` or `arweave`.

My intention of doing this is inspired by the original scaffold-eth [ipfs command](https://github.com/scaffold-eth/scaffold-eth/blob/master/packages/react-app/package.json#L96) in the react-app package to deploy the app on IPFS. 

Let me know if there are any changes needed somewhere. I am happy to talk about it in the PR. 
Happy buidl, guys!


rekpero:
Hey @sverps @technophile-04 I just rebased my branch with main. Let me know if there is anything needed to be updated. Would love to see this merged.

packages/nextjs/package.json
```diff
@@ -11,7 +11,8 @@
     "format": "prettier --write . '!(node_module|.next|contracts)/**/*'",
     "check-types": "tsc --noEmit --incremental",
     "vercel": "vercel",
-    "vercel:yolo": "vercel --build-env NEXT_PUBLIC_IGNORE_BUILD_ERROR=true"
+    "vercel:yolo": "vercel --build-env NEXT_PUBLIC_IGNORE_BUILD_ERROR=true",
+    "spheron": "next build && next export && spheron login --github && spheron publish"
```
sverps, line 15: Can we instead use `spheron upload --path ./out --protocol ipfs` and omit the `spheron.json` file? I prefer to keep the optional config files to a minimum to avoid potentially confusing users.

sverps, line 15: With the newly added local blockexplorer, `next export` fails since it doesn't work with `getServerSideProps`. So we'll have to think about a potential solution for this.

Maybe in favor of ipfs deployment options (and potentially other deployment options that require `next export`) we could opt to omit the code tab from the blockexplorer while thinkink of a different way to include it (since I think the value of that tab might be rather limited). What do you think @carletex @technophile-04

sverps:
Hey @rekpero , sorry for not picking this up sooner. Let's try and get this included, always nice to provide users with some options for deployment. Added some comments.

technophile-04:
Closing this for now, please feel free to re-open 

Thanks all 🙌

PR #304, opened 2023-04-12
Fix re-render whole app on theme switch 

technophile-04:
This PR might end up going to the trash, and this is just a suggestion/discussion: 

### Reason : 
I feel we should refrain from adding states that can be changed easily at the very top level of app like `_app` because changing them re-renders the whole app, and switching of theme causes the whole app to re-render feel a bit uncomfortable default to me and would delegate to end-user (**But this seems the only solution to keep rainbowkit theme inline with our app them**)

### Solution Proposed : 
 Use [Dark Mode support](https://www.rainbowkit.com/docs/theming#dark-mode-support) mentioned by rainbow kit which sets the rainbow kit theme based on OS default theme. 
 
 Advantage : 
 Solves the problem of re-rendering the whole app 

Disadvantage : 
It's independent of SE-2 theme and if the user selects the light se-2 theme he will still get the dark rainbowkit theme because of his OS preference. 

cc @sverps @damianmarti please free to suggest other options / directly close this 🙌

### DIFF
--- a/packages/nextjs/pages/_app.tsx
+++ b/packages/nextjs/pages/_app.tsx
@@ -1,10 +1,9 @@
-import { useEffect, useState } from "react";
+import { useEffect } from "react";
 import type { AppProps } from "next/app";
 import { RainbowKitProvider, darkTheme, lightTheme } from "@rainbow-me/rainbowkit";
 import "@rainbow-me/rainbowkit/styles.css";
 import NextNProgress from "nextjs-progressbar";
 import { Toaster } from "react-hot-toast";
-import { useDarkMode } from "usehooks-ts";
 import { WagmiConfig } from "wagmi";
 import { Footer } from "~~/components/Footer";
 import { Header } from "~~/components/Header";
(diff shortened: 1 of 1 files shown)


packages/nextjs/pages/_app.tsx
```diff
@@ -18,27 +17,27 @@ import "~~/styles/globals.css";
...
-  }, [isDarkMode]);
-
   return (
     <WagmiConfig client={wagmiClient}>
       <NextNProgress />
       <RainbowKitProvider
         chains={appChains.chains}
         avatar={BlockieAvatar}
-        theme={isDarkTheme ? darkTheme() : lightTheme()}
+        // theme={isClient && isDarkMode ? darkTheme() : lightTheme()}
```
technophile-04, line 35: Have commented out this, this is an alternate solution to https://github.com/scaffold-eth/se-2/pull/299#discussion_r1161985460

With this approach, we don't need an extra state. lol even this can be ignored was just trying to utilize [`useHooks`](https://usehooks-ts.com/) which was already in place and seems clean

sverps:
The high level state for the theme doesn't bother me, it's only triggered on theme switch, which is a user action and isn't expected to be triggered often. So at most causes a couple re-renders, hardly noticeable.

So I prefer the fix as it was, since not all platforms manage light/dark theme properly, or a user might have a different preference for se-2 than his OS default.

damianmarti:
> The high level state for the theme doesn't bother me, it's only triggered on theme switch, which is a user action and isn't expected to be triggered often. So at most causes a couple re-renders, hardly noticeable.
> 
> So I prefer the fix as it was, since not all platforms manage light/dark theme properly, or a user might have a different preference for se-2 than his OS default.

Yes, I agree.

A whole re-render when the user changes the theme doesn't seem a big issue for me, because we don't expect the user keep changing the theme all the time

Issue #301, opened 2023-04-10
Notification error not user friendly


KcPele:
The notification error that is been displayed is not user friendly
It should be formatted properly and only the reason for the error should be seen
![Screenshot from 2023-04-10 09-56-22](https://user-images.githubusercontent.com/64108987/230869792-38422f14-268a-4717-a1e8-841f3bd75c7f.png)

kevinjoshi46b:
I had raised similar concerns in #286 
@sverps maybe we can discuss the pros/cons here?

KcPele:
This is different. I am not talking about the theme. it is about the error itself. The one in #286 is about the theme

kevinjoshi46b:
Please do correct me if I am wrong but what you mean is to have a message shown as "Campaign deadline has not passed yet" / " Something went wrong" instead of this huge error json being presented in the notification right? If so then that is what I was talking about in this comment: https://github.com/scaffold-eth/se-2/issues/286#issuecomment-1499866944

KcPele:
If u check in #286 one of the comment is that another issue should be open that handles the way the error is been displayed. so this is the issue, not to worry its fine, i will close the issue after creating a pr for it

technophile-04:
Tysm for the issue 🙌

Actually parsing errors nicely with ethers v5 is really hard, check out the original comment -> https://github.com/scaffold-eth/se-2/pull/36#discussion_r1018310035

I think with ethers v6 this is solved somewhat checkout -> https://github.com/ethers-io/ethers.js/issues/2475#issuecomment-1428671151

But I think we may go on the path of [VIEM](https://www.viem.sh) instead of upgrading to v6(not clear yet) since WAGMI v1.0 is migrating from ethers to [VIEM](https://www.viem.sh). 

We already had this discussion about migration at #229, so marking this as `waiting` since currently migration is not top priority yet.

jxom:
this is what ethers v6 looks like right now: https://twitter.com/jakemoxey/status/1633240447761121281?s=20

KcPele:
I have been able to write a function that returns a better error message
![Screenshot from 2023-04-13 11-05-15](https://user-images.githubusercontent.com/64108987/231727171-341df6ea-271a-4d8b-bfc5-983117a9f002.png)
![Screenshot from 2023-04-13 11-06-20](https://user-images.githubusercontent.com/64108987/231727273-592269dd-ab14-4ec5-94a0-e26fa4309121.png)

[...]

PR #305, opened 2023-04-13
Notification display error

KcPele:
Open to feedback and better testing #301 
Screenshot from 2023-04-13 11-06-20
Screenshot from 2023-04-13 11-05-15


packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -160,6 +160,24 @@ const getParsedEthersError = (e: any): string => {
 
   return message;
 };
+const extractKeyValueErrorPairs = (errorMessage: any) => {
+  let keyValuePairs = {} as any;
```
sverps, line 164: Don't assert as `any`, be as strict as possible.
Also assigning an empty object as default value is probably not the best way to go, I think.

I'd remove this `variable` and declare it inside the `try { ... } catch` block. Handle the rest of the logic there.

If the logic fails (due to invalid JSON or for whatever other reason), the default fallback of this method should be to return the error message as is.

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -160,6 +160,24 @@ const getParsedEthersError = (e: any): string => {
...
+    keyValuePairs = JSON.parse(json);
+  } catch (error) {
+    console.error("Error parsing JSON:", error);
+  }
+  if (keyValuePairs?.reason) {
+    return keyValuePairs.reason;
+  } else {
+    const errorMessage = keyValuePairs.message.slice("[ethjs-query] while formatting outputs from RPC".length).trim();
+    const parsedErrorMessage = JSON.parse(errorMessage.substring(1, errorMessage.length - 1)) as any;
+    return parsedErrorMessage?.value.data.message;
```
sverps, line 178: Move this logic to inside the `try { ... }` block.

The `catch { ... }` block can then return the `errorMessage` in it's entirety as a fallback.

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -160,6 +160,24 @@ const getParsedEthersError = (e: any): string => {
 
   return message;
 };
+const extractKeyValueErrorPairs = (errorMessage: any) => {
```
sverps, line 163: I think the naming of this function could better reflect its purpose, e.g. `extractErrorMessage` or something like that

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
```diff
@@ -90,8 +90,10 @@ export const useTransactor = (_signer?: Signer): TTransactionFunc => {
       }
       // TODO handle error properly
       console.error("⚡️ ~ file: useTransactor.ts ~ error", error);
-      const message = getParsedEthersError(error);
-      notification.error(message);
+      const err = extractKeyValueErrorPairs(JSON.stringify(error)) as any;
```
sverps, line 93: Why are we stringifying the error here instead of using it directly as an error object? That would avoid the need for the JSON parsing inside, no?

Can't we just improve the parsing done by `getParsedEthersError` rather than recreating something similar?

sverps (changes requested):

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
KcPele: Greate

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
KcPele: I do not want to temper with what was there since i will be making use of it

technophile-04:
Hey @KcPele really thanks for this !! But currently at `main` its parses error correctly : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/9d7d1e12-5ea2-445f-81f8-80bbae1d14ea

But yeah I agree #301 is real because sometimes it does not parse it well its kind of hard to determine. Saying that we are currently working on #352 which hopefully will internally solve this issue checkout -> https://github.com/scaffold-eth/scaffold-eth-2/issues/301#issuecomment-1506215490

So closing this for now but we surely gonna come back at this if we face the same issue even after #352 🙌

PR #311, opened 2023-04-19
remove zksync-test-damu dir

technophile-04:
I think it was added by mistake at #302


carletex:
Thanks @technophile-04 

Please, let's be careful when reviewing/merging stuff.

Issue #251, opened 2023-03-22
Create a SE2 cli tool / NPX Install


carletex:
Since the core functionality of SE2 it's almost there (we are in the fixing bugs / small improvements phase), I think it's time to start working on a CLI tool for SE2. This is also related to #121 (NPX Install)

### Goal 

Having a tool that:
- Installs SE-2 (with NPX? it could use the CLI under the hood)
- Allows choosing between hardhat & foundry
- Allows adding «extensions» to SE-2 (SIWE, Graph, Backend, etc)
- Have random utils (e.g. remove the contract part of SE-2, for front-end only projects)

### Builder workflow

Just writing a rough example of what an end goal for this could look like:

```bash
npx create-se-2 my-project
# - It will check dependencies (nodejs >= 18, yarn & git installed)
> Do you want to install the dependencies? (Y/n)
# ......... yarn installing everything (including the CLI). See implementation section
# postinstall: run cli set up. See implementation section
> What smart contract dev framework do you want to use
(*) Hardhat
( ) Foundry
# ... extra config
```

Now SE2 is installed on `myproject`. Besides all the `chain / deploy / start`, you could (yarn could just wrap the se-cli calls)

```bash
yarn add-extension
> Which extension do you want to add (Select multiple)
(*) Sign-In with Ethereum
(*) The Graph
( ) Backend API Example
( ) Foundry
```

Also some utils:

```bash
yarn clean # Remove all the extra stuff (index page => empty, remove example UI, all unused stuff)
yarn remove-contracts # Remove the hardhat/foundry folder, an any component that interacts with it on the contract.
// ....
```

### Implementation

There are a bunch of options to go about this, throwing some ideas out there and its reasons.

**1. NPX install**
Could be a separate repo `create-se2-app` with a simple script that uses [simple-git](https://www.npmjs.com/package/simple-git) to clone the repo (+ check deps + run yarn install)

-> Reasons
- I don't like polluting this repo with the NPX/Cli stuff, so it's easier to contribute and manage.
- People could also just clone the SE2 repo as it's and will still work.

**2. Hardhat vs Foundry**
This is the main piece of configuration for the project, and it's a must-have.

I see two options:

1. Having both folders in the projects that generate the same `hardhat_contracts.ts` file (so the front-end doesn't care).
	- CONS: Bigger/Polluted repo. Can be confusing
	- PRO: Easy to manage and create (basically just create the folder and yarn commands)
[...]

PR #315, opened 2023-04-20
Add a subgraph package

damianmarti:
This package adds the feature to generate subgraphs using The Graph.

It's a proof of concept on how to integrate the subgraphs into SE-2.

There is a README with more information.

This branch should be updated when we have the installer script working and setting the subgraph package as an optional one.


carletex:
Thanks for this @damianmarti 

This is a perfect test for #251 besides foundry/hardhat, since this is more like an extension (in foundry/hardhat you need to choose between the two)

@damianmarti and I are actively working on #251 , and there might be several ways to integrate new extensions (packages). We'd have to choose between those taking into consideration scalability & maintainability. 

Posting a comment in #251.

packages/nextjs/components/example-ui/ContractData.tsx
```diff
@@ -1,4 +1,5 @@
 import { useEffect, useRef, useState } from "react";
+import { gql, useQuery } from "@apollo/client";
```
schmidsi, line 2: I'd use [URQL](https://github.com/urql-graphql/urql/tree/main/packages/react-urql) for querying The Graph/GraphQL in Next.js. There were some weird problems with SSR.

packages/nextjs/pages/_app.tsx
```diff
@@ -22,6 +23,12 @@ const ScaffoldEthApp = ({ Component, pageProps }: AppProps) => {
   const [isDarkTheme, setIsDarkTheme] = useState(true);
   const { isDarkMode } = useDarkMode();
 
+  const subgraphUri = "http://localhost:8000/subgraphs/name/scaffold-eth/your-contract";
```
schmidsi, line 26: Here it would be good to have some logic about dev/prod environments. I started to use `NEXT_PUBLIC_SUBGRAPH_ENDPOINT` env var which works pretty good.

packages/subgraph/build/YourContract/YourContract.wasm
schmidsi, line 1: Compiled files should not be checked into Git in my opinion

packages/subgraph/build/subgraph.yaml
schmidsi, line 1: Built files should not be checked into git in my opinion

packages/subgraph/generated/YourContract/YourContract.ts
schmidsi, line 1: Generated files should not be checked into Git in my opinion

packages/subgraph/graph-node/README.md
```diff
@@ -0,0 +1,79 @@
...
+  -e postgres_db=<DBNAME> \
+  -e ipfs=<HOST>:<PORT> \
+  -e ethereum=<NETWORK_NAME>:<ETHEREUM_RPC_URL> \
+  graphprotocol/graph-node:latest
+```
+
+### Example usage
+
+```sh
+docker run -it \
```
schmidsi, line 21: I like that. Some folks might not want to run docker-compose.

schmidsi: Though I highly recommend running docker-compose.

packages/subgraph/src/schema.graphql
```diff
@@ -0,0 +1,17 @@
+type Greeting @entity {
+  id: ID!
```
schmidsi, line 2: Would be great to follow recent best practices. Like "Bytes as IDs" as described here: https://medium.com/edge-node-engineering/two-simple-subgraph-performance-improvements-a76c6b3e7eac

schmidsi (commented):
Great stuff! Thanks for doing this. I left some minor comments.

schmidsi:
Also, you might want to consider using Graph Client: https://github.com/scaffold-eth/scaffold-eth-2/pull/117. Although it adds a bit of complexity on top. I'd suggest to make this two separate plugins.

packages/subgraph/generated/YourContract/YourContract.ts
damianmarti: Same that with subgraph.yaml. Why do you think is better to not add this file to the git repository?

damianmarti:
> Also, you might want to consider using Graph Client: #117. Although it adds a bit of complexity on top. I'd suggest to make this two separate plugins.

Thanks! I saw that!

damianmarti:
> Great stuff! Thanks for doing this. I left some minor comments.

Thanks for your review!!

Issue #319, opened 2023-04-24
Visibility of icons in the input


kevinjoshi46b:
# Problem
The icons present in various inputs are not that well visible in dark mode (looks good on light mode)

`*` here:
![image](https://user-images.githubusercontent.com/62877695/233971017-754673cf-5fb3-4bc9-90c5-c5c53b8dbabe.png)

`$, Ξ` here:
![image](https://user-images.githubusercontent.com/62877695/233971056-752d258a-8dfe-419d-b0a0-a2771038869c.png)

and `#` in byte input

# Solution
Change the css class to `text-gray-400` this will make the icons perfectly visible on both dark and light mode

kevinjoshi46b:
Solved by #320

PR #320, opened 2023-04-24
fix: visibility of icons in IntegerInput

kevinjoshi46b:
Solves #319

The input icons now look like this:

`*` here:
image
image

`$, Ξ` here:
image
image

and similar for `#` in byte input

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Input/Bytes32Input.tsx
+++ b/packages/nextjs/components/scaffold-eth/Input/Bytes32Input.tsx
@@ -22,7 +22,7 @@ export const Bytes32Input = ({ value, onChange, name, placeholder }: CommonInput
       onChange={onChange}
       suffix={
         <div
-          className="self-center cursor-pointer text-xl font-semibold px-4 text-accent"
+          className="self-center cursor-pointer text-xl font-semibold px-4 text-gray-400"
           onClick={convertStringToBytes32}
         >
           #
--- a/packages/nextjs/components/scaffold-eth/Input/Bytes32Input.tsx
+++ b/packages/nextjs/components/scaffold-eth/Input/Bytes32Input.tsx
@@ -22,7 +22,7 @@ export const Bytes32Input = ({ value, onChange, name, placeholder }: CommonInput
       onChange={onChange}
       suffix={
         <div
-          className="self-center cursor-pointer text-xl font-semibold px-4 text-gray-400"
+          className="self-center cursor-pointer text-xl font-semibold px-4 text-primary"
           onClick={convertStringToBytes32}
         >
           #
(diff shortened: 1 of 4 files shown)


damianmarti (approved):
LGTM!

carletex (changes requested):
@kevinjoshi46b  The idea of having daisy themes is to use those colors (primary, secondary, accent, neutral, etc) so they always look consistent if you change your theme colors in `tailwind.config.js`.

If we hardcode a color, like `text-gray-400`, you are "disconnecting" it from the theme.

From DaisyUI:
image.

So we need to use a color from the theme (maybe we need to re-think colors in there)

Also, for me, it's not that bad how they look currently:

image

cc @damianmarti @technophile-04

kevinjoshi46b:
> If we hardcode a color, like `text-gray-400`, you are "disconnecting" it from the theme.

Alright ya this is a good point.

> Also, for me, it's not that bad how they look currently:
> 
> image

Maybe the `*` is fine because its a more dense icon (but it would probably still go unnoticed maybe on a bigger screen) what about `Ξ` ?

> So we need to use a color from the theme (maybe we need to re-think colors in there)

Ya I do feel this might be needed since the colors are too close and not too contradicting.

Would like to know others opinions too.

technophile-04:
> Maybe the * is fine because its a more dense icon (but it would probably still go unnoticed maybe on a bigger screen)

I feel they are visible in `main` branch, like I have a kind of big monitor screen and looks great their too...we also had improved their asociality before check #170 

Screenshot 2023-04-25 at 9 41 22 PM

> what about Ξ ?

Yeah, as Carlos mentioned nice catch it can be improved and should have been `text-accent`....lol `text-neutral` a bit too contrasty and kind off misses the theme so I would go with `text-accent` but its just my personal preference 😂

kevinjoshi46b:
> I feel they are visible in `main` branch, like I have a kind of big monitor screen and looks great their too...

That's great not an issue there then

> `text-neutral` a bit too contrasty

Ohhh yes, its brighter then the input number itself :)

I majorly noticed this issue in `Ξ` and then just ended up changing the rest as well assuming they have a similar issue but ya it seems like just changing from `text-accent` -> `text-primary` is good enough.

carletex:
@kevinjoshi46b You did it the other way around.

You changed "text-accent" to "text-primary" in integer & bytes inputs, and left the Ξ as it was. Now I can barely see them.

primary:
image

accent:
image

It's the other way around. You just need to change the  Ξ color inside the EtherInput to "text-accent"

technophile-04:
>  from text-accent -> text-primary is good enough

I think it should be `text-primaryt` -> `text-accent` right 😅

Like We keep `IntegerInput` and `BytesInput` as it is because they look good in current `main` as discussed above and just change `EtherInput` from from `text-primary` -> `text-accent` 🙌

kevinjoshi46b:
> You changed "text-accent" to "text-primary" in integer & bytes inputs, and left the Ξ as it was. Now I can barely see them.

Oops, my bad sorry for the inconvenience

carletex (approved):

Issue #321, opened 2023-04-24
Native currency symbol as per target network


kevinjoshi46b:
# Problem

Currently the unit to show user's account balance is hardcoded to be `ETH` which is very misleading on other networks whose native currency might not be ETH.

# Solution

Use the native currency symbol provided for the target network by wagmi

kevinjoshi46b:
Solved by #322

PR #322, opened 2023-04-24
feat: native currency symbol as per target network

kevinjoshi46b:
Solves #321, #323

### DIFF
--- a/packages/nextjs/components/example-ui/ContractInteraction.tsx
+++ b/packages/nextjs/components/example-ui/ContractInteraction.tsx
@@ -4,12 +4,10 @@ import { DiamondIcon } from "./assets/DiamondIcon";
 import { HareIcon } from "./assets/HareIcon";
 import { ArrowSmallRightIcon, XMarkIcon } from "@heroicons/react/24/outline";
 import { useScaffoldContractWrite } from "~~/hooks/scaffold-eth";
-import { getTargetNetwork } from "~~/utils/scaffold-eth";
 
 export const ContractInteraction = () => {
   const [visible, setVisible] = useState(true);
   const [newGreeting, setNewGreeting] = useState("");
-  const configuredNetwork = getTargetNetwork();
 
   const { writeAsync, isLoading } = useScaffoldContractWrite({
     contractName: "YourContract",
@@ -79,7 +77,7 @@ export const ContractInteraction = () => {
 
           <div className="mt-4 flex gap-2 items-start">
             <span className="text-sm leading-tight">Price:</span>
-            <div className="badge badge-warning">0.01 {configuredNetwork.nativeCurrency.symbol} + Gas</div>
+            <div className="badge badge-warning">0.01 ETH + Gas</div>
           </div>
         </div>
       </div>
--- a/packages/nextjs/components/example-ui/ContractInteraction.tsx
+++ b/packages/nextjs/components/example-ui/ContractInteraction.tsx
@@ -4,10 +4,12 @@ import { DiamondIcon } from "./assets/DiamondIcon";
 import { HareIcon } from "./assets/HareIcon";
 import { ArrowSmallRightIcon, XMarkIcon } from "@heroicons/react/24/outline";
 import { useScaffoldContractWrite } from "~~/hooks/scaffold-eth";
+import { getTargetNetwork } from "~~/utils/scaffold-eth";
 
 export const ContractInteraction = () => {
   const [visible, setVisible] = useState(true);
   const [newGreeting, setNewGreeting] = useState("");
+  const configuredNetwork = getTargetNetwork();
 
   const { writeAsync, isLoading } = useScaffoldContractWrite({
     contractName: "YourContract",
@@ -77,7 +79,7 @@ export const ContractInteraction = () => {
 
           <div className="mt-4 flex gap-2 items-start">
             <span className="text-sm leading-tight">Price:</span>
-            <div className="badge badge-warning">0.01 ETH + Gas</div>
+            <div className="badge badge-warning">0.01 {configuredNetwork.nativeCurrency.symbol} + Gas</div>
           </div>
         </div>
       </div>
--- a/packages/nextjs/components/scaffold-eth/Balance.tsx
+++ b/packages/nextjs/components/scaffold-eth/Balance.tsx
@@ -1,4 +1,5 @@
 import { useAccountBalance } from "~~/hooks/scaffold-eth";
+import { getTargetNetwork } from "~~/utils/scaffold-eth";
 
 type TBalanceProps = {
   address?: string;
@@ -9,6 +10,7 @@ type TBalanceProps = {
  * Display (ETH & USD) balance of an ETH address.
  */
 export const Balance = ({ address, className = "" }: TBalanceProps) => {
+  const configuredNetwork = getTargetNetwork();
   const { balance, price, isError, isLoading, onToggleBalance, isEthBalance } = useAccountBalance(address);
 
   if (!address || isLoading || balance === null) {
@@ -39,7 +41,7 @@ export const Balance = ({ address, className = "" }: TBalanceProps) => {
         {isEthBalance ? (
           <>
             <span>{balance?.toFixed(4)}</span>
-            <span className="text-xs font-bold ml-1">ETH</span>
+            <span className="text-xs font-bold ml-1">{configuredNetwork.nativeCurrency.symbol}</span>
           </>
         ) : (
           <>
--- a/packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
@@ -6,7 +6,7 @@ import { getTargetNetwork } from "~~/utils/scaffold-eth";
 export function useAccountBalance(address?: string) {
   const [isEthBalance, setIsEthBalance] = useState(true);
   const [balance, setBalance] = useState<number | null>(null);
-  const price = useAppStore(state => state.ethPrice);
+  const price = useAppStore(state => state.nativeCurrencyPrice);
 
   const {
     data: fetchedBalanceData,
--- a/packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
@@ -19,7 +19,9 @@ export function useAccountBalance(address?: string) {
   });
 
   const onToggleBalance = useCallback(() => {
-    setIsEthBalance(!isEthBalance);
+    if (price > 0) {
+      setIsEthBalance(!isEthBalance);
+    }
   }, [isEthBalance]);
 
   useEffect(() => {
--- a/packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
@@ -22,7 +22,7 @@ export function useAccountBalance(address?: string) {
     if (price > 0) {
       setIsEthBalance(!isEthBalance);
     }
-  }, [isEthBalance]);
+  }, [isEthBalance, price]);
 
   useEffect(() => {
     if (fetchedBalanceData?.formatted) {
--- a/packages/nextjs/hooks/scaffold-eth/useNativeCurrencyPrice.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useNativeCurrencyPrice.ts
@@ -8,17 +8,17 @@ const enablePolling = false;
 
 /**
  * Get the price of ETH based on ETH/DAI trading pair from Uniswap SDK
- * @returns ethPrice: number
+ * @returns nativeCurrencyPrice: number
  */
-export const useEthPrice = () => {
+export const useNativeCurrency = () => {
   const provider = useProvider({ chainId: 1 });
-  const [ethPrice, setEthPrice] = useState(0);
+  const [nativeCurrencyPrice, setNativeCurrency] = useState(0);
 
   // Get the price of ETH from Uniswap on mount
   useEffect(() => {
     (async () => {
       const price = await fetchPriceFromUniswap(provider);
-      setEthPrice(price);
+      setNativeCurrency(price);
     })();
   }, [provider]);
 
@@ -26,10 +26,10 @@ export const useEthPrice = () => {
   useInterval(
     async () => {
       const price = await fetchPriceFromUniswap(provider);
-      setEthPrice(price);
+      setNativeCurrency(price);
     },
     enablePolling ? scaffoldConfig.pollingInterval : null,
   );
 
-  return ethPrice;
+  return nativeCurrencyPrice;
 };
--- a/packages/nextjs/hooks/scaffold-eth/useNativeCurrencyPrice.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useNativeCurrencyPrice.ts
@@ -10,15 +10,15 @@ const enablePolling = false;
  * Get the price of ETH based on ETH/DAI trading pair from Uniswap SDK
  * @returns nativeCurrencyPrice: number
  */
-export const useNativeCurrency = () => {
+export const useNativeCurrencyPrice = () => {
   const provider = useProvider({ chainId: 1 });
-  const [nativeCurrencyPrice, setNativeCurrency] = useState(0);
+  const [nativeCurrencyPrice, setNativeCurrencyPrice] = useState(0);
 
   // Get the price of ETH from Uniswap on mount
   useEffect(() => {
     (async () => {
       const price = await fetchPriceFromUniswap(provider);
-      setNativeCurrency(price);
+      setNativeCurrencyPrice(price);
     })();
   }, [provider]);
 
@@ -26,7 +26,7 @@ export const useNativeCurrency = () => {
   useInterval(
     async () => {
       const price = await fetchPriceFromUniswap(provider);
-      setNativeCurrency(price);
+      setNativeCurrencyPrice(price);
     },
     enablePolling ? scaffoldConfig.pollingInterval : null,
   );
--- a/packages/nextjs/hooks/scaffold-eth/useNativeCurrencyPrice.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useNativeCurrencyPrice.ts
@@ -7,7 +7,7 @@ import { fetchPriceFromUniswap } from "~~/utils/scaffold-eth";
 const enablePolling = false;
 
 /**
- * Get the price of ETH based on ETH/DAI trading pair from Uniswap SDK
+ * Get the price of Native Currency based on Native Token/DAI trading pair from Uniswap SDK
  * @returns nativeCurrencyPrice: number
  */
 export const useNativeCurrencyPrice = () => {
--- a/packages/nextjs/services/store/store.ts
+++ b/packages/nextjs/services/store/store.ts
@@ -10,11 +10,11 @@ import create from "zustand";
  */
 
 type TAppStore = {
-  ethPrice: number;
-  setEthPrice: (newEthPriceState: number) => void;
+  nativeCurrencyPrice: number;
+  setNativeCurrency: (newNativeCurrencyState: number) => void;
 };
 
 export const useAppStore = create<TAppStore>(set => ({
-  ethPrice: 0,
-  setEthPrice: (newValue: number): void => set(() => ({ ethPrice: newValue })),
+  nativeCurrencyPrice: 0,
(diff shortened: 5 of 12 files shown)


damianmarti:
Maybe add a getCurrencySymbol at utils/scaffold-eth?

kevinjoshi46b:
> Maybe add a getCurrencySymbol at utils/scaffold-eth?

Above the `getTargetNetwork` function there is a comment mentioning that it returns the extra network metadata so I thought it should be used here.

But ya maybe having a `getCurrencySymbol` that simply returns a string could also be neat.
If we were to implement this I would like to ask which of these two approaches are a better programming design decision and why?
1. Use `getTargetNetwork` function inside `getCurrencySymbol` function to get the data.
2. Get the data directly just like how it is implemented in `getTargetNetwork`

packages/nextjs/components/example-ui/ContractInteraction.tsx
```diff
@@ -77,7 +79,7 @@ export const ContractInteraction = () => {
 
           <div className="mt-4 flex gap-2 items-start">
             <span className="text-sm leading-tight">Price:</span>
-            <div className="badge badge-warning">0.01 ETH + Gas</div>
+            <div className="badge badge-warning">0.01 {configuredNetwork.nativeCurrency.symbol} + Gas</div>
```
carletex, line 82: This is just an example, inside example-ui. So no need to make it dynamic. You can remove it.

packages/nextjs/components/scaffold-eth/Balance.tsx
carletex, line 1: This is cool, but we need to handle the USD conversion in this component.

If people see 1 MATIC, and they click the Balance, they will get ~1900 USD (which is the ETH Price). So I see two options:

1. Fast: Disable the toggle funcionality when != ETH
2. Slow: Get the price from uniswap (as we are doing for ETH in `fetchPriceFromUniswap`) and display the correct conversion. We'd need to tweak a bunch of stuff
   - useEthPrice => useNativeCurrencyPrice();
   - state.setEthPrice / getEthPrice => state.setNativeCurrencyPrice / getNativeCurrencyPrice
   - Add the uniswap contract addresses in NETWORKS_EXTRA_DATA (if different that the ETH one).
   - etc...

Anyway up to you. Happy to merge the first option, and leaving the second one for another PR.

carletex (changes requested):
Thanks for this @kevinjoshi46b 

I left a couple of comments.

> Above the getTargetNetwork function there is a comment mentioning that it returns the extra network metadata so I thought it should be used here.

I think it's fine not having util (at least for now) since we are only going to use it one time now (after you remove the example UI).

packages/nextjs/components/scaffold-eth/Balance.tsx
carletex: Just saw #323 

I feel it should be a single PR. Feel free to merge it in this one.

packages/nextjs/components/example-ui/ContractInteraction.tsx
kevinjoshi46b: Ya but removing it from the example could be confusing right if they switch chains.
Why not keep the example to the point as well 🤷‍♂️

kevinjoshi46b:
> I think it's fine not having util (at least for now) since we are only going to use it one time now (after you remove the example UI).

If it could be a thing needed to be built in the future why not just add it right now. I just used getTargetNetwork because I thought the whole purpose of supplying the metadata in it was to use it in such cases.

carletex:
> If it could be a thing needed to be built in the future why not just add it right now.

Might not be needed in the future. We don't like early optimizations.

packages/nextjs/components/example-ui/ContractInteraction.tsx
carletex: It won't be confusing, having it dynamic might be more confusing. 

Maybe you are missing some context of how/why ExampleUI was created in the first place.

kevinjoshi46b: Ya that could be the case, can you please share any references?

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
```diff
@@ -1,15 +1,26 @@
...
-    return price;
-  } catch (error) {
-    console.error("useEthPrice - Error fetching ETH price from Uniswap: ", error);
+  const configuredNetwork = getTargetNetwork();
+  if (configuredNetwork.nativeCurrency.symbol == "ETH" || configuredNetwork.tokenAddress) {
+    try {
+      const DAI = new Token(1, "0x6B175474E89094C44Da98b954EedeAC495271d0F", 18);
+      const TOKEN = await Fetcher.fetchTokenData(
+        1,
+        configuredNetwork.tokenAddress ? configuredNetwork.tokenAddress : "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
```
carletex, line 12: `configuredNetwork.tokenAddress || "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"` is a more concise/clean way to write it.

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
```diff
@@ -1,15 +1,26 @@
...
-  try {
-    const DAI = new Token(1, "0x6B175474E89094C44Da98b954EedeAC495271d0F", 18);
-    const pair = await Fetcher.fetchPairData(DAI, WETH[DAI.chainId], provider);
-    const route = new Route([pair], WETH[DAI.chainId]);
-    const price = parseFloat(route.midPrice.toSignificant(6));
-    return price;
-  } catch (error) {
-    console.error("useEthPrice - Error fetching ETH price from Uniswap: ", error);
+  const configuredNetwork = getTargetNetwork();
+  if (configuredNetwork.nativeCurrency.symbol == "ETH" || configuredNetwork.tokenAddress) {
```
carletex, line 7: It should be `===`. 

Also the nesting is a bit ugly, let's use a guard clause.

```js
if (configuredNetwork.nativeCurrency.symbol !== "ETH" && !configuredNetwork.tokenAddress) {
  return 0;
}
```

packages/nextjs/services/store/store.ts
```diff
@@ -10,11 +10,11 @@ import create from "zustand";
  */
 
 type TAppStore = {
-  ethPrice: number;
-  setEthPrice: (newEthPriceState: number) => void;
+  nativeCurrencyPrice: number;
+  setNativeCurrency: (newNativeCurrencyState: number) => void;
```
carletex, line 14: `setNativeCurrencyPrice`

carletex (changes requested):
.

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
kevinjoshi46b: Oooo nice something new to learn. Thanks

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
kevinjoshi46b: Done

packages/nextjs/services/store/store.ts
kevinjoshi46b: Oops my bad

carletex:
Thanks @kevinjoshi46b , good job!!

Let us test a bit before merging.

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
```diff
@@ -1,15 +1,25 @@
...
+  if (configuredNetwork.nativeCurrency.symbol !== "ETH" && !configuredNetwork.tokenAddress) {
+    return 0;
+  }
   try {
     const DAI = new Token(1, "0x6B175474E89094C44Da98b954EedeAC495271d0F", 18);
-    const pair = await Fetcher.fetchPairData(DAI, WETH[DAI.chainId], provider);
-    const route = new Route([pair], WETH[DAI.chainId]);
+    const TOKEN = await Fetcher.fetchTokenData(
+      1,
+      configuredNetwork.tokenAddress || "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
```
technophile-04, line 14: I think this can be improved because its kind of confusing setting fallback here(which won't work and works only for one condition when on `mainnnet`) since kind of conflicting guard check above (am I missing something ? ) : 
```ts
if (configuredNetwork.nativeCurrency.symbol !== "ETH" && !configuredNetwork.tokenAddress) {
    return 0;
  }
```

Also I think for consistency, lets move WETH (0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2) address to `networks.ts`  and add : 

```ts
[chains.mainnet.id]: {
    color: "#ff8b9e",
     tokenAddress : "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
  },
```

And now, in this file we can do : 
```ts
export const fetchPriceFromUniswap = async (provider: Provider): Promise<number> => {
  const configuredNetwork = getTargetNetwork();
  if (!configuredNetwork.tokenAddress)  return 0;
  try {
      //  ...
      const TOKEN = await Fetcher.fetchTokenData(1, configuredNetwork.tokenAddress, provider);
     // ...
  }
  // ...
 }

```

packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -5,6 +5,7 @@ import scaffoldConfig from "~~/scaffold.config";
 export type TChainAttributes = {
   // color | [lightThemeColor, darkThemeColor]
   color: string | [string, string];
+  tokenAddress?: string;
```
technophile-04, line 8: Nitpick : Should we change it to `ethMainnetTokenAddress` ? 

Lol I am bad at names, but current one might be kind of confusing

packages/nextjs/hooks/scaffold-eth/useNativeCurrencyPrice.ts
```diff
@@ -8,28 +8,28 @@ const enablePolling = false;
 
 /**
  * Get the price of ETH based on ETH/DAI trading pair from Uniswap SDK
```
technophile-04, line 10: Maybe `Get the price of Native currency based on NativeToken/DAI...` please feel free to improve it 🙌

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
```diff
@@ -19,7 +19,9 @@ export function useAccountBalance(address?: string) {
   });
 
   const onToggleBalance = useCallback(() => {
-    setIsEthBalance(!isEthBalance);
+    if (price > 0) {
+      setIsEthBalance(!isEthBalance);
+    }
   }, [isEthBalance]);
```
technophile-04, line 25: I think we should add a `price` dependency here because even eslint sets the warning and makes sense to add it as a dependency

packages/nextjs/utils/scaffold-eth/networks.ts
carletex: `nativeCurrencyTokenAddress`? IDK, I suck at naming too haha

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
technophile-04: > It will work for every network that has ETH as their native currency, not only mainnet! There are a lot of them: goerli, sepolia, arbitrum, optimism, etc.

Ohhhhh, just noticed and makes sense,  lol I edited the comment at the same time 😆..... BTW then if we use the WETH address for `arbitrum`, `opitimism` etc again I think it might get misleading  

Yeah it's kindof of a PITA to add those addresses to each network :( 

BTW, what's our plan ? Do we plan to show a fallback, which returns the ETH price if it doesn't have `nativeCurrencyToken` address or we just don't allow the user to switch to USD if it's not available? 

depending on this we can use one of the both approach

technophile-04: > If I think about the Balance component... my vote will be to disable the switch to USD if it's not available. If we default ro the ETH price, it can be confusing, right ?

Yes 💯, So I think maybe we keep `nativeCurrencyTokenAddress` for each token in `NETWORKS_EXTRA_DATA ` (covers famous chain and that too big) making it as source of truth and if its not there then return 0 also leading to disabling switch ot USD functionality...

Reason : 
1. Having  `configuredNetwork.nativeCurrency.symbol !== "ETH"` might again mislead on chains like `arbitrium`, `optmism` showing actual ETH value 
2. It will kind of make it consistent (I think), and will also help the developer grasp it in one go if he wants to add his own chain

But cc @damianmarti for thoughts because even I don't feel too strong about the above approach and ok with anything

packages/nextjs/utils/scaffold-eth/networks.ts
technophile-04: `nativeCurrencyTokenAddress ` seems more intuitive  :), also I think we should add a small comment above it that help developers know why they need to add it maybe : 

"Used to show current native currency price" (can be improved a lot 😅)

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
kevinjoshi46b: > So I think maybe we keep `nativeCurrencyTokenAddress` for each token in `NETWORKS_EXTRA_DATA ` (covers famous chain and that too big) making it as source of truth and if its not there then return 0 also leading to disabling switch ot USD functionality...

Yup that is exactly how it has been implemented currently. You might want to read more about the current flow here https://github.com/scaffold-eth/scaffold-eth-2/pull/324#discussion_r1176580525

> 1. Having  `configuredNetwork.nativeCurrency.symbol !== "ETH"` might again mislead on chains like `arbitrium`, `optmism` showing actual ETH value

Can you elaborate on this more because I am unable to understand how will it be misleading?

packages/nextjs/utils/scaffold-eth/networks.ts
kevinjoshi46b: I do agree we should add a comment explaining the variable so that developers can use it for other chains if required.

And how about `nativeCurrencyMainnetAddress` for the variable name since it would be more precise explaining that its the chain's native currencies mainnet address.

packages/nextjs/hooks/scaffold-eth/useNativeCurrencyPrice.ts
kevinjoshi46b: Yup should have updated this thanks

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
kevinjoshi46b: Yup we should add it.

BTW just curious how did you get the eslint warning as in on running `yarn next:lint` it returned `No ESLint warnings or errors` for me?

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
technophile-04: So example if my targetNetwork is `arbitrum` (who's `nativeCurrency.symbol` is `ETH`) now in current situation we will first check : 
```ts

if (configuredNetwork.nativeCurrency.symbol !== "ETH" && 
//.   ^ This condition will be false and we don't go inside if 
!configuredNetwork.tokenAddress) {
    return 0;
  }
```

after that, we reach : 
```ts

// Here we will pick 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 (WETH) as token address instead the correct address should be 0xB50721BCf8d664c30412Cfbc6cf7a15145234ad1 (ARB token address)
const TOKEN = await Fetcher.fetchTokenData(
      1,
      configuredNetwork.tokenAddress || "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
      provider,
    );
```

So chains like `arbitrum` , `optimism`  etc whose symbold is `ETH` will always show `WETH / DAI` price istead they should have shown `ARB / DAI` etc which is misleading right ? Really sorry if I am missing something please feel to correct me 🙏

kevinjoshi46b: > So chains like `arbitrum` , `optimism` etc whose symbold is `ETH` will always show `WETH / DAI` price istead they should have shown `ARB / DAI` etc which is misleading right ? Really sorry if I am missing something please feel to correct me 🙏

If I am not wrong the native token of `arbitrum` is `ETH` itself and not `ARB` (`ETH` is used to pay gas fees and also fetching account balance will also return the amount of `ETH` and not `ARB` in the wallet) and hence `configuredNetwork.nativeCurrency.symbol` also return `ETH` which is correct

technophile-04: Ohh ohh, yes yes makes sense got confused with which balance we are showing tysm 🙌 resolving this then current situation makes more sense

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
technophile-04: My IDE was complaining :   
Screenshot 2023-04-28 at 12 51 32 PM

kevinjoshi46b: Ohhh its the eslint extension on VSC, I did have it installed thanks

packages/nextjs/utils/scaffold-eth/networks.ts
kevinjoshi46b: > I think `nativeCurrencyTokenAddress ` is suffice, the word `Token` adds some value 😆 but just my personal preference also we will be adding a comment about the above so will be easy to understand 🙌

ya fair, works

packages/nextjs/utils/scaffold-eth/fetchPriceFromUniswap.ts
kevinjoshi46b: Yes I myself was confused initially 😅

packages/nextjs/components/scaffold-eth/Balance.tsx
```diff
@@ -39,7 +41,7 @@ export const Balance = ({ address, className = "" }: TBalanceProps) => {
         {isEthBalance ? (
```
sverps, line 41: ~~Should we can rename this variable as well to something more fitting~~

On reading the changes further, I think we'd have to rename too many things... Even `scaffold-eth` technically mentions eth... So let's not go that route :sweat_smile:

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
```diff
@@ -19,8 +19,10 @@ export function useAccountBalance(address?: string) {
   });
 
   const onToggleBalance = useCallback(() => {
-    setIsEthBalance(!isEthBalance);
-  }, [isEthBalance]);
+    if (price > 0) {
```
sverps, line 22: Can the price be negative? Should we just check for it to be a number? Would `if (typeof price === 'number') {` achieve the same?

Or is this to avoid division by 0?

packages/nextjs/components/scaffold-eth/Balance.tsx
kevinjoshi46b: Are you suggesting undoing all the renaming already done?

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
kevinjoshi46b: I don't think it has got anything to do with division by 0.

But what I think is if the currency price ever falls down to 0 or maybe even below 0 which I feel can only happen when the currency has lost its value or there is something wrong with uniswap SDK then probably it isn't right to display the price anyways. What do you think?

kevinjoshi46b:
@technophile-04 rather than providing a disabled button with tooltip why don't we just hide the suffix button in such cases. What do you think?

technophile-04:
> @technophile-04 rather than providing a disabled button with tooltip why don't we just hide the suffix button in such cases. What do you think?

Yes even this is fine, I actually saw the disabled button while testing, and my first instinct was to show some message....but hiding it completely also feels fine to me since we are already showing errors in the console

technophile-04 (approved):

carletex (approved):
Nice work everyone!

PR #326, opened 2023-04-26
Save selected contract to local storage

Bogarne:


carletex:
Hey @Bogarne , thanks for working on this!

Instead of creating your own storage function, you can use `useLocalStorage` from `usehooks-ts`. There are a few examples in the codebase (e.g `useAutoConnect.ts: check how we use the hook and how we had a variable with the storage key.)

Could you refactor using it?

Thanks!

Bogarne:
@[carletex](https://github.com/carletex) thank you! 
Your guidance helped me get on the right track and complete my task.

packages/nextjs/pages/debug.tsx
```diff
@@ -1,12 +1,26 @@
...
+    selectedContractStorageKey,
+    contractNames[0],
+  );
+
+  useEffectOnce(() => setSelectedContract(selectedContractStorageValue));
+
+  const onSelectContract = (contractName: ContractName) => {
+    setSelectedContract(contractName);
+    setSelectedContractStorageValue(contractName);
+  };
```
sverps, line 23: I think a simpler approach would be to just change the `selectedContract` state from using `useState` to using `useLocalStorage`. There's no need to use 2 states, I believe.

sverps (changes requested):

packages/nextjs/pages/debug.tsx
Bogarne: Dear @sverps and @technophile-04, I wanted to share a different perspective on the current task.

`useLocalStorage` is a custom hook used to save and retrieve data from the local storage of a browser and is not a state of a React component. Access to the localStorage object becomes available after browser initialization and React component creation in the `componentDidMount` lifecycle method.

The `useState` hook becomes available in the `render` lifecycle method. The `render` method is a React component lifecycle method in which the `return` value of the component is rendered. The `render` method of a component is called before the `componentDidMount` method. A conditional expression is used in the `render` method: `contractName === selectedContractStorageValue.`
The initial state `selectedContractStorageValue` is set using `useState` and is available in the `render` method.

sverps: Hey @Bogarne, the [implementation](https://usehooks-ts.com/react-hook/use-local-storage) of `useLocalStorage` uses `useState` internally to actually hold the state (it doesn't simply retrieve/store it directly from localStorage, since that wouldn't trigger the expected rerenders when the value is modified). So there's really no point for us to manage a duplication of this state outside of the `useLocalStorage` hook.

(And a small side-note: we're using functional components. The old class component lifecycle methods aren't applicable here.)

sverps: Yeah, there's a bug in the usehooks-ts library that we use, see issue #340 
Let's discus options further there

damianmarti: @sverps It's using useLocalStorage now and the hydration error will be fixed when we merge https://github.com/scaffold-eth/scaffold-eth-2/pull/356

carletex: Ready to roll!

https://github.com/scaffold-eth/scaffold-eth-2/pull/356 has been merged

damianmarti: Please @sverps take a look when you can.

sverps: Glad we can finally get this working :sweat_smile:

PR #327, opened 2023-04-27
Update README: Add Vercel env var to disable type/lint checking

carletex:


technophile-04 (approved):
Yes, this was needed ! Looks nice to me 🙌

PR #333, opened 2023-04-30
fix: event history filter type

### LINKED ISSUE #332: useScaffoldEventHistory filter is of type any
The filters type should be improved to only allow valid filter params.

sverps:
Fixes #332


carletex:
Pushed https://github.com/scaffold-eth/scaffold-eth-2/pull/333/commits/fae9482eeedb735913ce95bc0addff8438fef02f making a indexed prop + updating the example.

What do you all think?

carletex:
Oh, and this is great @sverps . Love the autocompletion magic :)

Thanks

sverps:
Thanks for your addition @carletex 
Helps a lot in making clear how it should be used :+1:

PR #336, opened 2023-05-01
Added optional override for custom  gas limit when using useScaffoldContractWrite.

0xChijoke:
While hacking with Scaffold-Eth 2, I wanted to call my contract with 0.002 ETH which will be required to call another contract, I wanted a way to set a custom gas so I can set the gas and send 0.002 ETH without sending more ETH value.

Yes, I could just send in more ETH, but I didn't want to have to send in more 🥸 .

This PR is the improvise I used to add another optional override for setting a custom gas limit, which made sense to me👀.

Also tested to make sure this addition didn't break anything. 

Updated the README.md to include this usage in the useScaffoldWrite example. 

This is optional so if there is no need for a builder to customise their gas limit they can just ignore it.

The drawback to this was that I had to import BigNumber from ethers to handle the gas limit value which is a pretty heavy addition to the useScaffoldContractWrite hook.

Please take a look and let me know if this will be useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,8 @@ const { writeAsync, isLoading, isMining } = useScaffoldContractWrite({
   args: ["The value to set"],
   // For payable functions, expressed in ETH
   value: "0.01",
+  // An Optional overide for setting custom gasLimit.
+  gasLimit: "500000",
 });
 ```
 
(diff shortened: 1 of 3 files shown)


technophile-04:
Hey @0xChijoke thanks so much for raising the PR / issue, really appreciate it 🙌

I have raised a more detailed PR at #345 handling other `overrides` options too, because this PR just seems to only handle `gasLimit` 😅 and we were missing other `options` also like `nonce`, `blockTag` etc 

Closing this in favor of #345 but really appreciate for raising the PR and bringing the issue in notice 🙏

0xChijoke:
This is awesome and simple! Love it! Glad I could contribute to finding a solution.

PR #338, opened 2023-05-03
useAppStore => useGlobalState

carletex:
Whenever I'm hacking with @austintgriffith and we need to use `useAppStore`, we find it a bit confusing. I think this is especially true for iPhone users :D Where AppStore means something else.

I suggest changing it to something less confusing like `useAppState`. I like it because it resembles `useState`, but for the whole app. Another option we discussed was `useGlobalState`.

Thoughts?

### DIFF
--- a/packages/nextjs/components/Footer.tsx
+++ b/packages/nextjs/components/Footer.tsx
@@ -3,14 +3,14 @@ import { CurrencyDollarIcon } from "@heroicons/react/24/outline";
 import { HeartIcon } from "@heroicons/react/24/outline";
 import { SwitchTheme } from "~~/components/SwitchTheme";
 import { Faucet } from "~~/components/scaffold-eth";
-import { useAppStore } from "~~/services/store/store";
+import { useAppState } from "~~/services/store/store";
 import { getTargetNetwork } from "~~/utils/scaffold-eth";
 
 /**
  * Site footer
  */
 export const Footer = () => {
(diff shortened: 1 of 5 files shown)


sverps:
Looking at the explanation of this hook in the `store.ts`, it already mentions global state: `Think about it as a global useState.`, so I think `useGlobalState` best reflects what the hook does.

carletex:
> I think useGlobalState best reflects what the hook does

Austin likes that name too, and I don't feel too strong about any of the two, so I went and changed it to `useGlobalState`.

Thanks @sverps

technophile-04:
> I think useGlobalState best reflects what the hook does

Yup

**Nitpick:** Maybe `useGlobalStore`  because [zustand](https://github.com/pmndrs/zustand) seems to follow `use*Store` nomenclature in their README and docs but it's just nitpicking and even `useGlobalState` looks fine 🙌

carletex:
> Nitpick: Maybe useGlobalStore because [zustand](https://github.com/pmndrs/zustand) seems to follow use*Store nomenclature in their README and docs but it's just nitpicking and even useGlobalState looks fine raised_hands

Yep, this is why we use use*Store in the first place. I prefer `state` since `useState` vs `useGlobalState` makes a lot of sense to me.

sverps (approved):

technophile-04 (approved):

damianmarti:
Yes! useGlobalState sounds like the clearest and most explicit name!

Issue #273, opened 2023-03-30
README docs: hooks (+ components)


carletex:
Let's have a section on the README that explains

**Hooks**

Explain that are built on top of wagmi's, autocomplete / types magic, etc.
- useScaffoldContractRead
- useScaffoldContractWrite
- useScaffoldEventSubscriber
- useDeployedContractInfo

Just a one-liner with an example.

E.g.

**useScaffoldContractRead**: Use this hook to read a value from your deployed contracts.

```ts

  const { data: totalCounter } = useScaffoldContractRead({
    contractName: "YourContract",
    functionName: "userGreetingCounter",
    args: ["THE_ADDDRESS_AS_AN_ARGUMENT"],
  });
```

**Components**

Explain the main one with a screenshot of each one:

![image](https://user-images.githubusercontent.com/2486142/228882137-fcdd81a0-a5f1-4c59-ab92-4c3e61252ddf.png)

E.g.

**Address**: It displays the given address, with ENS resolution (name + avatar), copy icon and link to the blockexplorer
![image](https://user-images.githubusercontent.com/2486142/228882504-fed2301c-8464-40fd-9753-dc8ed4995b7f.png)

----

Maybe here:

![image](https://user-images.githubusercontent.com/2486142/228882636-bbb121a2-0643-4358-b42a-03acdac0f760.png)

We could add:
- A handful of components, hooks, and utils. See [link to section](#)

----

Anyway, my copys suck, this are just examples :D

carletex:
Related to #91

KcPele:
> Let's have a section on the README that explains
> 
> **Hooks**
> 
> Explain that are built on top of wagmi's, autocomplete / types magic, etc.
> 
> * useScaffoldContractRead
> * useScaffoldContractWrite
> * useScaffoldEventSubscriber
> * useDeployedContractInfo
> 
> Just a one-liner with an example.
> 
> E.g.
> 
> **useScaffoldContractRead**: Use this hook to read a value from your deployed contracts.
> 
> ```ts
>   const { data: totalCounter } = useScaffoldContractRead({
>     contractName: "YourContract",
>     functionName: "userGreetingCounter",
>     args: ["THE_ADDDRESS_AS_AN_ARGUMENT"],
>   });
> ```
> 
> **Components**
> 
> Explain the main one with a screenshot of each one:
> 
[...]

PR #339, opened 2023-05-03
Update readme: custom hooks section

edakturk14:
## Description

Update Custom Hooks section on README

## Additional Information

## Related Issues

Closes [273](https://github.com/scaffold-eth/scaffold-eth-2/issues/273)

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ const { writeAsync, isLoading, isMining } = useScaffoldContractWrite({
 });
 ```
 
-To send the transaction, you can call the writeAsync function returned by the hook. Here's an example usage:
+To send the transaction, you can call the `writeAsync` function returned by the hook. Here's an example usage:
 
 ```ts
 <button className="btn btn-primary" onClick={writeAsync}>
@@ -175,7 +175,7 @@ To send the transaction, you can call the writeAsync function returned by the ho
 </button>
 ```
 
-This example sends a transaction to the `YourContract` smart contract to call the `setGreeting` function with the arguments passed in `args`. The writeAsync function sends the transaction to the smart contract, and the isLoading and isMining properties indicate whether the transaction is currently being processed by the network.
+This example sends a transaction to the `YourContract` smart contract to call the `setGreeting` function with the arguments passed in `args`. The `writeAsync` function sends the transaction to the smart contract, and the `isLoading` and `isMining` properties indicate whether the transaction is currently being processed by the network.
 
 ### useScaffoldEventSubscriber:
 
@@ -259,7 +259,7 @@ const setGreeting = async () => {
 };
 ```
 
-This example uses the  `useScaffoldContract ` hook to obtain a contract instance for the  `YourContract ` smart contract. The data property of the returned object contains the contract instance that can be used to call any of the smart contract methods.
+This example uses the `useScaffoldContract` hook to obtain a contract instance for the `YourContract` smart contract. The data property of the returned object contains the contract instance that can be used to call any of the smart contract methods.
 
 ## Disabling type and linting error checks
 
--- a/README.md
+++ b/README.md
@@ -126,19 +126,22 @@ If you want to redeploy to the same production URL you can run `yarn vercel --pr
 
 ## Interacting with your Smart Contracts: SE-2 Custom Hooks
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -148,9 +151,11 @@ const { data: totalCounter } = useScaffoldContractRead({
 });
 ```
 
+This example retrieves the data returned by the `getGreeting` function of the `YourContract` smart contract. If the function accepts any arguments, they can be passed in the args array. The retrieved data is stored in the `data` property of the returned object.
```
technophile-04, line 154: Really love that we are mentioning the contract which we are using, should make it a markdown link like so -> [YourContract](https://github.com/scaffold-eth/scaffold-eth-2/blob/main/packages/hardhat/contracts/YourContract.sol) but even current one looks good

README.md
```diff
@@ -162,17 +167,19 @@ const { writeAsync, isLoading, isMining } = useScaffoldContractWrite({
...
-To make the actual call:
+To send the transaction, you can call the writeAsync function returned by the hook. Here's an example usage:
 
 ```ts
 <button className="btn btn-primary" onClick={writeAsync}>
   Send TX
 </button>
 ```
 
+This example sends a transaction to the `YourContract` smart contract to call the `setGreeting` function with the arguments passed in `args`. The writeAsync function sends the transaction to the smart contract, and the isLoading and isMining properties indicate whether the transaction is currently being processed by the network.
```
technophile-04, line 178: nitpick : maybe put isLoading, isMinning and writeAsync in backticks

README.md
```diff
@@ -246,6 +259,8 @@ const setGreeting = async () => {
 };
 ```
 
+This example uses the  `useScaffoldContract ` hook to obtain a contract instance for the  `YourContract ` smart contract. The data property of the returned object contains the contract instance that can be used to call any of the smart contract methods.
```
technophile-04, line 262: I think the extra space was added by mistake for `useScaffoldContract` and `YourContract`

technophile-04 (commented):
Really love the updations, it seems kind of more complete now to me 🙌

README.md
```diff
@@ -126,19 +126,22 @@ If you want to redeploy to the same production URL you can run `yarn vercel --pr
...
 
-- [useScaffoldContractRead](#usescaffoldcontractread)
-- [useScaffoldContractWrite](#usescaffoldcontractwrite)
-- [useScaffoldEventSubscriber](#usescaffoldeventsubscriber)
-- [useScaffoldEventHistory](#usescaffoldeventhistory)
-- [useDeployedContractInfo](#usedeployedcontractinfo)
-- [useScaffoldContract](#usescaffoldcontract)
+To help developers get started with smart contract interaction using Scaffold-ETH 2, we've provided the following custom hooks:
+
+- [useScaffoldContractRead](#usescaffoldcontractread): for reading public variables and getting data from read-only functions of your contract.
```
technophile-04, line 133: Not strong on adding descriptions here, like it seem repetitive to OG description(also a bit conservative about README length 😅) but yeah it makes sense to give quick one-liner about the hooks 

but would love to know other's thoughts

damianmarti: I think the quick one-liner about the hooks makes sense and helps to know better what the hooks actually do before going to the detailed section.

README.md
edakturk14: I actually would vote against not putting a link to the contract because it is for the example and also used in the other hook examples. So, think it would be more clear to show that it can be any contract and not specifically the one we are using. If there is general consensus to have it otherwise happy to make the change though :)

README.md
edakturk14: I wasn't sure either, seems like pretty self-explanatory, but also think the one liner can make it more explicit. On the readme and all the documentation I generally try to default to the more info the better.

damianmarti (approved):
I think that it's a very good improvement to the README! Thanks @edakturk14 !!

README.md
technophile-04: >So, think it would be more clear to show that it can be any contract and not specifically the one we are using

Ohh yes agreed and makes sense not to add a link

README.md
technophile-04: > I think the quick one-liner about the hooks makes sense and helps to know better what the hooks actually do before going to the detailed section.

> but also think the one-liner can make it more explicit.

Yup 💯💯

technophile-04 (approved):
Tysm Eda !! really love these changes, hooks docs seem to look strong now 🚀

carletex:
Thanks @edakturk14 looks great! And thank you all for the review.

At some point, we should start thinking about having a dedicated site for the docs. Maybe when the CLI is done?

PR #342, opened 2023-05-08
Fix daisyUI dropdown on IOS

technophile-04:
## Description

There seems some problem with dropdowns not working as expected on IOS mobile devices. 

**Reason** : 
We were using `button` instead of `lable`(suggested by [daisyUI](https://daisyui.com/components/dropdown/)) in their docs (faced this issue not only on safari but chrome too): 
Screenshot 2023-05-08 at 7 36 11 PM

**Before** 

https://user-images.githubusercontent.com/80153681/236846741-89fe0a40-29bd-4f37-9867-7e3cffb434fa.MP4

**After**

https://user-images.githubusercontent.com/80153681/236846788-6b2b41f8-afcc-4f31-8512-9ccd27dceb91.MP4

## Additional Information

## Related Issues

None


packages/nextjs/components/Header.tsx
```diff
@@ -58,14 +58,15 @@ export const Header = () => {
     <div className="sticky lg:static top-0 navbar bg-base-100 min-h-0 flex-shrink-0 justify-between z-20 shadow-md shadow-secondary">
       <div className="navbar-start w-auto lg:w-1/2">
         <div className="lg:hidden dropdown" ref={burgerMenuRef}>
-          <button
```
carletex, line 61: Do we also need to change it here?

It seems that we are not using the CSS magic from DaisyUI here, right? (tabIndex stuff) 

Here it looks like we are relying on react state (isDrawerOpen).

I personally prefer using React state (instead of CSS magic), but we are using it in some places (e.g. for modals). It feels comfy, because you don't have to manage any state.... but is it maybe a bit hacky? Maybe it's just me haha

What do you think, Shiv?

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
```diff
@@ -35,10 +35,10 @@ export const RainbowKitCustomConnectButton = () => {
               if (chain.unsupported || chain.id !== configuredNetwork.id) {
                 return (
                   <div className="dropdown dropdown-end">
-                    <button tabIndex={0} className="btn btn-error btn-sm dropdown-toggle">
+                    <label tabIndex={0} className="btn btn-error btn-sm dropdown-toggle">
```
carletex, line 38: In your second video, it seems that there is no separation between this label and the faucet button. Not sure if it's because of this change, but could we fix it?

carletex (commented):
Nice catch @technophile-04, thanks!!

Left a couple of comments

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
technophile-04: Actually, we are not showing `Wrong Network` and `Faucent Button` at the same time, like the `Faucet Button` is [hidden](https://github.com/scaffold-eth/scaffold-eth-2/blob/dd5082c400f07e7ce98879770290e4d6a49ddfa0/packages/nextjs/components/scaffold-eth/FaucetButton.tsx#L36) when you are on the `Wrong Network` / `Not Connected`. I hardcoded the `line 35` in this file to `true` in the above video for testing purposes. 

But found this on `main` branch : 
Screenshot 2023-05-10 at 1 57 36 AM


Added a minimal fix at [07df166](https://github.com/scaffold-eth/scaffold-eth-2/pull/342/commits/07df166c878e606a5009ce07c614949ab49a7ea5) : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/db4a70d7-1221-4d7f-bb48-9e9b03f50f77

But on too small devices it again merges :(, The way things are structured now leads me to add an [`xs` breakpoint ](https://tailwindcss.com/docs/screens#adding-smaller-breakpoints) to handle smaller devices  (this will also help responsiveness of footer) but not really sure if it's worth adding to `tailwind.config.js` and go that route.

packages/nextjs/components/Header.tsx
technophile-04: Just went through the code and it seems that we are just maintaining the state `isDrawerOpen ` to close dropdown list through `Hamburger icon` thats it. 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/ef58cb5d-9899-48cd-b6cc-8535ce993366



Rest all is handled by daisyUI like showing dropdown list in the `absolute` position, small animations and all, we used the `dropdown` class in the parent `div` so `tabIndex` was necessary to solve problem mentioned in the first video of the main comment. 

> I personally prefer using React state (instead of CSS magic),... but is it maybe a bit hacky?

Lol, even I have the same feeling! Should we go this route? I think we need to remove `dropdown` class and its all related classNames and add `absolute`  positioning and other classes ourselves not a big deal I think, also we will be more consistent on our source of truth for handling `opening`/ `closing` only depend on `state` unlike in current situation it is handled by both `css` and `js`, just wanted to confirm before I start making changes 🙌

carletex: Let's keep like this for now. But it's something that we might want to do in the future.

PR #345, opened 2023-05-13
allow `overrides` in useScaffoldContractWrite

technophile-04:
### `deployContracts.ts` is empty : 
This does problem does not show up if the `deployedContract.ts` is empty. 

###  Non empty `deployedContracts` : 
1.) It seems that typescript complains when you try to pass `overrides` to `useScafffoldContractWrite` : 
Screenshot 2023-05-13 at 3 26 52 PM

2.) Also another small problem is if people pass in `value` as  normal arg(ETH value string) and  `overrides : {gasLimit : BigNumber.from("210000")}` it will call `contract` with 0 `value` since we have omitted field `value` in `overrides`, checkout the logic below this line https://github.com/scaffold-eth/scaffold-eth-2/blob/dd5082c400f07e7ce98879770290e4d6a49ddfa0/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts#L42

Originally mentioned in #336 

### Proposed solution : 
**1). Fixing types** : 
Passed in "recklesslyUnprepared" argument to `UseContractWriteConfig` generic type : `UseContractWriteConfig<"recklesslyUnprepared">` 

Reason - It seems that wagmi kind of use Union discriminators for getting `UseContractWriteConfig` type checkout -> 
[this line on wagmi implementation](https://github.com/wagmi-dev/wagmi/blob/38f03dabde9c5aebfbe97565e18998c52a889bbc/packages/react/src/hooks/contracts/useContractWrite.ts#L19). 

Passing `"recklesslyUnprepared"` to generic helps typescript to correctly grab [`PartialBy<
      WriteContractUnpreparedArgs<TAbi, TFunctionName>,
      'abi' | 'address' | 'args' | 'functionName'
    >`](https://github.com/wagmi-dev/wagmi/blob/38f03dabde9c5aebfbe97565e18998c52a889bbc/packages/react/src/hooks/contracts/useContractWrite.ts#LL24C5-L27C6) this part of Union discriminator since that is what we want. 
    
**2.) Spreaded `overrides` and `resConfig` separately checkout [f32ebbf](https://github.com/scaffold-eth/scaffold-eth-2/pull/345/commits/f32ebbfd30d1d2eb03558bbd217f2626151c1d8b)

## Doubt 
I am kind of bit confused why TS does no complain on passing `overrides` when `deployedContract.ts` is empty. 

My guess is that its something to do with keys mentioned in `Omit` which is causing it to pick [`Prepared` branch](https://github.com/wagmi-dev/wagmi/blob/38f03dabde9c5aebfbe97565e18998c52a889bbc/packages/react/src/hooks/contracts/useContractWrite.ts#L20) of Union discriminator but don't know if that's true
https://github.com/scaffold-eth/scaffold-eth-2/blob/dd5082c400f07e7ce98879770290e4d6a49ddfa0/packages/nextjs/utils/scaffold-eth/contract.ts#L116

Reason for doubting above line since this the type being used when `deployedContract.ts` in non-empty. 

---

## Related Issues

None

**Please feel free to correct me if I missed something / close this PR in case of better approach 🙌**


technophile-04:
cc @sverps, can we merge this? I think the solution of  passing in "recklesslyUnprepared" argument to UseContractWriteConfig generic type makes sense and is minimal since we are deterministically using "recklesslyUnprepared" in our `useContractWrite` hook  🙌

sverps:
Yes, good fix :+1: 

(Will get resolved with the `viem` refactor, but we can merge this already in the meantime)

sverps (approved):

PR #346, opened 2023-05-14
add compatibility with SAFE

technophile-04:
## Description

It would be really great if we make SE-2 compatible with Gnosis SAFE by default as soon as it's deployed. 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/e928cb17-5acb-43e9-96d0-304df90076a0

Try adding it as a custom app on SAFE goerli ->  https://xx-ecru.vercel.app

**Note**: To make it work you have to deploy your Dapp first. It won't work if you are running `localhost`, we can make it work but we may need to configure CORS https://docs.safe.global/learn/safe-apps/get-started#cors and I think CORS options suggested in safe-docs are a bit harsh and don't like setting them as default in SE-2. 

I originally added it to [ETH-Splitter](https://github.com/Avelous/Eth-Splitter/pull/4) and thought it might be great if we provide SAFE compatibility out of the box , but feel free to close it 🙌

## Additional Information


packages/nextjs/hooks/scaffold-eth/useAutoConnect.ts
```diff
@@ -18,27 +19,30 @@ const getInitialConnector = (
...
+      // Don't connect if the user was previously connected to the burner wallet and burner is not allowed
+      if (previousWalletId === burnerWalletId && !allowBurner) return undefined;
 
-      const connector = connectors.find(f => f.id === previousWalletId);
-      return { connector };
+      return { connector: connectorMap.get(previousWalletId) };
     }
+  } else if (scaffoldConfig.walletAutoConnect && allowBurner) {
+    // If the user was not previously connected to a wallet, connect to the burner wallet if allowed
+    return { connector: connectorMap.get(burnerWalletId), chainId: defaultBurnerChainId };
```
damianmarti, line 45: It will be nice to have some unit test for this hook since following all the cases is not so easy. @technophile-04 what do you think?

austintgriffith:
I came here to post this same issue. It would be great if SE2 builds worked as safe apps out of the box.

packages/nextjs/hooks/scaffold-eth/useAutoConnect.ts
```diff
@@ -19,8 +22,14 @@ const getInitialConnector = (
   previousWalletId: string,
   connectors: Connector[],
 ): { connector: Connector | undefined; chainId?: number } | undefined => {
-  const targetNetwork = getTargetNetwork();
+  // Look for the SAFE connector instance and connect to it instantly if loaded in SAFE frame
+  const safeConnectorInstance = connectors.find(connector => connector.id === SAFE_ID && connector.ready);
 
+  if (safeConnectorInstance) {
+    return { connector: safeConnectorInstance };
+  }
```
technophile-04, line 30: We need to add this logic as Safe-sdk example says : 

> we want to automatically connect to the Safe if the app is loaded in the Safe Context. 
> ~ [doc](https://github.com/safe-global/safe-apps-sdk/tree/main/packages/safe-apps-wagmi)

technophile-04:
I think this PR is ready to merge 🙌, updated the logic so now its alot simpler and direct 

Here is the test app link : https://safe-goerli.vercel.app (can be tested on [gorerli safe](https://app.safe.global/home?safe=gor)


I also tested auto connect nicely just to be sure that we don't break anything 🙌

packages/nextjs/public/manifest.json
rin-st, line 1: why manifest is added here?

rin-st:
Gj!
Question: Should it show safe when I'm disconnected?

packages/nextjs/public/manifest.json
technophile-04: Actually while custom app to gnosis safe : 
Screenshot 2023-09-30 at 6 48 01 PM

it is mandatory to  expose `manifest.json`so that gnosis safe also can infer details to show in card, 

here is its mention in docs => https://docs.safe.global/safe-core-aa-sdk/safe-apps/get-started#manifest

technophile-04:
> Question: Should it show safe when I'm disconnected?

didn't get this 😅, like when you disconnect it will disconnect and when you click connect wallet it will show `Safe` wallet as an option(just pushed this changes) : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/48f0218d-a73c-49fd-953a-2e3a9e7a4b47

Also notice `Safe` wallet will only be shown if you open it in safe iframe, it wont show if you open https://safe-goerli.vercel.app/ in normal browser tab which is great 🙌

rin-st (approved):

technophile-04:
Tysm all for review merging this 🙌

PR #348, opened 2023-05-16
block confirmation options for useScaffoldContractWrite

technophile-04:
## Description
It would be really great if `useScaffoldContractWrite` could take in `callback` function which runs after a transaction is mined, also an option for the user pass in custom `blockConfirmations`. 

Added two optional properties for `useScaffoldContractWrite` : 
`onBlockConfirmation` : callback function which runs after `blockConfirmations = 1` mentioned
`blockConfirmations` : wait for this many block confirmations for the transaction

**Changes done** : 
- `useTransactor` already had option for callback, cleaned up `useTransactor` also added `blockConfirmations` optional arg
- Updated `useScaffoldContractWrite` types to take in optional properties mentioned above 

Demo : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/e1fce178-e3d7-4c38-b6b6-e09df0bb1313

## Additional Information


packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
```diff
@@ -69,7 +71,8 @@ export const useTransactor = (_signer?: Signer): TTransactionFunc => {
       notificationId = notification.loading(
         <TxnNotification message="Waiting for transaction to complete." blockExplorerLink={blockExplorerTxURL} />,
       );
-      transactionReceipt = await transactionResponse.wait();
+
+      const transactionReceipt = await transactionResponse.wait(options?.blockConfirmations);
```
damianmarti, line 75: If options?.blockConfirmations is undefined, will it use the default one block confirmation on wait?

damianmarti (approved):
LGTM, only a small question.

Thanks for the feature!!

packages/nextjs/hooks/scaffold-eth/useTransactor.tsx
technophile-04: Yes yes, here in docs -> https://docs.ethers.org/v5/api/providers/types/#providers-TransactionResponse

In implementation : 
https://github.com/ethers-io/ethers.js/blob/c785c1e5153d5a85a78f90f177577f2c8df15529/src.ts/providers/abstract-provider.ts#LL1017C61-L1017C61

So if its `undefined | null` or the arg is not passed it uses 1 as the default

carletex:
Great stuff @technophile-04 

This is complementary to what Wagmi offers

image

Ifechukwudaniel:
The font on the demo ☹️☹️☹️

PR #349, opened 2023-05-16
Memoize filters to avoid infinite rendering

moltam89:
## Description

Hey guys,

I've noticed that the Example UI page keeps re-rendering (see the console logs), because the filters object is used as a dependency of the useEffect hook.
Memoizing the object solved the issue for me.

Cheers,
Tamas

https://github.com/scaffold-eth/scaffold-eth-2/assets/1397179/acacd893-a745-4c8c-aaaa-aa335fa01f82

## Additional Information

Your ENS/address:
monyo.eth

### DIFF
--- a/packages/nextjs/components/example-ui/ContractData.tsx
+++ b/packages/nextjs/components/example-ui/ContractData.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 import Marquee from "react-fast-marquee";
 import { useAccount } from "wagmi";
 import {
@@ -38,6 +38,10 @@ export const ContractData = () => {
     },
   });
 
+  const filters = useMemo(() => {
+    return { greetingSetter: address };
+  }, [address]);
+
   const {
     data: myGreetingChangeEvents,
     isLoading: isLoadingEvents,
@@ -46,7 +50,7 @@ export const ContractData = () => {
     contractName: "YourContract",
     eventName: "GreetingChange",
     fromBlock: Number(process.env.NEXT_PUBLIC_DEPLOY_BLOCK) || 0,
-    filters: { greetingSetter: address },
+    filters: filters,
     blockData: true,
   });
 
--- a/packages/nextjs/components/example-ui/ContractData.tsx
+++ b/packages/nextjs/components/example-ui/ContractData.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useRef, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import Marquee from "react-fast-marquee";
 import { useAccount } from "wagmi";
 import {
@@ -38,10 +38,6 @@ export const ContractData = () => {
     },
   });
 
-  const filters = useMemo(() => {
-    return { greetingSetter: address };
-  }, [address]);
-
   const {
     data: myGreetingChangeEvents,
     isLoading: isLoadingEvents,
@@ -50,7 +46,7 @@ export const ContractData = () => {
     contractName: "YourContract",
     eventName: "GreetingChange",
     fromBlock: Number(process.env.NEXT_PUBLIC_DEPLOY_BLOCK) || 0,
-    filters: filters,
+    filters: { greetingSetter: address },
     blockData: true,
   });
 
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
@@ -109,6 +109,7 @@ export const useScaffoldEventHistory = <
     if (!deployedContractLoading) {
       readEvents();
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [
     provider,
     fromBlock,
@@ -118,7 +119,8 @@ export const useScaffoldEventHistory = <
     deployedContractData?.address,
     contract,
     deployedContractData,
-    filters,
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    JSON.stringify(filters),
     blockData,
     transactionData,
     receiptData,


technophile-04:
Hey, @moltam89 tysm for this and great catch !! 

Your suggested solution solves the issue, but I was wondering if it would be really great if we handle it inside `useScaffoldEventHistory` hook

Reason : We really don't want / expect from developers to always memoize the `filters` before passing it to `useScaffoldEvenHistory`, if they just pass in `filters` object (which is more intuative) they will always run into infinite renders. So its better we handle it internally what do you think ? 

A naive and easy solution coming to my mind is maybe we use [`JSON.stringify(filters)`](https://medium.com/suyeonme/react-lets-deep-dive-into-deps-array-of-useeffect-13ab96468db7) in hooks `useEffect`
 dependency : 
 https://github.com/scaffold-eth/scaffold-eth-2/blob/dd5082c400f07e7ce98879770290e4d6a49ddfa0/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts#L121
 
 But please feel free to suggest any other solution 🙌

moltam89:
Hey @technophile-04 ,

Hmm, this is a tough one. :)
ESLint doesn't like JSON.stringify, because it goes against the react-hooks/exhaustive-deps rule, but people still [recommend](https://javascript.plainenglish.io/a-less-known-trick-with-react-hook-dependencies-8dafaca7a150) it. 

I'll try to do some more research and get back to you.

Best,
Tamas

moltam89:
Hey @technophile-04 ,

I couldn't  really move forward with this, here are what I tried:

1. Use ethers' [encodeFilterTopics](https://github.com/scaffold-eth/scaffold-eth-2/commit/7507ff176adf0e8b42c9b1ed6da4039a1a113997). I think this could be a simpler solution, but we still have an array as a dependency, so it doesn't solve the original problem with the rerendering.

2. JSON.stringify - This one works correctly, but ESLint doesn't like it and I'm not sure if we want to disable it for this.

3. Using a custom [hook](https://github.com/sanjagh/use-custom-compare-effect) - this could solve the problem, but it requires external dependencies.

Best,
Tamas

carletex:
Thanks for the catch and the PR @moltam89  And thanks @technophile-04 for the fix suggestion. I think it could be the simple/best approach.

> JSON.stringify - This one works correctly, but ESLint doesn't like it and I'm not sure if we want to disable it for this.

IMO, ESLint rules are just there to give you some hints: "react-hooks/exhaustive-deps" is trying to warn you that you might be missing some dep.... which we are not! We are using JSON.stringify(dep), but the linter is not capable of inferring it. Using this method might be problematic on big objects / objects with circular references... but in this case, they will be small objects with one/few props.

My vote would be to add:
// eslint-disable-next-line react-hooks/exhaustive-deps

sverps:
Since the source of the rerenders is the creation of a new object, and not necessarily the values of the object keys changing, I think we could pass `...Object.values(filter)` to the dependency array. That feels a bit less hacky than the `JSON.stringify` approach.

To be confirmed if this solves the issue, haven't had time to play with it yet.

technophile-04:
Tysm @moltam89 for research 🙏

> 1. Use ethers' [encodeFilterTopics](https://github.com/scaffold-eth/scaffold-eth-2/commit/7507ff176adf0e8b42c9b1ed6da4039a1a113997). I think this could be a simpler solution, but we still have an array as a dependency, so it doesn't solve the original problem with the rerendering.

"array as a dependency, so it doesn't solve the original problem with the rerendering." yes :(, but it was really nice try 🙌

> 3. Using a custom [hook](https://github.com/sanjagh/use-custom-compare-effect) - this could solve the problem, but it requires external dependencies.

"but it requires external dependencies." yup :( 


> 2. JSON.stringify - This one works correctly, but ESLint doesn't like it and I'm not sure if we want to disable it for this.

Completely agree with Carols comment 💯

--- 


> I think we could pass `...Object.values(filter)` to the dependency array.

Thank @sverps 🙌 , I tried this but got this : 
Screenshot 2023-05-20 at 6 17 14 PM



--- 


So I think  doing `// eslint-disable-next-line react-hooks/exhaustive-deps` is ok and way to go with `JSON.stringify()` due to reasons mentioned by Carlos in https://github.com/scaffold-eth/scaffold-eth-2/pull/349#issuecomment-1555042745


@moltam89 can you please update, tysm 🙌

technophile-04 (commented):
Tysm @moltam89 for research 🙏

> 1. Use ethers' [encodeFilterTopics](https://github.com/scaffold-eth/scaffold-eth-2/commit/7507ff176adf0e8b42c9b1ed6da4039a1a113997). I think this could be a simpler solution, but we still have an array as a dependency, so it doesn't solve the original problem with the rerendering.

"array as a dependency, so it doesn't solve the original problem with the rerendering." yes :(, but it was really nice try 🙌

> 3. Using a custom [hook](https://github.com/sanjagh/use-custom-compare-effect) - this could solve the problem, but it requires external dependencies.

"but it requires external dependencies." yup :( 


> 2. JSON.stringify - This one works correctly, but ESLint doesn't like it and I'm not sure if we want to disable it for this.

Completely agree with Carols comment 💯

--- 


> I think we could pass `...Object.values(filter)` to the dependency array.

Thank @sverps 🙌 , I tried this but got this : 
Screenshot 2023-05-20 at 6 17 14 PM



--- 


So I think  doing `// eslint-disable-next-line react-hooks/exhaustive-deps` is ok and way to go with `JSON.stringify()` due to reasons mentioned by Carlos in https://github.com/scaffold-eth/scaffold-eth-2/pull/349#issuecomment-1555042745


@moltam89 can you please update, tysm 🙌

moltam89:
Hey guys,

Yeah, I was trying to solve the ESLint issue with `// eslint-disable-next-line react-hooks/exhaustive-deps`, but it only solves one of the 2 warnings.
Instead of 
```
112:6  Warning: React Hook useEffect has a missing dependency: 'filters'. Either include it or remove the dependency array.  react-hooks/exhaustive-deps
121:5  Warning: React Hook useEffect has a complex expression in the dependency array. Extract it to a separate variable so it can be statically checked.  react-hooks/exhaustive-deps
```
we get 
`112:6  Warning: React Hook useEffect has a missing dependency: 'filters'. Either include it or remove the dependency array.  react-hooks/exhaustive-deps`

I couldn't find a way to make ESLint ignore rules only for the current problematic hook.

Best,
Tamas

carletex:
Maybe like this? Having the `// disable` line twice:

image

Not very pretty, but for now...better than having an infinite loop :D

moltam89:
Oh yes, this is the way to get rid of both warnings!

moltam89:
I've forced pushed a rebased version of this, with @carletex 's solution.
Let me know if I should create a separate PR, or remove my initial idea from the commit history.

carletex (approved):

PR #350, opened 2023-05-19
fix: assert type as Abi to avoid ts error

### LINKED ISSUE #344: bug: type error in useScaffoldContract/EvenHistory when deployed multiple contracts

### Current Behavior

When you deploy two or more contracts and run `yarn next:check-types` you get `types error` for `useScaffoldContract/EventHistory` : 

 
Screenshot 2023-05-13 at 9 36 57 AM

### Expected Behavior

No type error should be thrown even when using multiple contracts

### Steps To Reproduce

1. Checkout to [`bug/useContract-eventHistory`](https://github.com/scaffold-eth/scaffold-eth-2/tree/bug/useContract-eventHistory) branch
2. `yarn chain`
3. `yarn deploy`
4. `yarn next:check-types`

### Anything else?

The problem is occurring due to type `abi` of wagmi's `useContract` hook mismatch.

According to me (I may be completely wrong, just a wild guess) : 
```ts
// wagmi is expecting only union of abiType | undefined 
abi : readonly [{...}] | undefined 

// but we are passing abiType | abiType | undefined
readonly[{...}] | readonly[{...}] | undefined
```

cc @sverps for the fix  🙌

sverps:
Fixes #344

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldContract.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldContract.ts
@@ -1,3 +1,4 @@
+import { Abi } from "abitype";
 import { ethers } from "ethers";
 import { useContract, useProvider } from "wagmi";
 import { useDeployedContractInfo } from "~~/hooks/scaffold-eth";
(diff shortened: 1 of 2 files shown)


technophile-04 (commented):
> I think a simple type assertion should suffice. Since we define our param and return types ourselves, we don't rely on any type inference inside the hook. Opened a PR for it.

Lol, Yup this makes sense from #344 comment and solves the issue ❤️ 

Can you please update it here too https://github.com/scaffold-eth/scaffold-eth-2/blob/dd5082c400f07e7ce98879770290e4d6a49ddfa0/packages/nextjs/hooks/scaffold-eth/useScaffoldContract.ts#L24

coz even this causes issue 🙌

PR #351, opened 2023-05-19
Block Explorer

portdeveloper:
## Description

This adds a Block Explorer page to Scaffold-eth 2.
This is my first ever serious pull request so bear with me.

I am open to any kind of feedback in any way.

## Additional Information

Your ENS/address: 0xA122A7Ed69597DBd77Fb2C539E13B7C3fB804637


packages/nextjs/pages/blockexplorer.tsx
```diff
@@ -0,0 +1,322 @@
...
+          <tbody className="bg-base-100 divide-y divide-primary-content text-base-content">
+            {Array.from(blocks.reduce((map, block) => map.set(block.hash, block), new Map()).values()).map(block =>
+              block.transactions.map((tx: TransactionWithFunction) => {
+                const receipt = transactionReceipts[tx.hash];
+
+                const shortTxHash = `${tx.hash.substring(0, 6)}...${tx.hash.substring(tx.hash.length - 4)}`;
+                const date = new Date(block.timestamp * 1000);
+                const month = String(date.getMonth() + 1).padStart(2, "0");
+                const day = String(date.getDate()).padStart(2, "0");
+                const timeMined = `${month}/${day} ${date.toLocaleTimeString("en-US", {
```
sverps, line 246: Since we have global users, best to let their OS/browser settings determine how to display date and timestrings. Forcing `month/day` and US timestring isn't the best approach.

I suggest simplifying this thing into `const timeMined = new Date(block.timestamp * 1000).toLocaleString()`

packages/nextjs/pages/blockexplorer.tsx
```diff
@@ -0,0 +1,322 @@
...
+                      {!receipt?.contractAddress ? (
+                        tx.to && <Address address={tx.to} />
+                      ) : (
+                        <span>
+                          Contract Creation:
+                          <Address address={receipt.contractAddress} />
+                        </span>
+                      )}
+                    </td>
+                    <td className="border px-4 py-2 text-base-content">{ethers.utils.formatEther(tx.value)} ETH</td>
```
sverps, line 285: native currency token instead of `ETH`

packages/nextjs/pages/transaction/[txHash].tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const [transaction, setTransaction] = useState<TransactionResponse | null>(null);
+  const provider = new ethers.providers.JsonRpcProvider("http://localhost:8545");
+
+  useEffect(() => {
+    if (txHash) {
+      const fetchTransaction = async () => {
+        const tx = await provider.getTransaction(txHash as string);
+        setTransaction(tx);
+        console.log("transaction page tx", tx);
```
Comment (sverps) [inline packages/nextjs/pages/transaction/[txHash].tsx:20]: Remove console.log

sverps (commented):
Nice work! Didn't find any major issues, just added some comments to polish it up a bit. Keep it up :rocket:

portdeveloper:
Thanks for the great feedback! I will work on the things you have suggested now.

packages/nextjs/pages/blockexplorer.tsx
portdeveloper: I fixed this but as this block explorer will only be used with a local hardhat node, the currency will always be ETH. It might be better to keep it as text. This is something trivial so we might decide it later as well.

packages/nextjs/components/blockexplorer/TransactionsTable.tsx
```diff
@@ -0,0 +1,139 @@
+import Link from "next/link";
+import { TransactionResponse } from "@ethersproject/abstract-provider";
+import { ethers } from "ethers";
+import { Address } from "~~/components/scaffold-eth";
+
+type TransactionWithFunction = TransactionResponse & { functionName?: string };
+
+// Define your types
+interface Block {
```
technophile-04, line 9: Maybe we can do this to be more accurate : 
```ts
export interface Block extends BlockWithTransactions {
  transactions: TransactionWithFunction[];
}
```

packages/nextjs/components/blockexplorer/TransactionsTable.tsx
```diff
@@ -0,0 +1,139 @@
...
+interface TransactionReceipts {
+  [key: string]: {
+    contractAddress: string;
+  };
+}
+
+interface TransactionsTableProps {
+  blocks: Block[];
+  transactionReceipts: TransactionReceipts;
+  configuredNetwork: ConfiguredNetwork;
```
technophile-04, line 31: Instead of passing it as props you could get in this component itself : 
```ts
const targetNetwork = getTargetNetwork()
``

File: packages/nextjs/components/blockexplorer/TransactionsTable.tsx [hunk 6]
```diff
@@ -0,0 +1,139 @@
...
+
+// Define your types
+interface Block {
+  hash: string;
+  transactions: TransactionWithFunction[];
+  timestamp: number;
+  number: number;
+}
+
+interface ConfiguredNetwork {
```
technophile-04, line 16: You could do to be more accurate: 
```ts
type ConfiguredNetwork = ReturnType<typeof getTargetNetwork>;
```

packages/nextjs/components/blockexplorer/TransactionsTable.tsx
```diff
@@ -0,0 +1,139 @@
...
+  nativeCurrency: {
+    symbol: string;
+  };
+}
+
+interface TransactionReceipts {
```
technophile-04, line 22: Since in `pages/blockexplorer.ts` we have defined state type as : 

```ts
interface TransactionReceipts {
  [key: string]: TransactionReceipt;
}
```

packages/nextjs/components/blockexplorer/TransactionsTable.tsx
```diff
@@ -0,0 +1,139 @@
...
+  };
+}
+
+interface TransactionsTableProps {
+  blocks: Block[];
+  transactionReceipts: TransactionReceipts;
+  configuredNetwork: ConfiguredNetwork;
+}
+
+export const TransactionsTable: React.FC<TransactionsTableProps> = ({
```
technophile-04, line 34: We are avoiding the use of `React.FC` checkout https://github.com/scaffold-eth/scaffold-eth-2/pull/36#discussion_r1015652259 and https://github.com/scaffold-eth/scaffold-eth-2/pull/35#discussion_r1008105245

Instead you can do : 

```ts
{ blocks, transactionReceipts } : {blocks: Block[];  transactionReceipts: TransactionReceipts; } 
```

Or checkout this -> https://github.com/scaffold-eth/scaffold-eth-2/blob/dd5082c400f07e7ce98879770290e4d6a49ddfa0/packages/nextjs/components/scaffold-eth/Address.tsx#L19

packages/nextjs/components/blockexplorer/TransactionsTable.tsx
```diff
@@ -0,0 +1,139 @@
...
+  blocks,
+  transactionReceipts,
+  configuredNetwork,
```
technophile-04, line 37: We could also avoid passing `configuredNetwork` and get it in this compoenent itsefl : 
```ts
const configuredNetwork = getConfiguredNetwork()
```

packages/nextjs/components/blockexplorer/TransactionsTable.tsx
```diff
@@ -0,0 +1,139 @@
...
+            <th
+              scope="col"
+              className="px-6 py-3 text-left text-xs font-medium text-primary-content uppercase tracking-wider"
+            >
+              Value ({configuredNetwork.nativeCurrency.symbol})
+            </th>
+          </tr>
+        </thead>
+        <tbody className="bg-base-100 divide-y divide-primary-content text-base-content">
+          {Array.from(blocks.reduce((map, block) => map.set(block.hash, block), new Map()).values()).map(block =>
```
technophile-04, line 89: Doubt : 
Why can't we directly use `blocks.map()` 😅 ?

Reason : 
```ts
  const blocks = [Block1, Block2, Block2]

  const blockObj = blocks.reduce(map, block) => map.set(block.hash, block), new Map<string, Block>())
  // blockObj = {
  // "0x1234": Block1,
  // "0x5678": Block2,
  // "0x9012": Block3,
  // }

  const values = blockObj.values()
  // values = MapIterator {Block1, Block2, Block3}

  const finalArr = Array.from(values)
  // finalArr = [Block1, Block2, Block3]
```
therefore finalArr == blocks, why don't we use `blocks.map` directly 😅


I am sure I might have missed something here, was just trying to keep intuitive and understand , but we can keep it as is

packages/nextjs/components/example-ui/ContractData.tsx
```diff
@@ -38,19 +37,19 @@ export const ContractData = () => {
...
-    isLoading: isLoadingEvents,
-    error: errorReadingEvents,
-  } = useScaffoldEventHistory({
-    contractName: "YourContract",
-    eventName: "GreetingChange",
-    fromBlock: Number(process.env.NEXT_PUBLIC_DEPLOY_BLOCK) || 0,
-    filters: { greetingSetter: address },
-    blockData: true,
-  });
+  // const {
```
technophile-04, line 40: I think its commented due to infinite render issue, just adding a comment so that we don't forgot to uncomment before merging 🙌

packages/nextjs/pages/blockexplorer.tsx
```diff
@@ -0,0 +1,172 @@
...
+// TODO: try useMemo to avoid unnecessary re-renders
+// TODO: Error handling for API calls
+// TODO: Add tests
+
+type TransactionWithFunction = TransactionResponse & { functionName?: string };
+
+const BLOCKS_PER_PAGE = 20;
+
+const Blockexplorer: NextPage = () => {
+  const [blocks, setBlocks] = useState<BlockWithTransactions[]>([]);
```
technophile-04, line 25: We should probably export type from -> https://github.com/scaffold-eth/scaffold-eth-2/pull/351/files#r1199794750 to use it for state `Block[]`
trying to be more consistent

packages/nextjs/pages/blockexplorer.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+  const configuredNetwork = getTargetNetwork();
+  const router = useRouter();
+
+  const provider = new ethers.providers.JsonRpcProvider("http://localhost:8545");
+
+  const fetchBlocks = async () => {
+    const blockNumber = await provider.getBlockNumber();
+    setTotalBlocks(blockNumber);
+    const blocks: BlockWithTransactions[] = [];
```
technophile-04, line 41: We should probably change its names because it clashes with state from above and might be confusing

```ts
  const [blocks, setBlocks] = useState<BlockWithTransactions[]>([]);
```

packages/nextjs/pages/blockexplorer.tsx
```diff
@@ -0,0 +1,172 @@
...
+    const receipts: { [key: string]: ethers.providers.TransactionReceipt } = {};
+    const startingBlock = blockNumber - currentPage * BLOCKS_PER_PAGE;
+
+    for (let i = 0; i < BLOCKS_PER_PAGE; i++) {
+      const blockNumberToFetch = startingBlock - i;
+      if (blockNumberToFetch < 0) {
+        break;
+      }
+
+      const block = (await provider.getBlockWithTransactions(blockNumberToFetch)) as BlockWithTransactions & {
```
technophile-04, line 51: We should avoid `as` assertion and find better way since `getBlockWithTransaction` transactions clearly won't have `functionName`property although we are dynamically asssiging key in below

Please ignore this comment, just leavning note here so that we can fix this before merging

packages/nextjs/pages/blockexplorer.tsx
```diff
@@ -0,0 +1,172 @@
...
+      for (const tx of block.transactions as TransactionWithFunction[]) {
+        const receipt = await provider.getTransactionReceipt(tx.hash);
+        receipts[tx.hash] = receipt;
+
+        if (tx.data.length >= 10) {
+          for (const [contractName, contractInterface] of Object.entries(interfaces)) {
+            try {
+              const decodedData = contractInterface.parseTransaction({ data: tx.data });
+              tx.functionName = `${contractName}: ${decodedData.name}`;
+              break;
```
technophile-04, line 73: Found an edge case with this logic when functions have same signature in different contract it shows the first contractName which matches : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/c093e54f-36ee-47d7-8fa4-a801f8c9e73e

technophile-04:
Hey @portdeveloper really awesome work !! Love this !! 

Really sorry for reviewing it while it was in draft but had some time and also exited for this 🚀, so went through the logic of `pages/blockExplorer` and found some nitpicking, we could also probably look at this in the end before merging

Really appreciate the PR and work, tysm 🙏....Also please feel free to ask for help to anyone if TS gets too complicated 🙌

packages/nextjs/pages/address/[address].tsx
```diff
@@ -0,0 +1,74 @@
...
+    block.transactions.some(
+      tx => tx.from.toLowerCase() === address.toLowerCase() || tx.to?.toLowerCase() === address.toLowerCase(),
+    ),
+  );
+
+  return (
+    <div className="m-10 mb-20">
+      <div className="flex justify-start mb-5">
+        <button
+          className="btn bg-primary text-primary-content hover:bg-accent hover:text-accent-content shadow-md transition duration-500 ease-in-out transform hover:-translate-y-1 hover:scale-110"
```
Comment (carletex) [inline packages/nextjs/pages/address/[address].tsx:44]: No need to go this fancy here :D 

We can do just `btn btn-primary` (or secondary?) to be consistent, and also it handles the hover state and everything.

Same with other buttons in this PR.

packages/nextjs/components/blockexplorer/TransactionsTable.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { Address } from "~~/components/scaffold-eth";
+import { getTargetNetwork } from "~~/utils/scaffold-eth";
+import { TransactionWithFunction, TransactionsTableProps } from "~~/utils/scaffold-eth/block";
+
+export const TransactionsTable = ({ blocks, transactionReceipts, isLoading }: TransactionsTableProps) => {
+  const targetNetwork = getTargetNetwork();
+
+  return (
+    <div className="overflow-x-auto shadow-lg">
+      <table className="min-w-full divide-y divide-primary shadow-lg rounded-lg bg-neutral">
```
carletex, line 12: Check the DaisyUI table component:

https://daisyui.com/components/table/

For all the tables, we should rely on it all we can.

P.S. I like the hover effect on rows (also included on DaisyUI), but maybe only for the `TransactionsTable`. I don't see its value in other places, like in the [txHash] table

packages/nextjs/components/example-ui/ContractData.tsx
carletex: Since you back-merged main here, this should be fixed now

packages/nextjs/components/scaffold-eth/Address.tsx
carletex, line 1: This happens to all the addresses.

This PR
image

Main:
image

packages/nextjs/pages/address/[address].tsx
```diff
@@ -0,0 +1,74 @@
...
+            <span className="font-bold text-lg">
+              <Address address={address} format="long" />
+            </span>
+          </h2>
+        </div>
+        <div className="flex-1 text-secondary-content p-4 rounded-lg shadow-lg flex items-center justify-center">
+          <h2 className="text-lg text-center">
+            Balance:{" "}
+            <span className="font-bold text-lg">
+              {parseFloat(balance).toFixed(2)} {targetNetwork.nativeCurrency.symbol}
```
Comment (carletex) [inline packages/nextjs/pages/address/[address].tsx:63]: image

We don't probably need both "cards" (address + balance) and we can group them into a single one. Maybe smaller and not centered.

carletex (commented):
Hey @portdeveloper thank you very much for this work. I'm very excited about this PR! Also thanks @technophile-04 & @sverps for the insightful review.

I gave it a first pass and left some comments. Let's keep iterating!

When we are almost there, we can also ask Andrea (the designer) to take a look and give us some feedback. Anyway, it's looking pretty good IMO.

packages/nextjs/pages/address/[address].tsx
Comment (carletex) [reply packages/nextjs/pages/address/[address].tsx:63]: Also, let's use the `<Balance>` component!

Comment (portdeveloper) [reply packages/nextjs/pages/address/[address].tsx]: Perhaps [address].tsx page could use a new look. I have updated it to look like debug page's contract details but it doesn't look right to me.

packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -93,6 +93,10 @@ export function getBlockExplorerTxLink(network: Network, txnHash: string) {
  */
 export function getBlockExplorerAddressLink(network: chains.Chain, address: string) {
   const blockExplorerBaseURL = network.blockExplorers?.default?.url;
+  if (network.id === 31337) {
+    return `http://localhost:3000/address/${address}`;
```
carletex, line 97: We need to update it to the new URL /blockexplorer/address

Also, would `/blockexplorer/address/${address}` work?  Without the localhost:3000. There might some cases when the dev env runs in a different port.

packages/nextjs/components/scaffold-eth/Address.tsx
```diff
@@ -76,6 +77,10 @@ export const Address = ({ address, disableAddressLink, format }: TAddressProps)
       </div>
       {disableAddressLink ? (
         <span className="ml-1.5 text-lg font-normal">{displayAddress}</span>
+      ) : blockExplorerAddressLink.includes("blockexplorer") ? (
```
carletex, line 80: I think we could have a less hacky way of doing this. We are already using `getTargetNetwork()` in this file, so we could rely on it (check if we are in the local network).

packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -93,6 +93,10 @@ export function getBlockExplorerTxLink(network: Network, txnHash: string) {
  */
 export function getBlockExplorerAddressLink(network: chains.Chain, address: string) {
   const blockExplorerBaseURL = network.blockExplorers?.default?.url;
+  if (network.id === 31337) {
```
carletex, line 96: For this, we have been doing

```if (network.id === hardhat.id)```

instead of hardcoding the chain Id.

File: packages/nextjs/components/blockexplorer/PaginationButton.tsx [hunk 22]
```diff
@@ -0,0 +1,48 @@
...
+    () => currentPage + 1 >= Math.ceil(totalItems / ITEMS_PER_PAGE),
+    [currentPage, totalItems],
+  );
+
+  const prevButtonClass = useMemo(
+    () => (isPrevButtonDisabled ? "bg-gray-200 cursor-default" : "btn btn-primary"),
+    [isPrevButtonDisabled],
+  );
+
+  const nextButtonClass = useMemo(
```
rin-st, line 24: nitpick: all the `useMemo`s are redundant

packages/nextjs/components/blockexplorer/TransactionHash.tsx
```diff
@@ -0,0 +1,37 @@
+import { useState } from "react";
+import Link from "next/link";
+import { CopyToClipboard } from "react-copy-to-clipboard";
+import { CheckCircleIcon, DocumentDuplicateIcon } from "@heroicons/react/24/outline";
+
+export const TransactionHash = ({ hash }: { hash?: string }) => {
```
rin-st, line 6: it seems that hash always exists, so it shouldn't be optional. And if it's optional `Link` will be broken

packages/nextjs/hooks/scaffold-eth/useContractLogs.ts
```diff
@@ -0,0 +1,37 @@
...
+    };
+
+    const handleLog = (log: ethers.providers.Log) => {
+      setLogs(prevLogs => [...prevLogs, log]);
+    };
+
+    const filter = { address: address };
+    provider.on(filter, handleLog);
+
+    fetchLogs();
```
rin-st, line 29: shouldn't it be before `provider.on....`?

packages/nextjs/pages/blockexplorer/address/[address].tsx
```diff
@@ -0,0 +1,177 @@
...
+  const [isContract, setIsContract] = useState(false);
+
+  const provider = useMemo(() => new ethers.providers.JsonRpcProvider("http://localhost:8545"), []);
+
+  useEffect(() => {
+    const checkIsContract = async () => {
+      const contractCode = await provider.getCode(address);
+      setIsContract(contractCode !== "0x");
+    };
+    console.log(Object.keys(deployedContracts[31337][0].contracts)[0]);
```
Comment (rin-st) [inline packages/nextjs/pages/blockexplorer/address/[address].tsx:43]: log

packages/nextjs/pages/blockexplorer/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+            </a>{" "}
+            instead
+          </p>
+        </>,
+      );
+    }
+  }, [error, ConnectedChain]);
+
+  return (
+    <>
```
rin-st, line 52: can be removed

rin-st:
hey @portdeveloper ! Awesome job! 🎉 

Added some small comments. Also, noticed that padding for data can be removed
Снимок экрана 2023-05-29 в 17 59 20

and font sizes for balance are different
Снимок экрана 2023-05-29 в 17 57 49

since it's mostly nitpicking and pr already large, you can change it in another pr or omit part of it

thank you!

damianmarti:
LGTM!!

We can make some improvements later from the feedback using it for sure...

Amazing new feature!!! Thanks @portdeveloper !!

packages/nextjs/components/scaffold-eth/Address.tsx
```diff
@@ -76,6 +78,10 @@ export const Address = ({ address, disableAddressLink, format }: TAddressProps)
       </div>
       {disableAddressLink ? (
         <span className="ml-1.5 text-lg font-normal">{displayAddress}</span>
+      ) : getTargetNetwork().id === chains.hardhat.id ? (
+        <span className="ml-1.5 text-lg font-normal">
+          <Link href={`/blockexplorer/address/${address}`}>{displayAddress}</Link>
```
technophile-04, line 83: We can use `blockExplorerAddressLink` directly since we have already added the logic to handle `hardhat` in `getBlockExplorerAddressLink` 

NITPICK : Also instead of `chains.hardhat.id` we can : 
```ts
import {hardhat} from "wagmi/chains"
hardhat.id
```
which might help in tree shaking

packages/nextjs/pages/blockexplorer/address/[address].tsx
```diff
@@ -0,0 +1,176 @@
...
+    "artifacts",
+    "build-info",
+  );
+
+  if (!fs.existsSync(buildInfoDirectory)) {
+    throw new Error(`Directory ${buildInfoDirectory} not found.`);
+  }
+
+  const chainId = chains.hardhat.id;
+  const contractPath = `contracts/${Object.keys(deployedContracts[chainId][0].contracts)[0]}.sol`;
```
Comment (technophile-04) [inline packages/nextjs/pages/blockexplorer/address/[address].tsx:171]: TODO : 

If there are multiple contracts, it will be showing the same code and assembly(first contract deployed) for other contracts too 

Maybe handle in different PR tackling all issues related with multiple contracts if it get too complicated

technophile-04 (commented):
Really awesome work @portdeveloper !!! Looking solid in both functionality and UI wise 💪🤩!! 

Found a few small changes and some nitpicks, please feel free ignore to nitpicks 😁

--- 

One small addition,  we can also add `/blockexplorer/tx/${txnHash}` in https://github.com/scaffold-eth/scaffold-eth-2/blob/c7af09a8b7717f9e7c7e2368e7a5cd9030ccedaa/packages/nextjs/utils/scaffold-eth/networks.ts#L63

So that we can point to local block explorer when doing transactions locally :) (Was suggested by Carlos ) 🙌

--- 

#### Next Steps( Handling in future PRs) : 
Adding a basic TODO so that I don't forget 

- [ ] Handling multiple contracts (bug: showing same `code` for  different contracts)

- [ ] Better handling when `targetNetwork` is not hardhat 
   - eg. Removing `Block Explorer` tab from navbar, **maybe** redirecting to the home page if people directly visit some dynamic route

- [ ] Finding a way to exclude `pages/blockexplorer` in build time 
   - Reason when you deploy due to `getServerSideProps` in `[address].tsx` Vercel uses `Node.js` runtime which is not unnecessary (This is not big deal but great if we can handle it)  
Screenshot 2023-05-30 at 5 52 46 PM
  - Possible solution [pageExtensions](https://nextjs.org/docs/pages/api-reference/next-config-js/pageExtensions) but need to dig a bit

packages/nextjs/pages/blockexplorer/address/[address].tsx
```diff
@@ -0,0 +1,176 @@
...
+  contractData: AddressCodeTabProps | null;
+};
+
+const AddressPage = ({ address, contractData }: PageProps) => {
+  const router = useRouter();
+  const { blocks, transactionReceipts, currentPage, totalBlocks, setCurrentPage, isLoading } = useFetchBlocks();
+  const [activeTab, setActiveTab] = useState("transactions");
+  const [isContract, setIsContract] = useState(false);
+
+  const provider = useMemo(() => new ethers.providers.JsonRpcProvider("http://localhost:8545"), []);
```
Comment (carletex) [inline packages/nextjs/pages/blockexplorer/address/[address].tsx:36]: We shouldn't hardcode the JSON provider here.

We have the `getLocalProvider` util

packages/nextjs/hooks/scaffold-eth/useContractLogs.ts
carletex, line 1: Do we need to create a new hook for this?

We already have two hooks for dealing with events / logs (useScaffoldEventHistory & useScaffoldEventSubscriber). Maybe we could use the first one allowing an empty "eventName"? (so it returns all?)

cc @technophile-04 @damianmarti

packages/nextjs/pages/blockexplorer/address/[address].tsx
Comment (carletex) [reply packages/nextjs/pages/blockexplorer/address/[address].tsx:36]: (same for other places where you do the same)

carletex:
We are heading in the right direction! Really nice work @portdeveloper 

Thanks @technophile-04 for the insightful review comments.

This PR got too big: too many components / utils / hooks & changes for a single PR :sweat_smile: 

It's already done, so let's focus on cleaning up as best as possible, trying not to merge bugs / messy code... and we can focus on tweaking / adding new functionality on later PRs.

I'll do a deeper review when all the current stuff is fixed.

Thanks!!

packages/nextjs/components/blockexplorer/TransactionsTable.tsx
```diff
@@ -0,0 +1,100 @@
...
+            <th className="bg-primary">Block Number</th>
+            <th className="bg-primary">Time Mined</th>
+            <th className="bg-primary">From</th>
+            <th className="bg-primary">To</th>
+            <th className="bg-primary text-end">Value ({targetNetwork.nativeCurrency.symbol})</th>
+          </tr>
+        </thead>
+        {isLoading ? (
+          <tbody>
+            {[...Array(20)].map((_, rowIndex) => (
```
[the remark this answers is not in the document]
portdeveloper: If I use `Array(20).map` instead of `[...Array(20)].map`, the skeleton UI does not work.

packages/nextjs/pages/blockexplorer/address/[address].tsx
Comment (portdeveloper) [reply packages/nextjs/pages/blockexplorer/address/[address].tsx]: I tried to use getLocalProvider but still added JsonRpcProvider as fallback because getLocalProvider may return undefined and that causes bunch of other issues.
Let me know if there are any better solutions because I couldn't find one.

Pabl0cks:
Awesome job @portdeveloper!

A nitpick UX improvement: In the block explorer transactions table, make links to Tx hash, To and From more visual, so users know they can click on them. Maybe [daisy link component](https://daisyui.com/components/link/) or  `hover:underline` can work.

packages/nextjs/components/blockexplorer/TransactionsTable.tsx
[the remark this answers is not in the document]
sverps: Yeah, `Array(20).map(...)` doesn't do the same, since `Array(20)` on its own creates an array of empties, it's a bit of a js quirk.

image

I usually do `Array.from({length: 20})`, not sure if there's any benefit/downside to either option.

packages/nextjs/hooks/scaffold-eth/useContractLogs.ts
technophile-04: `useScaffoldEventHistory ` is kind of stricter in terms of args like `contractName` and other stuff,  the place where `useContractLogs` is used has only access to `address`(which we are not accepting in `useScaffoldEventHistory`)...I mean we can tweak `useScaffoldEvenHistory` by just adding `if condtion` but might need to do some changes so that we don't lose autocompletion and dx

 Maybe we can handle it in separate PR 😅 ? But yeah I do see a overlap  between `useContractLogs` and `useScaffoldEventHistory`

technophile-04 (approved):
### Desc : 
- Added some minor general cleanup commits (tried not to be nitpicky), tried not touching the logic part since it looked pretty much solid 
- Just the UI was breaking when `deployedContracts` was `null` ....fixed at [5fda20e](https://github.com/scaffold-eth/scaffold-eth-2/pull/351/commits/5fda20e5780c822507e35d45717780d908ca0bb6) also made `deployedContract` back to null

Tested it out properly and everything works like a charm, also updated TODO's for future PRs in https://github.com/scaffold-eth/scaffold-eth-2/pull/351#pullrequestreview-1450327445

I think we are ready to merge once others do final final test/review 🚀🚀🚀, excited for this to get merged, really great work Port !!!!

packages/nextjs/hooks/scaffold-eth/useContractLogs.ts
carletex: Make sense! Thanks

carletex:
Hey all! I've been testing this, and everything seems to be working great. Amazing work @portdeveloper and @technophile-04. This is going to be a great addition.

I think we need to fix some styling before merging. I promise these will be the latest requests :D Since it's a really big PR, I'm sure we'll find some issues later, but we can handle them in a different PR.

### 1. Everything is huge

Most of the buttons, text, tabs, etc are really big compared to other elements of the page.

Eg: 
image
image

Let's remove all the `*-lg` stuff. It should be regular sizes (no modifier) and in some cases `*-sm` (buttons, maybe some texts)

### 2. Light mode

2.1 There is no background color here, all is floating on the page (dark mode looks great tho):
image

2.2 Not important but the table has no zebra style (maybe we need to tweak something in the tailwind styles?). Also, no need to tackle this here if we don't want to.
image

----

1 and 2.1 are bugging me, but other than that... everything looks great.

I'm happy to work on this too, if you guys are not sure about some of the stuff.

packages/nextjs/pages/index.tsx
```diff
@@ -45,6 +45,16 @@ const Home: NextPage = () => {
                 to build your own UI.
               </p>
             </div>
+            <div className="flex flex-col bg-base-100 px-10 py-10 text-center items-center max-w-xs rounded-3xl">
+              <MagnifyingGlassIcon className="h-8 w-8 fill-secondary" />
+              <p>
+                Explore your local hardhat node with the{" "}
```
carletex, line 51:
```suggestion
                Explore your local transactions with the{" "}
```

I know that "local hardhat node" is actually more accurate, but I feel this makes more sense for beginners. Anyway, it's just a soft suggestion :D

packages/nextjs/pages/blockexplorer/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+          </p>
+          <p className="mt-1 break-normal">
+            - Or you can change <code className="italic bg-base-300 text-base font-bold">targetNetwork</code> in{" "}
+            <code className="italic bg-base-300 text-base font-bold">scaffold.config.ts</code>
+          </p>
+        </>,
+      );
+    }
+
+    if (getTargetNetwork().id !== hardhat.id && !error) {
```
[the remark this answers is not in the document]
carletex: Let's add this as ToDo for another PR :D

carletex:
- Changed some more sizes
- Remove the pager if no pagination needed
- Tweak base-200 color 

There are some (little) issues still, but let's merge!

Great job here, this is a really cool feature.

Issue #216, opened 2023-02-27
useScaffoldPerpareContractWrite & useScaffoldContractWrite


technophile-04:
### Description 
I propose we should have two options for developers : 
1. **`useScaffoldPerpareContractWrite`** : An utility writes hook which uses wagmi's `usePrepareContractWrite` and `useContractWrite` both internally and returns the value, returned from both hook. 
2. **`useScaffoldContractWrite`** : An utility write hook that uses `useContractWrite` without `usePrepareContractWrite`. 

### Reason 
Although wagmi suggests  `usePrepareContractWrite`  should be always used with `useContractWrite` but considering se-2 will be used by beginner, and also to do quick scaffolding....It would be great if we have `useScaffoldContractWrite` which don't use `usePrepareContractWrite` because if you use `usePrepareContractWrite` then you have to handle `error's` properly returned by the hook and also display them, if not we may get misleading error / not so usefull error also hard to figure out errors. 

checkout #215 

That said errors are also handled pretty well by `useScaffoldContractWrite` because we are wrapping `writeAsync` inside our `useTransactor` hook. Which shows good error without developer needing to handle and display them in the UI. 

I 100% understand the necessity of `usePrepareContractWrite`  and all the things it does internally, but it might be PITA for the person doing quick scaffolding and he needs to handle all the errors on the input level properly to give good UX. 

### Example : 
1. useScaffoldPreparedContractWrite.ts
```js 
export const useScaffoldPreparedContractWrite = (contractName: string, functionName: string, args?: any[], value?: string) => {
   ...
  const writeTx = useTransactor();

  const { config, ...restPrepareValues } = usePrepareContractWrite({
    chainId: configuredChain.id,
    address: deployedContractData?.address,
    abi: deployedContractData?.abi,
    args,
    functionName,
    overrides: {
      value: value ? utils.parseEther(value) : undefined,
    },
  });

  const wagmiContractWrite = useContractWrite(config);

  const sendContractWriteTx = async () => {
   ....
  };

  return {
    ...wagmiContractWrite,
    ...restPrepareValues, 
    writeAsync: sendContractWriteTx,
  };
};
```

2. useScaffoldContractWrite.ts 
```js
export const useScaffoldContractWrite = (contractName: string, functionName: string, args?: any[], value?: string) => {
   ...  
  const writeTx = useTransactor();

  const wagmiContractWrite = useContractWrite({
    mode: "recklesslyUnprepared",
[...]

PR #352, opened 2023-05-24
Migration to wagmi v1 and viem

### LINKED ISSUE #229: Migration from ethers to viem (and wagmi 1.*)
What do you think about migrating to [Viem](https://viem.sh/), a new alternative to ethers.js and web3.js from the creators of wagmi?

[Why Viem](https://viem.sh/docs/introduction.html)

technophile-04:
## Plan of action : 
- [x] Migrate all the easy files ( File not using ethers / not deeply dependent on wagmi )

- [x] Migrate main file ( Deeply dependent on ethers and wagmi )
    - **Work in Progress**
      - [X] useTransactor  (@technophile-04 )
      - [x] useScaffoldContractWrite (@sverps)
      - [X] ContractUI (@sverps)
      - [X] useScaffoldEventHistory (@sverps)
      - [x] useBurnerWallet & BunerConnector inline with new Connector like `getWalletClient` instead of `getSigner` etc (@technophile-04)
      - [x] useScaffoldEventSubscriber (@sverps) 
      - [X] Fix `decodeTxData.ts` (@technophile-04 )
      - [X] blockexplorer all routes (@technophile-04)
      - [x] fetchPriceFromUniswap.ts
     
- [x] Lastly remove `ethers` as dependencies from `nextjs` workspace  
   - Some places like `@uniswap/sdk` still rely on `ethers` provider 
   
## Bugs Iteration : 
**Work In progress** : 

 -  [x]  `useContractWrite` not working with burner without prepare hook check -> https://github.com/scaffold-eth/scaffold-eth-2/pull/352#issuecomment-1584789003
  - [x]   The code was always shown on blockexplorer/address/${address} fixed at [8f07140](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/8f07140eeef7043793a4f7e66be8d1df88607e9e)
  - [x]  Fix `AddressInput` while using `useState("")`
  - [X]  Grabbing funds from the faucet breaks the UI checkout https://github.com/scaffold-eth/scaffold-eth-2/pull/352#issuecomment-1598209037 for more details ( @technophile-04, @sverps  )
  - [x]  Decimals break the UI while sending txn in `debug` page checkout https://github.com/scaffold-eth/scaffold-eth-2/pull/352#issuecomment-1598209037 for more desc (@sverps )
 
  
## Notes :
The [viem migration guide](https://viem.sh/docs/ethers-migration.html) and [wagmi migration guide](https://wagmi.sh/react/migration-guide#1xx-breaking-changes) are pretty robust and nice 

But here is the main difference between `ethers` and `viem` :
- In `viem` instead of `providers / signers` we have `clients`, so there are total 3 types of client
  - [Public client](https://viem.sh/docs/clients/public.html)
  - [Wallet client](https://viem.sh/docs/clients/wallet.html)
  - [Test client](https://viem.sh/docs/clients/test.html)
- Each client can be used to call different rpc methods( `viem` likes to call them `actions` )
- Example Public client can be used to call [Public actions](https://viem.sh/docs/actions/public/introduction.html#introduction-to-public-actions) ( rpc calls ) like `eth_blockNumber` etc 
- Example Wallet client can be used to call [Wallet actions](https://viem.sh/docs/actions/wallet/introduction.html#introduction-to-wallet-actions) ( rpc calls ) like `eth_sendTransaction` etc (There is no concept of `Signer` instead it calls the [`Accounts`](https://viem.sh/docs/ethers-migration.html#viem-10))

## Additional Information

## Related Issues

Fixes #229 

Your ENS/address: shivbhonde.eth , solidixen.eth


packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
```diff
@@ -14,15 +15,17 @@ export const useDeployedContractInfo = <TContractName extends ContractName>(cont
...
 
   useEffect(() => {
     const checkContractDeployment = async () => {
       if (!deployedContract) {
         setStatus(ContractCodeStatus.NOT_FOUND);
         return;
       }
-      const code = await provider.getCode((deployedContract as Contract<TContractName>).address);
+      const code = await publicClient.getBytecode({
+        address: deployedContract.address as Address,
```
technophile-04, line 27: Had to assert the type as `Address`  type imported from `viem` , It seems that due to `abitype` version mismatch in wagmi and viem we get the below error(even after configuring `AddressType`) : 
Screenshot 2023-05-25 at 2 43 44 AM

sverps: Isn't wagmi using viem? How come there's a mismatch in abitype?

technophile-04: > Install wagmi and its viem peer dependency.

Actually [Wagmi](https://wagmi.sh/react/getting-started#installation) explicitly mentions installing `viem` as peer dependency :( which causes the mismatch because `viem`  uses diff version of `abitype` checkout -> https://github.com/wagmi-dev/wagmi/pull/2421 for more details

sverps: Ah I understand, hope they can fix it by merging your PR 👍

technophile-04:
> I recall a discussion about the recklesslyUnprepared mode and not using usePrepareContractWrite, does anyone remember what was up with that?

#215 has the reason why it was removed, but were planning to add it again in #269 due to https://github.com/scaffold-eth/scaffold-eth-2/issues/216#issuecomment-1450295130

---

But in wagmi v1 its not mandatory also : 
Screenshot 2023-06-03 at 5 45 05 AM


Also, they removed have removed the problem mentioned in https://github.com/scaffold-eth/scaffold-eth-2/issues/216#issuecomment-1450295130 from https://wagmi.sh/react/prepare-hooks  page 

So I think we won't be needing #269 since the issue can be resolved without using `Prepare Hooks` in v1  🙌

sverps:
Ok, so, this branch is now finally in a workable state (except for maybe the transaction pages of the blockexplorer, that needs some more love :sweat_smile:, I just wanted to get the basics working again). We'll definitely need to iron out a lot of bugs. So we can slowly start playing with some contracts and our hooks to try and see what needs further fixing.

packages/nextjs/components/scaffold-eth/Contract/DisplayVariable.tsx
sverps, line 1: Nice, much cleaner approach! :+1:

technophile-04:
Tysm Samuel 🙌

I tried refactoring the Contract components again tried keeping it minimal and fixed most of the types issues utilizing `AbitType`, now all the Reads part of the Contract page works nicely 🙌

---
**BUG** : 
If we use `useContractWrite` without `prepare hooks` the transaction with burner wallet doesn't go through(works with other wallet), I tried spending a lot of time debugging it but no luck :( 

In below video for some reason `args` property `chainId` is undefined when you use `useContractWrite` without prepare hooks, which is then passed below and while converting it to `hex`and reading its `length` it throws error `cannot read length of undefined`

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/a1f53053-15ad-4e57-94f4-0c6a35ffc6ee

Because of this we had to use `usePrepare` hook but I think we should refrain from using it mainly because of #59 and other issues mentioned in https://github.com/scaffold-eth/scaffold-eth-2/pull/352#issuecomment-1574482292, we may use `enabled : false` to fix #59 but we may also need to apply some hack to make it work properly 

Potential bug might be also with `BunerWallet` config but please feel free to chime in if something clicks up 🙌, I will try to debug more and maybe also raise an issue on Viem's / Wagmi's github


---

**TLDR;**
- Lets migrate from `prepare hooks` to without `prepare` due to #59, #215 etc
- I also tried upgrading to [`viem v1`](https://viem.sh/docs/migration-guide.html#_1-x-x-breaking-changes) but seems like https://github.com/scaffold-eth/scaffold-eth-2/pull/352#discussion_r1204768900 has again appeared so lets wait until it solved :(

Also updated tracking list in https://github.com/scaffold-eth/scaffold-eth-2/pull/352#issue-1724746205

sverps:
@technophile-04 I'm really enjoying tackling this migration together :heart: 

I managed to upgrade `viem` to `1.0.2`. From what I saw, the remaining issue was in the `GenericContractsDeclaration` type on our end in `contracts.ts`.

I had a look at the `useContractWrite` without prepare issue and I saw no immediate fix for it. From the error it throws, I think the issue might be in `wagmi`. Needs some further investigation. The config we provide should work, from my understanding.

sverps:
I opened an issue on the `wagmi` lib concerning the failing contract write hook: https://github.com/wagmi-dev/wagmi/issues/2526

technophile-04:
> @technophile-04 I'm really enjoying tackling this migration together ❤️

++ 💯

> I managed to upgrade viem to 1.0.2. From what I saw, the remaining issue was in the GenericContractsDeclaration type on our end in contracts.ts.

Tysm 🙌, Actually for some reason `AddressType` from `viem` is resolving -> `0x{string}` and `AdderssType` from `wagmi`  -> `string` , so I had to assert at 2-3 place similar to [this comment](https://github.com/scaffold-eth/scaffold-eth-2/pull/352#discussion_r1204768900) but it might be an issue with my `eslint` server will have a look at it 🙌

> I had a look at the useContractWrite without prepare issue and I saw no immediate fix for it. From the error it throws, I think the issue might be in wagmi. Needs some further investigation. The config we provide should work, from my understanding.

> I opened an issue on the wagmi lib concerning the failing contract write hook: https://github.com/wagmi-dev/wagmi/issues/2526

Tysm for this !!🙏🙏

---

**Update**
Completed blockexplorer all routes

I think the whole app is now almost in running state with v1 wagmi and viem. Will start testing it out, clean up along the way and append bugs in https://github.com/scaffold-eth/scaffold-eth-2/pull/352#issue-1724746205 🙌

technophile-04:
### Update : 

- Got the `fetchPrice` working 

- Fixed all typescript errors now no more `git commit --no-verify` :D, there are very few places where we have any asseretion and @ts-expect-error

- Fixed merge conflicts  and did some minor clean up 

- Also fixed gh-actions-fails at [2eedb34](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/2eedb344298e51422666f6c44f669080f7b556b5) (lol spent almost 3hours debuging such small thing :( because of typescript encrypted error message )

<details>
  <summary>Removed `ethers` from `nextjs` workspace dependency, We are using `@ethersproject/provider` for BunerWallet and to be compliant with uniswapV3 sdk.</summary>  
&nbsp;

- Had to use [`EIP-1193`](https://eips.ethereum.org/EIPS/eip-1193) compliant provider for Connector in `BunerConnector` but seems like viem / wagmi does not expose it, We can write our own JsonRpcProvider as mentioned here -> [Custom transport](https://viem.sh/docs/clients/transports/custom.html) but I think its ok to rely on prebuilt providers. I have also tried using [`http`](https://viem.sh/docs/clients/transports/http.html) , using [`createPubliClient`](https://viem.sh/docs/clients/public.html#import) and none worked and it makes sense. I also tried finding other libraries which give [`EIP-1193`](https://eips.ethereum.org/EIPS/eip-1193) compliant providers but `@etheresproject/provider`  seems easy to use and also works nicely. 
 &nbsp;
- `uniswapV3/sdk` also requires passing in the provider and it internally also uses `@ethersproject/providers` to set types too  

</details>

--- 

### Next steps : 
- Start testing the functionality thoroughly
- Reviewing the code thoroughly for any betterments(we also don't want to break anything) 
- Find bugs

technophile-04:
### Update : 
**Fixed some bugs** : 

1. The `code` was always shown on `blockexplorer/address/${address}` fixed at [8f07140](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/8f07140eeef7043793a4f7e66be8d1df88607e9e) 

2. I think we collocated the types of `*Inputs` components, While testing I noticed that if you do 
```javascript
  const [address, setAddress] = useState("");
  <AddressInput value={address} onChange={newVal => setAddress(newVal)} />
 // Error: cannot assign type `undefined`  to type `string`
```

I tried fixing this at [48dec0b](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/48dec0bfd83701721def0395fdf73c86af8039c5) by making `onChange : (newValue : string) => void` previously it was `(newValue : T)` , I tried options but this seemed minimal without breaking others @sverps could you please verify [48dec0b](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/48dec0bfd83701721def0395fdf73c86af8039c5)? 

**Bugs yet to be fixed** : 

1.  When you send transaction on `debug` page with `decimal` as value the UI breaks but it should not break just show the error notification (This might be the issue on wagmi side since we have handled the error) : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/1d9c590a-5880-45ad-a376-41fbacc3f745

 2. Grabbing funds from the faucet breaks the UI : 
 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/0cf27ae4-a56a-4980-aede-6d0d7806b54f

sverps:
> 1. When you send transaction on `debug` page with `decimal` as value the UI breaks but it should not break just show the error notification (This might be the issue on wagmi side since we have handled the error) :

The issue was on our side. By not awaiting the result of `writeAsync` where we called it, the promise error was treated as uncaught. Fixed it by passing it as a callback, then calling it inside `writeTxn` and awaiting it there. Tricky one to debug :sweat_smile:

sverps:
> I tried fixing this at [48dec0b](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/48dec0bfd83701721def0395fdf73c86af8039c5) by making `onChange : (newValue : string) => void` previously it was `(newValue : T)` , I tried options but this seemed minimal without breaking others @sverps could you please verify [48dec0b](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/48dec0bfd83701721def0395fdf73c86af8039c5)?

I fixed it in a slightly different way. I think keeping the generic makes more sense, as it allows the Integer inputs to keep their value as a BigInt when it is one (rather than converting it to string). Feels more natural like that and then a user won't have to re-convert it to BigInt if he needs to use it as one.

sverps:
> 2\. Grabbing funds from the faucet breaks the UI :

Fixed it. Turns out `newBlock` only got tx hashes, so needed to tell the block watcher to also provide full transaction data.

technophile-04:
> The issue was on our side. By not awaiting the result of writeAsync where we called it, the promise error was treated as uncaught. Fixed it by passing it as a callback, then calling it inside writeTxn and awaiting it there. Tricky one to debug 😅

Tysm Samuel 🙌 !!! Lol even I spent yesterday a lot of time debugging it  was late night so thought I would update it in the morning but you already did it !!! Really sorry for wasting time I should have mentioned it yesterday itself 😅 won't happen next time! 

Btw I have created a test repo ->  https://github.com/technophile-04/wagmi-debug it has plain wagmi v1 nd rainbowkit  latest versions in NextJs along with `YourContract.sol` `abi` ...deployed to `Sepolia` so next time we are unsure if wagmi / viem has internal problem we can try it out in that plain repo and see 🙌 


> I fixed it in a slightly different way. I think keeping the generic makes more sense, as it allows the Integer inputs to keep their value as a BigInt when it is one (rather than converting it to string). Feels more natural like that and then a user won't have to re-convert it to BigInt if he needs to use it as one.


Great fix!! makes sense I will test other too 🙌

technophile-04:
#### Update : 
1. update parsing of error and small formatting tweak at [fcb9703](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/fcb9703299dd56b330713e006723fb1d73e621d5) : 

  Before V1     | After        
:------------:|:---------------:
 |  
|  |   |  



2. Got back the similar behaviour to `main` when values are too big they are shown as `Ξ 2.1` etc [cf63271](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/cf63271c1a8971a66e68d11a056e39b4a8c54e0d)

3.  There was a bug when you send small value like `1 wei` its was used to send `49 wei`  fixed at [d76328c](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/d76328cb84033c11e57657c8943294c4a1f78f1f)

4. Also fixed some todos

### BUG's Noticed : 

1. #391 

2. Handling passing of array / structs of inputs : 

We have this logic in `main`
```ts
      if (["array", "tuple"].includes(baseTypeOfArg)) 
```

But it seems now viem / wagmi does like to parse it as : 
Screenshot 2023-06-30 at 12 25 26 AM

We can maybe check if there includes `[` then its an array 🤔 need to think about it, will have a look at it tomorrow

Here is the example contract to test it out -> https://github.com/scaffold-eth/scaffold-eth-2/pull/156/commits/1f22545814be26d81897a42fc99c7e30863d4ab2 (look at previous changes)

packages/nextjs/components/scaffold-eth/Contract/utilsDisplay.tsx
```diff
@@ -22,8 +22,19 @@ export const displayTxResult = (
...
+      console.log("Number.MAX_SAFE_INTEGER", displayContent.toString());
+      const asNumber = Number(displayContent);
+      if (asNumber <= Number.MAX_SAFE_INTEGER && asNumber >= Number.MIN_SAFE_INTEGER) {
+        return asNumber;
+      } else {
+        return "Ξ" + formatEther(displayContent);
+      }
+    } catch (e) {
+      return "Ξ" + formatEther(displayContent);
+    }
```
sverps: I don't understand this change. A `bigint` display value can represent anything, not only an ether value. So I don't think we should do anything besides turning it into a string here.

technophile-04: Actually, this was in `main` and also in SE-1 so I added it : 
https://github.com/scaffold-eth/scaffold-eth-2/blob/123e6d84ac613e15e07fd854268b6869d7353615/packages/nextjs/components/scaffold-eth/Contract/utilsDisplay.tsx#L21

I think its there to nicely format stuff like `balanceOf` although `Ξ` is kind of misleading and every token might don't have `18` decimal places but it does its jobs pretty well.....since if the number is too large it's most likely representing some kind of decimal or balance 

But spawning @carletex here since I am also not too sure whether we should remove or keep it

sverps:
> @sverps could you please check for [1f46ed9](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/1f46ed9297514cf71a9ac7e5ff8004862d68c22f) .....It tries to remove the need for assertion since it was giving errors when a contract is not present
> 
> Lol it's bit hacky but please feel free to update it 🙌

Cleaned it up a bit. I also found a very nice utility type `Prettify`. Will go through the types to see where we could use it to provide nicer types for devs. (Check the resulting type also after removing the `Prettify<...>` wrapper if you want to see what it does)

technophile-04:
>  also found a very nice utility type Prettify. Will go through the types to see where we could use it to provide nicer types for devs. 

Omg love this, great find !! Such a great utility always had a hard time seeing just the Abstract type name instead of the resultant type, Yes we should it update wherever possible

packages/nextjs/components/example-ui/ContractData.tsx
```diff
@@ -33,8 +33,11 @@ export const ContractData = () => {
   useScaffoldEventSubscriber({
     contractName: "YourContract",
     eventName: "GreetingChange",
-    listener: (greetingSetter, newGreeting, premium, value) => {
-      console.log(greetingSetter, newGreeting, premium, value);
+    listener: logs => {
+      logs.map(log => {
+        const { greetingSetter, value, premium, newGreeting } = log.args;
+        console.log("📡 GreetingChange event", greetingSetter, value, premium, newGreeting);
+      });
```
technophile-04, line 40: @sverps any reason why we are not doing : 

```typescript
      const { args } = logs[0];
      const { greetingSetter, value, premium, newGreeting } = args;
      console.log("📡 GreetingChange event", greetingSetter, value, premium, newGreeting);

```
It was also suggested in [wagmi migration guide](https://wagmi.sh/react/migration-guide#usecontractevent)


I saw we changed it here : 
https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/439c4f2d50d94429e61a6d8b1e03fe58c79335a5#diff-4b13aae8a79eed1d29876f892dd1b17adeace865ad0ea5deb4baf53b00e479b1

technophile-04:
Fixed some bugs and clean up from [9ab3636](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/commits/9ab363624f4b021c953a0099bf9fd39fa3b688ef)... Also tested all the things nicely and everything seems to work as `main`

Marking this `Ready for review`  since everything looks cool now 🙌

We have just 2 todo's (can be conqured async with testing) : 
1. #391 
2. Using `Prettify` for better types hint as mentioned in https://github.com/scaffold-eth/scaffold-eth-2/pull/352#issuecomment-1614424272

packages/nextjs/components/example-ui/ContractData.tsx
sverps: The listener returns an array of all logs that occurred during the current [polling interval](https://viem.sh/docs/contract/watchContractEvent.html#pollinginterval-optional). So depending on how the client is configured, this could be multiple logs. To not miss logs in any scenario, best to just loop over all logs that are returned, rather than only take the first one in the array.

carletex:
We think the main stuff looks great (maybe some bugs will come up), so let's merge!!!

Really nice work @technophile-04 & @sverps

carletex (approved):

Issue #318, opened 2023-04-21
Foundry package


carletex:
We are working on a CLI tool (#251) that will allow SE-2 users to choose between hardhat / foundry when setting up their project.

The CLI is not ready yet, but we want to start building the foundry package (`packages/foundry`), setting the libs/config that you need to do the same things that we do with the hardhat package: 

- Deploy a smart contract
- Run a localhost node/rpc where you can deploy contracts and where we can point from our front end. 
- Have a sample contract (same) + a test
- Allow configuring networks to deploy to (foundry config file?)
- Generate the "deployedContracts.ts" file in the nextjs folder.
- Generate a burner/deployer account

Ideally, we want to have the same  root "CLI commands API" that we use for hardhat:

- `yarn chain`: start the local node / rpc
- `yarn deploy --network sepolia`: deploys your contract to the target network.
- `yarn generate`: generate a random account
- `yarn account`: show the QR for the generated account and the balances in all the configured networks

(maybe last two could be shared and don't need a new implementation for Foundry.)

supernovahs:
I would like to contribute

luloxi:
I want to collaborate in this change! 

PS: I don't see any good reason to do another implementation of "yarn generate" and "yarn account". Mnemonics are mnemonics

technophile-04:
For the deployments part I think we should keep an eye on [forge-deploy](https://github.com/wighawag/forge-deploy) it is created by the same developer who made [hardhat-deploy](https://github.com/wighawag/hardhat-deploy)(which is currently being used for deployments / managing deployments with abi in our Hardhat-based SE-2 also our [custom hooks](https://github.com/scaffold-eth/scaffold-eth-2#interacting-with-your-smart-contracts-se-2-custom-hooks) relies on the structure which it creates after deployments ). [forge-deploy](https://github.com/wighawag/forge-deploy) is trying to keep things similar to hardhat-deploy so might help us. 

Nevertheless, don't want to get ahead of ourselves also [forge-deploy](https://github.com/wighawag/forge-deploy) is **under active development** so might be a bit unstable. Just wanted to mention about it here 🙌

luloxi:
Just found this repo, might be useful for connecting Foundry with the other parts of SE2

[forge-exec](https://github.com/wighawag/forge-exec)

Btw, is there a branch of SE2 with foundry that I can try out, so I can see how the current implmentation is working?

jrcarlos2000:
[...]

PR #353, opened 2023-05-26
Foundry package

jrcarlos2000:
## Description

This PR adds the `foundry` package to SE2. 

Added `Deploy.s.sol` which mimics the functionality of ``hardhat/deploy/00_deploy_your_contract.ts``
Added helper function `DeployHelpers.s.sol` which: 
- reads a localhost account for deployment
- reads impersonates a {network} account using a provided PRIVATE_KEY
- saves a json of {"contractAddr" : "Name"}, this is meant to be used when deploying different contract name from the same factory
Added  helper `generateTsAbis.ts` which generates `deployedContracts.ts` in nextjs folder

## Additional Information

## Related Issues

_Closes #318


packages/foundry/foundry.toml
```diff
@@ -0,0 +1,9 @@
+[profile.default]
+src = 'src'
+out = 'out'
+libs = ['lib']
+fs_permissions = [{ access = "read-write", path = "./"}]
+[rpc_endpoints]
+localhost = "http://127.0.0.1:8545"
+mumbai = "https://rpc.ankr.com/polygon_mumbai"
```
carletex, line 8: Let's add some more endpoints (mainnet, polygon, optimism, sepolia, goerli...) Check `hardhat.config.ts`

packages/foundry/script/generateTsAbis.ts
```diff
@@ -0,0 +1,110 @@
+const fs = require("fs");
+const path = require("path");
+//@ts-expect-error  This script runs after `hardhat deploy --export` therefore its deterministic that it will present
```
carletex, line 3: Not hardhat

packages/foundry/src/YourContract.sol
```diff
@@ -0,0 +1,85 @@
...
+        _;
+    }
+
+    /**
+     * Function that allows anyone to change the state variable "greeting" of the contract and increase the counters
+     *
+     * @param _newGreeting (string memory) - new greeting to save on the contract
+     */
+    function setGreeting(string memory _newGreeting) public payable {
+        // Print data to the hardhat chain console. Remove when deploying to a live network.
```
carletex, line 51: hardhat => anvil?

packages/foundry/test/Counter.t.sol
carletex, line 1: This files shouldn't be called "Counter"

carletex: And also maybe write the real test? Like the one we have in hardhat

carletex (commented):
Hey @jrcarlos2000 

Thanks for this! Didn't test it yet, but did an initial review. Check the review comments.

The idea is to merge this into a `foundry` branch, until the CLI is ready (so people will be able to choose). We'll link it in the main README. So another change: let's remove the "hardhat" package here.

Thanks!

package.json
```diff
@@ -26,13 +26,22 @@
     "postinstall": "husky install",
     "precommit": "lint-staged",
     "vercel": "yarn workspace @se-2/nextjs vercel",
-    "vercel:yolo": "yarn workspace @se-2/nextjs vercel:yolo"
+    "vercel:yolo": "yarn workspace @se-2/nextjs vercel:yolo",
+    "chain:foundry": "cd packages/foundry && anvil --config-out localhost.json",
+    "deploy:foundry": "cd packages/foundry && forge build && forge script script/Deploy.s.sol --rpc-url ${0:-localhost} --broadcast && node script/generateTsAbis.ts",
+    "generate:foundry": "cd packages/foundry && node script/generateAccount.ts"
```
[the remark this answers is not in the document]
jrcarlos2000: do we want to migrate all the available commands in hardhat ? @carletex

packages/foundry/foundry.toml
```diff
@@ -0,0 +1,21 @@
...
+
+mainnet = "https://eth-mainnet.alchemyapi.io/v2/${providerApiKey}"
+sepolia = "https://eth-sepolia.g.alchemy.com/v2/${providerApiKey}"
+goerli = "https://eth-goerli.alchemyapi.io/v2/${providerApiKey}"
+arbitrum = "https://arb-mainnet.g.alchemy.com/v2/${providerApiKey}"
+arbitrumGoerli = "https://arb-goerli.g.alchemy.com/v2/${providerApiKey}"
+optimism = "https://opt-mainnet.g.alchemy.com/v2/${providerApiKey}"
+optimismGoerli = "https://opt-goerli.g.alchemy.com/v2/${providerApiKey}"
+polygon = "https://polygon-mainnet.g.alchemy.com/v2/${providerApiKey}"
+polygonMumbai = "https://polygon-mumbai.g.alchemy.com/v2/${providerApiKey}"
```
carletex, line 19: Would the ${providerApiKey} work here? How do I set up the .env var?

packages/foundry/script/Deploy.s.sol
```diff
@@ -0,0 +1,19 @@
+pragma solidity ^0.8.19;
+
+import "../src/YourContract.sol";
+import "./DeployHelpers.s.sol";
+
+contract DeployScript is ScaffoldETHDeploy {
+    // mapping(address => string) contractNames;
```
carletex, line 7: remove this?

packages/nextjs/generated/deployedContracts.ts
carletex, line 1: Don't need to delete this file, just leave it "empty". Check how it should be in the `main` branch

package.json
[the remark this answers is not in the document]
carletex: Ideally yes. But we add a ToDO and be in a different PR. `test` should be easy to add right?

carletex (changes requested):

package.json
```diff
@@ -5,17 +5,18 @@
   "workspaces": {
     "packages": [
       "packages/hardhat",
```
carletex, line 7: Did you see the comment here about hardhat?

https://github.com/scaffold-eth/scaffold-eth-2/pull/353#pullrequestreview-1464498252

(I think you also missed some comments from the last review)

packages/foundry/foundry.toml
jrcarlos2000: it wont work here, but a .toml file is not usually integrated with .env files. for now keeping it this way is the best. it wont break, you will need to replace your api key in this file.

package.json
jrcarlos2000: thats for `scripts`, shall i remove hardhat completely as a package ? @carletex

carletex: From https://github.com/scaffold-eth/scaffold-eth-2/pull/353#pullrequestreview-1464498252

> So another change: let's remove the "hardhat" package here.

So, yes, let's remove hardhat entirely here. This would be what people get after the choose "Foundry" in the CLI.

technophile-04:
Hey, @jrcarlos2000 first of all tysm for this, really appreciate it !!

I went through the flow and everything works !! The `deployedContracts.ts` in `nextjs` are also getting populated nicely, I even tried deploying multiple contracts on multiple chain and it is populated correctly 🙌

### TODO's I noticed : 
- [ ] Fix the small warning in deploy scripts and others like adding SPDX license and removing unwanted variables 
&nbsp;
- [ ] For some reason `yarn deploy:verify` fails to verify the contract check, After configuring `.env` I tried deploying to `sepolia` and the tried running `yarn deploy:verify` it's failing : 
Screenshot 2023-06-17 at 10 40 42 PM
(I also tried adding `sepolia` key under `[etherscan]` in `foundry.toml` but still facing the same issue)
&nbsp;
- [ ] We should probably setup `typescript` or convert `generateAccount.ts`, `genearteTsAbis.ts` -> `.js` cc @carletex for this
&nbsp;
- [ ] Maybe mention in `README.md` to the first setup `.env` if the answere to the below **Question** is `no`. Maybe we mention a small copy-paste shell script in README.md ? which people can run and it puts our `ALCHEMY_KEY` in their `.env` something like : 
Something similar was mentioned in [`forge-deploy`](https://github.com/wighawag/forge-deploy) for populating `foundry.toml`   
    ```shell
     (echo "DEPLOYER_PRIVATE_KEY=";  echo "ALCHEMY_API_KEY=oKxs-03sij-U_N0iOlrSsZFr29-IqbuF"; echo 
     "ETHERSCAN_KEY=DNXJA8RX2Q3VZ4URQIWP7Z68CJXQZSC6AW") >> packages/foundry/.env
    ```

----

**Question:** Is there a way were could use a default value if `env's` are not present? similar to [this](https://github.com/scaffold-eth/scaffold-eth-2/blob/9faa67540a264accf71cf03bd8dd89b06b6a4c36/packages/hardhat/hardhat.config.ts#L11)  where we default our own API key? Because if you don't configure `.env` initially I ran into the following problems : 

1. The first thing while running the repo is you need to set up `.env` file with at least `ALCHEMY_API_KEY` & `ETHERSCAN_API_KEY`. Reason is when I cloned the repo -> `yarn chain` -> `yarn deploy` I got this : Screenshot 2023-06-17 at 6 02 12 PM
(If I remove `polygonMumbai` from `foundry.toml` under `etherscan` it worked without configuring `.env` )
&nbsp;

---

tysm again @jrcarlos2000 🙌

jrcarlos2000:
I have addresses TODO's 1,2,4. cc @technophile-04 

- 1 ) added SPDX license to missing files
- 2 ) failed verification was also happening in my end, but it was resolved after doing ```foundryup``` 
- 3 ) waiting for comments on this @carletex
- 4 ) no, so i wrote it in the README

README.md
```diff
@@ -39,47 +39,60 @@ To get started with Scaffold-ETH 2, follow the steps below:
...
 cd scaffold-eth-2
 yarn install
+foundryup
 ```
 
-2. Run a local network in the first terminal:
+2. Create your .env file
+
+```
+(echo "DEPLOYER_PRIVATE_KEY=";  echo "ALCHEMY_API_KEY=oKxs-03sij-U_N0iOlrSsZFr29-IqbuF"; echo "ETHERSCAN_KEY=DNXJA8RX2Q3VZ4URQIWP7Z68CJXQZSC6AW") >> packages/foundry/.env
```
technophile-04, line 48: It should be -> `ETHERSCAN_API_KEY`  since that's what we have configured in `foundry.toml`

technophile-04:
Tysm @jrcarlos2000 🙌

Just some few changes  : 

1. Let's update `scaffold.config.ts` ->  `targetNetwork : chains.foundry`
2. Let's print the `address` of generated account when you do `yarn account` ? Because people will  know where to send funds. (In Future PR : print QR, we may go back to using .ts/js file but lets not worry too much about it )
3. Because of this https://github.com/scaffold-eth/scaffold-eth-2/blob/9faa67540a264accf71cf03bd8dd89b06b6a4c36/packages/nextjs/pages/blockexplorer/address/%5Baddress%5D.tsx#L160 the UI breaks at `/blockexplorer/address/${address}` ....Is there a way to access `assembly` and  `bytecode` of the  contract ? `hardhat` was use to store it at `hardhat/artifacts/build-info` 

    I mean if its get too complicated lets just comment on that part and maybe handle it in a different PR, we can comment `code` and `storage` tabs from `pages/[address].tsx` 
    
 ---- 
 
 > 2 ) failed verification was also happening in my end, but it was resolved after doing foundryup
 
 Yeah this worked for me !! But I think I am missing something `foundryup` install / updates all the required tools for development...So was there some problem with previous version of tools which caused `verification error` or do you have to do it always after cloning the repo  (No problem with this I mean we are just telling people to update to latest version of tools by doing `foundryup`) was just curious about it 
 
 Tysm @jrcarlos2000 again I think we are almost there !!

jrcarlos2000:
Resolved (1) , (2), and (3). 

About the verification, I will do some research on it. First time I saw it. @technophile-04

technophile-04:
Tysm @jrcarlos2000 !!!!, I think this looks great as the initial iteration 🚀🚀

---

#### TODO's for future PR's : 

- [ ] The `blockexplorer/address/${address}` `code` does not shows anything 
Screenshot 2023-06-26 at 2 44 41 PM

- [ ] We need to figure out better way for generating `.env` because I think it the current solution will have problem in `Windows` since we are using the forward slash (Maybe having a separate script to generate `.env` ) 

- [ ] Update the README.md properly (mainly deploying contracts to test / main networks)

- [ ] Maybe covert ListAccount.s.sol to .ts / .js file since `ListAccount.s.sol` takes time and throws the output at once.... we can show QR code when using `.js / .ts`.....(We should also remove `ethers` from dependency and use `viem` 🙂)

- [ ] Setup typescript, eslint / prettier for code formating of `.js /.ts` files

- [ ] Optional : Check for [forge-deploy](https://github.com/wighawag/forge-deploy) if it have features like `hardhat-deploy` eg. Not deploy a new contract if no changes have happen instead reuse prev one maybe it can also increase devX for deploying and handling deployments

PR #354, opened 2023-05-28
New commands to deploy on smart contract change

ByteAtATime:
## Description

Add the following commands, both of which watch `packages/hardhat/contracts` and `packages/hardhat/deploy` for changes:
- `deploy:watch` - on change, run `yarn deploy`
- `deploy:watch-reset` - on change, run `yarn deploy --reset`

Watching files is done through the [`watch`](https://npmjs.com/package/watch) package.

## Additional Information

## Related Issues

Minor QoL change, didn't file an issue

Your ENS/address: `0x3848457d61094F13Cbd6Fc4cf8D6383C48506232`

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,8 @@
     "chain": "yarn workspace @se-2/hardhat chain",
     "fork": "yarn workspace @se-2/hardhat fork",
     "deploy": "yarn workspace @se-2/hardhat deploy",
+    "deploy:watch": "yarn workspace @se-2/hardhat deploy:watch",
+    "deploy:watch-reset": "yarn workspace @se-2/hardhat deploy:watch-reset",
     "verify": "yarn workspace @se-2/hardhat verify",
     "compile": "yarn workspace @se-2/hardhat compile",
     "generate": "yarn workspace @se-2/hardhat generate",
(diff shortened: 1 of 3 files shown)


ByteAtATime:
I'm not sure what the CONTRIBUTING.md means by "tag a maintainer" so... @carletex?

carletex:
Hey @ByteAtATime 

Thanks for taking your time on this PR.

[Hardhat deploy](https://github.com/wighawag/hardhat-deploy) already have both flags, so ideally we don't want to bloat the repo with extra commands (and new libs) that we don't really need.

`yarn deploy --reset` works, but I noticed that `yarn deploy --watch` works (detects changes in contracts/ and deploy/)... BUT doesn't run the `generateTsAbis` script. It's because the `--watch` doesn't let the deploy command exit.

We should find a fix for that!

ByteAtATime:
Oh, sorry, I didn't know it had the command, probably should've taken more time with this. I'm going to close this PR and open a new one if I find the solution.

ByteAtATime:
@carletex  I believe that we can make a separate script that uses `fs.watch` (node stdlib). However, this might not offer the best experience. Here's an example I made:

```ts
const fs = require('fs');
const { run } = require('hardhat');

async function main() {
  // Deploy your contracts using Hardhat
  const contracts = await run('deploy');

  // Generate the ABI types
  await run('generateTypes');

  // Watch for changes and redeploy if necessary
  if (process.argv.includes('--watch')) {
    fs.watch('./contracts', { recursive: true }, async () => {
      console.log('Changes detected. Redeploying...');
      await run('deploy');
      await run('generateTypes');
    });
  }
}

main()
  .then(() => process.exit(0))
  .catch((error) => {
    console.error(error);
    process.exit(1);
  });
```

The problem with this is that there's likely many quality of life features that hardhat's watch has implemented, and using this would lose them.

PR #356, opened 2023-05-29
Patch useLocalStorage setting the right initial value to avoid a hydration error

### LINKED ISSUE #340: bug: hydration fails due to mismatch caused by usehooks-ts bug

### Current Behavior

When using `useLocalStorage`, in certain cases, Next throws a "hydration failed" error due to server/client mismatch.

### Expected Behavior

No hydration failures

### Steps To Reproduce

_No response_

### Anything else?

This is caused by a bug in `useLocalStorage` which is addressed in an [open PR](https://github.com/juliencrn/usehooks-ts/pull/251) on their end.

It appears, however, that `usehooks-ts` is no longer maintained. Should we look into alternatives? Options are:

- Create workarounds where needed (with risk that users aren't aware of these issues and run into them themselves)
- Ditch the dependency in favor of a newer hooks lib
- Ditch the dependency in favor of our own set of utility hooks

damianmarti:
fixes #340 (if we want to fix it with a patch to usehooks-ts)


technophile-04 (approved):
Tysm Damu !! 

This might be the problem with only my machine (even previously I had some issues picking the right dependencies) : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/f7a1d3ac-38b3-4543-b9e8-3fb226f92bb9

I have created the [useLocalStorage-patchTest](https://github.com/scaffold-eth/scaffold-eth-2/tree/useLocalStorage-patchTest) branch which branches from this branch and [added basic useLocalStorage](https://github.com/scaffold-eth/scaffold-eth-2/blob/a488ae7c0480d12630d9fe17012a674c73b7dbdf/packages/nextjs/pages/index.tsx#L8) to test

**Step to reproduce** : 
1. 
```shell
git pull
git switch useLocalStorage-patchTest
yarn install
```
2. click the greeting button on Home page( this will update `isHello` localStorage state to `false` ) and then refresh the page. 

--- 

Now this might be the problem with only my machine so would love it if someone could test it 

But, the changes look good to me 🙌

sverps (approved):

rin-st:
It doesn't work for me too. But it works if I copy-paste hook to hooks directory and change `storedValue` to `initialValue`

And it shows initial value for some milliseconds

https://github.com/scaffold-eth/scaffold-eth-2/assets/25638585/3999f6b8-2c78-4c6a-aecf-adeca00ffa56

damianmarti:
> @damianmarti yes
> 
> I found solution: [yarnpkg/berry#4231 (comment)](https://github.com/yarnpkg/berry/issues/4231#issuecomment-1470124509) this helped. If I change resolutions to `"usehooks-ts@^2.7.2": "patch:usehooks-ts@npm:^2.7.2#./.yarn/patches/usehooks-ts-npm-2.7.2-fceffe0e43.patch"` it works

I'm not sure about it. Maybe @sverps @carletex or @technophile-04 have something to say...

sverps:
I had the same issue as @rin-st , seems like the patch generated a faulty resolutions line (with a url-encoded `^` sign, rather than the actual one). I committed the fix from @rin-st , I think we can merge it like this :+1:

carletex:
It works! Thank you all!

PR #360, opened 2023-06-01
Solidity Optimizer options

moltam89:
## Description

Hey guys,

This PR is about the Solidity compiler's Optimizer.

It is enabled by default in the original scaffold-eth [repo](https://github.com/scaffold-eth/scaffold-eth/blob/fc2ae9213be121e9e813d0c6fcb02cbc21d3593f/packages/hardhat/hardhat.config.js#L283). 
I wasn't sure if I should just enable it for scaffold-eth-2 as well, or not. I decided to only add it as an option and update the readme, to let people know about this, before they deploy to a Live network.

Let me know what do you think.

Best,
Tamas

## Additional Information

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,12 +98,7 @@ The deployer account is the account that will deploy your contracts. Additionall
 
 You can generate a random account / private key with `yarn generate` or add the private key of your crypto wallet. `yarn generate` will create a random account and add the DEPLOYER_PRIVATE_KEY to the .env file. You can check the generated account with `yarn account`.
 
-3. Enable the Solidity Optimizer
-
-By default, the Optimizer is disabled. You can enable it in `hardhat.config.ts`.
-Here are the [benefits](https://docs.soliditylang.org/en/latest/internals/optimizer.html#benefits-of-optimizing-solidity-code) of enabling the Optimizer.
-
-4. Deploy your smart contract(s)
+3. Deploy your smart contract(s)
 
 Run the command below to deploy the smart contract to the target network. Make sure to have some funds in your deployer account to pay for the transaction.
 
@@ -111,7 +106,7 @@ Run the command below to deploy the smart contract to the target network. Make s
 yarn deploy --network network_name
 ```
 
-5. Verify your smart contract
+4. Verify your smart contract
 
 You can verify your smart contract on Etherscan by running:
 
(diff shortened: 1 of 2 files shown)


README.md
technophile-04, line 1: I think its fine we don't mention it in README, don't want people to do a side quest on `optimizer` while they are getting started 😅 so let's just remove this...We already provided useful links in `hardhat.config.ts` 🙌

technophile-04 (approved):
Tysm @moltam89 🙌,  cc @carletex , @sverps , @damianmarti for final thoughts before we merge this, here is the TLDR; -> https://github.com/scaffold-eth/scaffold-eth-2/pull/360#issuecomment-1581354916

carletex:
I'm up for enabling it by default!

A nitpick:

```js
// By default, the Solidity Optimizer is enabled
// https://docs.soliditylang.org/en/latest/internals/optimizer.html#benefits-of-optimizing-solidity-code
// https://docs.soliditylang.org/en/latest/using-the-compiler.html#optimizer-options
// https://hardhat.org/hardhat-runner/docs/config
```

Isn't this too much? haha

`// By default, the Solidity Optimizer is enabled` This seems redundant (easily inferred from optimizer: { enabled: true} :D). And maybe we don't need 3 links? I feel that the important part is the "runs" parameter, so some helper text with a link could help there (move the https://docs.soliditylang.org/en/latest/using-the-compiler.html#optimizer-options link right above "runs"?)

Anyway, everything looks good, just suggesting some tiny tweaks.

Thanks!

XOder64x:
After you fucked me, you want me to come back

lun., 12 iun. 2023, 19:11 Carlos Sánchez ***@***.***> a scris:

technophile-04:
> // By default, the Solidity Optimizer is enabled This seems redundant (easily inferred from optimizer: { enabled: true} :D). And maybe we don't need 3 links? I feel that the important part is the "runs" parameter, so some helper text with a link could help there (move the https://docs.soliditylang.org/en/latest/using-the-compiler.html#optimizer-options link right above "runs"?)

Yes yes, I agree with this 

Tysm @moltam89 just pushed the changes Carlos suggested at https://github.com/scaffold-eth/scaffold-eth-2/pull/360/commits/aef0a990626f40c18088cc34776d7090158f348e 

Also tested that deployment and verification work correctly 🙌 

https://sepolia.etherscan.io/address/0x786b4Dff292876aD3aDE9CCD534373AEa1Ad8eD7#code

moltam89:
Thank you guys!

PR #368, opened 2023-06-06
feat(generateTsAbis): generate from inside deploy script

### LINKED ISSUE #367: yarn deploy not working

### Current Behavior

yarn deploy
result in: 
yarn run v1.22.1
$ yarn workspace @se-2/hardhat deploy
warning package.json: No license field
$ hardhat deploy --export-all ./temp/hardhat_contracts.json "$@" && hardhat run scripts/generateTsAbis.ts
Error HH308: Unrecognized positional argument $@

For more info go to https://hardhat.org/HH308 or run Hardhat with --show-stack-traces
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed.
Exit code: 1
Command: C:\Program Files\nodejs\node.exe
Arguments: C:\Users\bansc\Documents\GitHub\scaffold-test\.yarn\releases\yarn-1.22.1.cjs deploy
Directory: C:\Users\bansc\Documents\GitHub\scaffold-test\packages\hardhat
Output:

info Visit https://yarnpkg.com/en/docs/cli/workspace for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

### Expected Behavior

_No response_

### Steps To Reproduce

Download project from github, run 
yarn install
yarn deploy

### Anything else?

I am on windows..
Yarn version: yarn run v1.22.1
Node version:
v18.16.0

what am i missing ?

sverps:
Fixes #354

## Description

Moves the generation of the TypeScript Abis to a deploy script. This removes the need to call a separate script after `hardhat deploy` and thus makes it easier to pass on any desired param (such as a watcher, etc.)

The implemented approach was needed, since during the deploy phase, the output file is not generated yet, so we need to create the deployment info ourselves from the hardhat runtime environment.

## Additional Information

## Related Issues

Fixes #354 
Will also (at least partially) fix #367


technophile-04:
I love this approach !! 

Just one thing I noticed while quick testing is that when you deploy to another network it overrides the previous one, while on `main` branch it adds the new `chainId` as key. 

**Steps to reproduce**
1. Deploy to `sepolia`
```ts
yarn deploy --network sepolia
``` 

2. Deploy to `goerli` 
```ts
yarn deploy --network goerli
```

3. Check `deployedContract.ts`
```ts
const contracts = {
   5 : [...]
} as const 
export default contracts;
```
---

We should ideally have : 
```ts
const contracts = {
   5 : [...],
   11155111: [...]
} as const 
export default contracts;
```

--- 

My wild guess is `deployments.all()` is just giving us the current deployed contracts mentioned in `--network` (I may be wrong)

This might be helpful [exporting-deployments](https://github.com/wighawag/hardhat-deploy#exporting-deployments), don't know if there is a way to latch on this programmatically, or we might need to iterate over `hardaht/deployments` (which has all the deployments) 

I will update here if I found a better solution 🙌

sverps:
> Just one thing I noticed while quick testing is that when you deploy to another network it overrides the previous one, while on `main` branch it adds the new `chainId` as key.

Yeah, I noticed it as well and have looked a bit on how to access all this info. I didn't find a way to access the export / export-all functionality from inside a running deployment, that would be nice.

Alternatively, we can do with your suggestion and fetch this info from inside the `deployments` folder. I'll see if I can add that. :+1:

sverps:
It turns out the deployments folder gets populated on the fly with the outputs for each deployment script, so with this new generate abi script being run last, it should always get access to the current deployed contracts as well directly from that folder. No need anymore to deduce this info from `hre`. Much cleaner :smile:

packages/hardhat/deploy/99_generateTsAbis.ts
carletex, line 1: Maybe a comment at the beginning of this file?

Explaining what is this file a why?

carletex (commented):
Thanks for this!!

This will allow the use of the `--watch` flag and also Windows users won't have issues.

It seems a _lil bit_ hacky but I think it's worth it. I'm up for merging this (left a couple of review comments / suggestions)

packages/hardhat/deploy/99_generateTsAbis.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  if (!fs.existsSync(TARGET_DIR)) {
+    fs.mkdirSync(TARGET_DIR);
+  }
+  fs.writeFileSync(
+    `${TARGET_DIR}deployedContracts.ts`,
+    prettier.format(`const contracts = {${fileContent}} as const; \n\n export default contracts`, {
+      parser: "typescript",
+    }),
+  );
```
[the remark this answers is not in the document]
technophile-04: ++....Maybe this emoji 📝 ? Lol sorry for nitpicking, even the current looks good!

packages/hardhat/deploy/99_generateTsAbis.ts
sverps: Done :+1:

Issue #369, opened 2023-06-07
bug: deployedContracts file being out of sync


kevinjoshi46b:
### Current Behavior

Currently the `deployments` folder in hardhat has been gitignored leading to all the data related to deployments being stored only on local computer.

The `deployedContracts` file in nextjs is being generated based on this `deployments` folder whenever a new deployment takes place.

Consider a scenario where you deployed the smart contract on polygon from your machine which updated the `deployedContracts` file with its address, abi etc. You committed and pushed all the changes to github. Now someone else working on the same project pulled the changes and deployed the contract on arbitrum, what happens is the `deployedContracts` file gets updated with the deployement data of only arbitrum and looses the data from polygon since the second person doesn't have the `deployements` folder containing data related to deployment on polygon chain on their computer.

### Expected Behavior

The expected behavior would be when the second person deploys the contract on arbitrum the `deployedContracts` file should be updated by appending the data related to arbitrum deployment and not loosing any previously existing deployment data related to other chains

### Anything else?

Removing the `deployments` folder in hardhat from gitignore solves this issue of inconsistency. I am not sure on why it was added but I am currently unable to see any reason for its existence if there is some please do enlighten me.

sverps:
Fair point. One remark, though, is that anything generated on a local chain is only present locally, so I think it's fine to keep `deployments/localhost` in `.gitignore` in any case.

Committing stuff deployed to other chains to a repo makes sense.

carletex:
Hey @kevinjoshi46b 

Thanks for posting this issue.

I noticed this when working with multiple people and deploying different smart contracts. As you said, the source of truth is the `deployments` folder inside hardhat (this is because we use the hardhat-deploy lib). 

I saw your PR #370 where un-gitignore the `deployments` folders. That will fix the issue but also will create a lot of noise with the `localhost` contracts. So maybe we want to ignore that one.

Related to this, It also makes me think that maybe we should have different files for the `generated` folder in NextJS.

1. `deployedContracts.ts`. All non-localhost contracts deployed from SE-2
2. `deployedLocalhostContracts.ts`. All localhost contracts deployed from SE-2 (gitignored)
3. `externalContracts.ts`. Allow people to add external contracts (existing contract, not deployed from your SE-2 repo) https://github.com/scaffold-eth/scaffold-eth-2/issues/271

[...]

PR #370, opened 2023-06-07
fx: deployedContracts data inconsistency

kevinjoshi46b:
Solves #369


kevinjoshi46b:
As per our discussion on separation of deployed contracts into two files, one of which being for local deployments which is gitignored, there are two working solutions that I was able to find:
1. Creating a postinstall script just like before that will generate the empty (returning null) `deployedLocalhostContracts.ts` file
2. importing using try catch and require statements

I have implemented and tested the second method and it works well too. We can have further discussion on what to go ahead with now... and also if there is any other method.

Reviews and opinions @sverps @carletex @damianmarti @technophile-04

kevinjoshi46b:
Wait why aren't my new commits showing up in this PR :)
Anyone has faced this problem or knows how to fix this?

I have pushed changes into [kevinjoshi46b:fx-deployedContracts-data-inconsistency](https://github.com/kevinjoshi46b/scaffold-eth-2/tree/fx-deployedContracts-data-inconsistency) branch you can check it out there or maybe I might have to just create another PR :)

technophile-04:
Closing this now, will try to create a fresh one later this week as discussed in #370 

But Tysm @kevinjoshi46b, really appreciate it !!!

PR #372, opened 2023-06-12
fix(contract.ts): improve missing config check

### LINKED ISSUE #371: bug: Argument of type '{ contractName: string; functionName: string; }' is not assignable to parameter of type 'never'.

### Current Behavior

**Description**:

I encountered a type error when attempting to build the Next.js frontend package using `yarn build`. The error message is as follows:

image

**Steps to reproduce**:

1. Clone the repository and navigate to the project directory.
2. Install the required dependencies using `yarn install`.
3. Run `yarn build` to build the Next.js frontend package.

**Expected behavior**:

The frontend package should build successfully without any type errors.

**Actual behavior**:

The build process fails with the following error message:

```
Argument of type '{ contractName: string; functionName: string; }' is not assignable to parameter of type 'never'.
```

**Environment**:

- OS: Windows 11
- Node.js version: 18.25.0
- Yarn version: 3.2.3
- Next.js version: 13.1.6

sverps:
## Description

Rather than just check if `deployedContracts` is `null`, this improves the behavior to check if there is a contract definition for the network selected in `scaffold.config.ts`. This results in fewer type errors when a contract might have been deployed on some network, but locally a user only has the file generated for another network.

## Related Issues

Fixes #371

### DIFF
--- a/packages/nextjs/utils/scaffold-eth/contract.ts
+++ b/packages/nextjs/utils/scaffold-eth/contract.ts
@@ -27,14 +27,17 @@ export type GenericContractsDeclaration = {
 
 export const contracts = contractsData as GenericContractsDeclaration | null;
 
-type IsContractsFileMissing<TYes, TNo> = typeof contractsData extends null ? TYes : TNo;
-type ContractsDeclaration = IsContractsFileMissing<GenericContractsDeclaration, typeof contractsData>;
+type ConfiguredChainId = (typeof scaffoldConfig)["targetNetwork"]["id"];
+
+type IsContractDeclarationMissing<TYes, TNo> = typeof contractsData extends { [key in ConfiguredChainId]: any }
+  ? TNo
+  : TYes;
+
+type ContractsDeclaration = IsContractDeclarationMissing<GenericContractsDeclaration, typeof contractsData>;
 
 export type Chain = keyof ContractsDeclaration;
 
-type SelectedChainId = IsContractsFileMissing<number, (typeof scaffoldConfig)["targetNetwork"]["id"]>;
-
-type Contracts = ContractsDeclaration[SelectedChainId][0]["contracts"];
+type Contracts = ContractsDeclaration[ConfiguredChainId][0]["contracts"];
 
 export type ContractName = keyof Contracts;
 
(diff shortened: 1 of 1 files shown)


technophile-04:
Tysm @sverps 🙌, but I think I am kind of neutral on this because I would love to see an error when I deploy my contract to `sepolia` and my frontend is pointing to `goerli`. 

With this fix I deployed my contract to `sepolia` and pointed my frontend to `mainnet` and I see no error for contractReads/Write even when you run `yarn next:types` no errors are shown this means while deploying too it would get easily deployed even though I am pointing to `mainnet` and don't have any contract on it. 

Yup this fix will be beginner friendly also the TS error message `{ contractName: string; functionName: string; }` is not good but I would prefer to get an error if I some doing something wrong 😅

---

**An idea** : What if we log a message like "Make sure you update the chain to seplolia in scaffold.config.ts" after you run deploy to avoid situations like #371, since now we run our custom script in the end : 
https://github.com/scaffold-eth/scaffold-eth-2/blob/e77565c1b731e4851db923a08c051c969ea4fd83/packages/hardhat/deploy/99_generateTsAbis.ts#L76

sverps:
When you run the frontend without a deployed contract on the chain you configured, it already properly informs you of it, right? For beginners, I think that is slightly preferred over some console warning or error that could easily get overlooked.

But it's not entirely clear-cut either. So I'm okay with exploring this discussion a bit further.

technophile-04:
**If you run frontend without a deployed contract on the chain you configured** 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/9c77cfed-2266-4bab-8bbc-a2aee6e4a0f7

People get this popup if they use `useScaffold` hooks 

---

**Regarding types error** 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/19c19aa0-c95b-46ea-8d58-53ba2cbfe93e

As we can see we don't get any error on this branch even if we have a missing deployed contract for that chain. 

---

> But it's not entirely clear-cut either.

+++

I mean I am completely fine with the current fix and it makes sense,  but I am kind of leaning towards showing good ts-error(which will be tough) instead of not showing the error itself but yeah current fix also makes sense!! since the ts-error is not beginner friendly

sverps:
#378 seems to have the same issue (awaiting confirmation from issue creator)

@technophile-04 I'd like to merge this PR to avoid blocking users with non-intuitive errors.

technophile-04 (approved):
> blocking users with non-intuitive errors

Yeah I agree, people might spend time in the wrong direction while debugging due to non-intuitive errors so lets merge

PR #375, opened 2023-06-14
fix(Balance): font-size of symbol should be relative to current font-size, not document root

sverps:
## Issue

When displaying Balance component with larger font-size, the ETH / $ symbol remains `text-xs`
image

## Fix

Use `text-[...em]` instead of `text-xs`. This sets an `em` value instead of a `rem` value, to ensure the symbol size is modified relative to the local font-size rather than that of the document root.
image


sverps:
This will allow users to better customize the size of the `Balance` component by setting a font-size on it (see #373).

rin-st (approved):

technophile-04 (approved):
Yup makes sense! Tested it and looks great !! Tysm @sverps 🙌

PR #377, opened 2023-06-15
fix(contract.ts): very long abi's now no longer go over type instantiation limit

### LINKED ISSUE #374: bug: When contracts ABI is generated, yarn build fails

### Current Behavior

After I run a `yarn deploy` (in Hardhat), I obtain this error if I try to run `yarn build` (in Nextjs):

```
$ yarn build                                                                                                                                                                                                                                                                                      ~/project/packages/nextjs
info  - Linting and checking validity of types .Failed to compile.

./hooks/scaffold-eth/useScaffoldContractRead.ts:37:39
Type error: Type instantiation is excessively deep and possibly infinite.

  35 |     watch: true,
  36 |     args,
> 37 |     enabled: !Array.isArray(args) || !args.some(arg => arg === undefined),
     |                                       ^
  38 |     ...(readConfig as any),
  39 |   }) as Omit<ReturnType<typeof useContractRead>, "data" | "refetch"> & {
  40 |     data: AbiFunctionReturnType<ContractAbi, TFunctionName> | undefined;
```

### Expected Behavior

It should be able to build. `yarn start` does not show errors.

### Steps To Reproduce

- Create a contract (I don't think it's related to my specific contracts)
- Compile to generate generated/deployedContracts.ts

Then the error will occur. If I revert the changes to that file, it works again.

### Anything else?

_No response_

sverps:
Fixes #374

## Description

Make a very long contract.ts file by just copy-pasting the `greeting` view function (no need to actually deploy or make a valid Abi). At a certain length, TypeScript starts returning errors when running `yarn next:check-types`.

## Findings

Contract abi's well over 2000 lines long can hit the typescript type instantiation limit (see [context of this limit](https://github.com/microsoft/TypeScript/issues/34933)). A workaround is described [here](https://stackoverflow.com/a/74292294/5000057).

Applying it does seem to resolve the issue.

### DIFF
--- a/packages/nextjs/utils/scaffold-eth/contract.ts
+++ b/packages/nextjs/utils/scaffold-eth/contract.ts
@@ -121,6 +121,10 @@ type RestConfigParam<TAbiStateMutability extends AbiStateMutability> = Partial<
   >
 >;
 
+type Expand<T> = T extends object ? (T extends infer O ? { [K in keyof O]: O[K] } : never) : T;
+
+type UnionToIntersection<U> = Expand<(U extends any ? (k: U) => void : never) extends (k: infer I) => void ? I : never>;
+
 type OptionalTupple<T> = T extends readonly [infer H, ...infer R] ? readonly [H | undefined, ...OptionalTupple<R>] : T;
 
 type UseScaffoldArgsParam<
@@ -129,7 +133,7 @@ type UseScaffoldArgsParam<
   TFunctionName extends ExtractAbiFunctionNames<ContractAbi<TContractName>, TAbiStateMutability>,
 > = TFunctionName extends FunctionNamesWithInputs<ContractAbi<TContractName>, TAbiStateMutability>
   ? {
-      args: OptionalTupple<AbiFunctionArguments<ContractAbi<TContractName>, TFunctionName>>;
+      args: OptionalTupple<UnionToIntersection<AbiFunctionArguments<ContractAbi<TContractName>, TFunctionName>>>;
     }
   : {
       args?: never;


technophile-04 (approved):
> by just copy-pasting the greeting view function (no need to actually deploy or make a valid Abi). At a certain length, TypeScript starts returning errors when running yarn next:check-types

Yeah it did break on `main` for me the breakpoint was copy-pasting 1100 `greeting` fragments and yes `yarn next:check-types` did pass when I switch to this branch !! 

Just had a small naive doubt : 

Lets suppose : 
```js
type AbiFunctionArguments = [`0x{string}`, BigInt];
```

What is a resultant type if I pass it through `UnionToIntersection ` : 
```js
type x = UnionToIntersection<AbiFunctionArguments>
//   ^? 
```

Because I thought `U` in `UnionToIntersection<U>` must be a `unions of object`  but here we are passing it just an `array`  

Sorry for being PITA 😅 just curious what am I missing 

But tested the solution out and it does work nicely 🙌!!

sverps:
The issue is that `AbiFunctionArguments<ContractAbi<TContractName>, TFunctionName>` takes generic params, so it's not simply a tupple, it's a union of all tupples for each different `TFunctionName` you could give.

So there's the issue, when there's too many different functions, TypeScript thinks it might be in an infinite type resolving loop.

I believe the fix works by resolving this type to `never` when `TFunctionName` is not set (so it doesn't do the unnecessary type resolving), and when actually using some hook with a defined `functionName`, the type works because then it's constrained enough by the chosen `functionName` to not get in trouble. 

It does feel a bit like dark magic, I must admit... :sweat_smile:

carletex:
Great job on this and great catch  @alvatar

Thanks!! :)

PR #256, opened 2023-03-24
feat: adds zksync era testnet and mainnet

codenamejason:

PR #383, opened 2023-06-20
zkSyncEra Testnet config

moltam89:
## Description

Configuration for deploying on zkSyncTestnet.
According to the [zkSync docs](https://era.zksync.io/docs/tools/hardhat/migrating-to-zksync.html#deploy-contracts), hardhat-deploy version ^0.11.26 supports deployments on zkSync Era.

Therefore I've upgraded hardhat-deploy to 0.11.26.
Also, zkSync uses their own "hardhat-zksync-solc" compiler, which has to be imported in hardhat.config.ts.

After these changes, I could deploy to zkSyncTestnet with **yarn deploy --network zkSyncTestnet** and I could interact with the demo YourContract when I switched the targetNetwork to chains.zkSyncTestnet in [scaffold.config.ts](https://github.com/scaffold-eth/scaffold-eth-2/blob/9faa67540a264accf71cf03bd8dd89b06b6a4c36/packages/nextjs/scaffold.config.ts#L16).

Let me know if you have any questions.

Best,
Tamas

## Additional Information

## Related Issues

https://github.com/scaffold-eth/scaffold-eth-2/pull/256

Your ENS/address: monyo.eth

### DIFF
--- a/packages/hardhat/hardhat.config.ts
+++ b/packages/hardhat/hardhat.config.ts
@@ -82,6 +82,11 @@ const config: HardhatUserConfig = {
       zksync: true,
       accounts: [deployerPrivateKey],
     },
+    zkSync: {
+      url: "https://mainnet.era.zksync.io",
+      zksync: true,
+      accounts: [deployerPrivateKey],
+    },
   },
   verify: {
     etherscan: {
--- a/packages/hardhat/hardhat.config.ts
+++ b/packages/hardhat/hardhat.config.ts
@@ -76,6 +76,11 @@ const config: HardhatUserConfig = {
       url: `https://polygon-mumbai.g.alchemy.com/v2/${providerApiKey}`,
       accounts: [deployerPrivateKey],
     },
+    zkSyncTestnet: {
+      url: "https://testnet.era.zksync.dev",
+      zksync: true,
+      accounts: [deployerPrivateKey],
+    },
   },
   verify: {
     etherscan: {
--- a/packages/hardhat/hardhat.config.ts
+++ b/packages/hardhat/hardhat.config.ts
@@ -4,6 +4,7 @@ import { HardhatUserConfig } from "hardhat/config";
 import "@nomicfoundation/hardhat-toolbox";
 import "hardhat-deploy";
 import "@matterlabs/hardhat-zksync-solc";
+import "@matterlabs/hardhat-zksync-verify";
 
 // If not set, it uses ours Alchemy's default API key.
 // You can get your own at https://dashboard.alchemyapi.io
@@ -81,11 +82,13 @@ const config: HardhatUserConfig = {
       url: "https://testnet.era.zksync.dev",
       zksync: true,
       accounts: [deployerPrivateKey],
+      verifyURL: "https://zksync2-testnet-explorer.zksync.dev/contract_verification",
     },
     zkSync: {
       url: "https://mainnet.era.zksync.io",
       zksync: true,
       accounts: [deployerPrivateKey],
+      verifyURL: "https://zksync2-mainnet-explorer.zksync.io/contract_verification",
     },
   },
   verify: {
--- a/packages/hardhat/hardhat.config.ts
+++ b/packages/hardhat/hardhat.config.ts
@@ -3,6 +3,7 @@ dotenv.config();
 import { HardhatUserConfig } from "hardhat/config";
 import "@nomicfoundation/hardhat-toolbox";
 import "hardhat-deploy";
+import "@matterlabs/hardhat-zksync-solc";
 
 // If not set, it uses ours Alchemy's default API key.
 // You can get your own at https://dashboard.alchemyapi.io
--- a/yarn.lock
+++ b/yarn.lock
@@ -278,6 +278,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@balena/dockerignore@npm:^1.0.2":
+  version: 1.0.2
+  resolution: "@balena/dockerignore@npm:1.0.2"
+  checksum: 0d39f8fbcfd1a983a44bced54508471ab81aaaa40e2c62b46a9f97eac9d6b265790799f16919216db486331dedaacdde6ecbd6b7abe285d39bc50de111991699
+  languageName: node
+  linkType: hard
+
 "@coinbase/wallet-sdk@npm:^3.5.4":
   version: 3.6.3
   resolution: "@coinbase/wallet-sdk@npm:3.6.3"
@@ -482,7 +489,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@ethersproject/contracts@npm:5.7.0":
+"@ethersproject/contracts@npm:5.7.0, @ethersproject/contracts@npm:^5.7.0":
   version: 5.7.0
   resolution: "@ethersproject/contracts@npm:5.7.0"
   dependencies:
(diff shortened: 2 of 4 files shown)


packages/hardhat/hardhat.config.ts
```diff
@@ -76,6 +77,11 @@ const config: HardhatUserConfig = {
       url: `https://polygon-mumbai.g.alchemy.com/v2/${providerApiKey}`,
       accounts: [deployerPrivateKey],
     },
+    zkSyncTestnet: {
+      url: "https://testnet.era.zksync.dev",
+      zksync: true,
+      accounts: [deployerPrivateKey],
+    },
```
carletex, line 84: Is there a mainnet also?

packages/hardhat/hardhat.config.ts
carletex, line 1: From the docs:

image

Do we have to do that? Not a fan of polluting all the other network configs

packages/hardhat/hardhat.config.ts
carletex: What about the `ethNetwork` prop that I see in the docs?

carletex (commented):
Hey @moltam89 

Thanks for this. I left a few comments/questions.

Also...does the deploy script work as it is?

The docs say that you need to use the Deployer from 
`import { Deployer } from "@matterlabs/hardhat-zksync-deploy";`

packages/hardhat/hardhat.config.ts
moltam89: Yeah, there is a mainnet as well, I think it works the same way and we would only need to add the proper configs according to [this](https://era.zksync.io/docs/reference/troubleshooting/important-links.html).
However, I didn't want to use real ETH to test it, and didn't want to add something that might be broken.

With ```ethNetwork``` we could deposit founds from the mainnet/goerli, but I it would make to code more complex, it's easier if we already bridged founds.

moltam89:
Hey @carletex 

> Also...does the deploy script work as it is?

I could deploy by **yarn deploy --network zkSyncTestnet**. Not sure if this is the proper answer, or you meant something else.



> The docs say that you need to use the Deployer from
> `import { Deployer } from "@matterlabs/hardhat-zksync-deploy";`

I think this would be only needed if we didn't use **hardhat-deploy**.
If we would import it, we could also deploy by "yarn hardhat deploy-zksync", but it would be different than how other networks are used.

damianmarti:
Thanks @moltam89 

I added the zkSync mainnet config and I was able to deploy the contract to this network using **yarn deploy --network zkSync**

https://explorer.zksync.io/tx/0x041947bfa70bea288655ce64f4438114f69050f27b2eeafef5f895eb6b26d572

damianmarti:
> Thanks @moltam89
> 
> I added the zkSync mainnet config and I was able to deploy the contract to this network using **yarn deploy --network zkSync**
> 
> https://explorer.zksync.io/tx/0x041947bfa70bea288655ce64f4438114f69050f27b2eeafef5f895eb6b26d572

And then I was able to set a new greeting from the debug page.

moltam89:
Hey @damianmarti ,

Great, thank you for testing this!

packages/hardhat/hardhat.config.ts
carletex: Checking the source code of the `@matterlabs/hardhat-zksync-solc` seems that it's not necesary as they say in the docs.

They always check it with something like this.
`if (hre.network.zksync !== true) {`

yarn.lock
```diff
@@ -990,6 +997,19 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@matterlabs/hardhat-zksync-solc@npm:^0.3.17":
+  version: 0.3.17
+  resolution: "@matterlabs/hardhat-zksync-solc@npm:0.3.17"
+  dependencies:
+    "@nomiclabs/hardhat-docker": ^2.0.0
```
damianmarti, line 1004: My only concern about this PR is how many libraries it adds to yarn.lock. It seems like hardhat-zksync-solc adds @nomiclabs/hardhat-docker as a dependency. I'm not sure why this requires this dependency...

damianmarti (approved):

technophile-04 (approved):
Tested it and works properly, Tysm @moltam89  looks a lot cleaner 🙌

I have added [hardhat-zksync-verify](https://era.zksync.io/docs/tools/hardhat/hardhat-zksync-verify.html) at [3d3ceab](https://github.com/scaffold-eth/scaffold-eth-2/pull/383/commits/3d3ceab626fa7966ad603716beee8ae0abd6e9cd)...Reason: We  already had `hardhat-etherscan` to which its goes in conjuction and it didn't required any `API keys` so thought its worth the `verify`   feature by default 

Here is the verified contract: https://goerli.explorer.zksync.io/address/0xC8796A82A7A03A333295938d54d8F7310d607be2#contract

PS: I tried deploying to sepolia `yarn deploy --network sepolia` and got some error the first time but after a second try it worked correctly, just can anyone just test once deploying to `sepolia`.

carletex:
> My only concern about this PR is how many libraries it adds to yarn.lock. It seems like hardhat-zksync-solc adds @nomiclabs/hardhat-docker as a dependency. I'm not sure why this requires this dependency...

Yep, I think this was also one of the reasons Shiv closed the initial PR. We can't automatically add every single thing that _could_ be useful on SE-2... we'd end with a massive project. That's why he suggested adding it as an extension in the CLI (#251)

In this particular case... I think ZkSyncEra is probably something that will be used and might be worth it to add it to the core.

Let me know what you all think!

technophile-04:
> Yep, I think this was also one of the reasons Shiv closed the initial PR. We can't automatically add every single thing that could be useful on SE-2... we'd end with a massive project. That's why he suggested adding it as an extension in the CLI (https://github.com/scaffold-eth/scaffold-eth-2/issues/251)

Yes !!  I was even not going to suggest creating a PR to `main` at all and instead have it on just another branch until CLI but then I saw a couple of builds from BG member which was deployed to ZkSync so thought it might be worth the add  to `main` 

> In this particular case... I think ZkSyncEra is probably something that will be used and might be worth it to add it to the core.

Yup

damianmarti:
@moltam89 @carletex There is another issue. If you want to replicate all the zkSync features (like a Paymaster) locally, you have to run a local zkSync node, which is not the same that running **yarn chain**. Maybe it's something we should add a note somewhere at least... https://era.zksync.io/docs/tools/hardhat/testing.html

packages/hardhat/hardhat.config.ts
```diff
@@ -76,6 +78,18 @@ const config: HardhatUserConfig = {
       url: `https://polygon-mumbai.g.alchemy.com/v2/${providerApiKey}`,
       accounts: [deployerPrivateKey],
     },
+    zkSyncTestnet: {
```
carletex, line 81: > There is another issue. If you want to replicate all the zkSync features (like a Paymaster) locally, you have to run a local zkSync node, which is not the same that running yarn chain. Maybe it's something we should add a note somewhere at least... https://era.zksync.io/docs/tools/hardhat/testing.html

Maybe something like?

```suggestion
    // If you want to run a local node with all zkSyncEra functionalities, check:
    // https://era.zksync.io/docs/tools/hardhat/testing.html
    zkSyncTestnet: {
```

damianmarti:
Thanks @moltam89!! Merged!

moltam89:
Hi @technophile-04 

> I have added [hardhat-zksync-verify](https://era.zksync.io/docs/tools/hardhat/hardhat-zksync-verify.html) at [3d3ceab](https://github.com/scaffold-eth/scaffold-eth-2/pull/383/commits/3d3ceab626fa7966ad603716beee8ae0abd6e9cd)...Reason: We already had `hardhat-etherscan` to which its goes in conjuction and it didn't required any `API keys` so thought its worth the `verify` feature by default

I tried to verify with **yarn verify --network zkSyncTestnet**, but I run into this error:
**Network with chainId: 280 not supported. You can specify the url manually via --api-url <url>.**

How did you run the contract verification?

Thanks,
Tamas

PR #385, opened 2023-06-22
update writeAsync to take in updated args while calling it

technophile-04:
## Description

Sometimes there is a situation where you need to calculate some stuff just before calling `writeAsync` and currently I think there is no way to pass that calculated value as args  before calling the function (are there any good ways of doing this ?)

Current non-working example : 
```javascript
const [uploadedItemHash, setUploadedItemHash] = useState("")

const { writeAsync: mintItem } = useScaffoldContractWrite({
  ...,
  args : [connectedAddress,uploadedItemHash] 
  // here the args are not updated while calling the writeAsync
});
  
const handleMintItem = async () => {
  const uploadedItem = await ipfsClient.add(JSON.stringify(currentTokenMetaData));
  setUploadedItemHash(uploadedItem.path)
  await mintItem();
 }  
```

New Working example : 
```javascript

const { writeAsync: mintItem } = useScaffoldContractWrite({...});
  
const handleMintItem = async () => {
  const uploadedItem = await ipfsClient.add(JSON.stringify(currentTokenMetaData));
  await mintItem({
    args: [connectedAddress, uploadedItem.path],
  });
};

```

## Additional Information

### DIFF
--- a/packages/nextjs/components/example-ui/ContractInteraction.tsx
+++ b/packages/nextjs/components/example-ui/ContractInteraction.tsx
@@ -66,7 +66,7 @@ export const ContractInteraction = () => {
                   className={`btn btn-primary rounded-full capitalize font-normal font-white w-24 flex items-center gap-1 hover:gap-2 transition-all tracking-widest ${
                     isLoading ? "loading" : ""
                   }`}
-                  onClick={writeAsync}
+                  onClick={() => writeAsync()}
                 >
                   {!isLoading && (
                     <>
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
@@ -49,7 +49,15 @@ export const useScaffoldContractWrite = <
     ...restConfig,
   });
 
-  const sendContractWriteTx = async () => {
+  const sendContractWriteTx = async ({
+    args,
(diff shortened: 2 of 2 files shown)


packages/nextjs/components/example-ui/ContractInteraction.tsx
```diff
@@ -66,7 +66,7 @@ export const ContractInteraction = () => {
                   className={`btn btn-primary rounded-full capitalize font-normal font-white w-24 flex items-center gap-1 hover:gap-2 transition-all tracking-widest ${
                     isLoading ? "loading" : ""
                   }`}
-                  onClick={writeAsync}
+                  onClick={() => writeAsync()}
```
technophile-04, line 69: We wont be able to `onClick={writeAsync}` anymore 😅 but I think its ok

sverps:
Makes sense, this basically exposes what wagmi already allows to our custom hook :+1:

packages/nextjs/components/example-ui/ContractInteraction.tsx
sverps: I don't see an issue either, the behavior will be even more aligned with wagmi's `writeAsync`

sverps (approved):

damianmarti (approved):
Thanks!! I think this change makes the hook works as the expected behavior for the user.

PR #395, opened 2023-06-30
autoConnect fix

### LINKED ISSUE #391: bug: autoConnect is not working

### Current Behavior

It seems that autoConnect does not works properly, even thought you have `autoConnect: true` in `scaffold.config`

### Expected Behavior

It should autoConnect if you have it enabled

### Steps To Reproduce

1. Change `targetNetwork`to `chains.sepolia`
2. Run frontend and connect wallet 
3. Refresh the page, it will tell you to again connect the wallet

### Anything else?

It seems that this was caused due to #356, if you check out to its parent commit there it works fine

Anurag-Raut:
## Description

walletAutoConnect declared in scaffold.config.ts was not passed down to WagmiClient .

In wagmiClient autoconnect is set to false. So even if we changed the value of walletAutoConnect in scaffold.config.ts changes were not reflected.

https://github.com/scaffold-eth/scaffold-eth-2/blob/main/packages/nextjs/services/web3/wagmiClient.tsx#L5

## Additional Information

## Related Issues

Closes #391 

Your ENS/address: 0x72aA8A79D02705444CC6918A6766Be03AD4C29F5


technophile-04:
Tysm @Anurag-Raut 🙌 

I actually tried the fix but it gave me a "Hydration Error" : 

Screenshot 2023-07-01 at 7 18 14 AM

**Steps to reproduce** : 
1. Make sure you have `hardhat` network selected in `scaffold.config.ts`
2. `yarn start`

Also if you switch to another chain than `hardhat` then there it works correctly, I also tried setting `bunerWallet.onlyLocal : false` and it works fine there too just the problem is with `hardhat`

Could you please confirm the above error, Anurag?  Thanks !

Anurag-Raut:
> 

I got this error even when I had first cloned the repo and hadn't changed anything.
I had also encountered this error while I was working on my prev project where I was using rainbowKit.

I got through this issue by -https://github.com/rainbow-me/rainbowkit/issues/944#issuecomment-1377147717
(which works in this case too).

I thought the problem was at my local setup.
(Does hydration error occur in freshly cloned repo as well ?)
I was going to raise a issue regarding this on rainbowKit repo. Not sure why enabling autoConnect caused the problem.

So the se-2 repo doesn't cause hydration error to other devs right? could you try in a prev working version if this error happens if you just changed the auto-connect variable to true in  https://github.com/scaffold-eth/scaffold-eth-2/blob/main/packages/nextjs/services/web3/wagmiClient.tsx#L5C16-L5C21

technophile-04:
> I thought the problem was at my local setup.
(Does hydration error occur in freshly cloned repo as well ?)
> So the se-2 repo doesn't cause hydration error to other devs right? could you try in a prev working version if this error happens if you just changed the auto-connect variable to true in

Yes the the current `main` branch state doesn't have hydration error 

I tried adding `autoConnect : true` to `createClient` in se-2 `main` branch and it shows hydration error when I am on `targetNetwork : chains.hardhat` in scaffold.config.ts 

As I mentioned in previous comment the problem is only on hardhat network and it does not give hydration error even when `autConnect: true` in `createClient`, I will have a look at it later 🙌 

Also I am not sure whats the perfect cause because as mentioned in this https://github.com/scaffold-eth/scaffold-eth-2/issues/391#issue-1781187456 it working fine for me before #356 soo need to look at it too

Anurag-Raut:
I noticed that when network hardhat is selected , autoconnect functionality is working perfectly due to  useAutoConnect hook in-
https://github.com/scaffold-eth/scaffold-eth-2/blob/123e6d84ac613e15e07fd854268b6869d7353615/packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx#L12C1-L12C1

but when we change auto connect to true in wagmiClient along with the hook , those two clashes with each other and hydration errors occurs.
however useAutoConnect hook doesn't work for other networks apart from hardhat.

so made changes to pass down the scaffoldConfig.walletAutoConnected to wagmi connector only when current network 
!= hardhat.

It now works for both hardhat and other networks .

could you look into it ? @technophile-04

technophile-04:
Thanks 🙌 Umm with this fix we will be missing some cases like and it seems like a bit hacky to me : 

**Steps to reproduce** 
 1. `git switch Anurag-Raut/Autoconnect-fix`
 2. Change `targetNetwork : chains.mainnet` and `burnerWallet.onlyLocal : false` in `scaffold.config.ts` and then `yarn start`
 3. Try connecting to `Metamask` and then refresh the page it will be get connected to Burner Wallet instead of connecting to metasmask wallet 
 
 
 **Idea Condition** 
 1. `git switch main`
 2. `git checkout c7af09a`
 3. Do step 3 & 4 , It will reconnect to metamask instead of connecting to BurnerWallet

Issue #394, opened 2023-06-30
Foundry package todo's


technophile-04:
We have all the basics setup at #353, you run local anvil chain, deploy your contracts, tinker with them on frontend all the commands are similar se-2 `main` branch....checkout out [README.md](https://github.com/scaffold-eth/scaffold-eth-2/tree/foundry#readme) of [`foundry`](https://github.com/scaffold-eth/scaffold-eth-2/tree/foundry#readme) branch for instruction. 

#### some issues / improvements noticed : 

- [x] The `blockexplorer/address/${address}` `code` does not shows anything 
![Screenshot 2023-06-26 at 2 44 41 PM](https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/b4fadc0f-a66c-44ff-8fd2-044731fe7277)

- [x] We need to figure out better way for generating `.env` because I think it the current solution will have problem in `Windows` since we are using the forward slash (Maybe having a separate script to generate `.env` ) (We have it already solved it CLI)

- [x] Update the README.md properly (mainly deploying contracts to test / main networks)

- [x] Maybe covert ListAccount.s.sol to .ts / .js file since `ListAccount.s.sol` takes time and throws the output at once.... we can show QR code when using `.js / .ts`.....(We should also remove `ethers` from dependency and use `viem` 🙂)

- [x] Setup typescript, eslint / prettier for code formating of `.js /.ts` files

- [ ] Optional : Check for [forge-deploy](https://github.com/wighawag/forge-deploy) if it have features like `hardhat-deploy` eg. Not deploy a new contract if no changes have happen instead reuse prev one maybe it can also increase devX for deploying and handling deployments

---

Please feel free to tinker around with [`foundry`](https://github.com/scaffold-eth/scaffold-eth-2/tree/foundry) and append issues / suggest any improvements   🙌

lucasia:
Excited to see a foundry version.  Just what I was looking for!  

I noticed a small issue on the foundry branch.  Shall I open an issue for it or add it to this thread?

In: packages/foundry/package.json 

yarn "deploy" tries to run node script/generateTsAbis.**t**s (Typescript) but the script is script/generateTsAbis**.j**s (javascript)

See my code fix: https://github.com/lucasia/se2-foundry/commit/3dcb93d498760e81ac6468d628ba9a7c926ab44c

technophile-04:
[...]

PR #341, opened 2023-05-07
Add Gnosis to scaffold-eth-2 configurations

0xMako:
## Description

_Concise description of proposed changes, We recommend using screenshots and videos for better description_

## Additional Information

- [ ] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [ ] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

_Closes #{issue number}_

_Note: If your changes are small and straightforward, you may skip the creation of an issue beforehand and remove this section. However, for medium-to-large changes, it is recommended to have an open issue for discussion and approval prior to submitting a pull request._

Your ENS/address:

PR #406, opened 2023-07-05
fix/update foundry branch

jrcarlos2000:
## Description

Addressing some errors , improvements in foundry branch

- [x] Fix bytecode opcode in `code` tab of blockexplorer.
- [x] Fix export deployment for custom networks ( networks that are not in stdChains of foundry )
- [x] Updated README
- [ ] 

## Additional Information

## Related Issues

_Closes #394


jrcarlos2000:
Custom Networks `exportDeployments()` fails because it uses a list of `stdChains` and if your custom chain is not in the list, the script breaks.

So far in the case of custom networks, the user needs to use `exportDeployments("CustomNetworkName")`, that's the best and fastest solution I have found so far, been playing around with the different tools of the foundry library, however there are other ways to avoid this and get back to a single function exportDeployments()
- Similar to `ListAccount.s.sol`, iterate over available network with forks and compare chainId, however this will be very slow.
- Forget about network names and only use chainIds, this would still work in /nextjs package.

will keep tinkering about this problem. cc: @technophile-04

technophile-04:
Tysm @jrcarlos2000 !! really appreciate it 🙌

> Fix bytecode opcode in code tab of blockexplorer.

Thanks, I was wondering if is there a way to put all contract's in `contracts` directory instead of `src`? We are trying hard to keep `foundry` branch similar to `hardhat`, So it would be great if we default all our contracts in `contracts` directory 🙌. maybe we update mapping `src = 'contracts'` in `foundry.toml` not sure 

---

> Custom Networks exportDeployments() fails because it uses a list of stdChains and if your custom chain is not in the list, the script breaks.

> So far in the case of custom networks, the user needs to use exportDeployments("CustomNetworkName"), that's the best and fastest solution I have found so far, been playing around with the different tools of the foundry library, however there are other ways to avoid this and get back to a single function exportDeployments()

Regarding this you mean to say if people add custom chain in `foundry.toml` which is not mentioned here [stdChains](https://github.com/foundry-rs/forge-std/blob/ff4bf7db008d096ea5a657f2c20516182252a3ed/src/StdCheats.sol#L230)   eg. `zetaChain` and then do `yarn deploy zetaChain` it will break? 

And if you want to work you need to call extra function after this : 
https://github.com/scaffold-eth/scaffold-eth-2/blob/d5c1366484c2808e8b35eecbe2d01d11cefce2d4/packages/foundry/script/Deploy.s.sol#L21

Like `exportDeployments("zetaChain")`? 

I would say let's keep this for other PR, could you just mention this problem in #394 and handle it in a separate PR? 

---

Also we recently migrated ListAccount.s.sol -> `.js` in #437, so I see there is some conflict in `packages/foundry/package.json` could you please solve it 🙌

jrcarlos2000:
> Thanks, I was wondering if is there a way to put all contract's in contracts directory instead of src? We are trying hard to keep foundry branch similar to hardhat, So it would be great if we default all our contracts in contracts directory 🙌. maybe we update mapping src = 'contracts' in foundry.toml not sure

Yes, that is possible !

Noted on the rest!

jrcarlos2000:
>Like exportDeployments("zetaChain")?
I would say let's keep this for other PR, could you just mention this problem in https://github.com/scaffold-eth/scaffold-eth-2/issues/394 and handle it in a separate PR?

The current implementation has overloaded functions so we can work on improving that mechanism in future PRs
cc: @technophile-04

technophile-04:
Tysm @jrcarlos2000 🙌, just tested and looks great !! 

- Added some comments for the `exportDeployments` function 
- Added gnosis chain for fixing #341 
- add the comments for YourContract.sol

PR #410, opened 2023-07-06
fix: type issues when multiple contracts deployed

### LINKED ISSUE #409: bug: type error in useScaffoldEventHistory

### Current Behavior

In `useScaffoldContractHistory` , `useEventSubscriber` : 

Screenshot 2023-07-06 at 4 53 37 PM

### Steps To Reproduce

Checkout this `https://github.com/scaffold-eth/scaffold-eth-2/tree/typeError-findNotCallable` branch

1. `yarn chain`

2.`yarn deploy`

3. `yarn next:check-type` 

### Anything else?

I am getting this error only when `MetaMultisigWallet` is paired with another contract, 

If you deploy only `MetaMultisigWallet` it works 

If you deploy only `YourContract` and other `MyContract`(this is any random contract) it works 

But if you deploy `YourContract`(or any) with `MetaMultisigWallet` it gives you this error 

@sverps  could you please look at it 🙌

sverps:
## Description

Typescript has issues to infer the type of the `part` variable in the `find` call because the array is `readonly [...] | readonly [...]`. Since we don't depend on the actual inferred type in this internal hook, we can just assert the abi to it's general `Abi` type.

The other type issue in `eventSubscriber` is similar. Typescript can't infer which of the two abis is to be used in the `useContractEvent` call, and thus doesn't understand that the `eventName` should always be valid. It's possible that a more complex method exists to make typescript infer a more narrow abi, but since it's just internals and is passed down to `useContractEvent`, I think the proposed fix more than suffices. The types that are exposed (param and return type) work fine, so shouldn't really be of concern to the users of the `useScaffoldEventSubscriber` hook) 

## Additional Information

## Related Issues

Fixes #409 

Your ENS/address: solidixen.eth

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
@@ -1,5 +1,5 @@
 import { useEffect, useState } from "react";
-import { AbiEvent, ExtractAbiEventNames } from "abitype";
+import { Abi, AbiEvent, ExtractAbiEventNames } from "abitype";
 import { Hash } from "viem";
 import { usePublicClient } from "wagmi";
 import { useDeployedContractInfo } from "~~/hooks/scaffold-eth";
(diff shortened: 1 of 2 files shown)


rin-st (approved):

technophile-04 (approved):
Yup, this makes sense Samuel !! Tysm 🙌

PR #384, opened 2023-06-20
added scaffold eth configuration doc

KcPele:
## Description
Added a section in the readme.md that explains how to make use of scaffold-eth config file

_Concise description of proposed changes, We recommend using screenshots and videos for better description_
Screenshot from 2023-06-20 13-08-29

## Additional Information

- [ ] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [x] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

_Closes #{issue number}_

_Note: If your changes are small and straightforward, you may skip the creation of an issue beforehand and remove this section. However, for medium-to-large changes, it is recommended to have an open issue for discussion and approval prior to submitting a pull request._

Your ENS/address:
kcpele.eth

PR #412, opened 2023-07-07
Fix/auto connect

### LINKED ISSUE #391: bug: autoConnect is not working

rin-st:
## Description
Fixes auto connect and hydration error. Removes `useAutoConnect` hook since auto connect now works without a hook. Please lmk if it is still needed

## Additional Information

## Related Issues

Closes #391 

Your ENS/address: 0xrinat


packages/nextjs/components/scaffold-eth/Faucet.tsx
```diff
@@ -74,7 +77,7 @@ export const Faucet = () => {
   };
 
   // Render only on local chain
-  if (!ConnectedChain || ConnectedChain.id !== hardhat.id) {
+  if (ConnectedChain?.id !== hardhat.id || !isMounted()) {
```
sverps, line 80: Don't think we need `isMounted` here. This line is only evaluated during rendering of this component, at which stage `isMounted()` should always return true. When the component isn't mounted, this code will not be executed.

(Checking `isMounted` is more for async stuff, for example when waiting for a response from an API, it's possible the user navigates away and the component that made the request gets unmounted.)

packages/nextjs/components/scaffold-eth/FaucetButton.tsx
```diff
@@ -40,7 +44,7 @@ export const FaucetButton = () => {
   };
 
   // Render only on local chain
-  if (!ConnectedChain || ConnectedChain.id !== hardhat.id) {
+  if (ConnectedChain?.id !== hardhat.id || !isMounted()) {
```
sverps, line 47: See my other coment abbout `isMounted`. I don't think it's doing anything here.

sverps (changes requested):
Nice cleanup @rin-st 

Added two small comments

packages/nextjs/components/scaffold-eth/Faucet.tsx
rin-st: `isMounted` is `false` by default and becomes `true` in `useEffect` after the first render. I added it because it fixes hydration error - `useNetwork` works differently for nextjs server and client

carletex: Yeah! I think it's because of NextJS pre-rendering. the `isMounted` logic makes that pre-rendering === first render. If not, we'd get a Hydration error.

technophile-04:
Also one thing I think now is we can have only maybe as `burnerWallet`  boolean instead of having it as an object and maybe rename it to `burnerWalletOnlyLocal`?    

we will be just needing to update it here : 
https://github.com/scaffold-eth/scaffold-eth-2/blob/50de54f54db133b5a06ae41f3f9d610127925843/packages/nextjs/services/web3/wagmiConnectors.tsx#L64

and update this : 
https://github.com/scaffold-eth/scaffold-eth-2/blob/50de54f54db133b5a06ae41f3f9d610127925843/packages/nextjs/services/web3/wagmi-burner/burnerWalletConfig.ts#L29


We have burnerWallet always on `hardhat` and disabled by default in other chains and if user makes that boolean true it will show on other chains too

packages/nextjs/components/scaffold-eth/Faucet.tsx
sverps: Aaah of course, should have realized in the context of this fix, you're right @rin-st

rin-st:
updated. We should mention changes on https://github.com/scaffold-eth/scaffold-eth-2/pull/384 after merge

sverps (approved):
Looks good

technophile-04 (approved):
Tysm @rin-st 🙌 It works nicely!!

PR #414, opened 2023-07-08
[txHash].tsx page UI improvements

portdeveloper:
## Description

Currently, if the value of a function argument is 0, it is not shown in txHash page because of the already existing logic:

image

By explicitly checking for  undefined, we let the 0 survive javascript's evaluation :). In Javascript 0 is a falsy value so it defaults to "" because of ||.

image

Also, when the function name and the arguments take too much space, the table looks a bit off due to the scrollbar.

Here is a video showing the two problems.

https://github.com/scaffold-eth/scaffold-eth-2/assets/108868128/563d7157-37cd-40ca-b07c-34ee922c98de

This PR solves that problem by utilizing tailwind-scrollbar, and using it in the Function called row in [txHash].tsx page.
Furthermore, I think tailwind-scrollbar can serve people using scaffold-eth 2 well when designing good-looking dApps.

Here is the video after implementing the solutions suggested by this PR.

Also increased the Data row's textarea's height to a fixed value of 150px

https://github.com/scaffold-eth/scaffold-eth-2/assets/108868128/315c751f-21c3-4149-ba6f-4a6913f16825

Let me know what you think about these improvements/fixes! I am open to any kind of feedback.

## Additional Information

- [ ] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [ ] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

_Closes #{issue number}_

Your ENS/address: portdev.eth

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -48,6 +48,7 @@
     "eslint-plugin-prettier": "^4.2.1",
     "postcss": "^8.4.16",
     "prettier": "^2.8.4",
+    "tailwind-scrollbar": "^3.0.4",
     "tailwindcss": "^3.1.8",
     "typescript": "^4.9.5",
     "vercel": "^28.15.1"
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -8,7 +8,7 @@
     "build": "next build",
     "serve": "next start",
     "lint": "next lint",
-    "format": "prettier --write . '!(node_module|.next|contracts)/**/*'",
+    "format": "prettier --write . '!(node_modules|.next|contracts)/**/*'",
     "check-types": "tsc --noEmit --incremental",
     "vercel": "vercel",
     "vercel:yolo": "vercel --build-env NEXT_PUBLIC_IGNORE_BUILD_ERROR=true"
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -48,7 +48,6 @@
     "eslint-plugin-prettier": "^4.2.1",
     "postcss": "^8.4.16",
     "prettier": "^2.8.4",
-    "tailwind-scrollbar": "^3.0.4",
     "tailwindcss": "^3.1.8",
     "typescript": "^4.9.5",
     "vercel": "^28.15.1"
--- a/packages/nextjs/pages/blockexplorer/transaction/[txHash].tsx
+++ b/packages/nextjs/pages/blockexplorer/transaction/[txHash].tsx
@@ -42,7 +42,7 @@ const TransactionPage: NextPage = () => {
         Back
       </button>
       {transaction ? (
-        <div className="overflow-x-auto">
+        <div>
           <h2 className="text-3xl font-bold mb-4 text-center text-primary-content">Transaction Details</h2>{" "}
           <table className="table w-full shadow-lg">
             <tbody>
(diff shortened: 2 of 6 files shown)


packages/nextjs/pages/blockexplorer/transaction/[txHash].tsx
```diff
@@ -94,14 +94,16 @@ const TransactionPage: NextPage = () => {
...
                 <td>
-                  {functionCalled === "0x" ? (
-                    "This transaction did not call any function."
-                  ) : (
-                    <>
-                      <span className="mr-2">{getFunctionDetails(transaction)}</span>
-                      <span className="badge badge-primary font-bold">{functionCalled}</span>
-                    </>
-                  )}
+                  <div className="w-full md:max-w-[600px] lg:max-w-[800px] overflow-x-auto whitespace-nowrap scrollbar-thin scrollbar-thumb-primary scrollbar-track-secondary">
```
Comment (rin-st) [inline packages/nextjs/pages/blockexplorer/transaction/[txHash].tsx:97]: what about sm/xs?

packages/nextjs/utils/scaffold-eth/decodeTxData.ts
```diff
@@ -51,7 +51,10 @@ export const getFunctionDetails = (transaction: TransactionType) => {
     transaction.functionArgs
   ) {
     const details = transaction.functionArgNames.map(
-      (name, i) => `${transaction.functionArgTypes?.[i] || ""} ${name} = ${transaction.functionArgs?.[i] || ""}`,
+      (name, i) =>
+        `${transaction.functionArgTypes?.[i] || ""} ${name} = ${
+          transaction.functionArgs?.[i] !== undefined ? transaction.functionArgs?.[i] : ""
```
rin-st, line 56: `transaction.functionArgs?.[i] ?? ""` ?

packages/nextjs/package.json
```diff
@@ -48,6 +48,7 @@
     "eslint-plugin-prettier": "^4.2.1",
     "postcss": "^8.4.16",
     "prettier": "^2.8.4",
+    "tailwind-scrollbar": "^3.0.4",
```
rin-st, line 51: Not sure about this lib. For me, if a scrollbar is used somewhere in the app, then all the scrollbars should be the same. Let's see what others think.

As I understand, it's impossible with this lib to make all scrolls the same from one place, am I right?

rin-st:
hey @portdeveloper ! Thanks for improvements! Added some comments

packages/nextjs/pages/blockexplorer/transaction/[txHash].tsx
Comment (portdeveloper) [reply packages/nextjs/pages/blockexplorer/transaction/[txHash].tsx:97]: The table in [txHash].tsx page overflows the containers when the viewport width is below 600px. I plan to fix it in another PR since this one introduces a package and changes the page already. 
But maybe we can add it as a commit. Thanks for pointing this out!

packages/nextjs/utils/scaffold-eth/decodeTxData.ts
portdeveloper: Much appreciated!

packages/nextjs/package.json
technophile-04: I think it's fine we use the default browser instead of using `tailwind-scrollbar`, because we don't want to set any extra default styles ourselves and rather leave it to the developers whether they want to customize the scroll bar(if yes then whether they want to do it through pure css or rely on any other library)

I know it looks a bit raw on Windows but I think it's fine, not the worth for addition of a library 😅 🙌

technophile-04:
Tysm @portdeveloper the fix looks great, 

I would rather remove `tailwind-scrollbar` and keep it as browser default because on Mac it doesn't look that bad : 
Screenshot 2023-07-12 at 5 35 47 PM

Always btw love the idea of showing the scroll bar on for function `function` instead of the whole container 🙌

portdeveloper:
Thanks for the feedback @technophile-04 ! Time for me to get a mac I guess lol

technophile-04:
Tysm @portdeveloper !! While testing I noticed a small bug which is present on `main` branch too : 

We are not showing the function name called on `/blockexplorer` page : 
Screenshot 2023-07-13 at 12 39 15 AM

Could you please look into it too ? Also please feel free to tackle it in a different PR if it's too serious and hard 🙌

portdeveloper:
> Tysm @portdeveloper !! While testing I noticed a small bug which is present on `main` branch too :
> 
> We are not showing the function name called on `/blockexplorer` page : Screenshot 2023-07-13 at 12 39 15 AM
> 
> Could you please look into it too ? Also please feel free to tackle it in a different PR if it's too serious and hard 🙌

Fixed by using decodeTransactionData in useFetchBlocks. Somehow we forgot it while migrating to viem. Thanks for noticing!

technophile-04:
> Fixed by using decodeTransactionData in useFetchBlocks. Somehow we forgot it while migrating to viem. Thanks for noticing!

Tysm great catch !!

PR #417, opened 2023-07-12
Fix bug with findFilesRecursiveSync

dgrcode:
## Description
Fix a bug that made `copyExtensionsFiles` work async, so when later code tried to access extension files, they weren't there yet. That caused a bug with `findFilesRecursiveSync`, where files that should've been found weren't returned.

### DIFF
--- a/src/tasks/copy-template-files.ts
+++ b/src/tasks/copy-template-files.ts
@@ -36,7 +36,7 @@ const copyExtensionsFiles = async (
   { extensions }: Options,
   targetDir: string
 ) => {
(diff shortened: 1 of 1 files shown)


carletex:
Nice catch @dgrcode !

Just tried this code and it works:

```ts
findFilesRecursiveSync(targetDir, (path) =>
  path.endsWith(".gitignore-keep")
).map((file) => {
  fs.renameSync(file, file.replace(".gitignore-keep", ".gitignore"));
});
```

We will not be using it for that rename (will be using a template file instead) but wanted to test. (The same code didn't work before)

PR #423, opened 2023-07-14
Use templates instead of renaming files

dgrcode:
## Description

Uses templates to avoid renaming specific files. During this PR we found templates and args needed the extension `.mjs` to work properly with any extension, so the RFC-extensions.md file has also been updated to reflect this point.


technophile-04:
Tysm @dgrcode !! Looks clean and nice 🙌 

I think we forget about converting [`.env-default`](https://github.com/scaffold-eth/scaffold-eth-2/blob/cli/templates/extensions/foundry/packages/foundry/.env-default) -> `.env.template.mjs`

I tested the rest out and works nicely!

dgrcode:
Yes! I forgot about the other file, my bad 🤦

I'll add a commit next time I get to my laptop. Monday morning at the latest.

Good catch! Thanks for pointing it out 🙌

dgrcode:
> I'll add a commit next time I get to my laptop

Done! And tested it works as expected. Thanks again @technophile-04 🙌

carletex (approved):
Good job @dgrcode & @technophile-04 

Tested and everything works perfectly!

PR #425, opened 2023-07-15
CLI/fix frontend network

technophile-04:
## Description : 

It tries to add templating logic for `scaffold.config.ts` because we want to change choose the `targeNetwrok` based on the option selected 

Would love to know if the logic can be improved, I think we can utilize `withDefaults` method but was a bit confused how to use it and would love to learn that so please feel to suggest betterment / clean ups 🙌


templates/base/packages/nextjs/scaffold.config.ts.template.mjs
```diff
@@ -11,7 +11,7 @@ export type ScaffoldConfig = {
 
 const scaffoldConfig = {
   // The network where your DApp lives in
-  targetNetwork: chains.hardhat,
+  targetNetwork: chains.${namedArgs?.default?.[0]?.chainName || "mainnet"},
```
technophile-04, line 14: Maybe this can improved ? 

Added this condition because people might select `None` as option while choosing template 

Is there a better way ? Can we pass some default value to this templating function ?

templates/extensions/foundry/packages/nextjs/scaffold.config.ts.args.mjs
```diff
@@ -0,0 +1 @@
+export default { chainName: "foundry" };
```
technophile-04, line 1: Please feel free suggest if there is an better way to export this 🙌

dgrcode (commented):
I'd recommend using a named export, like
```js
// file.args.mjs
export const chainName = "foundry"
```

Then you'd use it in the template like this, in combination with the utility `withDefaults`:
```js
// file.template.mjs
const contents = ({ chainName }) => `blah blah
  targetNetwork: chains.${chainName[0] || "mainnet"},
blah blah`

export default withDefaults(contents, ["chainName"])
```

Note a few differences with your existing code:
1. I'm using [parameter destructuring](1) (a bit lower in the page, an exampled named "Unpacking properties from objects passed as a function parameter"). That way you can skip drilling like `namedArgs.default[0].chainName`, and just use `chainName[0]` instead.

2. I'm using the utility `withDefaults`, which does two things:
  2.1. Gives you warnings if sent an argument that is not expected
  2.2. If no argument is sent, it just assigns an empty array. Thanks to this, `chainName[0]` would be `undefined` and the value `"mainnet"` used instead.
  Using `withDefaults` also remove the need for optional chaining, because using it warrantees we'll always have a value, even if just an empty array. More thoughts on this as follows.

When I built this utility I didn't have in mind the scenario where we want a default, and the name `withDefault` was just to set `[]` as the default value. However, looking at this use case, it kind of makes sense to use the following signature to set the default value from the template itself:
```js
// file.template.mjs
const content = ({ chainName }) => `... ${chainName[0]} ...`

export default withDefaults(contents, {
  chainName: 'mainnet'
})
```
Meaning: we expect the argument `chainName`, so give a warning if we get something else; also if no arguments are sent for `chainName`, use `'mainnet'` as the only sent argument.

@carletex @technophile-04 thoughts on this API for the `withDefaults`? I also think the name would make much more sense.

I think the resulting value sent to the `content` function should be wrapped in an array, even in the default scenario, to have the same handling when we have multiple args files writing to a template, or in any other scenario.

I think it would be cleaner this way

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#object_destructuring

templates/base/packages/nextjs/scaffold.config.ts.template.mjs
dgrcode: I've given a general answer for this question and your other comment. I'll mark the two as resolved to avoid having multiple threads, but feel free to continue the conversation in the main thread.

dgrcode:
☝️ just opened a quick draft https://github.com/scaffold-eth/scaffold-eth-2/pull/430 to show what would change if we decide to go with the `withDefaults` API change discussed above.

carletex:
Thanks for this @technophile-04!! And also @dgrcode thanks for the detailed explanation.

I definitely prefer this option. The API makes more sense to me (and I like having a default value, where the `withDefaults` name makes even more sense)

```ts
export default withDefaults(contents, {
  chainName: 'mainnet'
}
```

I guess we could also decide not to have a default value (null, undefined?)

Maybe we can test & merge #430 and update the code here?

cc @technophile-04

dgrcode:
I agree the changes in #430 make the code nicer in general.

As per the argument not having a default value, I guess we can still declare explicitly we want an empty string. Just for consistency.

Something like
```js
export default withDefaults(contents, {
  chainName: 'mainnet',
  empty: ''
}
```

The problem with relying on `null` and `undefined` is that they get type coerced into a string, and become the strings `"null"` and `"undefined"` respectively. We can still catch those two values and replace with an empty string if we want, but I think being explicit here could show better what the template expects.

Thoughts?

In the meantime I'll give #430 a manual test and check things work as expected. Also, a thought that comes to mind regarding this: how do we feel about adding automatic tests to utilities like this one? With something like jest it'd be pretty straight forward, but still it'd take waaay less time to manual test for one offs like this one.

technophile-04:
Tysm @dgrcode for the great explanation love it and everything makes sense !!! 

I like the idea of #430 and it's good! 

Btw I think this PR has very less changes, I was thinking maybe we do these changes in #430 itself and close this PR because with these changes in #430  we can also test it nicely with a live example... what do you guys say?  @carletex , @dgrcode 

> how do we feel about adding automatic tests to utilities like this one? With something like jest it'd be pretty straight forward, but still it'd take waaay less time to manual test for one offs like this one

Lol I have no good idea what's a setup for `tests` looks like but if it's minimal I think this makes a lot of sense because while playing with PR I was going through the whole `CLI` workflow to see how some `utility` functions were working

dgrcode:
@technophile-04 I'm actually using this specific template/args for testing, so it will come out of the box. I just pushed the changes with those files included.

I'm just finishing to update the RFC document explaining how to use `withDefaults` and I think after that it should be ready to merge.

carletex:
Fixed on #430 !

Thanks @dgrcode and @technophile-04 for the discussion :)

PR #430, opened 2023-07-18
Update withDefaults API to receive an object

dgrcode:
## Description

From https://github.com/scaffold-eth/scaffold-eth-2/pull/425, I left [a comment](https://github.com/scaffold-eth/scaffold-eth-2/pull/425#pullrequestreview-1534668142) where I mention an API change for `withDefaults`. This is what it'd take, although I haven't tested it yet.

If we think this makes sense, I'll run some tests to make sure this works as expected and would also update the RFC.

### DIFF
--- a/RFC-extensions.md
+++ b/RFC-extensions.md
@@ -129,9 +129,29 @@ When a package enforces commonjs imports, our templates created within those pac
 
 ### Default values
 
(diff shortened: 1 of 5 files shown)


dgrcode:
right now, if using an argument where a) a default was not provided and b) no args is sent, it'll become the string "undefined". That's because its default value is not sent, so it's `undefined`, and then it gets coerced into a string by JS. I'll see if there's an easy win here, or just set all defaults to an empty string if none is sent.

dgrcode:
Just added more info to the RFC to take into account the last comment I wrote here about args becoming "undefined" if not declared explicitly when using `withDefaults`.

PR #435, opened 2023-07-20
Change foundryup command

carletex:
`foundryup --version` make an http request and it can be slow. `foundry -h` solves the same problem and it's instant.


src/utils/render-outro-message.ts
```diff
@@ -1,5 +1,5 @@
-import chalk from "chalk";
 import type { Options } from "../types";
+import chalk from "chalk";
```
carletex, line 2: We need a prettier/linter in the CLI project :D

This was automatic for me since It's still pointing to the prettier/lint/config of SE-2

PR #436, opened 2023-07-20
CLI dev mode

dgrcode:
## Description

Add dev mode for the CLI.

The goal is to improve the DX when contributing changes to anything in the "templates/" folder.

### How to use
The way this should be used is by running `yarn cli project_name --dev`. By using the dev flag, the CLI turns into dev mode.

### What it does
When in dev mode, we create symlinks instead of copying files from "templates/". That way, the user would see as if the instance project (the project resulting from running the CLI) had been created. The user can also modify the files within the instance project, and see changes in realtime with the normal tooling (`yarn dev`, etc). However, since the files are actually symlinked to the source files within "templates/", they are actually modifying the source files.

Once the user is happy with the changes, they can run `git diff`, and git actually understand that the files changed were the source ones, so there's no extra copy/paste required to update source files.

### Caveat: special files
In the normal CLI mode, there are a few files that are not copied, but generated from the source files instead. Those files are `package.json` and `**.templates.mjs`.

Even though there are ways to automate those changes as well, it's quite complex to update the right files in the source. Because this is a proof of concept, we won't do any automation, and these special files have to be updated manually.

However, we'll generate sibling files to them with information that aims to make those manual changes easier. Think of those files as a human readable sourcemap for generated files. These files will have the name `resulting_file.ext.dev`, so they should live next to the resulting files, only with the suffix `.dev`.

## Tasks
 - [x] Add dev flag
 - [x] Create symlinks instead of copying when the dev flag has been sent
 - [x] Add relevant content to `package.json.dev` files
 - [x] Add relevant content to `**.dev` files generated out of templates.


dgrcode:
force pushed after rebasing on current cli, to get the templates from #430

dgrcode:
@carletex @technophile-04 I've given this a test locally and things seem to be working as expected.

There are improvements around the .dev file, but we can make them as we need them, which might be never haha.

carletex:
This is looking promising!! Thanks for the detailed explanation @dgrcode 

Still testing, but noticed a couple of things:

> @carletex @technophile-04 I've given this a test locally and things seem to be working as expected.

It seems that the `package.json` files inside hardhat/foundry are not being copied. Noticed when tried to run `yarn chain` and it failed `Usage Error: Workspace '@se-2/hardhat' not found`. Not sure how it can work for you, maybe it's specific to my OS?

> However, we'll generate sibling files to them with information that aims to make those manual changes easier. Think of those files as a human readable sourcemap for generated files. These files will have the name resulting_file.ext.dev, so they should live next to the resulting files, only with the suffix .dev.

It seems that these files are not being generated for me (again, my OS?).

dgrcode:
hmm interesting. I'll have a look at this after the issue with windows

dgrcode:
alright so there was a bug where the package.json were not created. I'm not sure what I tested. Probably tested before introducing the bug, and then only tested the symlinks and not the project.

Anyways, the interesting part is that after the package.json are created again, the package "@se-2/hardhat" exists, but npm says it's missing.

It also looks like the package.json from the root folder is symlinked (where it shouldn't), which might be causing npm to look at the packages under the source's "template/base/", instead of the instance's "packages/". It sounds weird that it behaves that way, but I guess there are a few things that can work funny using symlinks.

I'll see if not using symlinks for package.json files fixes it.

dgrcode:
I just pushed a couple of commits that made some packages "invisible", but I'm still having funky behavior when trying to run next

technophile-04:
Just tested it out and works like a charm ✨ !!! Tysm @dgrcode !!

I think we missed something during[a6610a5](https://github.com/scaffold-eth/scaffold-eth-2/pull/436/commits/a6610a5ab7ecbe59ce7d86b31aefff56f5a56d3a) so it breaks... added a [notion doc link]() in TG chat but if you checkout to [d5860a9](https://github.com/scaffold-eth/scaffold-eth-2/pull/436/commits/d5860a9a320eee5f7e7387fef98c2b7a67ae122b) everything works there!


### Some basic testing which I performed : 

- [X] Tried all the templates with `--dev` mode and everyone generates nicely, also all the commands like `yarn chai` etc work!  

- [X] I tried changing `YourContract.sol` and other frontend files and they were getting reflected in CLI templates files as expected, love it 🔥!

- [X] Tried adding a new package `react-icons` in frontend and it works as expected nothing breaks. As discussed in "Caveat: special files" of https://github.com/scaffold-eth/scaffold-eth-2/pull/436#issue-1813454800 I think it's ok we can update `package.json` manually since it will be very less changes 🙌

- [X] Also tested with `.template.mjs` checkout this branch -> [cli-dev-test](https://github.com/scaffold-eth/scaffold-eth-2/tree/cli-dev-test), here I tried templating `index.tsx` which updates page Heading with package name selected ... And it works nicely too !!! 

Just one question on generated `index.tsx.dev` file (this was generated for me) : 

```js
--- TEMPLATE FILE
templates/base/packages/nextjs/pages/index.tsx.template.mjs


--- ARGS FILES
	- undefined


--- RESULTING ARGS
	- pkgName:	[Hardhat]

```

Regarding `ARGS FILES` section I `undefined` was that expected? Or I should have got the `index.tsx.args.mjs` filePath there ? (I might be completely wrong here!!)


Nevertheless, everything works great !!! Let's fix the latest merge issue and maybe update CONTRIBUTING.MD a bit, Tysm Dani this is really nice !!!

dgrcode:
Thanks for the tests Shiv! 🙌 

> I think we missed something during[a6610a5](https://github.com/scaffold-eth/scaffold-eth-2/pull/436/commits/a6610a5ab7ecbe59ce7d86b31aefff56f5a56d3a) so it breaks

I'll check what's going on an try to push a fix.

> Regarding ARGS FILES section I undefined was that expected?

Nope, you're right. There we should have a path to the file instead of `undefined`. That should be easy to fix.

I'll see if I can get both those points done today.

Thanks again!

technophile-04:
The way I usually run CLI is I have two panes split : 
1. one running `yarn dev` so that it auto compiles and put the changes in `dist` directory 
2. `yarn cli` to test cli 

Screenshot 2023-08-08 at 9 58 18 PM

--- 

Additionally, if you try to do `yarn build` I get this error : 

Screenshot 2023-08-08 at 10 00 00 PM

---

> 2. I move to the parent folder cd ..
> 3. I run the cli from there node scaffold-eth-2/bin/create-dapp-se2.js test-dev --dev

Maybe I think you have the prvious build version which you are using here, 

could you do `yarn build` and then do `cd ..` and `node scaffold-eth-2/bin/create-dapp-se2.js test-dev --dev` ?

dgrcode:
oh yes, my bad! I forgot to build after pulling the latest changes!  😅🤦

I'll build and see what I find

dgrcode:
yep, reproducing now 😓

dgrcode:
Alright, found the issue with the `argsFilesPath`. I changed the variable name in the `cli` branch, and then used that variable (with old name) in the `cli-dev-mode` branch. The merge didn't find any conflict because that happened in different lines. Sneaky. [[reference](https://github.com/scaffold-eth/scaffold-eth-2/compare/686db1a..b5e3c86#diff-b466febb1b5c766b13a72b87529308ab7a50ee564c759d4a5bc7f91bd55a23daL134-R137)]
image

I think this has also fixed the "ARGS FILES" section being `undefined` in the `*.dev` files.

After fixing it I've tested:
 - Without framework. It works. I initially got an hydration error, but after deleting and restarting, it didn't happen again.
 - With hardhat. Working without any issue.
 - With foundry. Working without any issue.

I can't reproduce the .DS_Store issue. Maybe because I'm on a Mac?

I'll commit the fix for the rename. Could you check again if the .DS_Store went away in the next commit?

Thanks!

technophile-04:
Tysm @dgrcode, I just tested and works great even with `.DS_Store` files present 🙌
                                                                        
Umm I still got "ARGS FILES" section `undefined`, you can check it in `scaffold.config.ts.dev` since there we are already using templating logic            

Also, from https://github.com/scaffold-eth/scaffold-eth-2/pull/436#issuecomment-1648424258 this is still there : 
Screenshot 2023-08-09 at 4 34 48 AM

I see this dosen't cause any harm and everything works correctly but I found doing this solves this issue : 

```ts
import { promises } from "fs";

const { mkdir, link } = promises;
```
checkout this [stackoverflow link](https://stackoverflow.com/a/68731292/16480510)


I think are pretty close to merging this beast !!

---

cc @Pabl0cks whenever you have time could you just try this on `windows`  🙌

dgrcode:
Regarding "ARGS FILES" section, I just created the dev mode projects with hardhat, foundry, and no framework. I'm not getting `undefined`. My results:

<details>
  <summary>Hardhat</summary>

```
--- TEMPLATE FILE
templates/base/packages/nextjs/scaffold.config.ts.template.mjs


--- ARGS FILES
	- templates/extensions/hardhat/packages/nextjs/scaffold.config.ts.args.mjs


--- RESULTING ARGS
	- chainName:	[hardhat]
```
</details>

<details>
  <summary>Foundry</summary>

```
--- TEMPLATE FILE
templates/base/packages/nextjs/scaffold.config.ts.template.mjs


--- ARGS FILES
	- templates/extensions/foundry/packages/nextjs/scaffold.config.ts.args.mjs


--- RESULTING ARGS
	- chainName:	[foundry]

```
</details>

<details>
  <summary>No framework</summary>

```
--- TEMPLATE FILE
templates/base/packages/nextjs/scaffold.config.ts.template.mjs


--- ARGS FILES
(no args files writing to the template)


--- RESULTING ARGS
(no args sent for the template)

```
</details>

Regarding the unresolved dependency, I completely missed that one. I'll implement the fix you propose, thanks! 👍

damianmarti:
I tested it and everything is working fine!!

Anyway, I think that since this dev feature will be used mostly by us at least at the beginning, we can use it carefully, trying to find new bugs and new improvements to make.

I think we should update the CONTRIBUTING.MD file with clear instructions on how to use the cli in dev mode, test the cli on Windows, and if everything goes well, we can merge it!

Great work guys!!!

src/tasks/copy-template-files.ts
```diff
@@ -189,6 +256,32 @@ const processTemplatedFiles = async (
...
+      if (isDev) {
+        const hasCombinedArgs = Object.keys(combinedArgs).length > 0
+        const hasArgsPaths = argsFileUrls.length > 0
+        const devOutput = `--- TEMPLATE FILE
+templates/${templateFileDescriptor.source}${templateFileDescriptor.relativePath}
+
+
+--- ARGS FILES
+${hasArgsPaths
+  ? argsFileUrls.map(url => `\t- ${url.split('scaffold-eth-2/')[1]}`).join('\n')
```
technophile-04, line 269: > Regarding "ARGS FILES" section, I just created the dev mode projects with hardhat, foundry, and no framework. I'm not getting undefined. My results:

I think I found the bug since we are using here: "scaffold-eth-2/"  to split and I don't  it name "scaffold-eth-2" locally its named "cli" so it giving me undefined 

I tried logging the whole path and check this out : 
Screenshot 2023-08-10 at 9 15 26 PM


Here is my propose solution : 

```ts
argsFileUrls.map((url) => `\t- ${path.join("templates", url.split("templates")[1])}`).join("\n")

```

Splitting on the basis of `"templates`" because irrespective of how people name it locally we will have "templates" always in the path since `.args.mjs` will only be present in "templates"


But feel free update the logic 🙌

Pabl0cks:
I tried CLI dev mode with Windows and it worked nice. 

Tried selecting both Hardhat and Foundry, changing stuff on a few pages and into the contract and all got reflected on templates/ from the source when I did `git diff` or checked in GitHub Desktop.

Good job! 🙌

src/tasks/copy-template-files.ts
dgrcode: oh that makes sense. I like your proposal. I'll make a commit as soon as I find the time.

Thanks!

technophile-04:
> I tried CLI dev mode with Windows and it worked nice.

Awesome, tysm 🙌 !! 

I think we are just remaining on : 
- [X] Windows testing
- [ ] Mentioning this CONTRIBUTING.MD 
- [ ]  https://github.com/scaffold-eth/scaffold-eth-2/pull/436#discussion_r1290344558

We can update CONTRIBUTING.MD in different PR but since we already published a version after #483, we are not in a rush  🙌

src/tasks/copy-template-files.ts
dgrcode: Fixed!

dgrcode:
We moved CONTRIBUTING to "templates/base/", but that's going to be the CONTRIBUTING of the project created by people. Should we have a different CONTRIBUTING at the root? or maybe a copy so it's both in our project and the project created by users?

I'm thinking the same about the root's README.md. Should we rewrite both the README and CONTRIBUTING taking RFC-extensions into account?

technophile-04:
> We moved CONTRIBUTING to "templates/base/", but that's going to be the CONTRIBUTING of the project created by people. Should we have a different CONTRIBUTING at the root? or maybe a copy so it's both in our project and the project created by users?

We are probably gonna remove "template/base/CONTRIBUTING.md" since people can create their own according to their needs, So I think we should create our CONTRIBUTING.md at the root 

> I'm thinking the same about the root's README.md. Should we rewrite both the README and CONTRIBUTING taking RFC-extensions into account?

I was thinking something like this : 

#### Root `README.md` 
This will contain a small description of SE-2 and Quickstart similar to what we have in our `main` README.md and in the end, it will point to CONTRIBUTING.md. 

#### Root `CONTIRBUTING.md`
1. Some description like it's under active development, things to keep in mind before PR/issue like some points from  `main` CONTRIBUTING.md 
2. How to run the project locally like requirements 
3. A small desc about `--dev` and how to use it, a summarised version of https://github.com/scaffold-eth/scaffold-eth-2/pull/436#issue-1813454800
4. Also in the end or in the middle we can point them contributors/developer's guide. 

#### RFC-extension renamed to Conitburotrs/developer_guide.md : 
1. I see RFC-extension as a nice reference book for CLI it has everything in depth and internal working/structure, so I prefer not removing it maybe just updating it a bit. 
2. I think this has all stuff already nicely documented and we can keep updating it as we move into future 🙌, We can also link some of its header `Templated Files` etc in CONTRIBUTING.md wherever required 


Was just thinking out loud and would love to hear other's opinion 🙌

--- 

For this PR : 

Maybe we just create `CONTRIBUTING.md` at the root and just have a section "Getting Started" which tells you : 

1. Clone and cd into repo 
2. start a watcher by running `yarn dev` 
3. In new terminal run `yarn cli project-name --dev` mode 
After this point we just summarise https://github.com/scaffold-eth/scaffold-eth-2/pull/436#issue-1813454800 like what it does and how to make changes and and commit. 

We can cover other sections of  `CONTIRBUTING.md` other PR, what do you think?

dgrcode:
Yep that sounds great Shiv 🙌 

I'll put something together and push a commit

dgrcode:
I added an initial version. To see the changes made to the copied version of CONTRIBUTING.md, I recommend looking at the changes from the two last commits together: https://github.com/scaffold-eth/scaffold-eth-2/compare/bbff792..cli-dev-mode

I'll update the links to the resulting urls after having pushed them to the remote

CONTRIBUTING.md
```diff
@@ -0,0 +1,91 @@
...
+
+- We may ask questions, request additional information or ask for changes to be made before a PR can be merged. Please note that these are to make the PR clear for everyone involved and aims to create a frictionless interaction process.
+- As you update your PR and apply changes, mark each conversation resolved.
+
+Once the PR is approved, we'll "squash-and-merge" to keep the git commit history clean.
+
+## Developer Guide
+
+You can find a detailed guide on `contributors/DEVELOPER-GUIDE.md` (TODO add link once published
+)
```
dgrcode, line 91:
```suggestion
You can find a detailed guide on [`contributors/DEVELOPER-GUIDE.md`](https://github.com/scaffold-eth/scaffold-eth-2/blob/cli-dev-mode/contributors/DEVELOPER-GUIDE.md)
```

contributors/DEVELOPER-GUIDE.md
```diff
@@ -0,0 +1,60 @@
...
+node path-to-this-repo/bin/create-dapp-se2.js [project-name --flags]
+```
+
+You can send any option or flag to the CLI command, except for the project name, which is hardcoded to be "test-cli" in the package.json script. For example, a handy command is `yarn cli -s` to tell the CLI to skip installing dependencies.
+
+## Changes to the Resulting Project
+
+The source files for the instance projects can be found under `templates/`. You'll see there are two folders there: `base/` and `extensions/`. The `base/` folder has the source files that will be present in all instances, whereas `extensions/` hold the source files that will be added or not to the instances based on the user choices within the CLI tool.
+
+It's highly recommended that you go through RFC-extensions.md (TODO add link once published to the right folder) to understand the template API to create extensions. We use a custom template API to allow extensions to modify any file inside the `templates/` folder. While flexible and powerful, it requires developers to understand how it works. It's JS based, so there's no new technology needed to understand and use it.
```
dgrcode, line 37:
```suggestion
It's highly recommended that you go through [RFC-extensions.md](https://github.com/scaffold-eth/scaffold-eth-2/blob/cli-dev-mode/contributors/RFC-extensions.md) to understand the template API to create extensions. We use a custom template API to allow extensions to modify any file inside the `templates/` folder. While flexible and powerful, it requires developers to understand how it works. It's JS based, so there's no new technology needed to understand and use it.
```

dgrcode (commented):
I guess we'll have to update those links as we merge branches

CONTRIBUTING.md
```diff
@@ -0,0 +1,90 @@
...
+Once you submit your PR:
+
+- We may ask questions, request additional information or ask for changes to be made before a PR can be merged. Please note that these are to make the PR clear for everyone involved and aims to create a frictionless interaction process.
+- As you update your PR and apply changes, mark each conversation resolved.
+
+Once the PR is approved, we'll "squash-and-merge" to keep the git commit history clean.
+
+## Developer Guide
+
+You can find a detailed guide on [`contributors/DEVELOPER-GUIDE.md`](https://github.com/scaffold-eth/scaffold-eth-2/blob/cli-dev-mode/contributors/DEVELOPER-GUIDE.md)
```
dgrcode, line 90:
```suggestion
You can find a detailed guide on [`contributors/DEVELOPER-GUIDE.md`](contributors/DEVELOPER-GUIDE.md)

```

contributors/DEVELOPER-GUIDE.md
```diff
@@ -0,0 +1,60 @@
...
+node path-to-this-repo/bin/create-dapp-se2.js [project-name --flags]
+```
+
+You can send any option or flag to the CLI command, except for the project name, which is hardcoded to be "test-cli" in the package.json script. For example, a handy command is `yarn cli -s` to tell the CLI to skip installing dependencies.
+
+## Changes to the Resulting Project
+
+The source files for the instance projects can be found under `templates/`. You'll see there are two folders there: `base/` and `extensions/`. The `base/` folder has the source files that will be present in all instances, whereas `extensions/` hold the source files that will be added or not to the instances based on the user choices within the CLI tool.
+
+It's highly recommended that you go through [RFC-extensions.md](https://github.com/scaffold-eth/scaffold-eth-2/blob/cli-dev-mode/contributors/RFC-extensions.md) to understand the template API to create extensions. We use a custom template API to allow extensions to modify any file inside the `templates/` folder. While flexible and powerful, it requires developers to understand how it works. It's JS based, so there's no new technology needed to understand and use it.
```
dgrcode, line 37:
```suggestion
It's highly recommended that you go through [RFC-extensions.md](contributors/RFC-extensions.md) to understand the template API to create extensions. We use a custom template API to allow extensions to modify any file inside the `templates/` folder. While flexible and powerful, it requires developers to understand how it works. It's JS based, so there's no new technology needed to understand and use it.
```

dgrcode (commented):
It looks like github supports relative links!

contributors/DEVELOPER-GUIDE.md
```diff
@@ -0,0 +1,60 @@
...
+node path-to-this-repo/bin/create-dapp-se2.js [project-name --flags]
+```
+
+You can send any option or flag to the CLI command, except for the project name, which is hardcoded to be "test-cli" in the package.json script. For example, a handy command is `yarn cli -s` to tell the CLI to skip installing dependencies.
+
+## Changes to the Resulting Project
+
+The source files for the instance projects can be found under `templates/`. You'll see there are two folders there: `base/` and `extensions/`. The `base/` folder has the source files that will be present in all instances, whereas `extensions/` hold the source files that will be added or not to the instances based on the user choices within the CLI tool.
+
+It's highly recommended that you go through [RFC-extensions.md](contributors/RFC-extensions.md) to understand the template API to create extensions. We use a custom template API to allow extensions to modify any file inside the `templates/` folder. While flexible and powerful, it requires developers to understand how it works. It's JS based, so there's no new technology needed to understand and use it.
```
dgrcode, line 37:
```suggestion
It's highly recommended that you go through [RFC-extensions.md](RFC-extensions.md) to understand the template API to create extensions. We use a custom template API to allow extensions to modify any file inside the `templates/` folder. While flexible and powerful, it requires developers to understand how it works. It's JS based, so there's no new technology needed to understand and use it.
```

technophile-04 (approved):
Tysm @dgrcode for updating `.md` files, Love it as always !!!!!

I have just removed hardcoded `test-cli` from `package.json` leaving it upon on developer add path while testing, because I like to prefer creating "instance project" out of this repo 😅 just to be safe and avoid committing `test-cli` directory : 
<details>
<summary></summary>

Screenshot 2023-08-19 at 7 13 40 AM

</details>

We could have added `test-cli` in `.gitignore` but I think its fine if we let the developer decide 🙌, also update Developer_Guide reflecting these changes  at [8945977](https://github.com/scaffold-eth/scaffold-eth-2/pull/436/commits/89459776bdf641d16debc6f5ea836998b8046da6). But we can always iterate on it as well as other `.md` files later 


Mergin this, Thanks again !!! Also thanks to everyone for testing 🙌

dgrcode:
Yeah makes sense. I didn't think about that use case. Sounds good then! 🙌

PR #439, opened 2023-07-20
Add .env.example for nextjs env variables

Pabl0cks:
## Description

This PR adds .env.example file for nextjs environment variables, like we already have for hardhat.

Also adds some extra information on how we are managing API Keys in config/env files in both hardhat and nextjs .env.example files, so users can have a better understanding.

## Additional Information

Your ENS/address: pabl0cks.eth

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,11 +152,11 @@ The configuration parameters are described below, make sure to update the values
 
 - **alchemyApiKey**  
   Default Alchemy API key from Scaffold ETH 2 for local testing purposes.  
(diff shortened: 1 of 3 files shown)


carletex:
Great job @Pabl0cks !!

I pushed some little tweaks.

Also noticed that we use `NEXT_PUBLIC_DEPLOY_BLOCK` to improve performance in the `useScaffoldEventHistory` hook. I feel that we could move that env var from `ContractData.tsx` to the hook itself. Like use fromBlock if passed, if not use the env var, if not use 0. But we can discuss that in another issue. cc @technophile-04 @rin-st

carletex (approved):

Pabl0cks:
I think it's a great idea! So you don't have to manually set the `fromBlock` or add the `process.env.NEXT_PUBLIC_DEPLOY_BLOCK ?` logic in every `useScaffoldEventHistory` you use.

Issue #438, opened 2023-07-20
CLI: Update README


carletex:
Let's create the first version of the README for the CLI (PR against cli branch)

Main ideas:
- Change install methods (npx instead of cloning).
- Add foundry instructions (where it says hardhat / hardhat.config, etc)

Pabl0cks:
First version of CLI and Foundry Readme already merged into `cli` branch

PR #442, opened 2023-07-24
Adapt README to CLI and foundry

Pabl0cks:
Solves #438 

I've created another issue to keep iterating on this after we merge the PR and have a initial README for `cli` branch => #472


README.md
```diff
@@ -1,8 +1,11 @@
+> ⚠️ This branch is under active development.  
+> If you find any bug, please report as [issue](https://github.com/scaffold-eth/scaffold-eth-2/issues) or send a message in [🏗 scaffold-eth developers chat](https://t.me/joinchat/F7nCRK3kI93PoCOk)
+
 # 🏗 Scaffold-ETH 2
 
 🧪 An open-source, up-to-date toolkit for building decentralized applications (dapps) on the Ethereum blockchain. It's designed to make it easier for developers to create and deploy smart contracts and build user interfaces that interact with those contracts.
 
-⚙️ Built using NextJS, RainbowKit, Hardhat, Wagmi, and Typescript.
+⚙️ Built using NextJS, RainbowKit, Hardhat/Foundry, Wagmi, and Typescript.
```
rin-st, line 8: I think "Hardhat, Foundry," is better

Pabl0cks: Thx for the feedback! Updated! :)

Pabl0cks:
Moving this PR to review, could you take a look to it @dgrcode @technophile-04 ?

If you feel something is missing for the initial version, let me know and I'll add it to this PR 🙌

You can also add more sections that would be nice to add in the future in #472.

README.md
carletex, line 1: We might want to add "npx" as a dependency on the Requirements section

README.md
```diff
@@ -34,29 +37,40 @@ Before you begin, you need to install the following tools:
...
 
 ```
 yarn deploy
 ```
 
-This command deploys a test smart contract to the local network. The contract is located in `packages/hardhat/contracts` and can be modified to suit your needs. The `yarn deploy` command uses the deploy script located in `packages/hardhat/deploy` to deploy the contract to the network. You can also customize the deploy script.
+This command deploys a test smart contract to the local network. The contract can be modified to suit your needs. Is located in:
+
+- Hardhat => `packages/hardhat/contracts`
+- Foundry => `packages/foundry/src`
```
carletex, line 68: This is going to be `packages/foundry/contracts` when we merge this: https://github.com/scaffold-eth/scaffold-eth-2/pull/473

README.md
```diff
@@ -66,21 +80,25 @@ yarn start
 
 Visit your app on: `http://localhost:3000`. You can interact with your smart contract using the contract component or the example ui in the frontend. You can tweak the app config in `packages/nextjs/scaffold.config.ts`.
 
-Run smart contract test with `yarn hardhat:test`
+Run smart contract test with `yarn hardhat:test` or `yarn foundry:test` depending of your solidity framework.
 
-- Edit your smart contract `YourContract.sol` in `packages/hardhat/contracts`
+- Edit your smart contract:
+  - Hardhat => `YourContract.sol` in `packages/hardhat/contracts`
+  - Foundry => `YourContract.sol` in `packages/foundry/src`
```
carletex, line 87: same https://github.com/scaffold-eth/scaffold-eth-2/pull/442/files#r1281921252

README.md
```diff
@@ -66,21 +80,25 @@ yarn start
...
+  - Foundry => `packages/foundry/script`
 
 ## Deploying your Smart Contracts to a Live Network
 
 Once you are ready to deploy your smart contracts, there are a few things you need to adjust.
 
 1. Select the network
 
-By default, `yarn deploy` will deploy the contract to the local network. You can change the defaultNetwork in `packages/hardhat/hardhat.config.ts.` You could also simply run `yarn deploy --network target_network` to deploy to another network.
+By default, `yarn deploy` will deploy the contract to the local network. You can change the defaultNetwork in `packages/hardhat/hardhat.config.ts.` or `packages/foundry/foundry.toml` depending if you are using Hardhat or Foundry. You could also simply run `yarn deploy --network target_network` to deploy to another network.
```
technophile-04, line 99: Actually, there is a slight change in the deploy command for the `foundry` : 

We directly do `yarn deploy target_network` without specifying `--network`

eg. `yarn deploy sepolia`

README.md
```diff
@@ -66,21 +80,25 @@ yarn start
...
 
-Check the `hardhat.config.ts` for the networks that are pre-configured. You can also add other network settings to the `hardhat.config.ts` file. Here are the [Alchemy docs](https://docs.alchemy.com/docs/how-to-add-alchemy-rpc-endpoints-to-metamask) for information on specific networks.
+Check the `hardhat.config.ts` / `foundry.toml` for the networks that are pre-configured. You can also add other network settings to the `hardhat.config.ts` / `foundry.toml` file. Here are the [Alchemy docs](https://docs.alchemy.com/docs/how-to-add-alchemy-rpc-endpoints-to-metamask) for information on specific networks.
 
 Example: To deploy the contract to the Sepolia network, run the command below:
```
technophile-04, line 103: Same here as https://github.com/scaffold-eth/scaffold-eth-2/pull/442/files#r1282162633

We should also add here example here for `foundry` : 

`yarn deploy sepolia`

README.md
Pabl0cks: Sure! Forgot about it because I think it got auto-installed for me in the first execution. 

Is it always like that? Getting prompt if you want to install it, and just accept it, or we should link an [URL](https://www.npmjs.com/package/npx) in the README to install it manually.

This is in their README, but I'm not sure if that means npx itself always get auto-installed:
> By default, npx will check whether <command> exists in $PATH, or in the local project binaries, and execute that. If <command> is not found, it will be installed prior to execution.

carletex: > By default, npx will check whether exists in $PATH, or in the local project binaries, and execute that. If is not found, it will be installed prior to execution.

This refers to the package/command that you install **with npx** (like create-eth :D), not the npx package itself.

I think you need to install it globally, as with any other npm package (not sure how you got it auto-installed)

carletex: But it seems that it's installed by default with npm now!

> From npm@5.2.0, npm started installing a new binary alongside the usual npm called npx.

So maybe we don't need to add it (npm is also included with node....)

README.md
Pabl0cks: Thanks! Just pushed the update 🙌

README.md
Pabl0cks: Updated!

README.md
Pabl0cks: Updated in both places 👌

Pabl0cks:
Thanks for the review @carletex and @technophile-04! ♥ 

I've just pushed the tweaks on the README, let's wait for @dgrcode review to see if I have do some final tweaks 😊

README.md
carletex: This will change if we merge this tho! 

https://github.com/scaffold-eth/scaffold-eth-2/pull/475

It'd be the same as in hardhat

README.md
```diff
@@ -34,29 +37,40 @@ Before you begin, you need to install the following tools:
...
+1. Install from NPM Registry and follow the CLI instructions.
 
 ```
-git clone https://github.com/scaffold-eth/scaffold-eth-2.git
-cd scaffold-eth-2
-yarn install
+npx create-eth@latest
 ```
 
+> 💬 Hint: If you choose Foundry as solidity framework in the CLI, you must have Foundryup previously installed in your machine. Checkout: [Getfoundry.sh](https://getfoundry.sh)
```
dgrcode, line 46: > you must have Foundryup previously installed in your machine

I'd replace this with something like "you'll also need Foundryup installed in your machine". The current wording makes me think Foundryup is a requirement needed _before_ using the cli, but it can actually be installed after using the cli

README.md
```diff
@@ -66,27 +80,30 @@ yarn start
 
 Visit your app on: `http://localhost:3000`. You can interact with your smart contract using the contract component or the example ui in the frontend. You can tweak the app config in `packages/nextjs/scaffold.config.ts`.
 
-Run smart contract test with `yarn hardhat:test`
+Run smart contract test with `yarn hardhat:test` or `yarn foundry:test` depending of your solidity framework.
```
dgrcode, line 83: This is more a dev comment (rather than a doc comment), but should we rename all these scripts as `yarn chain:*` so they don't change with the framework selected?

cc @carletex

README.md
```diff
@@ -66,27 +80,30 @@ yarn start
...
+  - Foundry => `packages/foundry/script`
 
 ## Deploying your Smart Contracts to a Live Network
 
 Once you are ready to deploy your smart contracts, there are a few things you need to adjust.
 
 1. Select the network
 
-By default, `yarn deploy` will deploy the contract to the local network. You can change the defaultNetwork in `packages/hardhat/hardhat.config.ts.` You could also simply run `yarn deploy --network target_network` to deploy to another network.
+By default, `yarn deploy` will deploy the contract to the local network. You can change the defaultNetwork in `packages/hardhat/hardhat.config.ts.` or `packages/foundry/foundry.toml` depending if you are using Hardhat or Foundry. You could also simply run `yarn deploy --network target_network` to deploy to another network using Hardhat, or run `yarn deploy target_network` if you use Foundry.
```
dgrcode, line 99: this is also more "dev" than "doc", but thinking about consistency between different frameworks, should we create a wrapper script that takes the same shape for both frameworks and then makes the call with the right attributes based on the framework used?

So ideally a user would do:
```bash
yarn deploy target_network
```

And that runs a script that would be roughly something like:
```text
// pseudocode
if framework === hardhat:
  run `yarn deploy --network ${target_network}`
else if framework === foundry:
  run `yarn deploy ${target_network}`
```

cc @carletex

dgrcode (commented):
When we talk about `yarn chain` and `yarn deploy` we're assuming people will choose one of the frameworks, but they can select "None" if they didn't want to use a local blockchain.

Not sure where this info should go. Thoughts?

@carletex I've cc'd you in two of the comments that I think are more like a design thing. We could open issues to work on those if we wanted to implement why I suggest.

README.md
carletex: This will be fixed here: https://github.com/scaffold-eth/scaffold-eth-2/pull/475

Both hardhat/foundry will have the same signature

dgrcode: awesome! 🙌

README.md
Pabl0cks: Thx for the review, I've just pushed the update on that explanation 🙌

README.md
technophile-04: The command for both are the same 🙌 let's not forgot to update it  😅

README.md
technophile-04, line 99: Once we merge #483 

The command for both are same 🙌 lets not forgot to update it

README.md
```diff
@@ -66,27 +80,30 @@ yarn start
...
-Check the `hardhat.config.ts` for the networks that are pre-configured. You can also add other network settings to the `hardhat.config.ts` file. Here are the [Alchemy docs](https://docs.alchemy.com/docs/how-to-add-alchemy-rpc-endpoints-to-metamask) for information on specific networks.
+Check the `hardhat.config.ts` / `foundry.toml` for the networks that are pre-configured. You can also add other network settings to the `hardhat.config.ts` / `foundry.toml` file. Here are the [Alchemy docs](https://docs.alchemy.com/docs/how-to-add-alchemy-rpc-endpoints-to-metamask) for information on specific networks.
 
 Example: To deploy the contract to the Sepolia network, run the command below:
 
-```
-yarn deploy --network sepolia
-```
+- Hardhat => `yarn deploy --network sepolia`
+- Foundry => `yarn deploy sepolia`
```
technophile-04, line 106: Same here after #483 they both will be same 🙌

Pabl0cks:
I've just pushed the `yarn deploy --network target_network` for both hardhat and foundry, we can merge the Readme if you see it OK when #483 is merged.

technophile-04:
Looking great Pablo !! Just one thing I just noticed 😅 : 

For verify section : 
Screenshot 2023-08-12 at 8 50 17 PM


There is a slight change 😅, Like when using `hardhat` this steps are correct : 
1. `yarn deploy --network sepolia` 
2.  `yarn verify` 


But for `foundry` currently its either : 
1. `yarn deploy` -> To deploy contract without verification 
2. `yarn deploy:verify --network sepolia` -> This deploy's and verify's command 

Like you don't have the direct command `yarn verify` which you can run just after `yarn deploy` 

Also I we mention about `yarn deploy:verify` it would be great if we mention that it's kindof of unstable for other custom configured network but works great for `sepolia`, `goerli`

Pabl0cks:
> But for foundry currently its either :
>
> yarn deploy -> To deploy contract without verification
> yarn deploy:verify --network sepolia -> This deploy's and verify's command
> Like you don't have the direct command yarn verify which you can run just after yarn deploy
>
> Also if we mention about yarn deploy:verify it would be great if we mention that it's kindof of unstable for other custom configured network but works great for sepolia, goerli

Thx @technophile-04 !! Nice catch 🙌 I've just pushed the change in Verify section. 

Do you feel the need to add some kind of warning in the Deploy section so users don't deploy and then realize they need to redeploy it in order to verify? Or since both sections are one after the other, may be not needed.

technophile-04:
Tysm 🙌, did a minor update to `deploy:verfiy` warning at [fb1cfeb](https://github.com/scaffold-eth/scaffold-eth-2/pull/442/commits/fb1cfeb1b596556b9663459ff12fb51365da9d7f)

Btw this will `deploy:verfiy` and will be hopefully fixed in #489, we can remove it after that 🙌


Also [fa7cbae](https://github.com/scaffold-eth/scaffold-eth-2/pull/442/commits/fa7cbae92ab7b93da0234243ffbc26d60d781f3b) update Getfoundry.sh -> getfoundry.sh so that people don't feel it's like a fishy link (kind of nitpick 😅 srry)

Merging this 🙌, Thanks again !!

PR #440, opened 2023-07-22
Add address qrcode to rainbowkit custom connect button

portdeveloper:
## Description

This PR removes the popup when clicking the rainbowkit custom connect button and adds a dropdown instead.

Before:

https://github.com/scaffold-eth/scaffold-eth-2/assets/108868128/fa9feafd-9511-46db-9a34-461f67c89b0b

After: 

https://github.com/scaffold-eth/scaffold-eth-2/assets/108868128/fa2c0247-f3d5-432a-9a06-e9d648bcf4b9

## Additional Information

- [x] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [x] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

_Closes #237 

_Note: If your changes are small and straightforward, you may skip the creation of an issue beforehand and remove this section. However, for medium-to-large changes, it is recommended to have an open issue for discussion and approval prior to submitting a pull request._

Your ENS/address: portdev.eth

Issue #237, opened 2023-03-13
QR receiver code for connected wallet address


carletex:
It'll be handy to have a QR receiver code for the connected wallet in SE-2

Is this customizable in Rainbow kit? We could add it here as an button / direct qr embed (and also show more info about the connected wallet, as @technophile-04 suggested here https://github.com/scaffold-eth/se-2/issues/87#issuecomment-1375151789)

![image](https://user-images.githubusercontent.com/2486142/224693609-05e80a98-67a5-42c9-bc4c-8bd28004d4e0.png)

rin-st:
> Is this customizable in Rainbow kit?

according to [docs](https://www.rainbowkit.com/docs/introduction) and github [1](https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/src/components/AccountModal/AccountModal.tsx) [2](https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/src/components/ProfileDetails/ProfileDetails.tsx) there is no way to do that :(
we can configure only disclaimer and app info here

---
<img width="730" alt="Снимок экрана 2023-03-15 в 22 23 04" src="https://user-images.githubusercontent.com/25638585/225406571-2a3207fd-2acf-4153-82a2-7754a305d183.png">

carletex:
Let's hold on this for now. If we can't tweak the rainbow modal directly, we might think of another place to put it.

KcPele:
ok

technophile-04:
Closing this in favour of #446 🙌

Issue #441, opened 2023-07-24
Better notifications / dropdowns


carletex:
- I think the shadows in the notifications / dropdowns doesn't give them enough contrast.
-  I'm not sold on the rounded buttons inside the dropdown, something seems off.

I'll ask Andrea to see if we can make some improvements. I think we just need some subtle tweaks.

1.
![1](https://github.com/scaffold-eth/scaffold-eth-2/assets/2486142/aedad7cf-f0ea-4fea-a20e-40cdb94e5432)
---
2.
![2](https://github.com/scaffold-eth/scaffold-eth-2/assets/2486142/ebc45e82-9ba7-4129-84bb-d7d1dea9f329)
---
3.
![3](https://github.com/scaffold-eth/scaffold-eth-2/assets/2486142/ff1878ec-a3da-4ef4-8763-bba57b8d8e88)

PR #446, opened 2023-07-25
Qrcode receiver modal

portdeveloper:
## Description

This PR removes the QR code SVG from the dropdown list, and adds a button to open a modal. See #440 for more details.

https://github.com/scaffold-eth/scaffold-eth-2/assets/108868128/99af222d-08d8-48f7-87ba-5668951d833d

I had to remove the z-index from the Footer.tsx as I could only make it work like that. Open to any other suggestions.

## Additional Information

- [ ] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [ ] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

_Closes #237 

Your ENS/address: portdev.eth


austintgriffith:
this is great! 🔥🫡🔥 thanks sers!

austintgriffith:
(I like the version with the modal the best!)

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
```diff
@@ -62,24 +73,81 @@ export const RainbowKitCustomConnectButton = () => {
...
+                        <input className="h-0 w-0 absolute top-0 left-0" />
+                        <label
+                          htmlFor="qrcode-modal"
+                          className="btn btn-ghost btn-sm btn-circle absolute right-3 top-3"
+                        >
+                          ✕
+                        </label>
+                        <div className="space-y-3 py-6">
+                          <div className="flex space-x-4 flex-col items-center gap-6">
+                            <QRCodeSVG value={account.address} size={256} />
```
technophile-04, line 145: NITPICK : 

Maybe give `className="rounded-xl"` to `QRCodeSVG` ? So that it matches Modal & our kind of general rounded theme 

Screenshot 2023-07-25 at 9 43 56 PM

PS : I think we can also remove the parent's div `space-y-3`

technophile-04:
Tysm @portdeveloper looking great !! 🙌

Even I like this more as compared to #440 and this one will be more visible too while presenting

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
portdeveloper: Looks a lot better, thanks!

Pabl0cks:
I like this one better than #440, more compact and clean. I feel both use cases "Copy address" and "View QR code" work better with this design. Great job @portdeveloper !! 🙌

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
```diff
@@ -62,24 +73,81 @@ export const RainbowKitCustomConnectButton = () => {
...
+                    <label tabIndex={0} className="btn btn-secondary btn-sm pl-0 pr-2 shadow-md dropdown-toggle">
                       <BlockieAvatar address={account.address} size={24} ensImage={account.ensAvatar} />
                       <span className="ml-2 mr-1">{account.displayName}</span>
-                      <span>
-                        <ChevronDownIcon className="h-6 w-4" />
-                      </span>
-                    </button>
+                      <ChevronDownIcon className="h-6 w-4 ml-2 sm:ml-0" />
+                    </label>
+                    <ul tabIndex={0} className="dropdown-content menu p-2 mt-1 shadow-lg bg-base-100 rounded-box">
```
rin-st, line 88: I can't see where menu starts. What about changing bg to `bg-secondary`?
before:
Снимок экрана 2023-07-26 в 12 05 16
Снимок экрана 2023-07-26 в 12 10 02


after:
Снимок экрана 2023-07-26 в 12 04 32
Снимок экрана 2023-07-26 в 12 04 39

Also, I don't like gray hover and ability to click when address already copied. But it's daisyui bug and minor thing

rin-st:
Looks good!

Noticed a bug - when I click to the address on modal, blockexplorer is opened in background. But I expect that the address just be copied

https://github.com/scaffold-eth/scaffold-eth-2/assets/25638585/390a7e5b-1aea-4698-900b-f210ab666533

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
carletex: Yes! I noticed the same (but it feels like a more general issue in SE-2), so I created this issue:

https://github.com/scaffold-eth/scaffold-eth-2/issues/441

Not sure if a different bg, a border, or more shadow... I'll ask Andrea too!

carletex:
> Noticed a bug - when I click to the address on modal, blockexplorer is opened in background. But I expect that the address just be copied

Yep, not sure about this.

The <Address> component links to the chain blockexplorer for the displayed address. In localhost, it links to the local blockexplorer, but in live chains, it goes to the corresponding blockexplorer (which can be handy).

Options:
1. Leave it as it's
2. Using `disableAddressLink` in the Address component for that instance. (although it will also affect live chains)
3. Same as 2, but only for local networks.

rin-st:
I forgot that `<Address />` works that way :). Probably we need to discuss it in another issue, but I believe that
- click on address should copy it by default or probably do nothing, since we need to copy much more often than viewing on block explorer (or is it only for me?) and it's not obvious that it links to blockexplorer. 
- Small `View on block explorer` link below the address or some icon after "Copy", for example something like this
image

> 1. Leave it as it's
> 2. Using disableAddressLink in the Address component for that instance. (although it will also affect live chains)
> 3. Same as 2, but only for local networks.

So I think point 2 is good. But open modal -> go to blockexplorer is convenient too, so additional blockexplorer link/icon can help :). Or we can add "View on blockexplorer" option in main menu, so we'll not need it in modal.

rin-st (approved):

rin-st:
lgtm 👍

carletex:
Let's wait until #455 is merged. We should add the new shadows and the new menu item sizes here too

rin-st:
Looks great! Noticed one micro bug - when you click on copy address, it blinks when content of button changes. First one - click. Second - not.

https://github.com/scaffold-eth/scaffold-eth-2/assets/25638585/f6cbe9c0-a7f4-4986-a45d-12db3ced7e00

carletex:
Good eye @rin-st !! haha

It's the :active pseudo-class when using daisy's `menu` class. In some places where using some daisy SE-2 where we should probably benefit using tw directly (if not, you have to override, sometime with `!`)

Anyway, let's merge this, and maybe we can rethink using the menu / menu-item classes here in another PR

image

carletex:
Great job @portdeveloper ;)

And thank you all for the review!!

portdeveloper:
Thanks! I am honored :)

PR #450, opened 2023-07-25
fix empty debug page  when mismatch in localStorage and contractName

technophile-04:
## Description

Screenshot 2023-07-26 at 2 15 22 AM

We get a blank debug page, this happens when there is a value in localStorage that does not match any of newly deployed contractNames 

### Solution : 
Added a `useEffect` which checks if the current localStorage value is present in `contractNames`array and returns 0th contractName if it doesn't find any match


rin-st (approved):

rin-st:
tested, works great!

carletex:
Thanks !!! Working great.

PR #452, opened 2023-07-26
use targetNetwork instead of hardcoding dir name in template root

technophile-04:
## Description

The UI breaks for `blockexplorer/[address]` in `foundry` extension since we have hardcoded `hardhat` in `[address].tsx` 

Screenshot 2023-07-26 at 5 44 15 PM

Should we add this to `main` and `foundry` branch too 🤔?


carletex (approved):
Working great @technophile-04 Thanks!

I see that the `foundry` branch is already working with the hardcoded path, but this will fix the CLI too.

PR #454, opened 2023-07-26
fix eth_accounts calls on localNetwork

technophile-04:
## Description

Since we have `localWalletClient` as use effect dependency we get lot of `eth_accounts` because of this line : https://github.com/scaffold-eth/scaffold-eth-2/blob/6370b4c49440f4aab7eff6896b43d814d6d01650/packages/nextjs/components/scaffold-eth/Faucet.tsx#L35

You can easily notice lot of `eth_accounts` while using SE-2 through CLI but for some reason on `main` it sometimes doesn't happens directly and if you add `console.log("This is localWalletClient", localWalletClient)` below the above `permalink` it will start making `eth_accounts` (😅 even I am confused about this behavior on `main` but I noticed on it on main)

Nevertheless, I think it makes  sense not to pass objects as dependency to `useEffect` 

### Solution : 
Move `localWalletClient` outside of the component this we don't need to pass it as a dependency. 

JSON.stringify doesn't work because `uid` is changing in  `localWalletClient`

```
{"chain" :  ........... "uid":"0f45556b091"}
```

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Faucet.tsx
+++ b/packages/nextjs/components/scaffold-eth/Faucet.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
 import { useIsMounted } from "usehooks-ts";
 import { Address as AddressType, createWalletClient, http, parseEther } from "viem";
 import { useNetwork } from "wagmi";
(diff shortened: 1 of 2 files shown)


rin-st:
Good job! 👍 
Please do the same for `<FaucetButton />` or use the same wallet client if possible

technophile-04:
> Please do the same for <FaucetButton /> or use the same wallet client if possible

I think we are not using `useEffect` in `FaucetButton` 😅 but moved `localWalletClient` outside the component at  [9426ba2](https://github.com/scaffold-eth/scaffold-eth-2/pull/454/commits/9426ba2cd417185c19ae2b089cd13bfbbe696b43)  🙌

rin-st:
yes, but it created new client on every render :)

rin-st (approved):

carletex:
This fixes it! Since `createWalletClient` is not a hook I guess we are not supposed to use it as a dep because it changes every render.

Thank all!

PR #460, opened 2023-07-29
update the abi type for getContract in useScaffoldContract

### LINKED ISSUE #459: bug: <title> useScaffoldContract not exposing contract functions.

### Current Behavior

I tried to use the contract instance in my project using useScaffoldContract, but it came back undefined for a function that exists in the abi.

I assumed its an issue with the project but I have come across another builder with the same issue.

image

### Expected Behavior

Should be able to call contract functions from a useScaffoldContract instance.

### Steps To Reproduce

Can be reproduced by creating a contract instance with the useScaffoldContract hook.

  const { data: noob } = useScaffoldContract({ contractName: "Noob" });
  console.log("noob: ", noob);
  console.log("Noob supply: ", noob?.totalSupply)

### Anything else?

image

image

technophile-04:
### Description

This PR tries to add to autocompletion when using `useScaffoldContract` : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/a73e916e-891b-457b-a769-bf2959f11e19

### Small Catch : 
As you can see in the above video we haven't passed `walletClient` to `useScaffoldContract` but we are still able to access `yourContract?.write` and its all method. 

The reason I am telling this is because if you look at viem's [`getContract`](https://viem.sh/docs/contract/getContract.html#usage) it say's that: 
1.  if you pass only `publicClient` you see only  `youractContract.read.{all read methods}` is available 
2.  if you pass  only `walletClient` you see only `yourContract.write.{all write methods}` 
3.  if you pass both `publicClient` & `walletClient` you see both `youractContract.read.{all read methods}` &  `yourContract.write.{all write methods}` 

I wonder how would we achieve this ?   

### Ideal Situation : 

```typescript
 const { data: yourContract } = useScaffoldContract({ contractName: "YourContract" });
 console.log("yourContract: ", yourContract?.write.withdraw());
//                                             ^ Typescript should give error no "write" 
// property found because `walletClient` was not passed

/** Also if you try above code it will break frontend saying "`withdraw` does not exist on undefined"
this is also one reason we should this because it will give user false hope at compile time **/
```
`yourContract?.read.{all read methods}` this should always be available since internally in `useScffafoldContract` we are always passing `publicClient`

## Additional Information

## Related Issues

Closes #459 

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldContract.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldContract.ts
@@ -1,8 +1,8 @@
-import { Abi } from "abitype";
 import { getContract } from "viem";
+import { usePublicClient } from "wagmi";
 import { GetWalletClientResult } from "wagmi/actions";
 import { useDeployedContractInfo } from "~~/hooks/scaffold-eth";
-import { ContractName } from "~~/utils/scaffold-eth/contract";
+import { Contract, ContractName } from "~~/utils/scaffold-eth/contract";
 
 /**
  * Gets a deployed contract by contract name and returns a contract instance
@@ -18,16 +18,15 @@ export const useScaffoldContract = <TContractName extends ContractName>({
   walletClient?: GetWalletClientResult;
 }) => {
   const { data: deployedContractData, isLoading: deployedContractLoading } = useDeployedContractInfo(contractName);
-
-  // type GetWalletClientResult = WalletClient | null, hence narrowing it to undefined so that it can be passed to getContract
-  const walletClientInstance = walletClient != null ? walletClient : undefined;
+  const publicClient = usePublicClient();
 
   let contract = undefined;
   if (deployedContractData) {
     contract = getContract({
       address: deployedContractData.address,
-      abi: deployedContractData.abi as Abi,
-      walletClient: walletClientInstance,
+      abi: deployedContractData.abi as Contract<TContractName>["abi"],
+      walletClient: walletClient ? walletClient : undefined,
+      publicClient,
     });
   }
 


technophile-04:
I have added a solution that solves the issue discussed in https://github.com/scaffold-eth/scaffold-eth-2/pull/460#issue-1827577418 at [62f75d8](https://github.com/scaffold-eth/scaffold-eth-2/pull/460/commits/62f75d8d1220941d44ea864a8609e6566cbe964a)

Now it gives you an error for `yourContract.write` if you don't pass walletClient but there is again one catch 😅 : 


While passing `walletClient` now you need to do this : 

```ts
  const { data: walletClient } = useWalletClient();
  const { data: yourContract } = useScaffoldContract({
    contractName: "YourContract",
    walletClient: walletClient || undefined,
  });
  console.log("yourContract: ", yourContract?.write.withdraw());

```

We need to add `walletClient: walletClient || undefined,` because the type returned by `useWalletClient` is incompatible with getContract's arg "walletClient", There is a mismatch of "null" eg : 

getContrac'ts `walletClient` accepts type -> `WalletClient | undefined` 

The return type of `useWalletClient` -> `WalletClient | null`

Maybe we update `.d.ts` ? 

----


But there might be a completely different solution to tackle this! A good starting might be [41aacb3](https://github.com/scaffold-eth/scaffold-eth-2/pull/460/commits/41aacb361f75b35a3125b87009ee5a6faac62bfd) 🙌

rin-st:
I think `walletClient: walletClient || undefined` is not a big problem

rin-st (approved):

sverps:
Hey @technophile-04 , I pushed a small fix to the `useScaffoldContract` types to make it a bit more forgiving when using it. The change basically allows the `walletClient` to be `null` as well, which is OK, since internally that would be converted to undefined anyway when passing it to `getContract`.

sverps (approved):

technophile-04:
Thanks @sverps, but I think it again introduces the problem which I discussed in "small catch" of  https://github.com/scaffold-eth/scaffold-eth-2/pull/460#issue-1827577418

Screenshot 2023-08-02 at 4 53 47 PM

You can see `write` their even though we haven't passed `walletClient`

sverps:
> Thanks @sverps, but I think it again introduces the problem which I discussed in "small catch" of [#460 (comment)](https://github.com/scaffold-eth/scaffold-eth-2/pull/460#issue-1827577418)
> 
> Screenshot 2023-08-02 at 4 53 47 PM
> 
> You can see `write` their even though we haven't passed `walletClient`

Ah, my bad, I missed that comment. Let me have another look, this definitely feels like something we can fix with types.

sverps:
Found another way that seems to work better. We basically allow `null` as well for the walletClient, since this will be passed down to `getContract` just the same due to the ternary. The return type stays unchanged like this, so properly tells us we can't `write` when `walletClient` is undefined.

technophile-04:
>  it does feel a bit strange now that in the generic we specify it to exclude null, only to add it in the type for the param. So there might be a better way, but couldn't get anything else to play nice with the way the getContract generic types are set up

Yeah, it's a bit hacky but works nicely 🙌,  I tried different approaches too but no luck lol it took me a whole Saturday evening from coming [41aacb3](https://github.com/scaffold-eth/scaffold-eth-2/pull/460/commits/41aacb361f75b35a3125b87009ee5a6faac62bfd) -> https://github.com/scaffold-eth/scaffold-eth-2/pull/460/commits/62f75d8d1220941d44ea864a8609e6566cbe964a 

Merging this Thanks all for tinkering !!🙌

PR #463, opened 2023-08-01
Add an explanation and example usage when using `useScaffoldContractWrite` with state

ByteAtATime:
## Description

I got a bit stuck when trying to write to a contract with state. I don't use React that much, so I didn't know how hooks worked. This PR adds some documentation about the fact that the state changes after you initialize the hook.

## Additional Information

Your ENS/address: `0x598cb95773D9b66a27a5780DB5EED2d018685879`

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,22 @@ To send the transaction, you can call the `writeAsync` function returned by the 
 
 This example sends a transaction to the `YourContract` smart contract to call the `setGreeting` function with the arguments passed in `args`. The `writeAsync` function sends the transaction to the smart contract, and the `isLoading` and `isMining` properties indicate whether the transaction is currently being processed by the network.
 
+If you have any state, you can just pass it into the `args`, and it will automatically update when you call `writeAsync()`. Example:
+
+```ts
+const [address, setAddress] = useState("");
+
+const { writeAsync } = useScaffoldContractWrite({
+  contractName: "YourContract",
+  functionName: "calculateStuffWithAddress",
+  args: [address],
+});
+
+<input onChange={e => setAddress(e.target.value)} />
+
+<button onClick={writeAsync}> {/* This will submit with the current value of `address`! */}
+```
+
 ### useScaffoldEventSubscriber:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -240,6 +240,22 @@ To send the transaction, you can call the `writeAsync` function returned by the
 
 This example sends a transaction to the `YourContract` smart contract to call the `setGreeting` function with the arguments passed in `args`. The `writeAsync` function sends the transaction to the smart contract, and the `isLoading` and `isMining` properties indicate whether the transaction is currently being processed by the network.
 
+If you have any state, you can just pass it into the `args`, and it will automatically update when you call `writeAsync()`. Example:
```
sverps, line 243: The wording is a bit confusing. It's not so much the args that are updated when calling the `writeAsync`, it's the writeAsync` itself that get's updated when a state that is used by `useScaffoldContractWrite` has changed due to react rerender.

So this mechanism is just how react works, and isn't specific to this code.

sverps (commented):
Not entirely sure if and how this clarification should be included, since it's really more generally about how react works, and not specific to `scaffold-eth` or this custom hook.

@technophile-04, @edakturk14 wdyt?

carletex:
Thanks for this @ByteAtATime 

> Not entirely sure if and how this clarification should be included, since it's really more generally about how react works, and not specific to scaffold-eth or this custom hook

I agree with this.

We are planning on adding a "recipes" section in the new docs site. Maybe something like this could be included, but I don't see the point of adding it to the README.

I think we can close this (for now at least!)

README.md
edakturk14: agreed, since its more about how react works. Would suggest to keep it the way it is here instead of adding the clarification. As @carletex mentioned we can get into more details with the recipes.

edakturk14:
Thank you @ByteAtATime for your support!! 

Closing the PR for now as per the conv above. We have a WIP docs over here as well: https://github.com/scaffold-eth/se2-docs We'll be adding the recipes and some more details. If you'd like to follow along! :)

PR #465, opened 2023-08-01
Remove unused types

carletex:
While reviewing the types in `contract.ts` (the best way to learn TypeScript advanced features :D), I realized that some of the types aren't being used.


technophile-04 (approved):
> `contract.ts` (the best way to learn TypeScript advanced features :D)

++ 💯

Yes, I think we forgot to remove this before merging Wagmi & viem migration, merging this 🙌

PR #471, opened 2023-08-02
foundry contracts remapping to contracts dir 

technophile-04:
## Description

Update the remapping where the contracts live, keeping it similar to `hardhat` and it will also help us in frontend for blockexplorer for finding contract's metdata like bytecode in CLI

https://github.com/scaffold-eth/scaffold-eth-2/blob/ae2f4b654e105b5ec7eb603513d0d43c550dcf6c/packages/nextjs/pages/blockexplorer/address/%5Baddress%5D.tsx#L182

References #394 , #406 (comment)


technophile-04:
I have tested but just cc @jrcarlos2000 incase I missed something

Issue #468, opened 2023-08-01
catch when user hasn't run yarn generate on foundry


austintgriffith:
![files changed, compilation skipped](https://github.com/scaffold-eth/scaffold-eth-2/assets/2653167/673fde52-3493-4538-9079-548f48f90931)
****

you get this gnarly error when you haven't run yarn generate yet and you try to deploy

technophile-04:
![258219613-76ac3221-e2b7-42ab-b878-ebddf863e203](https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/a79047fc-60fe-48b2-b3c8-df0f9953f5f6)

We kindof added a better error message at #475 🙌

closing this for now 🙌

PR #475, opened 2023-08-02
Foundry Updates

### LINKED ISSUE #470: Making foundry stable
### Bugs noticed

1 . While doing `yarn deploy` I got this warning :

    > EIP-3855 is not supported in one or more of the RPCs used. Unsupported Chain IDs: 31337.
    > Contracts deployed with a Solidity version equal or higher than 0.8.20 might not work properly.
    > For more information, please see https://eips.ethereum.org/EIPS/eip-3855
    
    This might the reason why Austin said 0.8.20 was not working we should fix this

### Developer Experience :

1 . ~Lets have all contracts in `contracts` directory similar to `hardhat`~ (fixed at #471  )

2 . Lets add's intersection of all the networks present in `stdChains` & `hardhat.config.ts` default in `foundry.toml` 

3 . ~Let's add some comments so that the user does not delete out custom `exportDeploymentFunction`~ fixed at #406 
<details>
<summary>Sudo sample code</summary>

              function run() external {
                  uint256 deployerPrivateKey = setupLocalhostEnv();
            
                  vm.startBroadcast(deployerPrivateKey);
                  YourContract yourContract = new YourContract(
                      vm.addr(deployerPrivateKey)
                  );
                  console.logString(
                      string.concat(
                          "YourContract deployed at: ",
                          vm.toString(address(yourContract))
                      )
                  );
                  vm.stopBroadcast();
              
                   /**
                   * This function generates the file containing the contracts Abi definitions.
                   * These definitions are used to derive the types needed in the custom scaffold-eth hooks, for example.
                   * This function should be called last.
                   */
                   exportDeployments();
          
                  // If your chain is not present in foundry's stdChain, then you need to call function with chainName:
                  // exportDeployments("chiado")
      
             }
             

      
</details>

4 . Checkout Austin's issue here -> #468
Possible solution: Since we are inheriting from `ScaffoldETHDeploy` contract can we add nice checks in 
`ScaffoldETHDeploy` contract's `constructor` like checking `deployerKey` etc and will  using `revert`show nice error ? 
     
        
Also regarding this line :

```solidity
uint256 deployerPrivateKey = setupLocalhostEnv();
```

Maybe we can set a public variable in the `ScaffoldETHDeploy` contract named `envDeployerKey` and people who 
extend `ScaffoldETHDeploy` will have access to this.

5 . ~We recently added another utility similar to `exportDeployments` i.e you can now pass optional `networkName` : `exportDeployments("networkName")` this allows users to add `chains` eg: `chiado` which are not present in foundry's [`StdChain`](https://github.com/foundry-rs/forge-std/blob/f18e8aa3e72eef83518766eb34ad8c8d8e2aa0aa/src/StdChains.sol#L186) and generate abi definations for frontend without breaking~.

~Although this is a nice hack for now we should probably update it soon, let's find a way better way for this 🙌 such that user just need to define it in `foundry.toml` and just calling `exportDeployments()` without `networkName` does the job 🙌~ Fixed at #475 

6 . ~Get the deploy command similar to `hardaht` : `yarn deploy --network network_name`, this can maybe also help in pt.5 if there is way in foundry to access passed in flags cc @jrcarlos2000~  - Fixed at #475

jrcarlos2000:
## Description

- [x] use of unique function exportDeployments().
- [x] emulate hardhat package , allows passing a flag --network on deploy
- [x] displays readable error msg when  `DEPLOYER_PRIVATE_KEY` has not been set in `.env`.
- [x] added OZ contracts by default
- [x] added `remappings.txt` to  be able to read `import @openzeppelin` in `contracts`

## Additional Information

## Related Issues

Closes #470 #468 

### DIFF
--- a/packages/foundry/package.json
+++ b/packages/foundry/package.json
@@ -7,8 +7,8 @@
     "fork": "anvil --fork-url ${0:-mainnet} --config-out localhost.json",
     "compile": "forge compile",
     "generate": "node script/generateAccount.js",
-    "deploy": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${0:-default_network} --broadcast && node script/generateTsAbis.js",
-    "deploy:verify": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${0:-default_network} --broadcast --verify && node script/generateTsAbis.js",
+    "deploy": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${1:-default_network} --broadcast && node script/generateTsAbis.js",
+    "deploy:verify": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${1:-default_network} --broadcast --verify && node script/generateTsAbis.js",
     "lint": "forge fmt",
     "test": "forge test"
   },
--- a/packages/foundry/package.json
+++ b/packages/foundry/package.json
@@ -7,8 +7,8 @@
     "fork": "anvil --fork-url ${0:-mainnet} --config-out localhost.json",
     "compile": "forge compile",
     "generate": "node script/generateAccount.js",
-    "deploy": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${1:-default_network} --broadcast && node script/generateTsAbis.js",
-    "deploy:verify": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${1:-default_network} --broadcast --verify && node script/generateTsAbis.js",
+    "deploy": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${1:-default_network} --broadcast --legacy && node script/generateTsAbis.js",
+    "deploy:verify": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${1:-default_network} --broadcast --legacy --verify && node script/generateTsAbis.js",
     "lint": "forge fmt",
     "test": "forge test"
   },
--- a/packages/foundry/remappings.txt
+++ b/packages/foundry/remappings.txt
@@ -1,0 +1,10 @@
+ds-test/=lib/forge-std/lib/ds-test/src/
+erc4626-tests/=lib/openzeppelin-contracts/lib/erc4626-tests/
+forge-std/=lib/forge-std/src/
+openzeppelin-contracts/=lib/openzeppelin-contracts/
+@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts
+lib/forge-std:ds-test/=lib/forge-std/lib/ds-test/src/
+lib/openzeppelin-contracts:ds-test/=lib/openzeppelin-contracts/lib/forge-std/lib/ds-test/src/
+lib/openzeppelin-contracts:erc4626-tests/=lib/openzeppelin-contracts/lib/erc4626-tests/
+lib/openzeppelin-contracts:forge-std/=lib/openzeppelin-contracts/lib/forge-std/src/
+lib/openzeppelin-contracts:openzeppelin/=lib/openzeppelin-contracts/contracts/
--- a/packages/foundry/remappings.txt
+++ b/packages/foundry/remappings.txt
@@ -1,10 +1,1 @@
-ds-test/=lib/forge-std/lib/ds-test/src/
-erc4626-tests/=lib/openzeppelin-contracts/lib/erc4626-tests/
-forge-std/=lib/forge-std/src/
-openzeppelin-contracts/=lib/openzeppelin-contracts/
 @openzeppelin/contracts/=lib/openzeppelin-contracts/contracts
-lib/forge-std:ds-test/=lib/forge-std/lib/ds-test/src/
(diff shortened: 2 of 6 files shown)


packages/foundry/script/Deploy.s.sol
```diff
@@ -5,9 +5,15 @@ import "../contracts/YourContract.sol";
...
 
 contract DeployScript is ScaffoldETHDeploy {
+    error InvalidPrivateKey(string);
+
     function run() external {
         uint256 deployerPrivateKey = setupLocalhostEnv();
-
+        if (deployerPrivateKey == 0) {
+            revert InvalidPrivateKey(
+                "Make sure you have set DEPLOYER_PRIVATE_KEY in .env or use `yarn generate` to generate a new key"
```
carletex, line 14: Smal tweak to the copy.

```suggestion
                "You don't have a deployer account. Make sure you have set DEPLOYER_PRIVATE_KEY in .env or use `yarn generate` to generate a new random account"
```

packages/foundry/package.json
```diff
@@ -7,8 +7,8 @@
     "fork": "anvil --fork-url ${0:-mainnet} --config-out localhost.json",
     "compile": "forge compile",
     "generate": "node script/generateAccount.js",
-    "deploy": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${0:-default_network} --broadcast && node script/generateTsAbis.js",
-    "deploy:verify": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${0:-default_network} --broadcast --verify && node script/generateTsAbis.js",
+    "deploy": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${1:-default_network} --broadcast --legacy && node script/generateTsAbis.js",
```
carletex, line 10: Why is the legacy flag is needed?

carletex:
Thanks for this @jrcarlos2000 ! Really like having the same "--network" flag as in hardhat.

I left a couple of comments.

packages/foundry/package.json
jrcarlos2000: It is about [EIP 1559](https://eips.ethereum.org/EIPS/eip-1559), by default foundry will try to use EIP 1559 transactions, but some RPCs dont support it so transactions would fail. Has happened to me in some mumbai and custom network RPCs. For now it is more stable to have `--legacy`.

packages/foundry/remappings.txt
```diff
@@ -0,0 +1,10 @@
+ds-test/=lib/forge-std/lib/ds-test/src/
```
technophile-04, line 1: Any reason for adding remapping for other stuff? 

I think we can only keep this right? : 

```
@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts/
```

technophile-04: > those are by default, 

Yes, but I think we can remove other stuff right and just keep `@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts/` ? 

Because I just tried keeping only `@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts/` in `remappings.txt` and ran `forge remappings` and I got this : 
Screenshot 2023-08-03 at 11 00 14 PM

So I think it's picking other default stuff without us needing to mention it, Just trying to keep things minimal 

---

> Someone who is used to foundry will not use @openzeppelin but will use instead lib/openzeppelin

Yeah, but I was just going through [OZ docs](https://github.com/OpenZeppelin/openzeppelin-contracts#foundry-git) and found this : 

Screenshot 2023-08-03 at 10 55 30 PM


I mean in all their tutorials they use `@openzepplin`, also its important that we keep things very similar to `hardhat` and common practices

If `lib/openzeppelin` is too much popular then we add its remapping too, I think we add multiple remapping for same thing 

But its really important that we try to keep things as **minimal** as possible 🙌

technophile-04:
I just did some testing and it works great !!!

I deployed a contract to PGN testnet (which is not present in StdChains):[0x4c368c2D0b8B0fEC240AE0d0bcf382c263005476(https://explorer.sepolia.publicgoods.network/address/0x4c368c2D0b8B0fEC240AE0d0bcf382c263005476)

I just had to add `pgn = "https://sepolia.publicgoods.network/"` in `foundry.toml` and run `yarn deploy --network pgn` and it deployed correctly also generated it in frontend correctly 🙌 ( without needing to configure deploy script :) )

---

Regarding error  : 

Screenshot 2023-08-04 at 1 13 27 AM


Don't if it's possible to get only`Error message` , I think foundry likes to show traces which is kind of good but we can think about this in diff PR if there is an option, it's still better than previous error


---

Regarding default `OZ` lets clean `remapping` as discussed in `https://github.com/scaffold-eth/scaffold-eth-2/pull/475#discussion_r1283527793` and in TG chat for now lets keep only one `remapping` , I have also tested it out and it works nicely without other defaults because they are already inferred by `foundry`

---

Also noticed one thing, now if you do `yarn deploy seploia` without passing `--network` flag it defaults to default network configured in `foundry.toml` i.e (localhost) 

In hardhat if you do `yarn deploy sepolia` we get : 

Screenshot 2023-08-04 at 12 53 22 AM


I didn't explore any options for now, but maybe we can merge and think about it later 🙌

technophile-04:
Tysm @jrcarlos2000 !! Merging this, updated the readme a bit at  [f8906fa](https://github.com/scaffold-eth/scaffold-eth-2/pull/475/commits/f8906faa26d290409d5a8d334b36e20edd503248)

Let's try to iterate more regarding 2nd and 4th sections of this https://github.com/scaffold-eth/scaffold-eth-2/pull/475#issuecomment-1664567413 also update #470 🙌

PR #479, opened 2023-08-05
[Open For Discussion] - Enable the use of proxy contracts in useScaffoldContract components

escottalexander:
## Description

I am creating a branch that demonstrates how proxies work and it required these changes.

All this does is adds an optional proxyContractName param to each of the useScaffoldContract components. If the param is given then it uses the proxyContractName address but uses the original contractName ABI. Thus enabling the ability to call a proxy address with ABI methods defined in the contract.

Since the changes are small and they don't affect any existing functionality I was wondering if it made sense to add them to the base repo. Feel free to discuss here and close this PR if you decide it adds unnecessary complexity. I certainly don't want to hinder anyone's ability to learn but I also know that proxies are used all over the place and it might be nice to support it out of the box. @carletex @ZakGriffith @technophile-04 <-- Please tag others who may have opinions on this.

## Additional Information

Your ENS/address: gnole.eth


packages/nextjs/hooks/scaffold-eth/useScaffoldContract.ts
```diff
@@ -8,23 +8,31 @@ import { Contract, ContractName } from "~~/utils/scaffold-eth/contract";
  * Gets a deployed contract by contract name and returns a contract instance
  * @param config - The config settings
  * @param config.contractName - Deployed contract name
+ * @param config.proxyContractName - Deployed proxy contract name if you wish to use an address other than the one associated with the contractName (optional)
```
damianmarti, line 11: I think it is better to have something like "(Only for proxy contracts)" at the beginning of the comment, so it's clear at the very start that you don't need to worry about these parameters if you are not working with proxies.

damianmarti:
I like it!! 

I think that since it's an optional parameter and the feature does not affect the usage, it could be a nice new improvement to add to the core.

I added a small comment to reinforce the aspect that this param is only for using contracts through a proxy.

packages/nextjs/hooks/scaffold-eth/useScaffoldContract.ts
```diff
@@ -43,6 +51,6 @@ export const useScaffoldContract = <
 
   return {
     data: contract,
-    isLoading: deployedContractLoading,
+    isLoading: deployedContractLoading && proxyContractLoading,
```
rin-st, line 54: `deployedContractLoading || proxyContractLoading` ?

packages/nextjs/hooks/scaffold-eth/useScaffoldEventSubscriber.ts
```diff
@@ -17,13 +18,19 @@ export const useScaffoldEventSubscriber = <
...
+  // If no proxy is given then we default to the given contractName
+  if (!proxyContractName) {
+    proxyContractName = contractName;
+  }
   const { data: deployedContractData } = useDeployedContractInfo(contractName);
+  const { data: proxyContractData } = useDeployedContractInfo(proxyContractName);
 
   return useContractEvent({
-    address: deployedContractData?.address,
+    address: proxyContractData?.address,
```
rin-st, line 33: and again, is it ok that we're losing original contract events?

rin-st:
Hey @escottalexander , thanks for pr! Added some comments. And main question, why just not to pass proxy contract name to `contractName` field? What will change? Unclear errors if proxy contract deployed, but main was not?

technophile-04:
Tysm Alexander 🙌, We are planning to have the "Recipes" section in [docs.scaffoldeth.io](https://docs.scaffoldeth.io/) and I think we can tell people there that if you want to interact with Proxies you can just update `deployedContract.ts` with YourContarct address as proxy address and abi as implementation address : 

```diff
// deployedContract.ts
const contracts = {
  31337: [
    {
      chainId: "31337",
      name: "localhost",
      contracts: {
        YourContract : {
-          address : "<IMPLEMENTATION_ADDRESS>"        
+          address : "<PROXY_ADDRESS>",
          abi : [<IMPLEMENTATION_ABI>]
        }
     }
   ]
} as const;

export default contracts;
```

Now people can normally use our custom hooks. Personally, I think it's not worth adding complexity to custom hook when user's can easily handle one line code change 😅

Closing this for now, but others please feel free to reopen this and we can iterate more if it feels worth it 🙌

Thanks again, Alexander !

escottalexander:
Thanks for all of your thoughts. That sounds good to me!

PR #480, opened 2023-08-07
updated the UseScaffoldReadConfig and UseScaffoldWriteConfig to inclu…

patni11:
added address as an optional string to UseScaffoldReadConfig and UseScaffoldWriteConfig tyoes so that any contract address can be used to read and write. Also updated their respective hooks: useScaffoldContractRead and useScaffoldContractWrite

## Description

The change would allow factory contracts and their instances to work since users can use the address of instances as a field to read and write to that specific contract

## Additional Information

- [ True] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [True ] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

Your ENS/address: shubhpatni.eth

### DIFF
--- a/packages/nextjs/generated/deployedContracts.ts
+++ b/packages/nextjs/generated/deployedContracts.ts
@@ -1,3 +1,154 @@
-const deployedContracts = null;
+const contracts = {
+  31337: [
+    {
+      chainId: "31337",
+      name: "localhost",
+      contracts: {
+        YourContract: {
+          address: "0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512",
+          abi: [
+            {
+              inputs: [
+                {
+                  internalType: "address",
+                  name: "_owner",
+                  type: "address",
+                },
+              ],
+              stateMutability: "nonpayable",
+              type: "constructor",
+            },
+            {
+              anonymous: false,
+              inputs: [
+                {
+                  indexed: true,
+                  internalType: "address",
+                  name: "greetingSetter",
+                  type: "address",
+                },
+                {
+                  indexed: false,
+                  internalType: "string",
+                  name: "newGreeting",
+                  type: "string",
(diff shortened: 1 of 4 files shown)


sverps:
Hey, thank you for trying to help improve scaffold-eth! ❤️ 

A couple of remarks, since I don't entirely understand the use-case. The scaffold hooks are specifically designed to work with deployed contracts in the `deployedContracts` file. They provide auto-completion for those contracts. So I'm not sure there is any benefit in using these custom hooks when you dynamically deploy contracts with a contract factory, you're probably better off using the wagmi hooks directly for interacting with such contracts.

Please feel free to elaborate if I'm missing something.

In case there would be a benefit, we could simply not omit `address` from the [readConfig](https://github.com/scaffold-eth/scaffold-eth-2/blob/d60ff21e8d3d6f4c73c2d8bdeebcdee066fc9d82/packages/nextjs/utils/scaffold-eth/contract.ts#L137) rather than adding `{address?: string}`. Then no changes to the implementation are needed since `readConfig` would just overwrite the `address`. (And equivalent for the write hook).

technophile-04:
Closing this for now due to no reply, Also @patni11 please open a discussion / issue before crafting a direct PR as mentioned in PR template 🙌

Keref:
That saved me, I work on contract that deploy other contracts, and so I have addresses I have multiple addresses I need to map dynamically to contracts.

I had modified the useScaffold hooks accordingly and it all worked but being new to typescript I couldnt get the thing to check-type properly.

Unless I missed a functionality somehwere, the use case is that not all contracts are mapped at deployment time (includes clones, proxies, contract deploying things in constructors...)

PR #484, opened 2023-08-11
added "git checkout foundry" to readme

0xWildhare:
## Description

I think this is helpful...

Your ENS/address: 0xwildhare.eth


technophile-04:
Tysm @0xWildhare 🙌 

Updated it so that it clones and switches to `foundry` branch directly, also we had forgot to clone `submodules` so added it too

PR #489, opened 2023-08-14
Foundry v2

jrcarlos2000:
## Description

- [x] Export contracts to Next Js regardless of verification result after `yarn deploy:verify`
- [x] able to verify after deployment is done : cc @technophile-04 
## Additional Information

## Related Issues

_Closes #470 


packages/foundry/script/VerifyAll.s.sol
```diff
@@ -0,0 +1,104 @@
...
+import "forge-std/Vm.sol";
+
+contract VerifyAll is Script {
+    uint96 currTransactionIdx;
+
+    function run() external {
+        string memory root = vm.projectRoot();
+        string memory path = string.concat(
+            root,
+            "/broadcast/Deploy.s.sol/80001/run-latest.json"
```
technophile-04, line 14: I tried verifying contract which was deployed to sepolia and it gave me : 

Screenshot 2023-08-15 at 9 10 01 PM

I think its because of we have hardcoded `80001` ?

technophile-04:
I was trying to verify on `polygonMumbai` : 

Screenshot 2023-08-15 at 9 25 09 PM

https://mumbai.polygonscan.com/address/0x4FaE2E1Dd219400328edAC56B3bc9F81bD6EE190

I just say's you "submitted verification" and ends the script there, I think we should probably wait for contract verification and once we get a response from etherscan then output the message with verified etherscan URL 😅

I don't how would be able to make network request / wait for a network request if it's at all possible similar to se-2 hardhat

jrcarlos2000:
Yes, error logging is a TODO here. we can either run some checks after the script is done or while the script is done.

technophile-04:
Tysm @jrcarlos2000, I have been testing it a bit it seems work 🙌

But for some reason, I get this : 

Screenshot 2023-08-18 at 1 32 37 PM

As you can see logged the ERROR but the it was verified at the end 

--- 


Also if there was error like I used `Invalid API KEY` it shows this :   

Screenshot 2023-08-18 at 1 34 01 PM

I wonder if we could parse this nicely and show it on terminal ? Like Maybe just showing `stderr=Encountered an error verifying this contract` ? Like just showing nice messages which we alreay have in `stderr`, `Response`,  `Details` maybe log only this three things ? 

Also, it would be really great if we don't show "== Logs ==" if the thing failed 🙌

---

Also, I thinki there are warning like unsued `VerifyAll.sol` lets remove them 

But Thanks again !! The functionality works nicely lets clean it up a bit 🙌 !!

technophile-04:
Tysm @jrcarlos2000, 

Just a few things : 

1 . When we try to deploy intially we get warnings from `Verify.sol` : 
Screenshot 2023-08-22 at 11 45 47 AM


2 . I think we should log the etherscan URL for verified contract or at least the `address` of contract which just got verfied
Screenshot 2023-08-22 at 11 51 30 AM

☝️ We don't see any `address` or etherscan URL of the contract which just got verified 


When I was testing the verification of multiple contracts, it worked and also logged that one contract was already verified (since etherscan already found similar contract and verified it)
Screenshot 2023-08-22 at 12 20 30 PM

I like this logging !!, Would be really helpful if we also do something like this for point 2. 


3 . If the user has put invalid etherscan api key : 
Screenshot 2023-08-22 at 12 04 42 PM

We get this :( , Which is hard to debug. I think we should make it a lil verbose 🙌

packages/foundry/script/VerifyAll.s.sol
```diff
@@ -0,0 +1,152 @@
...
+
+        if (f.stderr.length != 0) {
+            console.logString(string(f.stderr));
+        } else {
+            console.logString(string(f.stdout));
+        }
+        return;
+    }
+
+    function nextTransaction(string memory content) external returns (bool) {
```
technophile-04, line 112: I think the warning is due we missed `view`

jrcarlos2000:
> 1 . When we try to deploy intially we get warnings from Verify.sol :
This has been corrected

> 2 . I think we should log the etherscan URL for verified contract or at least the address of contract which just got verfied
I have checked and i added a logging for the address being verified or verified, however is not possible to add the link to the verified contract unless we hardcode the URL values.

> 3 . If the user has put invalid etherscan api key :
I am not getting this error right now, if i put a wrong API KEY it would tell me it is invalid. seems to be a problem with just a few networks like goerli, if i put the wrong api key for mumbai, it shows me the error.
image


cc: @technophile-04

technophile-04:
Hey Tysm @jrcarlos2000 !! Looking really great now !! 

I tested it with a few different networks and working great! 

Just a small case that i noticed while testing is,For some reason this only happens only on `goerli` and `sepolia` :  

1. Add wrong etherscan API key in `.env` eg : `ETHERSCAN_API_KEY=fsdfdssdf`
2. Deploy the contract 
3. While verifying it gives : 
Screenshot 2023-09-04 at 7 58 34 PM

Instead of "Missing or invalid ApiKey"

Also, this problem is just with `goerli` and `sepolia` and maybe with `mainnet` but with other chains it gives nice error ! 


Nevertheless if it gets very hard to debug / require a very big change we can merge this and iterate over this issue on other PR, Because I think this PR already is looking alot better now & for chains like `goerli/sepolia` we already give etherscanAPI key so not big deal!!

technophile-04:
Merging this 🙌, Regarding `sepolia` & `goerli` I think since we already porvide API_KEY people won't get into this trap but neverthe less as discussed in TG chat let's open this issue on foundry repo 🙌 & keep track of it

PR #499, opened 2023-08-22
Update and typescript version

### LINKED ISSUE #492: Handling wagmi & viem vesions
### Context : 

We have configured [`AddressType`](https://abitype.dev/config#addresstype) to `string` instead of `0x{string}` for better developer experience [here in this file](https://github.com/scaffold-eth/scaffold-eth-2/blob/main/packages/nextjs/types/abitype/abi.d.ts). 

1. We are using `viem` and `wagmi` as our dependencies 

2. `wagmi` and `viem` both internally are using `abitype` as their dependency

Ideally, we shouldn't care about **`patch`** version releases but since `wagmi` and `viem` are under constant development they are pretty actively fixing bugs and publishing new versions. 

Since `wagmi` and `viem` are kind of the backbone of SE-2 and since we are a development template/starter kit and not a **product** , I think it makes sense to keep them updated obviously not every release but at least keep them checked after 1month or so because they add new chains, fix bugs, etc. <-- Would love to know what others thoughts on this

#### ->  The problem happens when there is a new **`patch`** version of `wagmi` and `viem` and we need to update them. 

Ideally, it should have been as easy as `yarn workspace @se-2/nextjs up wagmi viem` since we are just updating them to `**patch**` version. Checkout #419  "Steps to reproduce" section what chaos it causes when you do `yarn workspace @se-2/nextjs up wagmi viem` :(  
 
 Since we have `AddressType` as mentioned in begging this causes problems with types :(, We have faced this multiple times when trying to update `wagmi` 

This occurs whenever `wagmi's` internal version of `abitype` differs from `viem's` internal version of `abitype`. 

Checkout this issue https://github.com/wagmi-dev/wagmi/issues/1712#issue-1554554144 comments, We created this issue even before `viem` was launched but there was a version mismatch of `abitype` between `wagmi's` repo internal packages

And now since `viem` also uses `abitype` anytime there is a version mismatch (which there is currently). Checkout https://github.com/wagmi-dev/wagmi/issues/2736 because of which we have to lock version's of wagmi and viem at #420 

### Solutions : 

1. Make PR's / create issues to Wagmi repo to let them know and fix on their side (which we have been doing from the start). Not best solution because it might take time to resolve (like in current situtaion)

2. Keep updating `wagmi` normally as it is and update `viem` only if its internal `abitype` version matches with `wagmi`. In recent PR #491 I just updated `wagmi` keeping `viem` same because they both have `abittype` version `0.8.7` 

Obviously removing `AddressType`  configuration is not an option since it will decrease the developer experience for beginners a lot :( 

I don't know if this is something we could handle on our own like forcing typescript to use our configuration of `AddressType`  irrespective what `abitype` versions `wagmi` and `viem` are using, I tried playing and didn't find any good option, Would you love to hear from others 🙌

We can go with point 2. of the solution but there might be a point when we will break because in case in future if `wagmi`  might break with `viem` version we are using and we also won't be able to update `viem` since its `abitype` version will be different from wagmi

technophile-04:
## Description

So after a lot of digging, I landed on [how to two versions of same npm package](https://stackoverflow.com/questions/43770328/two-versions-of-same-npm-package-in-node-application) and for some reason, I am not able to find that particular link but people were trying to have kind of same typescript version. Which led to me thinking that the reason  `yarn next:check-types` might be checking `node_modules` as discussed in #492 is because `viem` is using typescript version `5+` 

So the steps I followed to fix #492  : 

1. Updated `viem` version to latest using `yarn workspace @se-2/nextjs up viem`

2. I did `yarn next:check-types` here I got `678 errors` mostly from `node_modules/viem`

3. `yarn workspace @se-2/nextjs up typescript` (For some reason this also updated hardhat typescript version which is ok I think nd kind of makes sense)

4. Again `yarn next:check-types` it now gave me only errors from Se-2 repo regarding `AddressType` 

5. Updated `abi.d.ts`, to also handle `viem/node_modules/abitype` and have `AddressType : string`

6. Again `yarn next:check-types` and lol no types error at all!!

---

Started with manual testing and found that nextjs app broke with "Cannot find `fs` module error", after some digging found this on [Rainbowkit installation Page](https://www.rainbowkit.com/docs/installation#additional-build-tooling-setup)  : 

Screenshot 2023-08-23 at 2 27 42 AM

So I added nextjs config which they linked on their [installation Page](https://www.rainbowkit.com/docs/installation#additional-build-tooling-setup) and everything started working. I don't why we didn't have to do it before. 

My wild guess is something heavily changed inside `viem` which kind of removed exposing `ethers` to outter world and we got that error because `viem` is internally using `ethers` [somewhere,  check its dependencies here](https://github.com/wagmi-dev/viem/blob/6a5688684a39e1ce24c8366debc65d9e1bd30132/package.json#L165). 

----

Lol I didn't particularly know what was actual cause, so explained all the steps if someone can infer something from it 🙌. 

I need to do some testing nicely, will do it tomorrow and would appreciate it if others could test as well 🙌

Also please feel free to share your thoughts !!

## Additional Information

## Related Issues

#492 

### DIFF
--- a/packages/hardhat/package.json
+++ b/packages/hardhat/package.json
@@ -43,7 +43,7 @@
     "solidity-coverage": "^0.8.2",
     "ts-node": "^10.9.1",
     "typechain": "^8.1.0",
-    "typescript": "^4.9.5"
+    "typescript": "^5.1.6"
   },
   "dependencies": {
     "@openzeppelin/contracts": "^4.8.1",
--- a/packages/nextjs/next.config.js
+++ b/packages/nextjs/next.config.js
@@ -9,6 +9,10 @@ const nextConfig = {
   eslint: {
     ignoreDuringBuilds: process.env.NEXT_PUBLIC_IGNORE_BUILD_ERROR === "true",
   },
+  webpack: config => {
+    config.resolve.fallback = { fs: false, net: false, tls: false };
+    return config;
+  },
 };
 
 module.exports = nextConfig;
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -31,7 +31,7 @@
     "react-hot-toast": "^2.4.0",
     "use-debounce": "^8.0.4",
     "usehooks-ts": "^2.7.2",
-    "viem": "1.2.1",
+    "viem": "^1.6.7",
     "wagmi": "1.3.10",
     "zustand": "^4.1.2"
   },
@@ -50,7 +50,7 @@
     "postcss": "^8.4.16",
     "prettier": "^2.8.4",
     "tailwindcss": "^3.3.3",
-    "typescript": "^4.9.5",
+    "typescript": "^5.1.6",
     "vercel": "^28.15.1"
   }
 }
--- a/packages/nextjs/types/abitype/abi.d.ts
+++ b/packages/nextjs/types/abitype/abi.d.ts
@@ -1,5 +1,11 @@
 import "abitype";
 
+declare module "viem/node_modules/abitype" {
+  export interface Config {
+    AddressType: string;
+  }
+}
+
 declare module "abitype" {
   export interface Config {
     AddressType: string;
--- a/yarn.lock
+++ b/yarn.lock
@@ -1171,6 +1171,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@noble/curves@npm:1.1.0":
+  version: 1.1.0
+  resolution: "@noble/curves@npm:1.1.0"
+  dependencies:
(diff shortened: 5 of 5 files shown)


rin-st (approved):

damianmarti (approved):
LGTM but I think it is better that @sverps looks at it before merging.

dgrcode (approved):
👍

sverps (approved):
If this solves the issues, we can merge this for now. I'll have a second look to see if there's an alternative, but no need to wait for that.

Issue #466, opened 2023-08-01
Example UI: event section


carletex:
Let's keep expanding the Example UI with more examples.

In this case, show how the Event hooks (useScaffoldEventHistory.ts && useScaffoldEventSubscriber ) work. Right now it's just a console.log in `ContractData.tsx`

In `example-ui.tsx` we can add a new component `<ContractEvents>`, that show in two columns: Event History (history hook) / New Event (subscriber hook) info about the events from the contract: Address (so we can showcase the `<Address>` component) + Greeting + Value

We can ask Andrea if we need some help with the design

rin-st:
Hey @carletex ! Could you ask Andrea for designs, please?

ceeriil:
Hey @carletex any update yet?

technophile-04:
Should we close since #576 ?

PR #507, opened 2023-08-26
Reusable TriggerModal component in RainbowKitCustomConnectButton and Faucet

0xChijoke:
## Description

Hello! This pull request introduces a reusable TriggerModal component and integrates it into the RainbowKitCustomConnectButton and Faucet components. This modification aims to improve developer/user experience and code reuseability.

https://github.com/scaffold-eth/scaffold-eth-2/assets/53488449/55b367aa-ae02-4c53-8ee6-3baef4c55f60

## Additional Information

Your ENS/address: chijoke.eth

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Faucet.tsx
+++ b/packages/nextjs/components/scaffold-eth/Faucet.tsx
@@ -1,4 +1,6 @@
 import { useEffect, useState } from "react";
+import Modal from "../TriggerWithModal/Modal";
+import Trigger from "../TriggerWithModal/Trigger";
 import { Address as AddressType, createWalletClient, http, parseEther } from "viem";
 import { useNetwork } from "wagmi";
 import { hardhat } from "wagmi/chains";
@@ -82,49 +84,42 @@ export const Faucet = () => {
 
   return (
     <div>
-      <label htmlFor="faucet-modal" className="btn btn-primary btn-sm px-2 rounded-full font-normal normal-case">
+      <Trigger id="faucet-modal" className="btn btn-primary btn-sm px-2 rounded-full font-normal normal-case">
         <BanknotesIcon className="h-4 w-4" />
         <span>Faucet</span>
-      </label>
-      <input type="checkbox" id="faucet-modal" className="modal-toggle" />
-      <label htmlFor="faucet-modal" className="modal cursor-pointer">
-        <label className="modal-box relative">
-          {/* dummy input to capture event onclick on modal box */}
-          <input className="h-0 w-0 absolute top-0 left-0" />
-          <h3 className="text-xl font-bold mb-3">Local Faucet</h3>
-          <label htmlFor="faucet-modal" className="btn btn-ghost btn-sm btn-circle absolute right-3 top-3">
-            ✕
-          </label>
-          <div className="space-y-3">
-            <div className="flex space-x-4">
-              <div>
-                <span className="text-sm font-bold">From:</span>
-                <Address address={faucetAddress} />
-              </div>
-              <div>
-                <span className="text-sm font-bold pl-3">Available:</span>
-                <Balance address={faucetAddress} />
-              </div>
+      </Trigger>
+
+      <Modal id="faucet-modal">
+        <h3 className="text-xl font-bold mb-3">Local Faucet</h3>
+        <div className="space-y-3">
+          <div className="flex space-x-4">
+            <div>
+              <span className="text-sm font-bold">From:</span>
+              <Address address={faucetAddress} />
             </div>
-            <div className="flex flex-col space-y-3">
-              <AddressInput
-                placeholder="Destination Address"
-                value={inputAddress ?? ""}
-                onChange={value => setInputAddress(value)}
-              />
-              <EtherInput placeholder="Amount to send" value={sendValue} onChange={value => setSendValue(value)} />
-              <button className="h-10 btn btn-primary btn-sm px-2 rounded-full" onClick={sendETH} disabled={loading}>
-                {!loading ? (
-                  <BanknotesIcon className="h-6 w-6" />
-                ) : (
-                  <span className="loading loading-spinner loading-sm"></span>
-                )}
-                <span>Send</span>
-              </button>
+            <div>
+              <span className="text-sm font-bold pl-3">Available:</span>
+              <Balance address={faucetAddress} />
             </div>
           </div>
-        </label>
-      </label>
+          <div className="flex flex-col space-y-3">
+            <AddressInput
+              placeholder="Destination Address"
+              value={inputAddress ?? ""}
+              onChange={value => setInputAddress(value)}
+            />
+            <EtherInput placeholder="Amount to send" value={sendValue} onChange={value => setSendValue(value)} />
+            <button className="h-10 btn btn-primary btn-sm px-2 rounded-full" onClick={sendETH} disabled={loading}>
+              {!loading ? (
+                <BanknotesIcon className="h-6 w-6" />
+              ) : (
+                <span className="loading loading-spinner loading-sm"></span>
+              )}
+              <span>Send</span>
+            </button>
+          </div>
+        </div>
+      </Modal>
     </div>
   );
 };
--- a/packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
+++ b/packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
@@ -1,4 +1,6 @@
 import { useState } from "react";
+import Modal from "../TriggerWithModal/Modal";
+import Trigger from "../TriggerWithModal/Trigger";
 import { ConnectButton } from "@rainbow-me/rainbowkit";
 import { QRCodeSVG } from "qrcode.react";
 import CopyToClipboard from "react-copy-to-clipboard";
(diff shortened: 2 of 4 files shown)


rin-st:
Hi @0xChijoke , thank you for the pr!

I believe we should ask designers for big changes on the example-ui page.

So in this pr let's polish new modal component and use it for components where we have modals already - `RainbowKitCustomConnectButton` and `Faucet`.

We can add modal for example-ui page later if needed

packages/nextjs/components/Modal.tsx
```diff
@@ -0,0 +1,44 @@
+/**
+ * Modal component: A reusable modal component.
+ */
+type ModalProps = {
+  trigger: JSX.Element;
+  modalTitle: string;
```
rin-st, line 6: We need to use that prop for `htmlsFor` and `id` attributes but not for title. So I think `name` or `id' is better.

Title should be in different place, because it can be empty or long with spaces and can break labels/ids.

packages/nextjs/components/Modal.tsx
```diff
@@ -0,0 +1,44 @@
...
+  modalContent: JSX.Element;
```
rin-st, line 7: what about just using children instead of this prop

packages/nextjs/components/Modal.tsx
```diff
@@ -0,0 +1,44 @@
...
+  action?: {
+    label: string;
+    onClick: () => void;
+  };
+};
+
+const Modal: React.FC<ModalProps> = ({ trigger, modalTitle, modalContent, action }) => {
```
rin-st, line 14: Basically, it's not only modal, but also trigger. So maybe it should be named like `TriggerWithModal`? Check prop types name too if you'll change component name.

packages/nextjs/components/Modal.tsx
```diff
@@ -0,0 +1,44 @@
+/**
+ * Modal component: A reusable modal component.
+ */
+type ModalProps = {
+  trigger: JSX.Element;
+  modalTitle: string;
+  modalContent: JSX.Element;
+  action?: {
```
rin-st, line 8: It can be two or more buttons, for example "cancel" and "save". So probably we need to include this in content too

packages/nextjs/components/example-ui/ContractInteraction.tsx
```diff
@@ -19,6 +20,45 @@ export const ContractInteraction = () => {
     },
   });
 
+  // Define the modal trigger button
```
rin-st, line 23: let's change this file when we'll have designs
https://github.com/scaffold-eth/scaffold-eth-2/issues/466

packages/nextjs/components/Modal.tsx
0xChijoke: Exactly, the actions can be included in the modal content.

In the exampe-ui the send is in the modalContent.

The action is optional cause I had a specific use for passing it in.

It can also be removed if there is no need to leave the option there.

packages/nextjs/components/Modal.tsx
0xChijoke: Sounds good.

Does TriggerModal work?

0xChijoke:
@rin-st  Let me know what you think about the current state of the component.

Should the optional action be removed because that can come with the children?

Would I need to create another PR to leave other files in there original state?

rin-st:
> Should the optional action be removed because that can come with the children?

Yes, we don't need it for now

> Would I need to create another PR to leave other files in there original state?

No, you can rename this pr and remove changes of other files. We can take those changes if needed from commits history.

When you complete the component, please change custom modals in `RainbowKitCustomConnectButton` and `Faucet`

Thanks!

packages/nextjs/components/TriggerModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+  trigger: JSX.Element;
+  name: string;
+  children: JSX.Element;
+  action?: {
+    label: string;
+    onClick: () => void;
+  };
+};
+
+const Modal: React.FC<TriggerModalProps> = ({ trigger, name, children, action }) => {
```
rin-st, line 14: name

packages/nextjs/components/TriggerModal.tsx
```diff
@@ -0,0 +1,42 @@
+/**
+ * Modal component: A reusable modal component.
```
rin-st, line 2: redundant comment

packages/nextjs/components/TriggerModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+ */
+type TriggerModalProps = {
+  trigger: JSX.Element;
+  name: string;
+  children: JSX.Element;
+  action?: {
```
rin-st, line 8: remove for now

0xChijoke:
Hello @rin-st,

While refactoring the RainbowKitConnectButton component, I encountered scenarios where rendering the trigger separately from the modal body would be beneficial. 

To address this, I've split the component into two parts and introduced a new folder called TriggerWithModal. Within this folder, you'll find the Trigger and Modal components, each serving their specific roles and enhancing modularity.

I've already utilized this structure in both the RainbowKitConnectButton and Faucet components. Your feedback on this approach and any potential adjustments would be greatly appreciated.

Thanks!

rin-st:
Hi @0xChijoke , good job!

I don't like two-component nature of `TriggerWithModal` but it seems in this case it's a good solution.

I've tested your branch and found that modal doesn't close on click outside, as it was before. Fix it, please

packages/nextjs/components/scaffold-eth/Faucet.tsx
```diff
@@ -82,49 +84,42 @@ export const Faucet = () => {
 
   return (
     <div>
-      <label htmlFor="faucet-modal" className="btn btn-primary btn-sm px-2 rounded-full font-normal normal-case">
+      <Trigger id="faucet-modal" className="btn btn-primary btn-sm px-2 rounded-full font-normal normal-case">
```
rin-st, line 87: and  it's better to move id to constant outside of component and use it in both trigger and modal

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
```diff
@@ -134,10 +136,10 @@ export const RainbowKitCustomConnectButton = () => {
                         )}
                       </li>
                       <li>
-                        <label htmlFor="qrcode-modal" className="btn-sm !rounded-xl flex gap-3 py-3">
+                        <Trigger id="qrcode-modal" className="btn-sm !rounded-xl flex gap-3 py-3">
```
rin-st, line 139: same here

0xChijoke: I've incorporated the changes you mentioned. 

I've also added an option for the modal to be closed on outside clicks(true by default), considering its reusability. 

This way, the modal can be used in various scenarios, allowing builders to choose the preferred behaviour.

Your patience and guidance are much appreciated. Looking forward to your feedback!

technophile-04:
Hey guys thanks, but I don't see much advantages of adding this component : 
 
 With [DaisyUI v3 modal](https://daisyui.com/components/modal/#dialog-modal-with-a-close-button-at-corner): 

```jsx
<button onClick={() => window.faucet_modal.showModal()} className="...">
   ...
  <span>Faucet</span>
</button>

<dialog id={"faucet_modal"} className="modal">
  <form method="dialog" className="modal-box">
    ...
  </form>
</dialog>
```


With this component : 
```jsx
<Trigger id={FAUCET_MODAL_ID} className="...">
  ...
  <span>Faucet</span>
</Trigger>

<Modal id={FAUCET_MODAL_ID}>
  ...
</Modal>
```


Since we are using a component library (daisyUI v3) I think it's not worth creating our own micro components unless we have good reason to or there is something that is not present / not working in daisyUI

Also, I think by creating this component we are just abstracting 1-2 lines of code, I think its we let user use daisy UI and learn/customize as they want 

What do you guys think?

0xChijoke:
> Hey guys thanks, but I don't see much advantages of adding this component :
> 
> With [DaisyUI v3 modal](https://daisyui.com/components/modal/#dialog-modal-with-a-close-button-at-corner):
> 
> ```js
> <button onClick={() => window.faucet_modal.showModal()} className="...">
>    ...
>   <span>Faucet</span>
> </button>
> 
> <dialog id={"faucet_modal"} className="modal">
>   <form method="dialog" className="modal-box">
>     ...
>   </form>
> </dialog>
> ```
> 
> With this component :
> 
> ```js
> <Trigger id={FAUCET_MODAL_ID} className="...">
>   ...
>   <span>Faucet</span>
> </Trigger>
> 
> <Modal id={FAUCET_MODAL_ID}>
>   ...
> </Modal>
> ```
> 
> Since we are using a component library (daisyUI v3) I think it's not worth creating our own micro components unless we have good reason to or there is something that is not present / not working in daisyUI
> 
> Also, I think by creating this component we are just abstracting 1-2 lines of code, I think its we let user use daisy UI and learn/customize as they want
> 
> What do you guys think?



> 

I share your perspective. While the initial intention was to simplify modal usage, it seems that introducing custom components might actually lead to unnecessary complexity.

technophile-04:
> I share your perspective. While the initial intention was to simplify modal usage, it seems that introducing custom components might actually lead to unnecessary complexity. 

Yeah, but tysm @0xChijoke for questing on this !! Also, Chijoke would really appreciate creating an issue or discussion before creating a new feature PR so that we can iterate and brainstorm on it if it is really required / others can put suggestions/improvements on it too 🙌

Really feel bad closing this PR but appreciate it !! Also, tysm to Rinat for reviewing this 🙌

rin-st:
> I think by creating this component we are just abstracting 1-2 lines of code, I think its we let user use daisy UI and learn/customize as they want

hmm, I'm not very agree, because
- why to force users to learn something if we can avoid it? yes it's a small thing, but it can take let's say 5-10 mins to configure it instead of two imports. Users can forget to add clickoutside and so on
- we have at least two modals for now, and if we want for example add shadow we need to do it in two places instead of one

Yes, if we want to add modal, we need to add most of the options from daisyui, and also add other components like `Button` and other simple things. It simplifies development but we need to think if we really need it.

Thanks @0xChijoke for your efforts and @technophile-04 for review!

Issue #509, opened 2023-08-28
daisyUI menu breaks on production


technophile-04:
### Description : 

It seem that daisy nextjs break on production : 
![Screenshot 2023-08-29 at 3 01 25 AM](https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/20bb94d7-9e70-47a9-9237-c8764798f3b7)

This dosen't happen while testing locally it only happens in production. 

### Reproduce : 

1. `yarn workspace @se2/nextjs build`

2. `yarn workspace @se2/nextjs serve`

It seems that there is some problem while bundling / nextjs build process, because while developing locally you could see this if you inspect link : 
![Screenshot 2023-08-29 at 2 21 10 AM](https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/640665ad-5b4a-4f2e-bd2c-5d6e4524aec2)

But if you build the app and the start, it doesn't generates this. 

### Some references : 

1 . https://github.com/saadeghi/daisyui/discussions/2183

2. https://github.com/saadeghi/daisyui/discussions/2272

PR #510, opened 2023-08-28
add grid and grid-flow class to li manually

technophile-04:
## Description

This is a hacky solution for #509

### DIFF
--- a/packages/nextjs/components/Header.tsx
+++ b/packages/nextjs/components/Header.tsx
@@ -85,7 +85,7 @@ export const Header = () => {
             </ul>
           )}
         </div>
-        <Link href="/" passHref className="hidden lg:flex items-center gap-2 ml-4 mr-6">
+        <Link href="/" passHref className="hidden lg:flex items-center gap-2 ml-4 mr-6 shrink-0">
           <div className="flex relative w-10 h-10">
             <Image alt="SE2 logo" className="cursor-pointer" fill src="/logo.svg" />
           </div>
--- a/packages/nextjs/components/Header.tsx
+++ b/packages/nextjs/components/Header.tsx
@@ -16,7 +16,7 @@ const NavLink = ({ href, children }: { href: string; children: React.ReactNode }
       passHref
       className={`${
         isActive ? "bg-secondary shadow-md" : ""
-      } hover:bg-secondary hover:shadow-md focus:!bg-secondary py-1.5 px-3 text-sm rounded-full gap-2`}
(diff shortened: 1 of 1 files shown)


damianmarti (approved):
Thanks!

rin-st (approved):

technophile-04:
Just pushed a few minor tweaks : 

1. [1551743](https://github.com/scaffold-eth/scaffold-eth-2/pull/510/commits/1551743781ec7c1f2e68780bb06322780fccc14e) -> Fixes Scaffol-eth logo wraps when not in full screen or in smaller screens : 
Screenshot 2023-08-29 at 12 56 28 PM

this has been bugging from long and lol also Austin has the thing broken always because his brwoser window is always smaller [while presenting](https://youtu.be/ahDEbcdVV-o?si=SkwbpBeNbPcXGv_4&t=572)

2. [3a99b66](https://github.com/scaffold-eth/scaffold-eth-2/pull/510/commits/3a99b66e922defaa2252cc7e82f34052bf25c10d) -> there was kind of flickering when you switch tabs : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/09d08fb2-203b-4ecc-9529-76007523b149

so now the behaviour is the same as we had in daisyUI v2 checkout this https://chall-accepted-builders.vercel.app/

3. [eaa9364](https://github.com/scaffold-eth/scaffold-eth-2/pull/510/commits/eaa9364ed6ccc5174d7946c03b42771b69a3c0a9) -> capitalized `eth` in logo since that what we are using eg in READEME.md name🙌

---

### Notes regarding hacky solution : 

1. Yeah it's PITA that we have to add these classes manually Instead they should been added by default similar to the way running on localhost but this seems problem with daisyUI. Created an issue on daisyUI -> https://github.com/saadeghi/daisyui/issues/2303

2. This problem does seem to get solved when you update the `next` version to the latest `13.4.X` and also `13.4.x` is backward compatible (this is what they say) but need to patiently see if there are any direct breaking and migrate slowly will research on this 🙌


Tysm guys!! Merging this for now, we can keep an eye on https://github.com/saadeghi/daisyui/issues/2303 if its get solved if not we can migrate to latest version of next slowly 🙌

carletex:
> this has been bugging from long and lol also Austin has the thing broken always because his brwoser window is always smaller [while presenting](https://youtu.be/ahDEbcdVV-o?si=SkwbpBeNbPcXGv_4&t=572)

THANK YOU. I've been having nightmares about this haha

PR #515, opened 2023-09-04
fix(AddressType): cleanup overrides

### LINKED ISSUE #514: Cleanup of abi Address type

### Current Behavior

To not have things break, we currently override the `AddressType` in `abi.d.ts` to be `string`. This resulted in broken types when viem and wagmi had versions that didn't match well.

Some additional things that are suboptimal:

- This could theoretically lead to run-time errors if some implementation of viem of wagmi changes.
- This might make scaffold-eth-2 users mislearn the API's of viem and wagmi.
- Slightly lower type safety by forcing AddressType to be less strict. 

### Expected Behavior

If we remove the type overrides, we should get following results:

- I think we shouldn't get conflicts between wagmi/viem anymore
- We can use the stricter `AddressType` in scaffold-eth-2 "internals" (bbetter type-safety) ...
- ... while still maintaining newcomer-friendly plain `string` address typing for the components we expect them to use. (better DX especially for newcomers) ...
- ... without leading them to believe recent wagmi and viem versions take plain strings if they end up using those libraries without our wrappers in some project.

### Steps To Reproduce

_No response_

### Anything else?

_No response_

sverps:
## Description

Cleanup of AddressType overrides

## Additional Information

## Related Issues

Fixes #514

Your ENS/address: solidixen.eth


sverps:
I tried updating viem and wagmi on this branch, and didn't run into any type issues. @technophile-04 would be good if you also had a look and check if the issue you used to encounter can happen here, maybe there's something I missed when trying to reproduce.

rin-st:
Hey @sverps !

Looks good, but `next:check-types` script gives me `node_modules/viem` errors again.

Found 684 errors in 42 files.

sverps:
> Looks good, but `next:check-types` script gives me `node_modules/viem` errors again.

Hmm, strange, did you do anything in particular, or did you just checkout the branch and then yarn install? Cause I didn't seem to get any errors on next:check-types

rin-st:
no, just checkout -> install -> next:check-types

sverps:
I tried a bunch of different things, but can't seem to reproduce this behavior. Can you try removing your node_modules and re-run yarn install?

rin-st:
> Can you try removing your node_modules and re-run yarn install?

It worked!

I'm wondering is it only for me or for other users too. Hope we don't need to force them to remove node_modules. Let's wait a bit and see how it works for others.

technophile-04:
> Current Behavior
To not have things break, we currently override the AddressType in abi.d.ts to be string. This resulted in broken types when viem and wagmi had versions that didn't match well.
~ From #514

> I also updated viem and wagmi on this branch, and didn't run into any type issues. ~From https://github.com/scaffold-eth/scaffold-eth-2/pull/515#issuecomment-1705333171

We actually already solved this at #499, So now using a different version of `viem` / `wagmi` should not cause any issues in the future 

The solution was: separately targeting `Viem's` AddressType as suggested in `abitype` [docs extending config](https://abitype.dev/config#overview) and we are now doing this at : 

https://github.com/scaffold-eth/scaffold-eth-2/blob/82d6dca9b7dd0f72d79fefaea1002087c522b88c/packages/nextjs/types/abitype/abi.d.ts#L3

--- 

> ... while still maintaining newcomer-friendly plain string address typing for the components we expect them to use. (better DX especially for newcomers) ... 
> ... without leading them to believe recent wagmi and viem versions take plain strings if they end up using those libraries without our wrappers in some project.
~ From #514 

> And could you please explain how it decrease dx? 
~ From https://github.com/scaffold-eth/scaffold-eth-2/issues/492#issuecomment-1688070350

Screenshot 2023-09-05 at 1 10 26 PM

The reason I don't like so much stricter type is because it does not go well with other tool's / libraries. As we can see in the above case even though I have the initial state set as "0x...." react's `useState` infers it as `string` which causes issue in `args` 

And now I have to assert it at useState : `useState<Address>`

The error at `args` gets solved but now I get an error at `input` : 
Screenshot 2023-09-05 at 2 02 42 PM


Another option might be instead of asserting it at `useState`  we do check at args : ` args: [isAddress(inputAddress) ? inputAddress : undefined],`

But beginners / quick tinkerers would rather just show the error on UI which wagmi / viem nicely gives you : 
Screenshot 2023-09-05 at 1 26 05 PM


Also if people use other libraries, for examle `Uniswap sdk` or `Safe SDK` it  have function [`getAddress`](https://docs.safe.global/reference/protocol-kit#getaddress) but still I need I need to do extra checks while passing it to args, etc 


LOL really sorry for ranting 😂 and over-exaggerated examples, But I feel since wagmi/viem natively support overriding ( [docs extending config](https://abitype.dev/config#overview) ) I think we can do and We can mention it in our docs.scaffoldeth.io about it that we have overridden it


But I completely get your points and makes full sense to me 🙌 I am up for overriding if we all come to a consensus 

--- 


Nevertheless, I like these changes that we are using a consistent `Address` type instead of keeping it `string` 

In case if we don't override `AddressType: string`, I think we can mention in our docs that we have overridden it, and If you want to be stricter with types you can just delete the `abi.d.ts` file and everything will work and people can also be stricter OR by default have them stricter and we can tell people that they can add `abi.d.ts` to have types beginner friendly 🙌

technophile-04:
> I'm wondering is it only for me or for other users too. Hope we don't need to force them to remove node_modules.

It worked for me directly without needing to remove `node_modules` 🙌,  I just checked out and did `yarn install` and `yarn next:check-types` and everything was passing 🙌

sverps:
Great points of discussion, @technophile-04 :+1: 

On thinking a bit further, I tried to boil it down for myself what I think would be ideal:

- We want good DX, targeting also people that are fairly new to TypeScript. So with that in mind, I agree with your points that letting them use addresses that are of type `string` would be good.
- The places where they should be allowed to do so are, in my opinion, mainly our wrapper hooks and custom scaffold-eth components.
- Using viem/wagmi directly, is probably already a step further in their learning process, and devs that do that likely have some better understanding of types already. So I'd expect them to better understand the `Address` type.
- Overriding types via `d.ts` files is an even more advanced concept, and I feel it's not something obvious even for slightly more advanced users. That's why I'd prefer not to do that and opt for keeping things simpler.

So to conclude, I'd try and find all touch points where users would use addresses in our custom hooks and components and ensure those places to allow `string`, then cast them to `Address` in our internal implementation (i.e. in the `[args]` example see my last commit).

So in any case, no need to rush anything, I probably still need to fix return types from our hooks to return it as string.

(Need to let this simmer a bit more, the stew isn't quite ready yet :stew: :sweat_smile: )

damianmarti:
> I just checked out and did `yarn install` and `yarn next:check-types` and everything was passing 🙌

I did the same and everything was passing here too!

technophile-04:
> The places where they should be allowed to do so are, in my opinion, mainly our wrapper hooks and custom scaffold-eth components.
> Using viem/wagmi directly, is probably already a step further in their learning process, and devs that do that likely have some better understanding of types already. So I'd expect them to better understand the Address type.
> Overriding types via d.ts files is an even more advanced concept, and I feel it's not something obvious even for slightly more advanced users. That's why I'd prefer not to do that and opt for keeping things simple

Yup yup, I agree and this makes complete sense to me !! Thanks Samuel 🙌

sverps:
Closed in favor of #630

Issue #528, opened 2023-09-08
bug: Block explorer skips transactions


portdeveloper:
### Is there an existing issue for this?

- [X] I have looked through the [existing issues](https://github.com/scaffold-eth/scaffold-eth-2/issues)

### Current Behavior

Currently block explorer skips transactions if they are coming too fast and it also is pretty slow.

https://github.com/scaffold-eth/scaffold-eth-2/assets/108868128/54a64bc6-95a9-49a7-9ecf-b740316f0e0a

### Expected Behavior

By upgrading to a websocket transfer from http, we get instant feedback in the ui, in addition to not skipping any transactions.

https://github.com/scaffold-eth/scaffold-eth-2/assets/108868128/be5b70c4-e67e-4e95-8777-a77b30c6088b

luksgrin:
Hello! I just noticed you were also struggling with the blockexplorer. [This issue](https://github.com/scaffold-eth/scaffold-eth-2/issues/524) is _kind of_ related... check it out!

It would be a universal benefit if the [`useFetchBlocks`](https://github.com/technophile-04/foundry-goerli-fork-524/blob/foundry-goerli-fork/packages/nextjs/hooks/scaffold-eth/useFetchBlocks.ts) hook gets optimized 😍 .

Anyways, these are my two cents 😅

portdeveloper:
> Hello! I just noticed you were also struggling with the blockexplorer. [This issue](https://github.com/scaffold-eth/scaffold-eth-2/issues/524) is _kind of_ related... check it out!
> 
> It would be a universal benefit if the [`useFetchBlocks`](https://github.com/technophile-04/foundry-goerli-fork-524/blob/foundry-goerli-fork/packages/nextjs/hooks/scaffold-eth/useFetchBlocks.ts) hook gets optimized 😍 .
> 
> Anyways, these are my two cents 😅

Thanks for letting me know of this. I am also thinking about not having the transactionReceipts as a seperate state variable. 
Since this PR only concerns the hardhat version I think we should merge it then we can check if it works with anvil, or we can directly optimize the block explorer for anvil use, since anvil is used more than hardhat these days.

technophile-04:
> Since this PR only concerns the hardhat version I think we should merge it then we can check if it works with anvil, or we can directly optimize the block explorer for anvil use, since anvil is used more than hardhat these days.

Yup, we can tackle it one by one 🙌

[...]

PR #529, opened 2023-09-08
useFetchBlocks hooks refactor

portdeveloper:
## Description

Solves #528 . Also I have added wallet and test actions to the client with the websocket connection.
I removed the isLoading state too, because it already is loading so fast with wss, using that isLoading state actually makes the experience worse.
I plan to optimize and make the code for useFetchBlocks a little bit better before merging. Feel free to correct me and share your ideas!

## Additional Information

## Related Issues

_Closes #528 _

Your ENS/address: portdev.eth


packages/nextjs/hooks/scaffold-eth/useFetchBlocks.ts
```diff
@@ -65,55 +60,57 @@ export const useFetchBlocks = () => {
...
         }
       } catch (err) {
         setError(err instanceof Error ? err : new Error("An error occurred."));
       }
     };
 
-    return client.watchBlocks({ onBlock: handleNewBlock, includeTransactions: true });
-  }, [blocks, client, currentPage]);
+    return extendedClient.watchBlocks({ onBlock: handleNewBlock, includeTransactions: true });
+  }, [blocks, currentPage]);
```
technophile-04, line 106: We can probably remove `blocks` from the dependency array ? since we don't to subscribe on every new block added right?

packages/nextjs/utils/scaffold-eth/client.ts
```diff
@@ -0,0 +1,12 @@
+import { createTestClient, publicActions, walletActions, webSocket } from "viem";
+import { hardhat } from "wagmi/chains";
+
+const transport = webSocket("ws://127.0.0.1:8545");
+
+export const extendedClient = createTestClient({
```
technophile-04, line 6: We can rename it to maybe `testClient`  also, maybe instead of creating new file we could collocate it inside `useFetchBlock` itself since its used only there and no other place

We can always abstract it in other file if its used at other places later 🙌

technophile-04:
Tysm @portdeveloper, this is looking great !! Added very minor comments 

But everything looks and works great !! 

---

Also noticed that `yarn fork`  does not show any transactions on blockexplorer now, it was used to show things earlier if you wait a lot as mentioned in this https://github.com/scaffold-eth/scaffold-eth-2/issues/528#issuecomment-1715039609 video

I think its because now we are using `websockets`  and if you inspect chrome developers tools `network` tab I see bunch of `"canceled"` `eth_blockNumber` request when you `yarn fork` and go to `blockexplorer` page but not completely sure if its because of `websockets`   

Nevertheless, I think these changes work great on Localchain and almost ready to merge 🙌

portdeveloper:
Thanks @technophile-04 for your valuable feedback. I have fixed the things you have pointed out. After this is merged I will look into test actions so that the users can have a more fine grained control over their local environment. `yarn fork` is an issue too. I have tested it myself with a websocket connection to infura and the problem is that the useFetchBlocks, with its current logic, tries to get all blocks and all transactions in them before showing any.

technophile-04 (approved):
Thanks @portdeveloper, I think it looks and works great now!

> After this is merged I will look into test actions so that the users can have a more fine grained control over their local environment. `yarn fork` is an issue too. I have tested it myself with a websocket connection to infura and the problem is that the `useFetchBlocks`, with its current logic, tries to get all blocks and all transactions in them before showing any.

Yup, it would be really helpful if we could fix that so we only make requests for the latest `20` blocks and fetch their corresponding tx receipts & details that's it we shouldn't make request for other blocks

PR #532, opened 2023-09-13
fix ETH price on sepolia network showing 0

technophile-04:
## Description

It seems wagmi chain `sepolia` has `nativeCurrencySymbol` as "SEP" instead of "ETH" and we had this guard condition in `fetchPriceUniswap` :
```ts
if (configuredNetwork.nativeCurrency.symbol !== "ETH" && !configuredNetwork.nativeCurrencyTokenAddress) {
    return 0;
  }
```

Which was returning the price as 0 for Sepolia chain originally encountered here -> https://github.com/scaffold-eth/se-2-challenges/pull/84#issuecomment-1716159180


rin-st (approved):

technophile-04:
Tysm Rinat for review, Merging this 🙌

carletex:
If we have to add another symbol in the future, we might want to use an array to store them ["ETH", "SEP", etc] with a good descriptive name + use `!includes` in the condition.

Thanks for fixing this!

PR #540, opened 2023-09-22
Add indexed args to events

rin-st:
## Description

Started [here](https://github.com/scaffold-eth/se-2-challenges/pull/93)

**Problem:**
After migrating to viem and wagmi 1+ access to event parameters changed.

Before (ethers.js):
All the parameters of the event were accessible by index of argument in [listener function](https://github.com/scaffold-eth/scaffold-eth-2/pull/352/files#diff-4b13aae8a79eed1d29876f892dd1b17adeace865ad0ea5deb4baf53b00e479b1L36) or basically by index in args array.

After. We have two cases:
- parameters of event in solidity contract are named, for example: `SomeEvent(uint256 arg1, uint256 arg2)`. We can access that event parameters on frontend **only** by param name:  `const {arg1, arg2} = event.args`
- parameters not named, for example: `SomeEvent(uint256, uint256)`. We can access that event parameters on frontend **only** by index:  `const [arg1, arg2] = event.args` or the same `const arg1 = event.args[0]; const arg2 = event.args[1];`

so writing frontend we're depending on how contract events are written and can't write universal code

**Solution:**
Add indexed parameters for frontend for named event parameters too. So for case with named params `SomeEvent(uint256 arg1, uint256 arg2)` we can access event params by param name OR by index of param.
 And it means we can always access event parameters at least by index

## Additional Information

Your ENS/address:
0xrinat


technophile-04 (approved):
Tysm @rin-st 🙌, Looking great to me ! 

Just fixed types for EventSubsriber because now people can also access through the index : 

Screenshot 2023-09-26 at 5 18 13 PM

Regarding `useScaffoldEvenHistory` currently in the `main`  branch too `data : any[]` we should fix it to other PR  

Let's see what others have to say about indexed args 🙌

rin-st:
> Just fixed types for EventSubsriber

Nice! 👍

damianmarti (approved):
LGTM and it seems a nice improvement. Thanks @rin-st !!

rin-st:
@technophile-04 your types doesn't work for other contract, could you check it?
https://github.com/scaffold-eth/se-2-challenges/pull/93#issuecomment-1736290799

technophile-04:
Actually Rinat I think we missed updating `contract.ts` in that PR looking at files changed => https://github.com/scaffold-eth/se-2-challenges/pull/93/files

Also I think we updated the types just for `useScaffoldEventSubscriber` and I see in `stakings.tsx` we are using `useScaffoldEventHIstory` : 

I just tried adding it to `contract.ts` and updated viem version (yes we need to do that in challenges), it seems to work for `useScafffoldEventSubscriber` : 
Screenshot 2023-09-29 at 9 37 07 PM

PR #444, opened 2023-07-24
Upgrade to pnpm

technophile-04:
## Description

Updated the pkgManager to pnpm for trial as discussed in #419 

~For deploying contract you need to use `pnpm hardhat:deploy` instead of just `pnpm deploy` because [`deploy`](https://pnpm.io/cli/deploy) seem reserved keyword~
~Will try to find if there exist any solution to above problem 🙌....But I think its ready to test and give a trial~ 

This is solved please check https://github.com/scaffold-eth/scaffold-eth-2/pull/444#issuecomment-1667241333 🙌

Issue #419, opened 2023-07-13
bug: Vercel deployment fails


technophile-04:
### Is there an existing issue for this?

- [X] I have looked through the [existing issues](https://github.com/scaffold-eth/scaffold-eth-2/issues)

### Current Behavior

Running : `yarn vercel` 

![Screenshot 2023-07-14 at 12 16 18 AM](https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/36c05f2d-5920-4b54-a6f5-2b519b1e43a2)

### Steps To Reproduce

**Alternatively to reproduce locally** : 
Update viem and wagmi to latest version : 
1.`yarn workspace @se-2/nextjs up wagmi viem`

Run check-types : 
2.`yarn next:check-types`

However through above steps you get different type errors then `vercel` : 
![Screenshot 2023-07-14 at 1 11 15 AM](https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/c1ef1a49-ae53-47f1-bafc-ff83ad6f2784)

### Anything else?

**I think** the reason its failing on vercel deployement and not locally is because since we have `^` for `viem` and `wagmi` in `package.json` because of which on Vercel while `yarn install` it gets latest `patch` version of both `viem` and `wagmi` and it seems that in latest `viem` and `wagmi` its broken. 

I think this problem is again related to -> https://github.com/wagmi-dev/wagmi/pull/2421#issue-1724776783

carletex:
Thanks for reporting this Shiv.

What I don't understand is why Vercel installs a different version of the packages. Even if we have `^` in `package.json`, it should take the exact version from `yarn.lock` (that's the goal of this file!)

technophile-04:
> What I don't understand is why Vercel installs a different version of the packages. Even if we have ^ in package.json, it should take the exact version from yarn.lock (that's the goal of this file!)

Yeah this makes sense !! 

I tried going through the deployment logs and found this : 
![Screenshot 2023-07-14 at 2 35 47 PM](https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/0bdc33b1-72d8-4ad3-8fd8-1a1a91af6baf)

Also if you look at the souce uploaded it dosen't have `yarn.lock` (I think because we are just uploading nextjs workspace and `yarn.lock` is present at root ?)

[...]

PR #541, opened 2023-09-24
add Prettify type helper locally

technophile-04:
## Description

It seems that NextJs deployments fails since in latest version `viem`(1.12.x) they have moved `Prettify` type from "viem/dist/types/types/utils" => "viem/types/utils"

Now due to #419, our version viem exports from "dist"  but when we upload to vercel it uses new viem where the location from which it is exported is changed 

Reported https://github.com/BuidlGuidl/PWA-burner-wallet/pull/1#issuecomment-1732340873 & https://github.com/scaffold-eth/scaffold-eth-2/pull/444#issuecomment-1732584888

--- 

As mentioned by Rinat https://github.com/scaffold-eth/scaffold-eth-2/pull/444#issuecomment-1732584888 adding`Prettify` locally works and also it makes sense to localize this "helper type"  since it is famous and we also shouldn't rely on Viem to export this for us because they might move to another place in future/change name/remove without mentioning it since its not their main api


packages/nextjs/utils/scaffold-eth/contract.ts
```diff
@@ -8,11 +8,20 @@ import {
...
 
+/**
+ * @description Combines members of an intersection into a readable type.
+ * @example
+ * Prettify<{ a: string } & { b: string } & { c: number, d: bigint }>
+ * => { a: string, b: string, c: number, d: bigint }
+ */
+type Prettify<T> = {
+  [K in keyof T]: T[K];
+} & unknown;
```
technophile-04, line 23: Using `unkown` here instead of `{}` since we have eslint rule setup : 

Screenshot 2023-09-24 at 9 23 08 PM


Also  using `unknown` is completely fine too as  suggested here : 
Screenshot 2023-09-24 at 9 47 23 PM

OG post => https://www.totaltypescript.com/concepts/the-prettify-helper

rin-st (approved):

Pabl0cks (approved):

carletex:
> Now due to https://github.com/scaffold-eth/scaffold-eth-2/issues/419, our version viem exports from "dist" but when we upload to vercel it uses new viem where the location from which it is exported is changed

Yeah, we really need to tackle this issue, where that deps local version != deployed version. https://github.com/scaffold-eth/scaffold-eth-2/pull/444 is a solution, but I'm not 100% sold on it. I'll comment soon.

The other option is locking the version in `package.json` but I feel you guys don't like that? I don't see any issues taking that approach. When we want to update we just update the version manually (which arguably makes us more conscious about updating / the version we are updating to), or even temporarily adding `^` before updating a given dependency and locking it later. Maybe not as convenient as just running `yarn update` on a given dependency... by it's definitely better than having a version mismatch between the local & prod env. Am I missing something?

----

Also for this PR: Shouldn't we just update viem (so it saves in the .lock file executed locally). It'll still fail for SE-2 forks/clones, but it'll be fixed on fresh installs.

Thanks!

technophile-04:
> The other option is locking the version in package.json but I feel you guys don't like that? I don't see any issues taking that approach. When we want to update we just update the version manually (which arguably makes us more conscious about updating / the version we are updating to), or even temporarily adding ^ before updating a given dependency and locking it later. Maybe not as convenient as just running yarn update on a given dependency... by it's definitely better than having a version mismatch between the local & prod env. Am I missing something?

Doing some very basic research, I don't see any big drawbacks of removing "^"  so I am completely up for it 🙌

but would love to know other thoughts if there are any big drawbacks that I might have missed also cc @damianmarti @sverps maybe too

rin-st:
> The other option is locking the version in package.json but I feel you guys don't like that? I don't see any issues taking that approach

The main difference I see is that our dependencies (for example "our-dep": "1.0.0") have their package.json with their dependencies inside it (for ex "child-dep": "^2.0.0"), and that child deps will be different without package.lock. It's a small chance that one child dep will break our dep, but we have hundreds of child deps and because of that chances of breaking again someday are not too small.

---

regarding this pr fix, we should avoid importing like that, because this type is not even exported from viem

damianmarti:
I always prefer to have a fixed version of everything, I really don't like ^ too much because sometimes there are some breaking changes between minor versions (if the .lock is used, this is not an issue because the version is fixed there). 

So, I don't see any issue with having a version fixed at the package.json (this library or whatever), and being explicit when we want to update it to another fixed version.

You guys are great! You are on all the little details! :-)

carletex:
Thanks @technophile-04 @rin-st @damianmarti for your view on this <3

I'll move the discussion to the pnpm PR, since If we go with fixed versions, we could rethink pnpm.

I'll prepare my allegations soon haha

PR #548, opened 2023-09-29
remove `parseEther` from useScaffoldContractWrite

technophile-04:
## Description

As discussed in https://github.com/scaffold-eth/scaffold-eth-2/discussions/516, I agree with the discussion and also makes sense now 🙌

also btw the type of `value` that we had was a bit too harsh as it was typed as `${number}` lol which was a PITA to get it right until you write your own validate function helper or assert it with `as` or just pass in literal value like "0.1".

Here is an example that might frustrate users: 
Screenshot 2023-09-29 at 3 46 33 PM

Nevertheless, this PR removes `parseEther` from `useScaffoldContractWrite` so we leave it to users, which solves all the issue 🙌

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
@@ -1,6 +1,5 @@
 import { useState } from "react";
 import { Abi, ExtractAbiFunctionNames } from "abitype";
-import { parseEther } from "viem";
 import { useContractWrite, useNetwork } from "wagmi";
 import { getParsedError } from "~~/components/scaffold-eth";
 import { useDeployedContractInfo, useTransactor } from "~~/hooks/scaffold-eth";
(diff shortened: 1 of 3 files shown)


packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
```diff
@@ -73,7 +72,7 @@ export const useScaffoldContractWrite = <
           () =>
             wagmiContractWrite.writeAsync({
               args: newArgs ?? args,
-              value: newValue ? parseEther(newValue) : value && parseEther(value),
+              value: newValue ? newValue : value && value,
```
rin-st, line 75: newValue || value :)

technophile-04: lol yup! tysm 🙌

Btw I actually updated it to `??` because checkout this condition : 

```tsx
  const { writeAsync, isLoading } = useScaffoldContractWrite({
    contractName: "YourContract",
    functionName: "setGreeting",
    args: [newGreeting],
    value: parseEther("1"),
  });
  <button onClick={() => writeAsync({ value: parseEther("0") })}>btn</button>;
```

Here we want to respect `0n` and use it instead of the previous value passed to hook

carletex (approved):
Yep, glad we are merging this in!

PR #550, opened 2023-09-30
added feature to give input suggestions for address

FilipHarald:
## Description

Instead of having to look around in the UI and code for the address, you could instead get a drop-down for addresses that's relevant.

https://github.com/scaffold-eth/scaffold-eth-2/assets/6756284/bdd0d3e0-ae93-4dfe-941c-1bd68860bdd1

## Additional Information

## Related Issues

I have not created an issue, but tried to get some opinions [through this discussion](https://github.com/scaffold-eth/scaffold-eth-2/discussions/506). I got a few :heart: , but no comments :smile: .

Your ENS/address: harmont.eth


rin-st:
Interesting idea! Some thoughts:
- owner doesn't exist in most of the contracts, so there is no need to hardcode it. It's better to get all the public `address` values from the contract.
- book icon should be optional
- addresses list wrapper should be absolute
- not clear how it will look with short inputs

Let's see what others think
cc @carletex @technophile-04 @damianmarti

FilipHarald:
Thank you for your feedback @rin-st .
- I agree that this can be problematic on larger lists of addresses. I'll look on a solution for this.
- What do you mean that the book icon should be optional? That the whole feature should be optional?
- What does it mean that the wrapper should be absolute?
- What do you mean what do you mean with short inputs? This is only intended for addresses which are all 34 chars long, right?

rin-st:
> That the whole feature should be optional?

Yes. Sometimes it's clear that we don't need addresses from address book.

> What does it mean that the wrapper should be absolute?

[CSS position](https://developer.mozilla.org/en-US/docs/Web/CSS/position) . So all the blocks below will not jump on opening

> This is only intended for addresses which are all 34 chars long, right?

Yes, but in some cases input can be short, for example on mobiles.

----

And I'm not really sure we need that feature. Let's wait for other opinions

FilipHarald:
@rin-st thank you for taking your time to give feedback. I completely understand that we might not need, or even want, the feature. And that is completely fine.

But I want to practice and I think your remarks makes sense to start with :+1:

technophile-04:
Hey, thanks @FilipHarald ! 

In the first go it looks good feature to add to increase DX while testing, but when you think of more broader perspective I don't think it won't be that helpful 

- as Rinat mentioned the list of public address can really get huge, since we not only fetching a particular contract `public` address but all the deployed contracts public address (Also people like to inherit from other contracts which makes it more long)

- Umm, I see this feature only seems good for the Debug contact page because if you deploy your app to prod you really don't want user using you app to mention public addresses of your contracts instead you can fetch it internally and prefill it for them (don't know if that makes sense 😅) 

So I kinda vote for not having this feature at all(thinking of more from general perspective ), but really sorry for missing out on the created discussion 🙏 and thanks @FilipHarald for crafting this PR since now it more clear how implementation looks 🙌

packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
```diff
@@ -51,34 +62,74 @@ export const AddressInput = ({ value, name, placeholder, onChange, disabled }: C
     [onChange],
   );
 
+  const addressBook = useAddressBook();
```
technophile-04, line 65: I don't want to run this hooks logic always, we can maybe pass `showAddressBook` to this hook and make RPC calls only if `showAddressBook` is true since as discussed there can be a lot of public addresses and we should avoid unnecessary RPC calls 🙌

PR #535, opened 2023-09-14
App router migration

Anurag-Raut:
## Description

Migrated from Page router to new Approuter .
upgraded to Next v13.5.2 since the app router is stable from v13.4 +

## Additional Information

- [x] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [x] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

_Closes #526 _

Your ENS/address: 0x72aA8A79D02705444CC6918A6766Be03AD4C29F5

PR #551, opened 2023-09-30
updated footer to use BuidlGuidl logo

FilipHarald:
## Description

Updated the footer with BG-logo. 
Screenshot from 2023-09-30 15-26-26
Screenshot from 2023-09-30 15-26-15

https://github.com/scaffold-eth/scaffold-eth-2/assets/6756284/d24910a9-fe21-4bd6-93d7-c1741179a1d2

## Additional Information

## Related Issues

Logotypes and ideas taken from [this discussion](https://github.com/scaffold-eth/scaffold-eth-2/discussions/549).

Your ENS/address: harmont.eth


packages/nextjs/components/Footer.tsx
```diff
@@ -42,18 +45,19 @@ export const Footer = () => {
...
-                  className="underline underline-offset-2"
-                >
-                  BuidlGuidl
-                </a>
+                Built with <HeartIcon className="inline-block h-4 w-4" /> at
               </p>
+              <div className="relative w-20 h-4 pb-5">
+                <a href="https://buidlguidl.com/" target="_blank" rel="noreferrer">
+                  {isDarkMode ? (
+                    <Image alt="SE2 logo" className="pb-1 cursor-pointer" fill src="/bg-logo-light.svg" />
```
rin-st, line 55: Change svg to tsx component, see for example `/components/example-ui/assets`. Use https://react-svgr.com/playground/?exportType=named to convert attributes to react-compatible. Change `fill="#color` to `fill="currentColor"`, so it will take color of the parent. So you'll have one component and don't need `useDarkMode` anymore and also use whatever color you want.

Also I don't like the space between logo and text, but I think we need to ask designers to change that

rin-st: I think it's better to take only icon without text here

ZakGriffith: If we want to keep the text as well, I can quickly increase the spacing between the logo and the text.   Here is a quick go at it:  https://www.figma.com/file/AD1r8jWVZrCke80K0kgfH8/Untitled?type=design&node-id=0%3A1&mode=design&t=ufPl843Dm68mBax2-1

ZakGriffith: Can also use a separate file for the logo and for the text so then y'all can work your magic.

FilipHarald: Thank you for your help @rin-st @ZakGriffith .

@rin-st , I did your suggested SVG-magic, renamed the file and changed to only use one file, neat!

I don't agree that there should be more spacing of the logo. But I tried some different approaches and attach their printscreens below. Let me know what you think.

Only icon and with underlined text.
Screenshot from 2023-10-01 10-30-46

Icon + text with extra spacing from v2 given by @ZakGriffith [here](https://www.figma.com/file/AD1r8jWVZrCke80K0kgfH8/Untitled?type=design&node-id=0%3A1&mode=design&t=ufPl843Dm68mBax2-1).
Screenshot from 2023-10-01 10-26-26

Icon + text from [original from here](https://www.figma.com/file/jPaXdmHNfbTSwBWwENr6Iq/BG-Media-Kit?type=design&node-id=1-29&mode=design&t=kUu7PIZFibkQ8mrz-0).
Screenshot from 2023-10-01 10-23-30

rin-st: Voting for first. Best spacing and it's clear that you can click it. 2-3 are good too, but I think with spacing like in 1 it could be better.

> Can also use a separate file for the logo and for the text so then y'all can work your magic.

I think we don't need it because devs can break base proportions between Icon and text. Probably better to have different versions of (Icon_spacing_text) if proportions or spacings are different.

But here since we use it just on footer we need just one version or Icon separately if we decide to go with first version.

FilipHarald: OK, I updated code with only icon (first pic). I can just revert it if we need to change a again.

packages/nextjs/components/Footer.tsx
```diff
@@ -42,18 +43,19 @@ export const Footer = () => {
...
+                Built with <HeartIcon className="inline-block h-4 w-4" /> at
               </p>
+              <a
+                className="flex justify-center items-center gap-1"
+                href="https://buidlguidl.com/"
+                target="_blank"
+                rel="noreferrer"
+              >
+                <BuidlGuidlLogo className="w-3 h-5 pb-1" />
+                <div className="underline">BuidlGuidl</div>
```
technophile-04, line 57: NITPICK : Maybe convert it to `span` 😅 just to me more semantic

technophile-04 (approved):
Looking really nice and clean !! Tysm @FilipHarald for the PR  !! also thanks to Rinat and Zak for review 🙌

packages/nextjs/public/assets/BuidlGuidlLogo.tsx
carletex, line 1: Since this is not an SVG file anymore, we shouldn't expose this file in the `public` folder.

We have a `Spinner.tsx` in `components`, we could add it there.... or maybe move both to `components/assets`? cc @rin-st @technophile-04 

This will change with when we merge the App Router tho.

packages/nextjs/components/Footer.tsx
FilipHarald: Alright, I don't know who is allowed to resolve this thread. But the current PR is with nbr 1. 

@ZakGriffith @rin-st

packages/nextjs/public/assets/BuidlGuidlLogo.tsx
FilipHarald: @carletex when you say "both" - do you mean the logo and the spinner?

I can move both to `components/assets`. Also, I can move both to wherever they should be when "App Router" is merged. And then we merge this PR after that.

If we want to go for the second option - which PR/issue is that?

FilipHarald: I moved both to `components/assets` now. Let me know if I should wait for "App Router" instead. :+1:

carletex: @FilipHarald

> when you say "both" - do you mean the logo and the spinner?

Yep, both. Having the Spinner in components, and the BG logo in components/assets won't let me sleep at night haha. I'd personally go with `components/assets`

> I can move both to wherever they should be when "App Router" is merged. And then we merge this PR after that.

No worries, we don't need to wait to merge this. We can merge and then solve it in the PR https://github.com/scaffold-eth/scaffold-eth-2/pull/535 when we back-merge main. 

Thanks!!

packages/nextjs/components/Footer.tsx
carletex: Looking good!

carletex:
Good stuff @FilipHarald ! And thanks all for the review.

- Pushed a little tweak to match the other links in the footer: https://github.com/scaffold-eth/scaffold-eth-2/pull/551/commits/6fd539fb54607eea8ad42fa0038c1099e73d1f83

**One improvement that we could make in another PR:** I miss a transition when hovering the link (like changing the opacity to 90% with a 300ms transition). We could just add to the `.link:hover` selector... or create a custom class in `tailwind.config.js`. Also, maybe changing the footer link to use the `link` class? (instead of manually setting the underline. I like the offset, so we might want to set the default for the `link` class... or the new custom class.)

FilipHarald:
Thank you!

OK, I'll have a look at it when I have time :)

PR #531, opened 2023-09-12
foundry: add chain id 31337 while forking chain

jrcarlos2000:
## Description

Addresses issues #524 

## Additional Information

- [x] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [x] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

_Closes #524

Your ENS/address: jrcarlos2000.eth

PR #555, opened 2023-10-03
Update CLI with latest main

carletex:
`main` Back-merge from https://github.com/scaffold-eth/scaffold-eth-2/commit/02dd02203300c4f4f74f04ba8d853e20fb962574 to https://github.com/scaffold-eth/scaffold-eth-2/commit/b33a87d8e023617faedcb8c5f0a930f0f5365097. Check commits.

I created a `cli-backmerge` branch from `cli`, where I merged `main` and fixed conflicts. That way we can create this PR and check & test the PR before merging to CLI. Maybe we could add this to DEV_GUIDE md file

**Note:** Until main === cli, let's do this (back-merge main) once a week. If not, it gets messy.


technophile-04:
Was trying to test it and when I do `yarn next:check-types` I get bunch of react types error  : 
Screenshot 2023-10-03 at 7 43 20 PM

also `yarn next:lint` doesn't give me any error so might be something to do with TS version or `types/react`, will try to debug 🙌

Steps to reproduce => select `hardhat` as option and choosed yes for installation

templates/base/packages/nextjs/package.json
```diff
@@ -39,7 +39,6 @@
     "@trivago/prettier-plugin-sort-imports": "~4.1.1",
     "@types/node": "~17.0.45",
     "@types/react": "~18.0.21",
```
technophile-04, line 41: Changing this to `~18.2.24` fixes the issue

technophile-04:
Curious about why this happened, I did some digging, and  here's my finding : 

SE-2 `main` branch `yarn.lock` has this : 
Screenshot 2023-10-03 at 8 26 01 PM


When you create a new template from this PR, its `yarn.lock` contains this : 
Screenshot 2023-10-03 at 8 30 34 PM

notice that it create two blocks one with version `18.2.24` and other with `18.0.38` is this what causing the problem ? 

I also tried removing "~"  from `~18.0.21` `package.json` and regenrating the app but still it makes two block 

is that some other package is dependent on latest version of `@types/react` because of which it is generating two blocks? I might be wrong here and lol I have a knowledge gap on how this works 

is this related somewhere to @rin-st comments => https://github.com/scaffold-eth/scaffold-eth-2/pull/541#issuecomment-1734204842 (might be unrealted)
--- 

For now, the solution seems https://github.com/scaffold-eth/scaffold-eth-2/pull/555#discussion_r1344299719 but just curious why this happened and this affects us in future

carletex:
Thanks for the investigation @technophile-04 !

Some more stuff that I found (`yarn why` is great): 

This is the **CLI**:
image

This is **main**:
image

In CLI we have 2 diff versions because:
- newer version of the VERCEL package (28.15.1 => 28.15.7)
- newer version of types/react-copy-to-clipboard (5.0.4 => 5.0.5)

These two use a newer version of types/react (18.2.24).

-----

What I did then is to remove the `~` in the package.json file (only for next), so it installs the same version as in the lockfile (at least the top dependency!):

image

So, it the vercel dep is gone, but not the clipboard one. I think this is exactly what @rin-st was talking about here: https://github.com/scaffold-eth/scaffold-eth-2/pull/541#issuecomment-1734204842

Even if we install the same top dep (5.0.4) it uses a newer version of `@types/react`. Why? Because package.json file of `@types/react-copy-to-clipboard`:

```
    "dependencies": {
        "@types/react": "*"
    },
```

:cry: 

----

So not having a lockfile is tricky.

We can remove `~`, but things can always break in the future because a new release of sub-dependency updates and breaks something.

A decent solution:
- Remove `~`
- Have some kind of automated testing that runs every day, installs SE-2 with different options, runs some commands, and checks errors. That way we can identify issues early.

Not ideal, but better than nothing.

Thoughts?

templates/base/packages/nextjs/package.json
carletex: Maybe having `^` instead of `~` or a fixed version for all the `@types` packages could solve this issue (and future issues too?)

carletex: It seems pretty standard on all the sub-deps in `@types` packages.

technophile-04: Yup 🙌 I think adding `^` seems fair solution, I tried fixing @types/react to `18.0.21` and it didn't work 

Also btw I think we should make a quick another patch PR cli  because `npx create-eth@latest` is broken now because of this

carletex:
@technophile-04 I think I understand what you are saying.

doing `sum : latest` is not a very good practice for any lib (but it seems common for @types libs). But yeah, the risk is there

> It will be really hard for us to debug too right ? Since A might using sum very deep down which causes some of A api's break which we are using in SE-2 also this will be hard to catch in automated tests?

If we have good automated tests, I guess we could detect that something is not working (deploy, start, checking types, etc), right? And the we can investigate the root of it and fix it (it'll prolly be a versions that needs to be updated)

I think I'm gonna update @types to ^ for now.

> Also btw I think we should make a quick another patch PR cli because npx create-eth@latest is broken now because of this

Maybe we can merge this one? I'll send the  update in a bit.

carletex:
This is working for me @technophile-04 

If it works for you.. let's merge and publish!

rin-st:
Great investigations! Experimented with it too but didn't find something better

>  I think this is exactly what @rin-st was talking about here

I thought: child dep breaks our dep -> our dep breaks our app. But it's even worse. child dep breaks our app 🤷‍♂️ . I didn't know about that problem before and thought package managers can manage it and use different versions needed for our dep and child dep. But it's not always true.

Also I thought `"*"` takes already installed libs with priority to our deps. Again it's not true. `"lib": "*"` is almost equal `"lib": "latest"`.

> I think I'm gonna update types to ^ for now.

It works for me, but I believe we also need to think later about removing that `"*"` deps. Because it will break cli after every major update of `@types/react` and `^` for types will not help. It seems not so difficult to change that `react-copy-to-clipboard`. For example there is a hook in [usehooks-ts](https://usehooks-ts.com/react-hook/use-copy-to-clipboard) (I think we need to move from usehooks-ts too, but it's another story). 

> Have some kind of automated testing that runs every day, installs SE-2 with different options, runs some commands, and checks errors. That way we can identify issues early.

Not sure how to do it, but if it's possible why not.

@technophile-04 sorry I don't know the answer to your question. Will try to read it again tomorrow 😄 

Thanks!

ps. FYI, if you didn't see it yet https://github.com/facebook/react/issues/24304#issuecomment-1094565891

rin-st (approved):

technophile-04:
> doing sum : latest is not a very good practice for any lib (but it seems common for https://github.com/types libs).

Yup yup 💯💯 I rarely see lib using `latest` or `*` for their dependencies and yes it seem common with @types which makes sense

Tysm for answering 🙌

> @technophile-04 sorry I don't know the answer to your question. Will try to read it again tomorrow 😄

No worries Rinat 😆 Just ignore it trust me it was very highly kind of unreal situation cooked up in my head and lol just dumped it in that comment, but yeah as Carlos mentioned libraries mentioning `latest` or `*` for their internal dependencies is very rare

---

I tested it and works nicely also tried doign `yarn vercel` and it worked nicely => https://test-cli-h.vercel.app

Just pushed a small tweak at foundry at [4f6fcb5](https://github.com/scaffold-eth/scaffold-eth-2/pull/555/commits/4f6fcb56fd91c71bda607c959ec264874c178c76) which we did at #531 

and added chandeset 🙌

Tysm guys merging this!

rin-st:
> We are using library A which has internal dependency as sum : * (which is expected to return sum of given array )
Now lets suppose there is another library we are using B which has an internal dependency sum : latest (since this is latest version API has changed and instead it returns object with sum inside eg. {sum : 54} )
This will affect A right ? and A internal working will break right ?

`sum: *` is the same as `sum: latest`. So `sum` from `A` will break `A`. I hope no one does `"*"` or  `latest` for non-type deps. 

If you meant by `*` some exact version as I understand it depends. Let's say dep in `A` has version `^1.0.0` and latest version is `1.1.0`. Package manager will resolve it for one version 1.1.0 since it fits both versions. But if it's not resolvable like `^1.0.0` and `2.0.0` package manager creates two versions or sometimes create one and show warnings about wrong versions in console.

In our case it creates two versions. The question for me how and why types from `react-copy-to-clipboard` seen and used at the top level

> It will be really hard for us to debug too right ?

Not sure. I think mostly same like other error. See stack trace -> find what's wrong

carletex:
Thank y'all for this discussion. Learning a lot!

I think we should have this discussion open (maybe a new issue?) because this is something that we are going to face. I guess the base issue is not having a `lockfile` (which we could use in our advantage to allow people to use their desired package manager, as Shiv mentioned.

Something that we could explore in some cases => [yarn resolutions](https://yarnpkg.com/cli/set/resolution):

```
"resolutions": {
   "@types/react": "18.2.21"
}
```

technophile-04:
> I hope no one does "*" or latest for non-type deps.
> Let's say dep in A has version ^1.0.0 and latest version is 1.1.0. Package manager will resolve it for one version 1.1.0 since it fits both versions. But if it's not resolvable like ^1.0.0 and 2.0.0 package manager creates two versions or sometimes create one and show warnings about wrong versions in console.

Yup 💯💯, Tysm Rinat 🙌

After reading alot of issues/discussions : 

> Also I thought "*" takes already installed libs with priority to our deps. Again it's not true. "lib": "*" is almost equal "lib": "latest".

Yeah "*" considers newest version but only in case of `yarn` checkout this => https://github.com/yarnpkg/yarn/issues/4489#issuecomment-342538350

If we would have used `npm` or `pnpm` then it uses our dependency as mentioned here in this table => https://github.com/Shopify/ui-extensions/issues/324#issuecomment-1137198706


lol `@types/react` seems pretty common when you don't have `.lock` file present. 

Also this @types/react happes only when there are multiple versions of `@types/react` as mentioned here => https://github.com/software-mansion/react-native-svg/issues/1741#issuecomment-1093812197

Looking at solutions, people seem rely on yarn resolution for this and we can do something like this : 

```ts
"resolutions": {
   "@types/react": "18.2.25"
}
```

I think we are good using `^` for now since in Carlos comment first image there is one @remix/.. which uses `^`  and if you do `^` latest minor version is used and we will still have only one `@types/react` version 

> Because it will break cli after every major update of @types/react and ^ for types will not help

I think there were will long way before major version of `@types/react` is released 😅 because `@types/packageName` will only release Major version if `packageName` releases major version as mentioned [here](https://stackoverflow.com/a/63751342/16480510), and lol React V19 seems long way from now but yeah this will be the case 

But maybe by then we give people option to select different package manager like pnpm or bun 🙌

rin-st:
> I think we are good using ^ for now since in Carlos comment first image there is one @remix/.. which uses ^ and if you do ^ latest minor version is used and we will still have only one @types/react version

Agree

> I think there were will long way before major version of @types/react is released 😅 because @types/packageName will only release Major version if packageName releases major version as mentioned [here](https://stackoverflow.com/a/63751342/16480510), and lol React V19 seems long way from now but yeah this will be the case

😄 yes, but we can forget about this issue when it will happen. And I'm afraid it can happen not only for react, but for smaller packages too. Hope I'm wrong

> If we would have used npm or pnpm then it uses our dependency

Thanks! Chatgpt deceived me 😄 . It said that it works for every package manager.

> But maybe by then we give people option to select different package manager like pnpm or bun

Tried to test it, pnpm resolves * and takes 18.0.21 so it works 🙌.
But bun behaves like yarn, so the same problem for now.

PR #559, opened 2023-10-08
Faucet refactor

coderrick:
## Description

Refactored the faucet workflow. Now faucet component contains everything dealing with faucet to hopefully make all available actions more straightforward.

- Removed FaucetButton from the Header. 

- Added FaucetButton to Faucet in the Footer

- Updated FaucetButton to the default address its pointing to


technophile-04:
Hey thanks @coderrick but please read CONTRIBUTING.md before making PR, I think the main reason to have Faucet button is you can quickly grab 1eth to get initial funds (and it disappears after you balance is above it)

Hiding it in Faucet component doesn't makes sense 

closing this for now, but please free to open discussion or issue before creating PR 🙌

PR #565, opened 2023-10-10
Memo history events

rin-st:
## Description

Return `data` from `useScaffoldEventHistory` is not memoized. So because of `.map` it returns new array on every render even when `events` doesn't change

This pr fixes it

## Additional Information

Your ENS/address:
0xrinat


technophile-04 (approved):
> So because of .map it returns new array on every render even when events doesn't change

Yup agree, Tysm @rin-st  looks good 🙌

PR #569, opened 2023-10-11
Extract headerMenuLinks from Header

### LINKED ISSUE #568: Extract header main menu link to a more accesible place.
Rn, to tweak the menu links, you have to go to `Header.tsx` and alter the `navLinks` var inside the Header component. It's a bit too deep.

I think it should be similar to this:

```ts
const mainMenuLinks = [
  {
    name: "Home",
    href: "/",
  },
  {
    name: "Debug Contracts",
    href: "/debug",
    icon: <BugAntIcon className="h-4 w-4" />,
  },
  // ...
];
```

and then dynamically generate the needed code.

Not sure if it should be in the same file at the top, or a separate component like `HeaderMenuLinks.tsx` (with this name it'd be next to `Header.tsx`). Having it in a separate file might have some benefits when the CLI install modifies it (template file will be `HeaderMenuLinks.tsx`)

carletex:
Fixes #568

Makes tweaking header menu links more accessible (+ easier to tweak with the CLI). Iteration 0.1. Open to ideas on how to do it better!

Should we mention it somewhere in the code? Or just in the docs?


packages/nextjs/components/headerMenuLinks.tsx
carletex, line 1: I guess the name should start with a lowercase, right? Since we are not exporting a react component, just an array.

rin-st: I'm not sure it should be in `components` folder's top level if exporting only links array. Only if it's better for cli.
 
I think it's better to move `<NavLink />` and `headerMenuLinks.map...` to this file too, so we will have exported component
```tsx
export const HeaderMenuLinks = () => (
  <>
    {headerMenuLinks.map(({ label, href, icon }) => (
      <li key={href}>
        <NavLink href={href}>
          <span>{icon}</span>
          <span>{label}</span>
        </NavLink>
      </li>
    ))}
  </>
);
```

But in general I think we need to add a section to the docs describing for newcomers what to update/delete when creating new project. So they won't need to find these places youself

technophile-04: >  think it's better to move <NavLink /> and headerMenuLinks.map... to this file too

Yup agree, also maybe remove the abstraction of NavLink and just inline it in the HeaderMenuLinks component so that we don't clutter the headerMenuLinks file, but looks fine to me 🙌 :  

```tsx
export const HeaderMenuLinks = () => {
  const router = useRouter();

  return (
    <>
      {headerMenuLinks.map(({ label, href, icon }) => {
        const isActive = router.pathname === href;
        return (
          <li key={href}>
            <Link
              href={href}
              passHref
              className={`${
                isActive ? "bg-secondary shadow-md" : ""
              } hover:bg-secondary hover:shadow-md focus:!bg-secondary active:!text-neutral py-1.5 px-3 text-sm rounded-full gap-2 grid grid-flow-col`}
            >
              {icon}
              <span>{label}</span>
            </Link>
          </li>
        );
      })}
    </>
  );
};

```

Also I was thinking maybe rename the file from `Header*Links` to `Navbar*Links` lol it a bit personal but I always do cmd + p for `Navbar`  instead of Header

carletex: Thanks for the feedback!

In https://github.com/scaffold-eth/scaffold-eth-2/pull/569/commits/502540ea9dbf5ae6a52bb920db2f7ba30c52f7a5 I moved the "navlink" to the separate file and inlined it.

> Also I was thinking maybe rename the file from Header*Links to Navbar*Links lol its a bit personal but I always do cmd + p for Navbar instead of Header

The only reason I chose that name was because it shows next to the header. But open to change it.

image

---

Also created a PR with a variation: https://github.com/scaffold-eth/scaffold-eth-2/pull/570 (all in `Header.tsx`)

so we can decide which one we like better (thinking about users, what is easier for them to locate / tweak / understand)

technophile-04: I think I like #570 more 🙌

rin-st: also approved https://github.com/scaffold-eth/scaffold-eth-2/pull/570

carletex:
Closing this one in favor of https://github.com/scaffold-eth/scaffold-eth-2/pull/570

Thanks Shiv & Rinat!

PR #574, opened 2023-10-13
Move Block Explorer to footer

carletex:
For the sake of cleaning up the UI a bit (Example UI might be next :D Will create an issue soon), I'm testing moving the Block Explorer from the main menu to the footer. You still have access in any page (+ in the TX notifications) but in a less intrusive way. It will only show on dev env (local network), like the Faucet.

image
image

---

Anyway, just tinkering. Let me know what you think!

### DIFF
--- a/packages/nextjs/components/Footer.tsx
+++ b/packages/nextjs/components/Footer.tsx
@@ -19,12 +19,12 @@ export const Footer = () => {
   return (
     <div className="min-h-0 py-5 px-1 mb-11 lg:mb-0">
       <div>
-        <div className="fixed flex justify-between items-center w-full z-10 p-4 bottom-0 left-0 pointer-events-none ">
+        <div className="fixed flex justify-between items-center w-full z-10 p-4 bottom-0 left-0 pointer-events-none">
           <div className="flex flex-col md:flex-row gap-2 pointer-events-auto">
             {nativeCurrencyPrice > 0 && (
               <div>
-                <div className="btn btn-primary btn-sm font-normal cursor-auto gap-0">
-                  <CurrencyDollarIcon className="h-4 w-4 mr-0.5" />
+                <div className="btn btn-primary btn-sm font-normal normal-case gap-1 cursor-auto">
+                  <CurrencyDollarIcon className="h-4 w-4" />
                   <span>{nativeCurrencyPrice}</span>
                 </div>
               </div>
@@ -32,8 +32,8 @@ export const Footer = () => {
             {isLocalNetwork && (
               <>
                 <Faucet />
-                <Link href="/blockexplorer" passHref className="btn btn-primary btn-sm font-normal gap-0 normal-case">
-                  <MagnifyingGlassIcon className="h-4 w-4 mr-0.5" />
+                <Link href="/blockexplorer" passHref className="btn btn-primary btn-sm font-normal normal-case gap-1">
+                  <MagnifyingGlassIcon className="h-4 w-4" />
                   <span>Block Explorer</span>
                 </Link>
               </>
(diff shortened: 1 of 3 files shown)


packages/nextjs/components/Footer.tsx
```diff
@@ -12,21 +14,32 @@ import { getTargetNetwork } from "~~/utils/scaffold-eth";
  */
 export const Footer = () => {
   const nativeCurrencyPrice = useGlobalState(state => state.nativeCurrencyPrice);
+  const isLocalNetwork = getTargetNetwork().id === hardhat.id;
```
carletex, line 17: I think we should probably move this util to `network.ts`, since we are using it in a bunch of places.

We could create and issue + another PR.

packages/nextjs/components/Footer.tsx
```diff
@@ -12,21 +14,32 @@ import { getTargetNetwork } from "~~/utils/scaffold-eth";
...
               </div>
             )}
-            {getTargetNetwork().id === hardhat.id && <Faucet />}
+            {isLocalNetwork && (
+              <>
+                <Faucet />
+                <Link href="/blockexplorer" passHref className="btn btn-primary btn-sm font-normal gap-0 normal-case">
+                  <MagnifyingGlassIcon className="h-4 w-4 mr-0.5" />
+                  <span>Block Explorer</span>
+                </Link>
```
carletex, line 38: A few variations here that we could try too:

- Only the Glass Icon
- Only :mag: Explorer
- Switch places with the Faucet...

rin-st: - 🔍 ScaffoldEthScan  (or Se2-Scan) 😄

rin-st (approved):
I like the way it looks now 👍

Pabl0cks (approved):
Looking great to me 🙌

packages/nextjs/components/Footer.tsx
Pabl0cks: I like it with the current icon + text 👌

packages/nextjs/components/Footer.tsx
technophile-04: Yup agree ! Will create a PR for it 🙌

packages/nextjs/components/Footer.tsx
technophile-04: Yeah even I like current sequence and icon + text 🙌

technophile-04 (approved):
Really looks clean !! and also now we don't have to manually remove `Block Explorer` it will be automatically removed if switched to other network 🙌

Tysm Carlos !! 

Nitpick : 

Screenshot 2023-10-14 at 6 38 54 PM
 
 Maybe we should keep the same spacing b/w text icons for all buttons?

carletex:
> Maybe we should keep the same spacing b/w text icons for all buttons?

Changed to use the same classes (and removed some that weren't doing anything). Visually it doesn't look 100% aligned, because the 16x16 icons have some air around and the "bankNote" one is wider than the other two... also, the "1" gives the feeling of being more on the right than "F" or "B"

Anyway, this is how it looks:

image

carletex:
Thanks all for the review!

Feel free to merge when checking the last commit.

Pabl0cks:
Looks great to me! But I'll let Shiv's with his hawk-eye have the last word 🧐

technophile-04:
>  Visually it doesn't look 100% aligned, because the 16x16 icons have some air around and the "bankNote" one is wider than the other two... also, the "1" gives the feeling of being more on the right than "F" or "B"

Ohh yes, that makes sense! I think it looks fine 🙌


Also, SE-2 will be mostly viewed on mobile if it's in prod (i.e no `hardhat` network) so both Facuet and Blockexplorer shouldn't show up on mobile 🙈 lol sorry for my OCD kicking in 


Tysm Carlos 🙏 look great to me 🙌

PR #581, opened 2023-10-21
Update RainbowKitProvider to use coolMode

tantodefi:
Update RainbowKitProvider to use coolMode

## Description

We're the coolest kids working on the coolest web3 starter repo and we should be using coolMode :)

## Additional Information

Your ENS/address: 0x9B2FB7a8d227cDaA8002f80e8c8A99a19Bb1b969


technophile-04:
I think setting it as default may not be a great idea because some people might not like it.....we could delegate it to the user if they want to enable `cool` mode or not since it is pretty straightforward to add 

But thanks @tantodefi 🙌

PR #584, opened 2023-10-24
fix: incorrect error when sending txn with insufficient funds 

technophile-04:
## Description

In summary, we get this error : 

Screenshot 2023-10-24 at 3 18 46 PM
 
 While sending `value` i.e greater than our wallet balance, Ideally it should have given us "Sender doesn't have enought fund...." error  

Check out https://github.com/wagmi-dev/wagmi/issues/3155 for full context

Removing `chainId` from `useContractWrite` solves this problem, We could also wait until https://github.com/wagmi-dev/wagmi/issues/3155 is fixed and close this 🙌

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx
@@ -32,7 +32,6 @@ export const WriteOnlyFunctionForm = ({ abiFunction, onChange, contractAddress }
     isLoading,
     writeAsync,
   } = useContractWrite({
-    chainId: getTargetNetwork().id,
     address: contractAddress,
     functionName: abiFunction.name,
     abi: [abiFunction] as Abi,
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
@@ -35,7 +35,6 @@ export const useScaffoldContractWrite = <
   const configuredNetwork = getTargetNetwork();
 
   const wagmiContractWrite = useContractWrite({
-    chainId: configuredNetwork.id,
     address: deployedContractData?.address,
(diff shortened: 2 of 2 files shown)


damianmarti (approved):
Nice catch @technophile-04 !!!

The fix looks good, but maybe we can wait a little if Wagmi fixes the bug.

packages/nextjs/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx
```diff
@@ -32,7 +32,6 @@ export const WriteOnlyFunctionForm = ({ abiFunction, onChange, contractAddress }
     isLoading,
     writeAsync,
   } = useContractWrite({
-    chainId: getTargetNetwork().id,
```
rin-st, line 35: will it work on other networks?

technophile-04: Yup, if `chainId` is not passed viem / wagmi infers chain from walletClient sending the transaction checkout this => https://github.com/wagmi-dev/viem/blob/0e928cfe04407a2c974aa8a925dacc7687e78c32/src/actions/wallet/sendTransaction.ts#L134C26-L134C26

Aslo in Wagmi docs its optional https://wagmi.sh/react/hooks/useContractWrite#chainid-optional

rin-st (approved):

packages/nextjs/components/scaffold-eth/Contract/WriteOnlyFunctionForm.tsx
rin-st: Thanks! Approved

carletex:
Thanks @technophile-04 ! And thanks for the review.

Yep, maybe we don't need `chainId` in the write hook after all: https://github.com/scaffold-eth/scaffold-eth-2/pull/584#discussion_r1371101002

I guess we do need it on the Read hooks (in case no wallet is connected) and ofc in the ENS ones (where we always want to look up mainnet)

Issue #567, opened 2023-10-11
Hardhat ethers => Hardhat viem


carletex:
![image](https://github.com/scaffold-eth/scaffold-eth-2/assets/2486142/cfa67d2b-ddd2-4cb8-a51a-2f2208efe72f)
https://twitter.com/HardhatHQ/status/1712126192935985312

---

Since we are already using `viem` on the NextJS package, I think, for consistency, we should move the Hardhat package to use `viem` too.

carletex:
ofc @rin-st !!

I assigned it to you, but anytime feel free to assign it to yourself!

rin-st:
@carletex @technophile-04 should we close this for the [same reason](https://github.com/scaffold-eth/scaffold-eth-2/issues/632#issuecomment-2072354741)? And because I didn't find working solution for now too 😄

carletex:
Yep yep! We can always come back to this if needed.

PR #587, opened 2023-10-24
[WIP]: Migration to hardhat-viem

rin-st:
## Description

Migration to hardhat-viem

Todos:
- [x] use `viem` an `hardhat-viem` in scripts and deploy files
- [x] generating `deployments` folder.
- [ ] fix verification
- [ ] fix type errors
- [ ] refactor listAccounts
- [ ] migrate tests 

Notes: 
- since `hardhat-viem` is a "new shiny lib (c)", some things is not working as expected, for example types. Also we need to replace `hardhat-deploy-ethers` and I think some other plugins too, but I'll try to go without it (probably [Ignition](https://hardhat.org/ignition/docs/getting-started#overview) can help, but we need to wait until they'll add viem support. Currently only ethers v6)
- after removing `hardhat-ethers/hardhat-deploy-ethers` deployments folder isn't generated automatically. So I needed to do it in order to create `next/generated/deployedContracts` folder. It caused adding `generateDeployments()` to `deployYourContract`. I'll think again and try to find alternatives. Feel free to add thoughts and ideas about it.

## Additional Information

## Related Issues

_Closes #567_

Your ENS/address: 0xrinat


carletex:
Thanks for working on this, Rinat!

Yeah, I guess waiting for Ignition to support Viem could be the way to go (rolling our own stuff might be too hardcore)

Maybe we just need to wait until hardhat-viem is more stable + ignition supports viem? If not, it seems that we'd need to have ethers & viem at the same time, or roll our own deployment workflow.

What do you think?

rin-st:
Basically, things mostly works already. But
- as I said already, I need to manually create `deployments` folder in order to create `generated/deployedContracts` for nextjs
- [hardhat-verify](https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-verify) also works, but from scripts or using cli with adding deployed address and params.

So it's just workarounds, not something convenient and stable. Ignition will help with first issue and hopefully with second too. I'll continue to experiment with it, but for now yes, looks like we need to wait for ignition and more stable versions

rin-st:
https://github.com/scaffold-eth/scaffold-eth-2/issues/567

PR #590, opened 2023-10-27
Add copy to clipboard to TxReceipt

damianmarti:
## Description

Add copy to clipboard to transaction receipt to allow to copy the transaction data easily.

localhost_3002_debug

## Additional Information

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/TxReceipt.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/TxReceipt.tsx
@@ -10,12 +10,8 @@ export const TxReceipt = (
   const [txResultCopied, setTxResultCopied] = useState(false);
 
   return (
-    <div className="flex-wrap collapse collapse-arrow mb-2">
-      <input type="checkbox" className="min-h-0 peer" />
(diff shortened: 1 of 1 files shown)


technophile-04 (approved):
Nice !! LGTM ! 

Thanks @damianmarti 🙌

carletex:
Nice! Thanks @damianmarti 

The only thing that I'd explore is to position the copy icon in a different place (rn It feels a bit in the middle of nowhere)

The standard seems to be on the top right, but since we already have the arrow, might look a bit weird:
image

Another option could be next to the title:
image

We'd have to tweak a bit since that whole "row" collapses the result on click.

---

Anyway, just a suggestion, not a big deal if we keep it as it is!

damianmarti:
@carletex yeah, I know, it's not the best place.

Next to the title is the best option, but we should remove the row collapse by clicking the header and moving it to only the arrow.

I will try it!

damianmarti:
@carletex what do you think now?

localhost_3002_debug (1)

The receipt is copied by clicking the copy icon and the receipt is expanded if you click anywhere else.

localhost_3002_debug (2)

rin-st (approved):
lgtm!

carletex:
Thanks @damianmarti !!!

I think this looks better.

Issue #271, opened 2023-03-30
External contract


carletex:
We have really nice hooks (on top of wagmi's) for interacting with contracts deployed from SE-2 (leveraging on the `hardhat_contracts.ts` file).

Let's find the best way to do something similar with external contracts. E.g. You want your dApp to interact with the DAI contract. SE1 use a `external_contracts.js` file.

Ideally, as a builder, I'd like to use it the same way we use our custom hooks (same autocomplete / types magic).

```ts
  const { writeAsync } = useScaffoldContractWrite({
    contractName: "DaiContract",
    functionName: "transfer",
    args: [address, amount],
  });
```

I guess we could do that right now if we make the builder paste (in the right format) the contract info in `hardhats_contracts.ts`. But that's not very friendly.

So maybe a `external_contracts.ts` with a more clear format? Using wagmi's cli?

Check out this comment from @technophile-04 https://github.com/scaffold-eth/se-2/pull/200#issuecomment-1485414504.

KcPele:
the hook should accepts the ABI and address of the contract right?

sverps:
I have made some progress on this (https://github.com/sverps/se-2/tree/feature/271-external-contracts) before my holidays. I'll see if I can find some time to pick this up again.

PR #592, opened 2023-10-29
better handling of deployedContracts and external contract

technophile-04:
## Description

Tries to simply `deployedContracts` structure so that it's more intuitive. 

### Reason : 

This is a kind of RFC which we can close for sure if it does not make sense 🙌

**Current Structure** : 

<details>
<summary>
Generic Strucutre : 
</summary>

```ts
{
  chainID : Array<Network>
}

```

</details>

<details>
<summary>Example : </summary>

```ts
{
  "31337": 
  [
    {
      "name": "hardhat",
      "chainId": "31337",
      "contracts": { 
        YourContract : {
          address : "..."
          abi : [...]
        }
      }
    }
  ]
}

// We need to access contracts like this ->  deployedContrats[31337][0].contracts. <-- Notice [0] here
```
The reason why `hardhat-deploy` have `chainId: Array` instead of `chainId : Object` directly is because of this -> https://github.com/wighawag/hardhat-deploy/issues/325#issuecomment-1137343020

Since now we are generating this manually we are actually missing the edge-case mentioned in the above comment i.e having same `chainId` but different network names for eg:  `goerli-stage` , `goerli-prod`

https://github.com/scaffold-eth/scaffold-eth-2/blob/e88419a97baedd81fceb37a2553bf49e80ad981b/packages/hardhat/deploy/99_generateTsAbis.ts#L43

Ideally we should have done `output[chainId].push({...})` but I think it's fine since It is a very rare case that people do the above thing having same `chainId` but diff name

</details>

<details>
<summary>
Redundant/non-intuitive part which we are not utilising in frontend : 
</summary>

```diff
{
  "31337": 
-  [
    {
-      "name": "hardhat",
-      "chainId": "31337",
-      "contracts": { 
        YourContract : {
          address : "..."
          abi : [...]
        }
-      }
    }
-  ]
}
```

</details>

Therefore final struct proposal : 

```ts
{
  "31337" : {
    "YourContract" : {...}
  }
}
```

~~Or We can even remove `contracts` key from the object.~~

EDIT: Removed `contracts` filed at [bd4c916](https://github.com/scaffold-eth/scaffold-eth-2/pull/592/commits/bd4c916c90cdadef3b9eec2cbc256197cf22c8e6) since it made more sense, but please let me know if we should keep it 🙌

### How will this help us and why? 

The reason we have `deployedContract.ts` files is because SE-2 hooks are able to provide autocompletion by looking at this file. But if people already have contracts deployed or using any generic contract like `DAI` or they choose `None` option in `CLI` then this makes our custom hooks non-usable. 

Now regarding the name `generated/deployedContract.ts`  we should change it since this change tries to blur the line between external contracts and deployedContracts maybe we could do  `contracts/index.ts`? (please feel free to suggest an alternate name)

We have also discussed this in #271 and TG chat we tried`@wagmi/cli` but it seems not that feasible  :(. Regarding separate `external_contract.ts`  it looks hard to get typescript inference from multiple files (TBH I am saying this without trying 😅 but looks hard) and even with having `external_contract.ts` people will still manually need to paste the address and abi why not do it directly in `contracts/index.ts`? 

### Next Steps : 
In case we plan to take this forward then : 

~~1. Maybe remove `contracts` key too from structure (would love everyone's suggestion here)~~ removed at [bd4c916](https://github.com/scaffold-eth/scaffold-eth-2/pull/592/commits/bd4c916c90cdadef3b9eec2cbc256197cf22c8e6)
2. Choose a better name instead of `contracts/deployedContracts.ts`
3. Mention in docs maybe in Recipe section how to add your already deployed contract. 

Thanks, Also it is completely fine if we don't go with this approach !!🙌


packages/nextjs/generated/deployedContracts.ts
```diff
@@ -1,3 +1,15 @@
...
+ *   31337: {
+ *    YourContract: {
+ *      address: "0x..."
+ *      abi: [...],
+ *    }
+ * } satisfies GenericContractsDeclaration | null;
+ */
+const deployedContracts = null satisfies GenericContractsDeclaration | null;
 
 export default deployedContracts;
```
carletex, line 15: Could we do this? (I think is valid)

```ts
const deployedContracts = null;

export default deployedContracts satisfies GenericContractsDeclaration | null;
```

I think it's more legible like this, right?

We'd need to change `99_generateTsAbis.ts` + change the example.

carletex, line 15: We are using `deployedContracts` here but `contracts` from `99_generateTsAbis.ts`.

I like deployedContracts (More on the naming later)

packages/nextjs/generated/deployedContracts.ts
```diff
@@ -1,3 +1,15 @@
...
+/**
+ * @example
+ * const deployedContracts = {
+ *   31337: {
+ *    YourContract: {
+ *      address: "0x..."
+ *      abi: [...],
+ *    }
+ * } satisfies GenericContractsDeclaration | null;
+ */
```
carletex, line 12: What about adding something like:

```ts
/**
 * This file is populated when deploying contracts from Scaffold-ETH 2.
 * You can also add external (or existing??) with the following structure
 * .....
 */
``` 

About the example, maybe we could use a "real example"

```
/**ts
 * @example
 * const deployedContracts = {
 *  // chainId
 *   1: {
 *    // Contract name
 *    DAI: {
 *      address: "0x..."
 *      abi: [...],
 *    }
 * };
 */
```

IDK, just thinking out loud!

carletex (commented):
Really great job @technophile-04, amazing PR and context to it <3

I'm all in with this. Left a few comments.

> Now regarding the name generated/deployedContract.ts we should change it since this change tries to blur the line between external contracts and deployedContracts maybe we could do contracts/index.ts? (please feel free to suggest an alternate name)

We definitely need to change `generated/` to `contracts/`.  `deployedContracts` I think it covers both: "external" contracts + contracts deployed with SE-2. Both are deployed contracts, right? :D

`index.ts` follows the convention. The only reason I don't like it is because there are a bunch of index.ts files in the project (barrel files most of them)... and it'd be hard to navigate (at least how I do it: CTRL + P, start typing "deploy..." and enter). But maybe `contracts/deployedContracts.ts` is too redundant.

IDK, I'd love to hear what others think too.

packages/hardhat/deploy/99_generateTsAbis.ts
```diff
@@ -68,9 +62,13 @@ const generateTsAbis: DeployFunction = async function () {
   }
   fs.writeFileSync(
     `${TARGET_DIR}deployedContracts.ts`,
-    prettier.format(`const contracts = {${fileContent}} as const; \n\n export default contracts`, {
-      parser: "typescript",
-    }),
+    prettier.format(
+      `import { GenericContractsDeclaration } from "~~/utils/scaffold-eth/contract"; \n\n
+ const contracts = {${fileContent}} satisfies GenericContractsDeclaration | null ; \n\n export default contracts`,
```
carletex, line 67: Will need to change this if we are doing https://github.com/scaffold-eth/scaffold-eth-2/pull/592#discussion_r1376002175
(if not, ignore :D)

sverps:
I like this approach of simplifying the contract definition file.

Downside of putting everything in the same file is running `yarn deploy` seems to overwrite the whole file, and thus it removes any manually added external contracts. So this approach requires changing `99_generateTsAbis.ts` to parse the existing file and ensure external contracts are kept. This would require some sort of deep merge, which might not always do exactly what the user expects (e.g. a user that changes a contract name might still see his old contract show up in addition to the newly named one, since our deepMerge would not know if an existing definition is an old deployment, or a user added external contract).

So with that in mind, it might be beneficial to keep a 2-file approach. What the types for this would look like can be seen [here](https://github.com/scaffold-eth/scaffold-eth-2/compare/main...sverps:se-2:feature/271-external-contracts#diff-9c79294d6ec334fac50d0ac22f2ae002ac0a25b987b7123110cff9e9674b9a1e). If we would go this route, I can definitely help adding this.

technophile-04:
> We definitely need to change generated/ to contracts/. deployedContracts I think it covers both: "external" contracts + contracts deployed with SE-2. Both are deployed contracts, right? :D .... index.ts follows the convention. The only reason I don't like it is because there are a bunch of index.ts files in the project (barrel files most of them).

Yup 💯

---

> Downside of putting everything in the same file is running yarn deploy seems to overwrite the whole file, and thus it removes any manually added external contracts. So this approach requires changing 99_generateTsAbis.ts to parse the existing file and ensure external contracts are kept. This would require some sort of deep merge, which might not always do exactly what the user expects (e.g. a user that changes a contract name might still see his old contract show up in addition to the newly named one, since our deepMerge would not know if an existing definition is an old deployment, or a user added external contract).

Yup yup agree it's really bad that external contract will be overridden :( 

> So with that in mind, it might be beneficial to keep a 2-file approach.

This is the most optimal / best-of-all approach if we get it working, 

As I understand the problem we are facing with 2-file approach is Merging two objects at the type level and also make sure the typescript doesn't widen's the  type while merging 

From above branch : 
https://github.com/sverps/se-2/blob/ee4207adf0cd62e67e84ca2a4681ad90ddeaab6c/packages/nextjs/utils/scaffold-eth/contract.ts#L50

What if we do instead of `any` : 

```ts
 return { ...external, ...combined } as MergeDeepRecords<typeof contractData, typeof externalContractsData>;
```

Here is [MergeDeepRecords](https://github.com/sindresorhus/type-fest/blob/9960ba4d9fe88757c306f0aa2aaeb0d37c4d1189/source/merge-deep.d.ts#L54)  type or we can do any type helper which hopefully does this job, btw found that [Ramda](https://ramdajs.com/docs/#mergeDeepWithKey) function `mergeDeepWithKey` does merge two objects both at runtime and type level lol not sure if its worth adding an library.....Just thinking out loud


Thanks guys for the great suggestions !! Will try to explore the 2-files approach since I kind of got the path thanks to Samuel's finding 🙌

carletex:
> Downside of putting everything in the same file is running yarn deploy seems to overwrite the whole file, and thus it removes any manually added external contracts. So this approach requires changing 99_generateTsAbis.ts to parse the existing file and ensure external contracts are kept. This would require some sort of deep merge, which might not always do exactly what the user expects (e.g. a user that changes a contract name might still see his old contract show up in addition to the newly named one, since our deepMerge would not know if an existing definition is an old deployment, or a user added external contract).

This is a great point.

If we can figure out the Types for the 2-files approach, it's definitely the way to go (`deployedContracts.ts` & `externalContracts.ts`). 

Thanks @technophile-04 & @sverps !!

sverps:
> As I understand the problem we are facing with 2-file approach is Merging two objects at the type level and also make sure the typescript doesn't widen's the type while merging
> 
> From Samuel's branch : https://github.com/sverps/se-2/blob/ee4207adf0cd62e67e84ca2a4681ad90ddeaab6c/packages/nextjs/utils/scaffold-eth/contract.ts#L50

There is actually no type widening in the deepMerge function, since the return type is declared rather than inferred. I'm guessing the `as any` was just temporary while wip (but don't recall for sure since it's been a while that I wrote that), and can indeed, for clarity, be replaced by the declared returntype `MergeContractDefinitions<typeof externalContractsData, typeof contractsData>`

> Here is [MergeDeepRecords](https://github.com/sindresorhus/type-fest/blob/9960ba4d9fe88757c306f0aa2aaeb0d37c4d1189/source/merge-deep.d.ts#L54) type or we can do any type helper which hopefully does this job, btw found that [Ramda](https://ramdajs.com/docs/#mergeDeepWithKey) function mergeDeepWithKey does merge two objects both at runtime and type level lol not sure if its worth adding an library.....Just thinking out loud

Could work, don't know enough about this specific rambda function, so we'd need to test it out. The `MergeContractDefinitions` util that I wrote does basically this, but in a way custom for 2 contract definition files. If we can get the ramda function to do the same job, it might be good to use it in order to reduce complexity and not have to worry about maintaining additional TypeScript magic :sweat_smile: .

I pushed your suggestion to my branch and fixed a small error for `null` contract definitions.
Let me know if you need anything. :+1:

damianmarti:
Great work @technophile-04 !! 

I don't have much to say. I agree that having 2 files is the best option, and the names should be contracts/deployedContracts.ts and contracts/externalContracts.ts or maybe contracts/deployed.ts and contracts/external.ts

packages/nextjs/utils/scaffold-eth/common.ts
```diff
@@ -1,3 +1,29 @@
...
+export const replacer = (_key: string, value: unknown) => (typeof value === "bigint" ? value.toString() : value);
+
+export const deepMerge = <D extends Record<PropertyKey, any>, S extends Record<PropertyKey, any>>(
+  destination: D,
+  source: S,
+) => {
+  const result: Record<PropertyKey, any> = {};
+  for (const key in destination) {
+    if (key in source) {
+      if (typeof destination[key] === "object" && typeof source[key] === "object" && !Array.isArray(source[key])) {
```
technophile-04, line 14: Summary how this function works is it recursively merges the value of the key if its an `object` and will directly override `destination` key if its value is primitive type by the value of the same key in the `source`

The reason I have `!Array.isArray(source[key])` here is because there might be an edge case where suppose : 

`deployedContrac.ts`: 

```ts
{
  1: {
    YourContract : {address : "0xabcd", abi : [YourContract_ABI] }
  }
}
```

`externalContract.ts`: 

```ts
// Bad guy trying to break SE-2 
{
  1: {
    YourContract : {address : "daiAddress", abi : [DAI_ABI] }
  }
}
```

So the final result will be, Instead of recursively merging Abi's we just replace with `externalContract` abi : 

```ts
{
  1: {
    YourContract : {address : "daiAddress", abi : [DAI_ABI] }
  }
}
```

 If we don't do that it will break ABI's array : 

```ts
{
  1: {
    YourContract : {address : "daiAddress", abi : [DISASTER] }
  }
}
```

The reason for `DISASTER` is because since `typeof Array` is `object` this function will try to recursively merge each element of both ABI's 

Hence if its array type we just override `destination` key with `source` key's array value similar to what we do with other primitive types

packages/nextjs/utils/scaffold-eth/common.ts
```diff
@@ -1,3 +1,29 @@
...
+    } else {
+      result[key] = destination[key];
+    }
+  }
+  for (const key in source) {
+    if (!(key in destination)) {
+      result[key] = source[key];
+    }
+  }
+  return result as MergeDeepRecord<D, S, { arrayMergeMode: "replace" }>;
```
technophile-04, line 28: This type-helper follows the same logic which is described in https://github.com/scaffold-eth/scaffold-eth-2/pull/592/files#r1377673945

technophile-04:
> Could work, don't know enough about this specific rambda function, so we'd need to test it out.

😅 My bad Ramda dosen't work nicely tried it :( 

The new approach I followed is using `deepMerge` along with asseretion [`MergeDeepRecords`](https://github.com/sindresorhus/type-fest/blob/9960ba4d9fe88757c306f0aa2aaeb0d37c4d1189/source/merge-deep.d.ts#L54) from [`type-fest `](https://github.com/sindresorhus/type-fest), Added `type-fest` as dev dependency 😅 I felt its worth it since it some great type-helpers such `Prettify` which we added it has same but with diff name as `Simplify`  etc 


Also renamed `generated` => `contracts`  dir

Just a thought what keep name `generated/deployedContract.ts` (denoting its `generated` and not to be touched) and move `externalContracts.ts` to root of NextJS, I think similar we had in SE-1 but it might be wierd open for suggestion 🙌

This PR is ready for test 🙌 I tested it normally works great but will also do through pass tomorrow 🙌

Tysm all

carletex:
Thanks @technophile-04 for pulling this off <3

I'll do a pass soon too.

> Just a thought what keep name generated/deployedContract.ts (denoting its generated and not to be touched) and move externalContracts.ts to root of NextJS, I think similar we had in SE-1 but it might be wierd open for suggestion 🙌

I like both being in the same folder (`contracts/`). What if we add something like this at the top of the `deployedContracts.ts` file?

```ts
// This file is autogenerated by Scaffold-ETH.
// You should not edit it manually or your changes might be overwritten.
```

packages/nextjs/utils/scaffold-eth/common.ts
rin-st: If I understand right, you're merging objects only on first two levels - networkId and contractNames. So, can we use something simpler? Not sure about types though. Something like that
```ts
  const result: Record<PropertyKey, any> = {};
  const allKeys = Array.from(new Set([...Object.keys(source), ...Object.keys(destination)]));
  for (const key of allKeys) {
    result[key] = { ...destination[key], ...source[key] };
  }
  return result as MergeDeepRecord<D, S, { arrayMergeMode: "replace" }>;
```

rin-st:
Great job Shiv! 👍 
Left one comment about merging

> maybe contracts/deployed.ts and contracts/external.ts

like!

packages/nextjs/utils/scaffold-eth/common.ts
technophile-04: I actually wrote generic logic in line with how `MergeDeepRecord` works but I think it not needed,  and  I really love the proposed solution minimal, inline with `MergeDeepRecord`   and handles our required edge-cases too !

Updated at https://github.com/scaffold-eth/scaffold-eth-2/pull/592/commits/f47535cf0a007ac39df864a98bb4217da62c829a  tysm Rinat 🙌

sverps:
> The new approach I followed is using `deepMerge` along with asseretion [`MergeDeepRecords`](https://github.com/sindresorhus/type-fest/blob/9960ba4d9fe88757c306f0aa2aaeb0d37c4d1189/source/merge-deep.d.ts#L54) from [`type-fest `](https://github.com/sindresorhus/type-fest), Added `type-fest` as dev dependency 😅 I felt its worth it since it some great type-helpers such `Prettify` which we added it has same but with diff name as `Simplify` etc but please let me know if we should just copy-paste the only required types and remove it as dependency

I like using this dependency, it seems well established and widely used, so no harm in that. I liked writing and exploring the typescript magic I created custom, but from a maintainability perspective, it wasn't ideal. If we can externalize it this way, it's even better :+1: 

PR looks nice. I agree with placing generated and manual contract files in same location, we can explain everything about it in the readme and in comments.

technophile-04:
Tested it nicely just wanted to make sure that types don't behave in weird way(like it kind of did in #377 related issue) when there is large amount of merging or things get way too big. 

But everything works great 🙌!!

Tysm all for great suggestion really appreciate it and thanks again @sverps for chiming in 🙌!!

rin-st (approved):
Lgtm! Thanks @technophile-04 !

Pabl0cks:
Hats off to everyone for this PR! Really nice context, explanations and research @technophile-04. It's been a pleasure to read the discussion with everyone's contributions and research about this, GJ 🙌

I think it will be very user friendly having this in two separate files with these names, Tested it and looks good to me 👌

Pabl0cks (approved):

carletex (approved):
Amazing stuff, it works great.

Thank you all <3

PR #593, opened 2023-11-02
Weekly CLI backmerge

carletex:
Updating the CLI branch with:

- remove chainId from useContractWrite (#584) 
- Add copy to clipboard to TxReceipt (#590)
- Add event logs to transaction page (#591)
- deployedContracts & externalContracts (#592)

PR #594, opened 2023-11-02
fix: don't throw error just log if artificats are not present

technophile-04:
## Description

If you don't have `artifacts` folder in `pacakges/hardhat` when running `yarn start` you get this error log in your next dev server terminal : 

Screenshot 2023-11-02 at 6 34 34 PM

This is not that above obvious when clonning SE-2, but when you do `npx create-eth@latest` and switch scaffold.config `targetNetwork` to other network eg `sepolia` it you get constant error log there because of this logic we have in CLI 

https://github.com/scaffold-eth/scaffold-eth-2/blob/f9d27d128b519af1d224d30b598475de83eae9f8/templates/base/packages/nextjs/pages/blockexplorer/address/%5Baddress%5D.tsx#L173

noticed this while testing #593

### DIFF
--- a/packages/nextjs/pages/blockexplorer/address/[address].tsx
+++ b/packages/nextjs/pages/blockexplorer/address/[address].tsx
@@ -173,7 +173,8 @@ export const getServerSideProps: GetServerSideProps = async context => {
   );
 
   if (!fs.existsSync(buildInfoDirectory)) {
-    throw new Error(`Directory ${buildInfoDirectory} not found.`);
+    console.error(`Directory ${buildInfoDirectory} not found.`);
+    return { props: { address, contractData: null } };
   }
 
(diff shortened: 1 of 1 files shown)


carletex:
Hey @technophile-04 thanks for reporting this!

How can I reproduce this? I tried in main and also installing with the CLI (choosing hardhat). In both cases I did:

- change network to sepolia on scaffold.config
- yarn start
- go to the site

I don't get any errors in the next dev console. What should I do to get the error you are encountering?

Thanks!

technophile-04:
Ohh shit! 😅 Actually I had a tab open in the background with this block explorer URL `http://localhost:3000/blockexplorer/address/0x5FbDB2315678afecb367f032d93F642f64180aa3` which was causing Nextjs to render `[address]` page and I didn't realize its open in the background while I was testing homepage in new tab

PR #597, opened 2023-11-06
Add pooling interval to EventHistory hook

damianmarti:
## Description

Add a new poolingInterval optional parameter to useScaffoldEventHistory to be able to get the new events. 

If we merge this PR, I will open a PR to update the doc.

## Additional Information

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
@@ -52,7 +52,7 @@ export const useScaffoldEventHistory = <
   const publicClient = usePublicClient();
   const configuredNetwork = getTargetNetwork();
 
-  const readEvents = async () => {
+  const readEvents = async ({ append }: { append?: boolean }) => {
     try {
       if (!deployedContractData) {
         throw new Error("Contract not found");
@@ -92,7 +92,7 @@ export const useScaffoldEventHistory = <
                 : null,
           });
         }
-        if (events) {
+        if (events && append) {
           setEvents([...events, ...newEvents]);
         } else {
           setEvents(newEvents);
@@ -110,7 +110,7 @@ export const useScaffoldEventHistory = <
 
   useEffect(() => {
     if (!deployedContractLoading) {
-      readEvents();
+      readEvents({});
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [
@@ -131,7 +131,7 @@ export const useScaffoldEventHistory = <
   useInterval(
     async () => {
       if (!deployedContractLoading) {
-        readEvents();
+        readEvents({ append: true });
       }
     },
     watch ? (configuredNetwork.id !== chains.hardhat.id ? scaffoldConfig.pollingInterval : 4_000) : null,
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
@@ -1,5 +1,6 @@
 import { useEffect, useMemo, useState } from "react";
 import { Abi, AbiEvent, ExtractAbiEventNames } from "abitype";
+import { useInterval } from "usehooks-ts";
 import { Hash } from "viem";
 import { usePublicClient } from "wagmi";
 import { useDeployedContractInfo } from "~~/hooks/scaffold-eth";
@@ -21,6 +22,7 @@ import {
  * @param config.blockData - if set to true it will return the block data for each event (default: false)
  * @param config.transactionData - if set to true it will return the transaction data for each event (default: false)
  * @param config.receiptData - if set to true it will return the receipt data for each event (default: false)
+ * @param config.poolingInterval - if set to a number, the events will be updated every poolingInterval milliseconds (default: no pooling)*
  */
 export const useScaffoldEventHistory = <
   TContractName extends ContractName,
@@ -28,6 +30,7 @@ export const useScaffoldEventHistory = <
   TBlockData extends boolean = false,
   TTransactionData extends boolean = false,
   TReceiptData extends boolean = false,
+  TPoolingInterval extends number | null = null,
 >({
   contractName,
   eventName,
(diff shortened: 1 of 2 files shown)


carletex:
Hey @damianmarti , thanks for this!!

I always have to use the history hook + subscriber hook to get this behavior. So this might be better & simpler!

So we have `pollingInterval` in `scaffold.config.ts` which all the wagmi hooks with `watch` use. So to be more consistent... maybe it would be better to use that `pollingInterval` and just pass a `watch` flag to the hook?

(Just thinking out loud)

damianmarti:
> Hey @damianmarti , thanks for this!!
> 
> I always have to use the history hook + subscriber hook to get this behavior. So this might be better & simpler!
> 
> So we have `pollingInterval` in `scaffold.config.ts` which all the wagmi hooks with `watch` use. So to be more consistent... maybe it would be better to use that `pollingInterval` and just pass a `watch` flag to the hook?
> 
> (Just thinking out loud)

Makes sense!! Changed! Thanks!!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -21,13 +23,15 @@ import {
  * @param config.blockData - if set to true it will return the block data for each event (default: false)
  * @param config.transactionData - if set to true it will return the transaction data for each event (default: false)
  * @param config.receiptData - if set to true it will return the receipt data for each event (default: false)
+ * @param config.watch - if set to true, the events will be updated every poolingInterval milliseconds set at scaffoldConfig (default: false)
```
rin-st, line 26: polling :)

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -36,30 +40,37 @@ export const useScaffoldEventHistory = <
   blockData,
   transactionData,
   receiptData,
-}: UseScaffoldEventHistoryConfig<TContractName, TEventName, TBlockData, TTransactionData, TReceiptData>) => {
+  watch,
+}: UseScaffoldEventHistoryConfig<TContractName, TEventName, TBlockData, TTransactionData, TReceiptData, TWatch>) => {
   const [events, setEvents] = useState<any[]>();
   const [isLoading, setIsLoading] = useState(true);
   const [error, setError] = useState<string>();
+  const [fromBlockUpdated, setFromBlockUpdated] = useState<bigint>(fromBlock);
```
rin-st, line 48: as I understand with your updates changing `fromBlock` parameter from outside of hook will not affect results.

So if I wanted let's say get events history from block `1000` first, and changed it to `2000`, hook gave me only events from block `2000`. But now it's just default state, so it still will show events from block `1000`. Or am I missing something?

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -36,30 +40,37 @@ export const useScaffoldEventHistory = <
...
 
-        const event = (deployedContractData.abi as Abi).find(
-          part => part.type === "event" && part.name === eventName,
-        ) as AbiEvent;
+      const event = (deployedContractData.abi as Abi).find(
+        part => part.type === "event" && part.name === eventName,
+      ) as AbiEvent;
 
+      const blockNumber = await publicClient.getBlockNumber({ cacheTime: 0 });
+      if (blockNumber >= fromBlockUpdated) {
```
rin-st, line 64: always true?

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
damianmarti: Great question! I didn't test it. fromBlock is on the useEffect hook, so readEvents will be called and then I don't know what will happen :-)

I will take a look at this! Thanks!

packages/nextjs/utils/scaffold-eth/contract.ts
```diff
@@ -232,6 +233,7 @@ export type UseScaffoldEventHistoryConfig<
   blockData?: TBlockData;
   transactionData?: TTransactionData;
   receiptData?: TReceiptData;
+  watch?: TWatch;
```
rin-st, line 236: it seems just `boolean` works fine here

damianmarti: I defined a type to be similar to the other parameters:

`  TBlockData extends boolean = false, TTransactionData extends boolean = false, TReceiptData extends boolean = false,`

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
damianmarti: Yes, if you change fromBlock when watching you will still get the events from the first fromBlock value. It doesn't make sense to change the fromBlock value when using the watch feature, maybe we can add something about it to the documentation.

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
damianmarti: Fixed!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -108,6 +126,15 @@ export const useScaffoldEventHistory = <
     receiptData,
   ]);
 
+  useInterval(
+    async () => {
+      if (!deployedContractLoading) {
+        readEvents();
+      }
+    },
+    watch ? scaffoldConfig.pollingInterval : null,
```
technophile-04, line 135: Maybe we make it `5000` if `targetNetwork`  is `hardhat` because it uses `30000` while on `hardhat` too, similar to what we have at : 

https://github.com/scaffold-eth/scaffold-eth-2/blob/44f31c17088d09f8428f0b2c6d51f8f5046b16a2/packages/nextjs/services/web3/wagmiConnectors.tsx#L39

damianmarti: I'm not sure, because if you want faster polling when using hardhat you just change the value at the scaffoldConfig file. 

If you use some special value for hardhat and then the user starts using a real blockchain, and the events start loading slowly, maybe the user doesn't understand why they were loading faster before.

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
rin-st: 🤔  looks like strange behavior. I think on `fromBlock` change it should reload all events (or missing ones). Let's see what others think

packages/nextjs/utils/scaffold-eth/contract.ts
rin-st: yes, but other params are used later in `UseScaffoldEventHistoryData` type

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
carletex: > I'm not sure, because if you want faster polling when using hardhat you just change the value at the scaffoldConfig file.

The thing is that we don't use the scaffoldConfig polling value when we are on hardhat / foundry (check the wagmiConnectors file Shiv linked), since we want to be as fast as possible. The default seems to be 4000 ms https://wagmi.sh/react/providers/configuring-chains#pollinginterval-optional

> If you use some special value for hardhat and then the user starts using a real blockchain, and the events start loading slowly, maybe the user doesn't understand why they were loading faster before.

Yep, this might be true! But it's how we are doing it rn :/ We have a warning in scaffoldConfig tho:

```
  // The interval at which your front-end polls the RPC servers for new data
  // it has no effect on the local network
```

So for the sake of consistency, it might be a good idea to set 4000 for local chains?

What do you all think?

damianmarti: Makes sense! Changed! Thanks!!

packages/nextjs/utils/scaffold-eth/contract.ts
damianmarti: You are right! Changed! Thanks!!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
damianmarti: Please take a look at the proposed change. Now only the events are appended when polling. 

I think it works as the expected behavior now. Thank you for noticing that!

rin-st: Thanks! I updated it a little bit https://github.com/scaffold-eth/scaffold-eth-2/pull/600, since with `append` param it didn't change `fromBlock` behavior. I'm not sure about hooks deps but fast testing worked fine. Could you take a look?

damianmarti: @rin-st Thanks!! Merged!

rin-st (approved):
lgtm!

carletex (commented):
Tested it and works great! Thanks all!

Now I guess there is some overlapping happening between useScaffoldEventHistory & useScaffoldEventSubscriber.

I'll create an issue or PR in the docs repo!

carletex:
https://github.com/scaffold-eth/se2-docs/pull/38

PR #609, opened 2023-11-17
Add multiplyBy1e18Enabled flat to IntegerInput component

damianmarti:
## Description

Add multiplyBy1e18Enabled flat to IntegerInput component. 

When this flag is set to false, the button to multiply the number by 1e18 is not shown. This is to be used when doesn't make sense to multiply the user input by that amount, like when it's used to enter a regular number.

The default value is true, so the default behavior is not changed.

## Additional Information

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Input/IntegerInput.tsx
+++ b/packages/nextjs/components/scaffold-eth/Input/IntegerInput.tsx
@@ -3,6 +3,7 @@ import { CommonInputProps, InputBase, IntegerVariant, isValidInteger } from "~~/
 
 type IntegerInputProps = CommonInputProps<string | bigint> & {
   variant?: IntegerVariant;
+  multiplyBy1e18Enabled?: boolean;
 };
 
 export const IntegerInput = ({
@@ -12,6 +13,7 @@ export const IntegerInput = ({
   placeholder,
   disabled,
   variant = IntegerVariant.UINT256,
+  multiplyBy1e18Enabled = true,
 }: IntegerInputProps) => {
   const [inputError, setInputError] = useState(false);
   const multiplyBy1e18 = useCallback(() => {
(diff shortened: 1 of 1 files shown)


packages/nextjs/components/scaffold-eth/Input/IntegerInput.tsx
```diff
@@ -3,6 +3,7 @@ import { CommonInputProps, InputBase, IntegerVariant, isValidInteger } from "~~/
 
 type IntegerInputProps = CommonInputProps<string | bigint> & {
   variant?: IntegerVariant;
+  multiplyBy1e18Enabled?: boolean;
```
sverps, line 6: [nit]: I personally prefer to work with pre-fixes for these boolean modifier flags. Makes it a bit clearer when reading the variable name (or when autocompleting props on a component) that it is a boolean flag. So something like `enableMultiplyBy1e18`?

sverps (commented):
Nice idea :+1:

packages/nextjs/components/scaffold-eth/Input/IntegerInput.tsx
sverps: Hmmm (some further nitting, sorry :sweat_smile: ), since setting a flag to true can be done like this `<IntegerInput enableMultiplyBy1e18 />`, it might make more sense to make the default value `false`. Otherwise devs would need to write `<IntegerInput enableMultiplyBy1e18={false} />`. To keep the desired default behavior, we could invert the name into `disableMultiplyBy1e18`.

```
<IntegerInput /> // default
<IntegerInput disableMultiplyBy1e18 /> // hides the asterisk
```

rin-st: I hid that asterisk in some apps too. And since it hides suffix I just named prop `hideSuffix`. Also I like Samuel's variant `disableMultiplyBy1e18`

damianmarti: @sverps @rin-st thanks for the feedback!! 

I changed the flag to disableMultiplyBy1e18 (I think it's more self-explanatory than hideSuffix).

sverps (approved):

PR #610, opened 2023-11-17
Fix AbiFunctionReturnType removing [0]

damianmarti:
## Description

When a contract function returns an array only the first element type was returned when using the useScaffoldContractRead hook.

Now all the types from the array are returned.

I'm not sure if this breaks something else... AbiFunctionReturnType is only used on useScaffoldContractRead hook and everything seems to be working fine.

## Additional Information

### DIFF
--- a/packages/nextjs/utils/scaffold-eth/contract.ts
+++ b/packages/nextjs/utils/scaffold-eth/contract.ts
@@ -81,7 +81,7 @@ export type AbiFunctionOutputs<TAbi extends Abi, TFunctionName extends string> =
 
 export type AbiFunctionReturnType<TAbi extends Abi, TFunctionName extends string> = IsContractDeclarationMissing<
   any,
-  AbiParametersToPrimitiveTypes<AbiFunctionOutputs<TAbi, TFunctionName>>[0]
+  AbiParametersToPrimitiveTypes<AbiFunctionOutputs<TAbi, TFunctionName>>
 >;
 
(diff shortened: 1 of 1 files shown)


sverps:
Hey @damianmarti. Nice find that this poses issues for array return values. This change does seem to break the return type for common cases. Check for example the type of `data` in the default `YourContract` greeting. It types it as `[string]`, whereas it actually returns a simple `string`, so the type seems to be wrong in that case.

I took the liberty of making both scenarios work by basically checking the array length. Feel free to amend or make further changes if needed.

carletex:
Great catch @damianmarti!!

First I tried returning an array (like a `uint[]`) and that was working, but as @sverps said it was happening when returning multiple values (which It's also an array I guess :D)

I tried with:

```solidity
contract YourContract {
	function getHardcodedValues() public pure returns (uint, address) {
		return (100, 0xd8b934580fcE35a11B58C6D73aDeE468a2833fa8);
	}
}
```

and

```ts
  const { data } = useScaffoldContractRead({
    contractName: "YourContract",
    functionName: "getHardcodedValues",
  });

  if (!data) return null;
  const a = data[0];
  const b = data[1];
  const c = data[2];
```

and before the PR everything was a bigint (instead of and array of bigint, address). 

image

after the changes:
image
image

Beautiful.

Everything else seems to be working as usual after Samuel's TS dark magic :P

Thank you both!!

carletex (approved):

carletex (approved):

PR #615, opened 2023-11-20
Feat/multi chain

### LINKED ISSUE #599: Allow multiple target networks

- See also https://github.com/scaffold-eth/scaffold-eth-2/discussions/429

### Current Behavior

In scaffold.config.ts, object scaffoldConfig.targetNetwork is a single value.  When the user is using a non-default wallet (e.g. MetaMask) and connected to a network that is not targetNetwork, a "Wrong Network" error is displayed, even if the dApp is deployed on the network the user is currently connected to.

### Expected Behavior

The targetNetwork value can be an array. If the network the user is currently connected to is in the array, the dApp behavior is the same as if that network were the single value, and the dApp works.  When the user changes to another network where the contracts are also deployed, the dApp continues to work as if that new network were the value of targetNetwork.

For backwards compatibility, if a single value is provided, it is interpreted as a single-element array.  The sample code is also updated to make the default network a single-element array to more easily communicate how this can be expanded as the dApp is deployed on additional networks.

### Steps To Reproduce

1. Do rapid development with the default Hardhat network.
2. Deploy to a public chain.  
3. Switch to that chain and test to verify correct operation.
4. Want to switch back and forth between chains quickly for demo purposes.
5. Find that doing so requires editing the config file.

### Anything else?

This will not change the one-network-at-a-time contract deployment workflow, only simplify use.

sverps:
Fixes #599 

Updates `scaffold.config.ts` to allow configuring multiple target networks like so:

```TypeScript
const scaffoldConfig = {
  targetNetworks: [chains.hardhat, chains.sepolia],
  ...rest
}
```

The first value is the default and will be the network selected upon launching the app.

- Switching can be done from the top menu
- Faucet only shown for local network
- When multiple networks configured, pollingInterval will for all networks be the one from the config (if anyone has an idea to keep polling interval to 4000 while on hardhat, while at the same time use a slower polling for other networks, let me know)
- Works for external wallets as well as for burner wallet

multi-network-1
multi-network-2
multi-network-3


packages/nextjs/scaffold.config.ts
```diff
@@ -30,7 +30,7 @@ const scaffoldConfig = {
   walletConnectProjectId: process.env.NEXT_PUBLIC_WALLET_CONNECT_PROJECT_ID || "3a8170812b534d0ff9d794f19a901d64",
 
   // Only show the Burner Wallet when running on hardhat network
-  onlyLocalBurnerWallet: true,
+  onlyLocalBurnerWallet: false,
```
carletex, line 33: I think we want this to be true by default.

I see we have the same _issue_ here as with"pollingInterval": where multiple chains are configured, we can't use the burner wallet in hardhat even if this is true.

I think it's ok. Live dApps shouldn't have hardhat enabled anyway.

packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -1,15 +1,19 @@
+import { useEffect, useState } from "react";
 import * as chains from "viem/chains";
+import { useNetwork } from "wagmi";
 import scaffoldConfig from "~~/scaffold.config";
 
-export type TChainAttributes = {
+export type ChainAttributes = {
```
carletex, line 6: For consistency, we probably want to remove all the other `T` prefixes in the codebase

Opened: https://github.com/scaffold-eth/scaffold-eth-2/issues/617

carletex:
This is looking amazing! Great job @sverps. Left a few review comments/open questions.

> When multiple networks configured, pollingInterval will for all networks be the one from the config (if anyone has an idea to keep polling interval to 4000 while on hardhat, while at the same time use a slower polling for other networks, let me know)

I think this is ok. Live apps shouldn't have hardhat/foundry configured, so you'll only want that on your dev env (where you can tweak the file for dev purposes)

Also, thank you @blahkheart!! Sorry for the miscommunication with the PRs. As @sverps said, we'll give you attribution on merging, since this PR took some inspiration from yours.

packages/nextjs/pages/_app.tsx
```diff
@@ -49,8 +49,14 @@ const ScaffoldEthApp = ({ Component, pageProps }: AppProps) => {
...
   );
 };
 
-export default ScaffoldEthApp;
+const ScaffoldEthAppWithWagmi = (props: AppProps) => (
+  <WagmiConfig config={wagmiConfig}>
+    <ScaffoldEthApp {...props} />
+  </WagmiConfig>
+);
+
```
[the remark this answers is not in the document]
sverps: Because the `useTargetNetwork` hook (which internally uses wagmi's `useNetwork` hook) is used inside the native price hook, so this must be inside the wagmi context. I might revisit some things in `_app.tsx` to make it a bit cleaner.

packages/nextjs/utils/scaffold-eth/networks.ts
```diff
@@ -105,11 +109,28 @@ export function getBlockExplorerAddressLink(network: chains.Chain, address: stri
...
+ * @returns targetNetworks array containing networks configured in scaffold.config including extra network metadata
  */
+export function getTargetNetworks(): ChainWithAttributes[] {
+  return scaffoldConfig.targetNetworks.map(targetNetwork => ({
+    ...targetNetwork,
+    ...NETWORKS_EXTRA_DATA[targetNetwork.id],
+  }));
+}
+
+export function useTargetNetwork(): ChainWithAttributes {
```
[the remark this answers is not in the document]
sverps: Yeah when merging @blahkheart's branch, I refactored a bunch of things. Indeed makes more sense to be in the hooks dir.

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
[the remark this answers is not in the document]
sverps: Didn't go for that option due to limited space on mobile devices.

header

Additionally, it tries to take the network from the wallet client, and switching can be done there also, so didn't feel like too big of an issue to put it in that dropdown.

packages/nextjs/scaffold.config.ts
sverps: oops, committed that by accident it seems

sverps:
Still a couple of things to check

- [ ] refactor network switch options (it's currently duplicated)
- [ ] revisit `_app.tsx` for some cleanup if possible
- [ ] ?

packages/nextjs/hooks/scaffold-eth/useTargetNetwork.ts
```diff
@@ -0,0 +1,28 @@
...
+import { useGlobalState } from "~~/services/store/store";
+import { ChainWithAttributes } from "~~/utils/scaffold-eth";
+import { NETWORKS_EXTRA_DATA } from "~~/utils/scaffold-eth";
+
+export function useTargetNetwork(): { targetNetwork: ChainWithAttributes } {
+  const { chain } = useNetwork();
+  const { targetNetwork, setTargetNetwork } = useGlobalState(({ targetNetwork, setTargetNetwork }) => ({
+    targetNetwork,
+    setTargetNetwork,
+  }));
```
technophile-04, line 13: I think we could simply this 😅 : 

```ts
  const { targetNetwork, setTargetNetwork } = useGlobalState(state => ({
    targetNetwork: state.targetNetwork,
    setTargetNetwork: state.setTargetNetwork,
  }));
```

If we are concerned with referential equality maybe we could use pass `shallow` comparison to `useGlobalState`  or get each prop individually (which look fine too)

checkout this discussion : https://github.com/pmndrs/zustand/discussions/1916

packages/nextjs/hooks/scaffold-eth/useTargetNetwork.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  useEffect(() => {
+    const newSelectedNetwork = scaffoldConfig.targetNetworks.find(targetNetwork => targetNetwork.id === chain?.id);
+    if (newSelectedNetwork && newSelectedNetwork.id !== targetNetwork.id) {
+      setTargetNetwork(newSelectedNetwork);
+    }
+  }, [chain, setTargetNetwork, targetNetwork.id]);
```
technophile-04, line 20: maybe we do `chain?.id` ?

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
```diff
@@ -38,7 +37,7 @@ export const useDeployedContractInfo = <TContractName extends ContractName>(cont
     };
 
     checkContractDeployment();
-  }, [isMounted, contractName, deployedContract, publicClient]);
+  }, [isMounted, contractName, deployedContract, publicClient, targetNetwork]);
```
technophile-04, line 40: we could also instead do `targeNetwork.id`

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
```diff
@@ -91,20 +102,42 @@ export const RainbowKitCustomConnectButton = () => {
...
+                            }}
+                          >
+                            <ArrowsRightLeftIcon className="h-6 w-4 ml-2 sm:ml-0" /> Switch to{" "}
+                            <span
+                              style={{
+                                color: getNetworkColor(network, isDarkMode),
+                              }}
+                            >
+                              {network.name}
+                            </span>
```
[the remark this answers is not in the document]
sverps: fixed :+1:

packages/nextjs/hooks/scaffold-eth/useTargetNetwork.ts
sverps: Don't think it matter here (I don't believe the reference to `chain` will ever change without the `id` changing).

But I'll update it for consistency :blush:

packages/nextjs/hooks/scaffold-eth/useDeployedContractInfo.ts
sverps: I think it doesn't even need to be in the dependency array, not sure how it got there :sweat:

packages/nextjs/hooks/scaffold-eth/useTargetNetwork.ts
sverps: I split it up, looks cleaner like that

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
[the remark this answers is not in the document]
technophile-04: We recently did multichain stuff in the hackathon and kept the network switch at the footer : 

Screenshot 2023-11-21 at 5 59 36 PM

checkout here => https://footy-dao.vercel.app

No strong feelings about it, just wanted to put the idea here

technophile-04:
Really awesome job @sverps and excited about this! and also nice work making burner switch chains  !!! 


Extending todo's from this https://github.com/scaffold-eth/scaffold-eth-2/pull/615#issuecomment-1820885043: 

So currently I see we are assuming that the same contracts are present on others chains as  `scaffoldConfig.targetNetworks[0].id`. 

Like if for eg: `31337: YourContract`  & `1115511 : YourContract, DAI`
1. We are will only give ts autocompletion for `YourContract` and `DAI` is not present in autocompleteions. 
2.  We are not showing `DAI` contract on debug page if the user switches the chain to `sepolia`. 

I mean ideally 70% of the time people will have the same contracts on all configured chains but it might be a chance that they don't or have an extra contract on a particular chain. 

I was just thinking about TS autocompletion what if we do this : 

```ts
type ConfiguredChainIds = (typeof scaffoldConfig)["targetNetworks"][number]["id"];
export type UseScaffoldWriteConfig<
  TChainId extends ConfiguredChainIds = (typeof scaffoldConfig)["targetNetworks"][0]["id"],
  TContractName extends ContractName<TChainId>,
  TFunctionName extends ExtractAbiFunctionNames<ContractAbi<TContractName>, WriteAbiStateMutability>,
> 
```

So we tell people to pass in chain ID, and then constrain `contractName` for autocompletion based on the chain id passed by the user (if the user doesn't pass the chain we give him autocompletion for 0th chain). similar stuff we can do for other custom hooks, what do you think? 

but again really thanks for this,  its already looking good, also thanks to @blahkheart  !! Gonna test it more over the week 🙌

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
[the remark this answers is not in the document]
carletex: The footer option is another alternative! It might be a bit too far from the wallet / network stuff in the header tho.

> Didn't go for that option due to limited space on mobile devices.

The second option I proposed above could work on mobile. So it's something to consider / test.

> How about having a down arrow (▼) next to the name that opens a dropdown (when multiple chains configured?). (The text is small, so the hitbox might be too little)

packages/nextjs/pages/_app.tsx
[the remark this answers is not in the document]
carletex: That makes sense!

We could also have `ScaffoldEthAppWithProviders` where we have all the declared providers.

packages/nextjs/contracts/deployedContracts.ts
carletex, line 1: Let's revert this

packages/nextjs/generated/hardhat_contracts.ts
carletex, line 1: This file doesn't exist anymore. Remove

carletex:
@sverps  A few things changed (like now we are using global state) since my last review and I think there is still some cleanup to do.

So I'll wait until you request for review again.

Thanks!

hcheng826:
Thanks a lot to making this change! Looking forward to this so much!

sverps:
> So we tell people to pass in chain ID, and then constrain `contractName` for autocompletion based on the chain id passed by the user (if the user doesn't pass the chain we give him autocompletion for 0th chain). similar stuff we can do for other custom hooks, what do you think?

I gave it a go, and it got complex quite quickly. :see_no_evil: 

If we just make it a generic type, then for it to have a default value, it can't be the first generic. So that implies users would need to specify the `TContractName` and `TFunctionName` as well, which is a bit weird and hard to explain.

When they do so, it will only do autocompletion based on the given chainId, but it's really only typescript, so internally the hook will continue to use whichever chain the user is actually on.

So in order to make those two things work, the hook should really accept an optional `chainId` param, but that needs to be propagated through all types all the way to the top. I started doing that, and things started breaking all over. In the end I was blocked by an error thrown by abitype's `ExtractAbiFunctionNames` that I didn't manage to resolve.

Maybe there's a way, but it started feeling like it wasn't really worth it... :sweat_smile: 

I think, for now, it is an acceptable trade-off if users have autocompletion based on the default chain. Multi-chain projects that use different contract names and functions on each chain are probably advanced enough that devs working on that will understand how to use wagmi directly, I suppose.

(In trying out this idea, I did find a small issue with how the scaffold config file type was defined, not sure why it didn't pop up earlier.)

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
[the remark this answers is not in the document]
sverps: > How about having a down arrow (▼) next to the name that opens a dropdown (when multiple chains configured?). (The text is small, so the hitbox might be too little)

Currently, clicking the displayed number toggles between ETH and $ value, so adding in an arrow button likely won't work without additional changes (possibly removing the ETH <-> $ toggle functionality?).

I'd personally keep it like this, since it feels like a logical place to check for a network switch button, as it's part of the "wallet" menu. We can always ask some opinions in one of the Telegram channels to get some more input?

packages/nextjs/pages/_app.tsx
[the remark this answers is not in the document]
sverps: Cleaned it up a little with your suggestion :+1:

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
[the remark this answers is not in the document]
carletex: I was talking about clicking the network itself, not the number ($ toggle functionality is great). Something like:

image.

> How about having a down arrow (▼) next to the name that opens a dropdown (when multiple chains configured?). (The text is small, so the hitbox might be too little)

That's the only downside I can think of. 

> I'd personally keep it like this, since it feels like a logical place to check for a network switch button, as it's part of the "wallet" menu. We can always ask some opinions in one of the Telegram channels to get some more input?

But yes! we can keep it like this and iterate if we get some feedback

carletex:
Thanks for raising the autocompletion issue @technophile-04 and @sverps for looking into it.

If Samuel says it gets too complex in TS, I 100% trust him haha

> I think, for now, it is an acceptable trade-off if users have autocompletion based on the default chain. Multi-chain projects that use different contract names and functions on each chain are probably advanced enough that devs working on that will understand how to use wagmi directly, I suppose.

Unless there was a simple solution (which it seems there is not)... I agree with this!

> (In trying out this idea, I did find a small issue with how the scaffold config file type was defined, not sure why it didn't pop up earlier.)

What is the issue? :nerd_face:

sverps:
> > (In trying out this idea, I did find a small issue with how the scaffold config file type was defined, not sure why it didn't pop up earlier.)
> 
> What is the issue? 🤓

`targetNetworks` was typed as `chains.Chain[]`, which meant accessing the first element with a utility type like so `type ConfiguredChainId = (typeof scaffoldConfig)["targetNetworks"][0]["id"]` was actually equaling an intersection of the options.

In other words, setting target networks as `[chains.hardhat, chains.sepolia]` in the config turned the `ConfiguredChainId` type into `31337 | 11155111`, rather than the expected `31337`. And I got all sorts of errors in the `contract.ts` type file.

Fix was to turn the targetNetworks into a tuple (which necessitated a custom `Tuple` utility type, since the length can vary).

Hope that makes sense :sweat_smile:

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton.tsx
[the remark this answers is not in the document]
sverps: > I was talking about clicking the network itself, not the number

Yeah (as you already mentioned) I think part of the issue is the hitbox would be very small. On top of that, it would be too close to hitbox of the the balance toggle and (especially on mobile) might be annoying to not be able to hit is properly.

What could work is a network toggle using the network icon (like uniswap), that might have a width small enough to fit on mobile. We could reduce the font-size of the balance and address for mobile if needed to create some more space. But I suggest gathering some feedback first and tackling it in another issue, if needed.

technophile-04:
> In other words, setting target networks as [chains.hardhat, chains.sepolia] in the config turned the ConfiguredChainId type into 31337 | 11155111, rather than the expected 31337. And I got all sorts of errors in the contract.ts type file.
> 
> Fix was to turn the targetNetworks into a tuple (which necessitated a custom Tuple utility type, since the length can vary).

Ohh this was strange ideally it shouldn't have happened right? Since we were using `satisfies`  (which I thought worked similarly to `as const`) and we should have gotten literal `31337` instead of unioning it with `11155111`. 

I tried doing some research and found another approach without using a custom Tuple utility and keeping `chains.Chain[]` we can just do this assertion in the end  `as const satisfies ReadonlyDeep<ScaffoldConfig>` where `ReadOnlyDeep` comes from `type-fest` (which we are already using )

---

> https://github.com/scaffold-eth/scaffold-eth-2/pull/615#issuecomment-1828583594

Tysm @sverps for exploring and great explainantion!! Completely agree that it is not in the scope of this PR and may require a lot of changes and brainstorming on changes to the custom hooks API design. 

The reason I mentioned that point was there is a slight difference b/w [cross chain vs multi-chain](https://blog.chain.link/cross-chain-vs-multi-chain/) and people tend to use [cross chain](https://chain.link/education-hub/cross-chain-smart-contracts) smart contracts more and probability of having different Abi's / name in cross-chain stuff is more,  where one chain smart-contract holds more functionalities than others.

Will create a new issue with a more detailed explanation and there we can discuss what's the best approach 🙌 

Also, I think we are already handling mult-chain stuff in this PR nicely 🙌 !!! 


Tysm again Samuel !! Let's get the merge conflict fixed, I think we are almost there merging this 🙌

sverps:
> I tried doing some research and found another approach without using a custom Tuple utility and keeping `chains.Chain[]` we can just do this assertion in the end `as const satisfies ReadonlyDeep<ScaffoldConfig>` where `ReadOnlyDeep` comes from `type-fest` (which we are already using )

Nice find, I'll try it out. Sounds like a better solution than the custom `Tuple` type (that is indeed limited to a certain max amount of chains)

sverps:
@technophile-04 @carletex @rin-st 
Conflicts resolved and final change made. Let me know if I missed anything :blush:

packages/nextjs/scaffold.config.ts
sverps: Ugh, I really need to find a system to prevent that... :sweat_smile:

technophile-04 (approved):
Awesome stuff nicely tested out and works like a charm,  Tysm Samuel !!! 

---

Just found two minor bugs : one with `useScaffoldContract` and other with `useScaffoldEventHistory` 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/300bff9e-e1eb-4770-bd95-2ef1d524977a

Regarding `useScaffoldContract` : If you look at the above video the `read` from the contract instance of `useScaffoldContract` gives `undefined`  ideally it should have given value from `0th` chain if the wallet is not connected (similar to what it's doing when we use `useScaffoldContractRead`)......I think the contract instance we are getting from `useScaffoldContract` is always `undefined` for some reason. 

Regarding `useScaffoldEvenHistory` : It also kind of broken I think when you switch chain, if you look at the video when I switched from sepolia (which had `6` events ) to goerli (which had only `1` event) it showed me `7` I think it appended the `1` goerli event on top of sepolia. 

Here are steps to reproduce this locally : 

1 . Update `targetNetwork : [chains.sepolia, chains.goerli]`

<details>
<summary>
2. Copy paste this in `externalContracts.ts` : 
</summary>

```ts

const externalContracts = {
  5: {
    MyContract: {
      address: "0x5bB28a827e250F6BF45D75809Ff21733e7c6704f",
      abi: [
        {
          inputs: [
            {
              internalType: "address",
              name: "_owner",
              type: "address",
            },
          ],
          type: "constructor",
        },
        {
          anonymous: false,
          inputs: [
            {
              indexed: true,
              internalType: "address",
              name: "previousOwner",
              type: "address",
            },
            {
              indexed: true,
              internalType: "address",
              name: "newOwner",
              type: "address",
            },
          ],
          name: "OwnershipTransferred",
          type: "event",
        },
        {
          inputs: [],
          name: "increaseCounter",
          outputs: [],
          stateMutability: "payable",
          type: "function",
        },
        {
          inputs: [],
          name: "owner",
          outputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "renounceOwnership",
          outputs: [],
          type: "function",
        },
        {
          inputs: [],
          name: "totalCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "address",
              name: "newOwner",
              type: "address",
            },
          ],
          name: "transferOwnership",
          outputs: [],
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          name: "userGreetingCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "withdraw",
          outputs: [],
          type: "function",
        },
        {
          stateMutability: "payable",
          type: "receive",
        },
      ],
      inheritedFunctions: {
        owner: "@openzeppelin/contracts/access/Ownable.sol",
        renounceOwnership: "@openzeppelin/contracts/access/Ownable.sol",
        transferOwnership: "@openzeppelin/contracts/access/Ownable.sol",
      },
    },
    YourContract: {
      address: "0x1a2de031A50Fe9918f0BCe71Cb060C8CA3911f55",
      abi: [
        {
          inputs: [
            {
              internalType: "address",
              name: "_owner",
              type: "address",
            },
          ],
          type: "constructor",
        },
        {
          anonymous: false,
          inputs: [
            {
              indexed: true,
              internalType: "address",
              name: "greetingSetter",
              type: "address",
            },
            {
              indexed: false,
              internalType: "string",
              name: "newGreeting",
              type: "string",
            },
            {
              indexed: false,
              internalType: "bool",
              name: "premium",
              type: "bool",
            },
            {
              indexed: false,
              internalType: "uint256",
              name: "value",
              type: "uint256",
            },
          ],
          name: "GreetingChange",
          type: "event",
        },
        {
          inputs: [],
          name: "greeting",
          outputs: [
            {
              internalType: "string",
              name: "",
              type: "string",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "owner",
          outputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "premium",
          outputs: [
            {
              internalType: "bool",
              name: "",
              type: "bool",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "string",
              name: "_newGreeting",
              type: "string",
            },
          ],
          name: "setGreeting",
          outputs: [],
          stateMutability: "payable",
          type: "function",
        },
        {
          inputs: [],
          name: "totalCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          name: "userGreetingCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "withdraw",
          outputs: [],
          type: "function",
        },
        {
          stateMutability: "payable",
          type: "receive",
        },
      ],
      inheritedFunctions: {},
    },
  },
  11155111: {
    MyContract: {
      address: "0xd7036EDDDA3743Aa577DC24ca817f3F85099bDbD",
      abi: [
        {
          inputs: [
            {
              internalType: "address",
              name: "_owner",
              type: "address",
            },
          ],
          type: "constructor",
        },
        {
          anonymous: false,
          inputs: [
            {
              indexed: true,
              internalType: "address",
              name: "previousOwner",
              type: "address",
            },
            {
              indexed: true,
              internalType: "address",
              name: "newOwner",
              type: "address",
            },
          ],
          name: "OwnershipTransferred",
          type: "event",
        },
        {
          inputs: [],
          name: "increaseCounter",
          outputs: [],
          stateMutability: "payable",
          type: "function",
        },
        {
          inputs: [],
          name: "owner",
          outputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "renounceOwnership",
          outputs: [],
          type: "function",
        },
        {
          inputs: [],
          name: "totalCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "address",
              name: "newOwner",
              type: "address",
            },
          ],
          name: "transferOwnership",
          outputs: [],
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          name: "userGreetingCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "withdraw",
          outputs: [],
          type: "function",
        },
        {
          stateMutability: "payable",
          type: "receive",
        },
      ],
      inheritedFunctions: {
        owner: "@openzeppelin/contracts/access/Ownable.sol",
        renounceOwnership: "@openzeppelin/contracts/access/Ownable.sol",
        transferOwnership: "@openzeppelin/contracts/access/Ownable.sol",
      },
    },
    YourContract: {
      address: "0xE009aea21af005e6B531B5f4a8f909C64A0c596d",
      abi: [
        {
          inputs: [
            {
              internalType: "address",
              name: "_owner",
              type: "address",
            },
          ],
          type: "constructor",
        },
        {
          anonymous: false,
          inputs: [
            {
              indexed: true,
              internalType: "address",
              name: "greetingSetter",
              type: "address",
            },
            {
              indexed: false,
              internalType: "string",
              name: "newGreeting",
              type: "string",
            },
            {
              indexed: false,
              internalType: "bool",
              name: "premium",
              type: "bool",
            },
            {
              indexed: false,
              internalType: "uint256",
              name: "value",
              type: "uint256",
            },
          ],
          name: "GreetingChange",
          type: "event",
        },
        {
          inputs: [],
          name: "greeting",
          outputs: [
            {
              internalType: "string",
              name: "",
              type: "string",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "owner",
          outputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "premium",
          outputs: [
            {
              internalType: "bool",
              name: "",
              type: "bool",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "string",
              name: "_newGreeting",
              type: "string",
            },
          ],
          name: "setGreeting",
          outputs: [],
          stateMutability: "payable",
          type: "function",
        },
        {
          inputs: [],
          name: "totalCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          name: "userGreetingCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "withdraw",
          outputs: [],
          type: "function",
        },
        {
          stateMutability: "payable",
          type: "receive",
        },
      ],
      inheritedFunctions: {},
    },
  },
} as const;

```


</details>


<details>
<summary>
3 . Copy pasted this in index.tsx
</summary>

```tsx
import { useEffect, useState } from "react";
import { parseEther } from "viem";
import { useAccount, useWalletClient } from "wagmi";
import { MetaHeader } from "~~/components/MetaHeader";
import { Address, InputBase } from "~~/components/scaffold-eth";
import {
  useScaffoldContract,
  useScaffoldContractRead,
  useScaffoldContractWrite,
  useScaffoldEventHistory,
} from "~~/hooks/scaffold-eth";

  const { address: connectedAddress } = useAccount();

  const [inputGreetings, setInputGreetings] = useState("");

  const { writeAsync: writeSetGreetings } = useScaffoldContractWrite({
    contractName: "YourContract",
    functionName: "setGreeting",
    args: [inputGreetings],
    value: parseEther("0.01"),
  });

  const { data: totalCounter } = useScaffoldContractRead({
    contractName: "YourContract",
    functionName: "totalCounter",
    watch: true,
  });

  const { data: walletClient } = useWalletClient();
  const { data: YourContract } = useScaffoldContract({
    contractName: "YourContract",
    walletClient,
  });

  useEffect(() => {
    const fetchTotalCounter = async () => {
      console.log("YourContract", YourContract);
      const value = await YourContract?.read.totalCounter();
      console.log("totalCounter", value);
    };

    fetchTotalCounter();
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);

  const { data: eventHistory } = useScaffoldEventHistory({
    contractName: "YourContract",
    eventName: "GreetingChange",
    filters: { greetingSetter: connectedAddress },
    fromBlock: 0n,
    watch: true,
  });
  console.log("eventHistory", eventHistory);

  return (
    <>
      <MetaHeader />
        <div className="text-center mb-8">
          <span className="block text-2xl mb-2">The connected address is</span>
        </div>
        <div className="px-5 space-y-4">
          <h1 className="text-center mb-4">
            <span className="block text-3xl mb-2">With Hooks</span>
          </h1>
          <h1 className="text-center mb-4">
            <span className="block text-2xl mb-2">The totalCounter is</span>
            <span className="block text-4xl font-bold">{totalCounter?.toString()}</span>
          </h1>
          <InputBase onChange={setInputGreetings} value={inputGreetings} placeholder="Set greetings" />
          <button className="btn btn-primary btn-sm ml-12 mb-2" onClick={() => writeSetGreetings()}>
            Set Greetings
          </button>
          <h1 className="text-center mt-8">
            <span className="block text-3xl mb-2">Imperative</span>
          </h1>
          <button
            className="btn btn-primary btn-sm"
            onClick={async () => {
              await YourContract?.write.setGreeting([inputGreetings], { value: parseEther("0.01") });
            }}
          >
            Set Imperative Greetings
          </button>
        </div>
      </div>
    </>
  );
};

export default Home;


```

</details>

If it get too complicated to handle maybe we can handle it in a different PR and even @damianmarti can help us with the issue with `useScffoldEventHistory` there 🙌

---


So the TODO's after this PR might be : 

- [ ] Placement of network switch button 
- [ ] Handle bugs for `useScaffoldContract` and `useScaffoldEventHistory`
- [ ] Handle autocompletions even if there are different contract names  

Thanks  again Samuel and others for reviews 🙌

sverps:
> Regarding useScaffoldContract : If you look at the above video the read from the contract instance of useScaffoldContract gives undefined ideally it should have given value from 0th chain if the wallet is not connected (similar to what it's doing when we use useScaffoldContractRead)......I think the contract instance we are getting from useScaffoldContract is always undefined for some reason.

`useScaffoldContract` works with a loading state (because internally in `useDeployedContractInfo`, it checks the bytecode to see if it is deployed), so the contract is indeed undefined when initializing. If you add `YourContract` to the dependency array of the `useEffect`, you'll see it reads the value properly as soon as the check resolves. (The same behavior occurs on our current `main` branch)

> Regarding useScaffoldEvenHistory

Good find :+1: 
I fixed it by clearing the internal state when the targetNetwork changes

technophile-04:
> If you add YourContract to the dependency array of the useEffect, you'll see it reads the value properly as soon as the check resolves.

Ohh shit, my bad...yup it works nicely 🙌

> I fixed it by clearing the internal state when the targetNetwork changes

Yup makes sense, but still sometimes it still works a bit weirdly : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/e517f421-bbee-4488-b5ad-17c8d406f3c0

I think maybe it's worth handling in different PR nicely, will create an issue for it 

--- 

Tysm all merging this 🙌 !!

carletex:
This is a great job @sverps. This feature was a long time coming!

Thanks all

PR #621, opened 2023-11-22
Add requiredFilters param to useScaffoldEventHistory hook

damianmarti:
## Description

Add requiredFilters param to useScaffoldEventHistory hook to avoid getting wrong events when setting a filter that is not yet populated with the right value.

If one of the requiredFilters is not set, the hook returns an error.

For example, if I get the filter value from the router query, the first value is undefined, and the hook will return all the events, without this filter applied.

To fix it, I can use a useEffect hook, something like this:

```
const { roomHash } = router.query as { roomHash: 0x${string} };

const [roomHashFilter, setRoomHashFilter] = useState<string>("aa");

  useEffect(() => {
    if (roomHash) {
      setRoomHashFilter(roomHash);
    }
  }, [roomHash]);

  const {
    data: roomJoinEvents,
    isLoading: isLoadingRoomCreateEvents,
    error: errorReadingRoomCreateEvents,
  } = useScaffoldEventHistory({
    contractName: "BettingRoom",
    eventName: "RoomJoin",
    fromBlock: 1n,
    filters: { roomHash: roomHashFilter },
    watch: true,
  });
```

But if I use the requieredFilters params I only need the useScaffoldEventHistory hook:

```
const { roomHash } = router.query as { roomHash: 0x${string} };

  const {
    data: roomJoinEvents,
    isLoading: isLoadingRoomCreateEvents,
    error: errorReadingRoomCreateEvents,
  } = useScaffoldEventHistory({
    contractName: "BettingRoom",
    eventName: "RoomJoin",
    fromBlock: 1n,
    filters: { roomHash: roomHash },
    watch: true,
    requiredFilters: ["roomHash"],
  });
```

Or maybe there is another better option that I'm missing :-)

## Additional Information

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
@@ -26,6 +26,7 @@ import {
  * @param config.transactionData - if set to true it will return the transaction data for each event (default: false)
  * @param config.receiptData - if set to true it will return the receipt data for each event (default: false)
  * @param config.watch - if set to true, the events will be updated every pollingInterval milliseconds set at scaffoldConfig (default: false)
+ * @param config.requiredFilters - if a required filter is not set, no events are returned and the error is set (default: empty)
  */
 export const useScaffoldEventHistory = <
   TContractName extends ContractName,
@@ -42,6 +43,7 @@ export const useScaffoldEventHistory = <
   transactionData,
   receiptData,
   watch,
+  requiredFilters,
 }: UseScaffoldEventHistoryConfig<TContractName, TEventName, TBlockData, TTransactionData, TReceiptData>) => {
   const [events, setEvents] = useState<any[]>();
   const [isLoading, setIsLoading] = useState(false);
@@ -62,6 +64,13 @@ export const useScaffoldEventHistory = <
       const event = (deployedContractData.abi as Abi).find(
         part => part.type === "event" && part.name === eventName,
       ) as AbiEvent;
+
+      if (requiredFilters && requiredFilters.length > 0) {
+        const missingFields = requiredFilters.filter(field => !filters || !filters[field]);
+        if (missingFields.length > 0) {
+          throw new Error(`Missing required fields: ${missingFields.join(", ")}`);
+        }
+      }
 
       const blockNumber = await publicClient.getBlockNumber({ cacheTime: 0 });
 
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
@@ -43,7 +43,7 @@ export const useScaffoldEventHistory = <
   transactionData,
   receiptData,
   watch,
-  requiredFilters,
+  enabled = true,
 }: UseScaffoldEventHistoryConfig<TContractName, TEventName, TBlockData, TTransactionData, TReceiptData>) => {
   const [events, setEvents] = useState<any[]>();
   const [isLoading, setIsLoading] = useState(false);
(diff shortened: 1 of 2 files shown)


rin-st:
what about adding `enabled` flag like for example [here](https://wagmi.sh/react/hooks/useContractRead#enabled-optional)? So if we have that flag is false then return `undefined` or empty array (why it can be `undefined` btw?)

damianmarti:
> ```ts
> router.query
> ```

Thanks for the feedback!

I think this is a good option too! Not sure what is better ;-)

roomHash is undefined before it's hydrated with the router query data.

sverps:
I like the `enabled` flag, because it follows the way that wagmi hooks do it, so should be somewhat familiar to devs.

damianmarti:
@rin-st @sverps Changed to use enabled flag!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -59,6 +61,10 @@ export const useScaffoldEventHistory = <
         throw new Error("Contract not found");
       }
 
+      if (!enabled) {
+        throw new Error("Hook disabled");
```
rin-st, line 65: Thanks for update!

But I think it shouldn't be error. If I assign `enabled = false` I'm expecting that there's no data and there's no error. I think 
```ts
if (!enabled) {
  return;
}
```
is ok.

Because let's say you handle errors in your react component

```ts
const {data, error} = useScaffoldEventHistory(...);

if (error) {
  return <p className='text-red-500'>{error}</div>
}

return {data?.map(...)}

```
Your example when filters is empty until query is parsed will show error first and then error will disappear. I think it's not good

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -26,6 +26,7 @@ import {
  * @param config.transactionData - if set to true it will return the transaction data for each event (default: false)
  * @param config.receiptData - if set to true it will return the receipt data for each event (default: false)
  * @param config.watch - if set to true, the events will be updated every pollingInterval milliseconds set at scaffoldConfig (default: false)
+ * @param config.requiredFilters - if a required filter is not set, no events are returned and the error is set (default: empty)
```
rin-st, line 29: forgot to remove

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
damianmarti: Yeah! Makes sense. Thanks!

Maybe we need:

```
if (!enabled) {
  setEvents(undefined);
  return;
}
```

To clean up the previous events when enabled true is swtiched to enabled false, right?

rin-st: Not sure, because if you remove events you also need to watch `fromBlockUpdated` and update it accordingly. If I understand right, returning `undefined` is better. For example in case when your `enabled` becomes `true`->`false` -> `true`, hook just adds events from last `fromBlockUpdated` and not reloading everything.

damianmarti: I added this https://github.com/scaffold-eth/scaffold-eth-2/pull/621/files#diff-0efe8962fd8674c8a04310503b57e5e0ff36eb7016ae11764e1cc83472bc7646R122 to reload all the events when enabled change. I think that makes sense that if the hook is disabled, events is undefined.

rin-st: you're right, approve!

rin-st (approved):

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
damianmarti: Removed and added enabled

carletex:
Thanks @damianmarti for making the hooks better.

And thanks all for the review!!!

Issue #617, opened 2023-11-21
Consistency codebase refactor


carletex:
Let's do some refactoring to have a more consistent codebase:

- [x] It seems that we are more inclined to use types without `T` prefix. Let's unify for consistency. https://github.com/scaffold-eth/scaffold-eth-2/pull/627
- [x] `0x${string}` to viem's `Address`? [WIP @sverps ]
- [x] TS: interfaces -> types https://github.com/scaffold-eth/scaffold-eth-2/pull/627
- [ ] JSDoc tweaks: see https://github.com/scaffold-eth/scaffold-eth-2/pull/627#issuecomment-1828163974

Check TS guide: https://github.com/scaffold-eth/scaffold-eth-2/issues/232

What else? cc @technophile-04 @rin-st @sverps

rin-st:
- TS: interfaces -> types
- [components/scaffold-eth/Contract](https://github.com/scaffold-eth/scaffold-eth-2/tree/main/packages/nextjs/components/scaffold-eth/Contract) utilsContract.tsx, utilsDisplay.tsx -> utils.ts. Also refactor it, for example move `getParsedError` to `utils/scaffold-eth`
- remove barrels?)

carletex:
Updated OP with new stuff. Also opened #627 

I also think we should have all the desired consistency documented up somewhere (with conclusions from https://github.com/scaffold-eth/scaffold-eth-2/issues/232 too).

Not sure where the best place could be: Github wiki? SE-2 docs -> contributing?

---

> remove barrels?

Yep, this is an interesting discussion too. I remember seeing some issues in the NextJS 14 migration with HMR.

technophile-04:
> Not sure where the best place could be: Github wiki? SE-2 docs -> contributing?

Didn't knew about Github Wiki! I think it looks a fare place 🙌


> > remove barrels?
> 
> Yep, this is an interesting discussion too. I remember seeing some issues in the NextJS 14 migration with HMR.

I am kind of 60%(remvoing barrels) and 40% keeping them because I think in new Next version there is option to mark barrels file and optimize them => https://vercel.com/blog/how-we-optimized-package-imports-in-next-js but don't know if it actually work and 100% assured solution will be removing barrels files

> - [ ] JSDoc tweaks: see https://github.com/scaffold-eth/scaffold-eth-2/pull/627#issuecomment-1828163974

Is anybody already working on this TODO ? Would love to tackle it 🙌

[...]

PR #630, opened 2023-11-29
fix(address type): allows user to set his preference in scaffold config

### LINKED ISSUE #514: Cleanup of abi Address type

sverps:
I aligned all addresses to use the Address type of viem (see #617).

Also added a config option for a dev to use more strict Address if he wants additional typesafety (default is current behavior where addresses are typed as `string`).

(This also basically fixes #514 by making it optional, with a sensible default)

### DIFF
--- a/packages/nextjs/components/blockexplorer/AddressStorageTab.tsx
+++ b/packages/nextjs/components/blockexplorer/AddressStorageTab.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useState } from "react";
-import { createPublicClient, http, toHex } from "viem";
+import { Address, createPublicClient, http, toHex } from "viem";
 import { hardhat } from "viem/chains";
 
 const publicClient = createPublicClient({
@@ -7,7 +7,7 @@ const publicClient = createPublicClient({
   transport: http(),
 });
 
-export const AddressStorageTab = ({ address }: { address: string }) => {
+export const AddressStorageTab = ({ address }: { address: Address }) => {
   const [storage, setStorage] = useState<string[]>([]);
 
   useEffect(() => {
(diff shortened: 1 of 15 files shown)


technophile-04:
> Also added a config option for a dev to use a more strict Address if he wants additional typesafety (default is current behavior where addresses are typed as string).

Love this idea, best approach !!!!!

rin-st (approved):
👍

technophile-04 (approved):
Nice !! Tested it thoroughly and works like a charm !!! Tysm @sverps !

Just a small discussion question should we keep `useStrictAddressType: true` by default (so that we are inlined with wagmi / viem by default and people can tweak it to `false` if they want ?)

I would kind of vote for keeping it `useStrictAddressType: false` (which is already current state) but not completely sure and would love to know others opinion 🙌


But yes really love the approach and already is ready for merge 🙌

sverps:
As much as I prefer stricter types, I'd go for the newbie-friendly approach as that's a big group of the target users.

PR #638, opened 2023-12-02
Fix displaying of custom solidity errors

technophile-04:
## Description

This problem occurs when you have custom solidity errors and you try to read/write from Debug page : 

<details>
<summary> Test YourContract.sol </summary>

```solidity
//SPDX-License-Identifier: MIT
pragma solidity >=0.8.0 <0.9.0;

// Useful for debugging. Remove when deploying to a live network.
import "hardhat/console.sol";

error NOT_OWNER();

error MY_CUSTOM_ERROR(string message, uint value);

contract YourContract {
	// State Variables
	address public immutable owner;
	string public greeting = "Building Unstoppable Apps!!!";
	uint256 public totalCounter = 0;

	constructor(address _owner) {
		owner = _owner;
	}

	modifier isOwner() {
		if (msg.sender != owner) {
			revert NOT_OWNER();
		}
		_;
	}

	function revertWithCustomErrror(bool _shouldRevert) public view {
		if (_shouldRevert) {
			revert MY_CUSTOM_ERROR(
				"Error with my custom message",
				totalCounter
			);
		}
	}

	function revertWithRequire(bool _shouldRevert) public pure {
		require(!_shouldRevert, "Error with require message");
	}

	function setGreeting(string memory _newGreeting) public payable {
		// Change state variables
		greeting = _newGreeting;
		totalCounter += 1;
	}

	function withdraw() public isOwner {
		(bool success, ) = owner.call{ value: address(this).balance }("");
		require(success, "Failed to send Ether");
	}

	receive() external payable {}
}

```

</details>

### Before (`main` branch) : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/7dd91f02-89e5-4041-b4c2-30a3b61567ce

### After : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/f1ccc5dd-a79d-4c85-aea2-2fdf5e4d4158

## Cause : 

When passing `abi` to wagmi's native hooks we were just passing that function abi eg: `abi: [functionABI]` because of which wagmi was not able to parse the errors.

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
@@ -25,7 +25,7 @@ const getParsedError = (e: any): string => {
       message = e.details;
     } else if (e.shortMessage) {
       message = e.shortMessage;
-      const cause = e.cause as any as { data?: DecodeErrorResultReturnType };
+      const cause = e.cause as { data?: DecodeErrorResultReturnType };
       // if its not generic error, append custom error name and its args to message
       if (cause.data && cause.data?.abiItem?.name !== "Error") {
         const customErrorArgs = cause.data.args?.toString() ?? "";
--- a/packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
@@ -1,5 +1,5 @@
 import { AbiFunction, AbiParameter } from "abitype";
-import { BaseError as BaseViemError } from "viem";
+import { BaseError as BaseViemError, DecodeErrorResultReturnType } from "viem";
 
 /**
  * Generates a key based on function metadata
@@ -18,14 +18,22 @@ const getFunctionInputKey = (functionName: string, input: AbiParameter, inputInd
  * @param e - error object
  * @returns {string} parsed error string
  */
-const getParsedError = (e: any | BaseViemError): string => {
-  let message = e.message ?? "An unknown error occurred";
-
+const getParsedError = (e: any): string => {
+  let message: string = e.message ?? "An unknown error occurred";
   if (e instanceof BaseViemError) {
     if (e.details) {
       message = e.details;
     } else if (e.shortMessage) {
       message = e.shortMessage;
+      const cause = e.cause as any as { data?: DecodeErrorResultReturnType };
+      if (cause.data && cause.data?.abiItem?.name !== "Error") {
+        const customErrorArgs = cause.data.args?.toString() ?? "";
+        message = `${message.replace(/reverted\.$/, "reverted with following reason:")}\n${
+          cause.data.errorName
+        }(${customErrorArgs})`;
+        console.log("The cause full data", cause.data);
+        console.log("The message is:", message);
+      }
     } else if (e.message) {
       message = e.message;
     } else if (e.name) {
--- a/packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
+++ b/packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
@@ -26,6 +26,7 @@ const getParsedError = (e: any): string => {
     } else if (e.shortMessage) {
       message = e.shortMessage;
       const cause = e.cause as any as { data?: DecodeErrorResultReturnType };
+      // if its not generic error, append custom error name and its args to message
       if (cause.data && cause.data?.abiItem?.name !== "Error") {
         const customErrorArgs = cause.data.args?.toString() ?? "";
         message = `${message.replace(/reverted\.$/, "reverted with following reason:")}\n${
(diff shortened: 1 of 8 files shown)


packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -18,14 +18,21 @@ const getFunctionInputKey = (functionName: string, input: AbiParameter, inputInd
...
-  let message = e.message ?? "An unknown error occurred";
-
+const getParsedError = (e: any): string => {
+  let message: string = e.message ?? "An unknown error occurred";
   if (e instanceof BaseViemError) {
     if (e.details) {
       message = e.details;
     } else if (e.shortMessage) {
       message = e.shortMessage;
+      const cause = e.cause as any as { data?: DecodeErrorResultReturnType };
```
technophile-04, line 28: Had to do this :( , I think there is some problem with `viem's`  [`BaseError`](https://viem.sh/docs/glossary/errors.html#errors) type although on runtime it `e.cause.data` is present when function always revert but `viem` type doesn't have typing for it and is typed as `unknown`. 


I also tried an approach by narrowing the type with `ContractFunctionExecutionError` instead of `BaseViemError` but it too does not have proper typing for `e.cause.data`

The reason we are trying to get `e.cause.data` because it is the only place from where you can get details about Custom error checkout this : https://github.com/wevm/viem/discussions/589#discussioncomment-6028456

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
```diff
@@ -18,14 +18,21 @@ const getFunctionInputKey = (functionName: string, input: AbiParameter, inputInd
...
+      // if its not generic error, append custom error name and its args to message
+      if (cause.data && cause.data?.abiItem?.name !== "Error") {
+        const customErrorArgs = cause.data.args?.toString() ?? "";
+        message = `${message.replace(/reverted\.$/, "reverted with following reason:")}\n${
```
technophile-04, line 32: example message: `The contract function "withdraw" reverted.`

Since the message contains a double quote, had to use `/reverted\.$/` regex, simply doing `message.replace("reverted.", "...")` doesn't work. 

Am I missing something? 

```ts

// 
message.replace("reverted." , "...")

```

technophile-04: Ohh yeah it works with `.replace("reverted.")` , actually my initial logic while debugging was a bit messier and might have made a mistake there because of which `.replace("reverted.")` was not working I thought it was due to presence of double quotes in `message` but I am wrong here

> I like your approach, where you check that the "reverted." string is at the end of the message.

TBH it was github copilot :), but yeah checking "reverted." is very end of string make more sense so keeping regex 🙌

carletex: > TBH it was github copilot :)

Like 90% of my coding these days haha

damianmarti (approved):
Great improvement @technophile-04 !!

Everything looks great to me!

One small comment that is related to this change, but it really affects all the notifications. The text is cut in an unfriendly way at the end of each line, like for example on the second video:

github com_scaffold-eth_scaffold-eth-2_pull_638

"reas" -> "on" and "cu" -> "stom". Maybe changing the text-wrap CSS style is enough to fix it. Maybe we can create a new issue for this...

technophile-04:
> The text is cut in an unfriendly way at the end of each line, like for example on the second video:

Ohh yes ! lol it has been bugging me too for a while, just removed word break `break-all` `className` at [fb2d440](https://github.com/scaffold-eth/scaffold-eth-2/pull/638/commits/fb2d4408ace04e92fdc343f01e2015eaf47579ee) : 

| Before | After | 
| ------ | ------ | 
| Screenshot 2023-12-07 at 9 37 36 AM | Screenshot 2023-12-07 at 9 34 34 AM | 

Like the `Before` one looks a bit better and smaller but I think it makes more sense to have `After` due to its nice readability. 

 I tried playing with some basic styles like adding positioning `absolute` to the "X" icon but still it didn't look that great maybe I think its better if we tackle it in another issue and tinker more with styles 🙌

packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx
sverps: Since `e.cause` is of type `unknown`, you should be able to cast it directly to the desired type without the `as any` in the middle.

So this should work:
```ts
const cause = e.cause as { data?: DecodeErrorResultReturnType };
```

sverps (approved):
Looks good. Added 1 small comment about removing the `as any`. Besides that, it looks good to go :+1:

carletex:
> break-words should work, but it translates into overflow-wrap: break-word which for some reason, doesn't work.

It seems that `overflow-x-hidden break-words` make it work

technophile-04:
> The problem with this is that big strings without spaces go over the edge
> It seems that `overflow-x-hidden break-words` make it work

Ohh yes! Tysm for the solution it works great !!, updated at [f7e960a](https://github.com/scaffold-eth/scaffold-eth-2/pull/638/commits/f7e960a2eb4751d9600ef2f9c55ea410beeb8c44)

Screenshot 2023-12-07 at 7 48 23 PM

PR #643, opened 2023-12-03
Contract Deploy - Check for exact path segment when extracting contract path in artifacts

Corantin:
### Error

I ran into some problems when importing the open-zepline  ERC721 contract.
Trying to import this:
image

Having this error when deploying at the moment it runs the 99_generateTsAbis.ts script:
image

### Explication:
In these lines, it tried to look for ERC721 in available entries (sources), however, it found IERC721 before the actual ERC721.
So he resolved with the path of IERC721 instead.

### Fix
Looking for the exact path segment using '/' as prefix for includes argument did the job:
E.g. Looking for `/ERC721` in 
[
"@openzeppelin/contracts/interfaces/IERC721.sol",
"@openzeppelin/contracts/token/ERC721.sol"
]

Only the second entry matched this time (expected one).

## Additional Information

Your ENS/address: Gossman.eth

### DIFF
--- a/packages/hardhat/deploy/99_generateTsAbis.ts
+++ b/packages/hardhat/deploy/99_generateTsAbis.ts
@@ -60,7 +60,7 @@ function getInheritedFunctions(sources: Record<string, any>, contractName: strin
   const inheritedFunctions = {} as Record<string, any>;
 
   for (const sourceContractName of actualSources) {
-    const sourcePath = Object.keys(sources).find(key => key.includes(sourceContractName));
+    const sourcePath = Object.keys(sources).find(key => key.includes(`/${sourceContractName}`));
     if (sourcePath) {
       const sourceName = sourcePath?.split("/").pop()?.split(".sol")[0];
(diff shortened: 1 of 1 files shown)


carletex:
Thanks for reporting this @Corantin!!

Weird that this has happened before! In my case `ERC721.sol` seems to always be before `IERC721.sol`, but it's not something we can rely on.

carletex (approved):

technophile-04:
> Weird that this hasn't happened before! In my case ERC721.sol seems to always be before IERC721.sol, but it's not something we can rely on.

++, and also same for me ERC721 appears before IERC721

but Tysm @Corantin for reporting and tackling it, the solution looks fine to me  🙌

Corantin:
> > Weird that this hasn't happened before! In my case ERC721.sol seems to always be before IERC721.sol, but it's not something we can rely on.
> 
> ++, and also same for me ERC721 appears before IERC721
> 
> but Tysm @Corantin for reporting and tackling it, the solution looks fine to me 🙌

No problem, happy to have 1 line for this repo. Congrats for the amazing job <3

PR #648, opened 2023-12-07
Weekly CLI backmerge 

technophile-04:
## Description

- fix(address type): allows user to set his preference in scaffold config (#630)
- patch scaffoldConfig type error (#641)
- Check for exact path segment for inheritedFunctions sources (#643)
- Fix displaying of custom solidity errors (#638)
- Check `cause?.data` on getParsedError (#649)

PR #649, opened 2023-12-07
patch: add check before accessing `cause.data` in getParsedError

technophile-04:
## Description

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/7bb60bf3-ddb8-4b16-a527-d0988ee62ef7

Noticed while testing #648 and was introduced in #638 my bad 🤦‍♂️ 

I will also raise an issue on VIem so that in the future we don't need to assert typing manually and get proper typing from it directly when narrowed through `ContractFunctionExecutionError`


carletex (approved):
Thank you!

PR #656, opened 2023-12-11
Daisy UI v4

### LINKED ISSUE #629: Upgrade to DaisyUI v4
It seems that DaisyUI v4 is already out : 

Upgrade Guide : https://github.com/saadeghi/daisyui/discussions/2507

Changelog : https://daisyui.com/docs/changelog/#400-2023-11-12

technophile-04:
## Description

Fixes #629


Pabl0cks (commented):
GJ @technophile-04 !! This is looking good to me 🙌

Only found something a bit weird with the button background colors on hover, I'll send a couple of videos (very bad quality after compressing, can send uncompress in tg):

In the rainbowkit and faucet buttons is easier to see.

Buttons_hover bg-color-2
Buttons_hover bg-color-1

technophile-04:
> Only found something a bit weird with the button background colors on hover, 

Ohh yes didn't notice it initially, actually this behavior was updated in daisy UI v4 by default  :( 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/240dd6ae-6559-4801-ad78-12044e901129


Left side is V4 and the right side is v3. 

---

Tried digging in and the problem seems that DaisyUI v4 has removed the `*-focus`  i.e they are also no more generating `--sf` (secondary-focus) color and `--pf` (primary-focus) color instead they are suggesting to manually do it with color-mix checkout the docs here -> [Breaking changes-colorTheme](https://daisyui.com/docs/changelog/#breaking-changes)

| Daisy V4 hover classes | Dasiy V3 hover classes |
| ----------------------- | ----------------------- | 
|  Screenshot 2023-12-12 at 9 42 51 PM |  Screenshot 2023-12-12 at 9 39 50 PM | 


---
### Possible Solution : 

We can add something like this `tailwind.config.ts` : 

```ts
".btn-secondary:hover": {
  backgroundColor: "color-mix(in oklab, oklch(var(--s)), black 7%)",
  borderColor: "color-mix(in oklab, oklch(var(--s)),black 7%)",
},
```

but then we have to do it for `.btn-primary:hover` etc too and not sure if we should tweak daisy UI defaults, also the above snipped doesn't seem to generate exact color as dasiUI v3 `--sf` :( 

So not sure whether to keep it default as DasiyUI default or we tweak it ? cc @carletex @rin-st @sverps 

will keep digging if we could find a better way to tweak it

Pabl0cks:
Awesome research @technophile-04 !! 

Knowing this is an intended change from Daisy v4, I would probably just leave it like it is, and let users define the hover bg color on their apps (if they don't like the default hover effect).

Pabl0cks (approved):

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton/AddressInfoDropdown.tsx
```diff
@@ -45,7 +45,10 @@ export const AddressInfoDropdown = ({
   return (
     <>
       <details ref={dropdownRef} className="dropdown dropdown-end leading-3">
-        <summary tabIndex={0} className="btn btn-secondary btn-sm pl-0 pr-2 shadow-md dropdown-toggle gap-0 !h-auto">
+        <summary
+          tabIndex={0}
+          className="btn btn-secondary uppercase btn-sm pl-0 pr-2 shadow-md dropdown-toggle gap-0 !h-auto"
```
carletex, line 50: Now that I see the `uppercase` here a couple of things:

1. Should we remove the uppercase here? It was how it was displayed before... but it might make more sense to have it normal: 0x VS. 0X and also the checksum stuff (0x40…16a9 VS. 0X40…16A9). Even for ENS seems more natural. Just my opinion!
2. Since buttons are not uppercased by default, let's remove all the `normal-case` classes

carletex (commented):
Looking great @technophile-04.Thanks for this!

Left a comment with a couple of little things

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton/AddressInfoDropdown.tsx
technophile-04: > but it might make more sense to have it normal: 0x VS. 0X and also the checksum stuff (0x40…16a9 VS. 0X40…16A9). Even for ENS seems more natural. Just my opinion!

Yup makes sense, makes it visually consistent too with other wallets like MM !! updated at [81c22cc](https://github.com/scaffold-eth/scaffold-eth-2/pull/656/commits/81c22cc8d37c78d5d08e10c40bea73d540227962) 

>Since buttons are not uppercased by default, let's remove all the normal-case classes
Ohh yes,  tysm updated at [f5b2e57](https://github.com/scaffold-eth/scaffold-eth-2/pull/656/commits/f5b2e57a581e46634ccaa9bf6c4eee08765e798c)

PR #660, opened 2023-12-14
Update wagmi to latest version

carletex:
It looks lke the LedgerConnector (@ledgerhq/connect-kit-loader) has been compromised

Context:

https://twitter.com/wevm_dev/status/1735289737185837303
https://twitter.com/bantg/status/1735279127752540465

----

We don't use it on SE2 (we just use Rainbow's kit LedgerWallet) so we should be fine.

In any case, updating wagmi with the last version (where they remove the dependency): https://github.com/wevm/wagmi/commit/53ca1f7eb411d912e11fcce7e03bd61ed067959c

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -31,7 +31,7 @@
     "use-debounce": "^8.0.4",
     "usehooks-ts": "^2.7.2",
     "viem": "1.19.9",
-    "wagmi": "1.4.7",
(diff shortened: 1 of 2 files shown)


carletex:
We should create the NPX back-merge after merging this.

technophile-04 (approved):
Tysm !! tested it nicely and everything seems smooth, merging this 🙌

PR #653, opened 2023-12-09
Fix externalContracts object example code

Pabl0cks:
## Description

Fixing syntax error in the code example for `externalContracts` object.

## Additional Information

- [X] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [X] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

Closes #652

PR #661, opened 2023-12-14
Weekly CLI backmerge

technophile-04:
## Description

- Update wagmi to latest version: (#660)
- Fix externalContracts object example code (#653)
- Add question for installation method in template Issue (#651)

### DIFF
--- a/.changeset/polite-radios-happen.md
+++ b/.changeset/polite-radios-happen.md
@@ -1,0 +1,7 @@
+---
+"create-eth": patch
+---
+
+- Update wagmi to latest version: (#660)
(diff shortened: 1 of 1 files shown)


carletex:
Working great. Thanks @technophile-04 

---

BTW, now that I see the tweak on `ISSUE_TEMPLATE/bug_report.yml` I wonder if that should be excluded from the template.

Like we want to have that for people contributing on SE-2... but not sure if people doing NPX installs want to have that on their repo (with git clone it's inevitable, but for NPX might make sense to exclude it)

PR #667, opened 2023-12-19
use default values for safeConnector

technophile-04:
This allows any app to iframe in SE-2 and connect. 

For example : [impersonator.vision](https://impersonator.vision/)

Also tested it nicely even with safe iframe and works great everywhere. This PR just removes extra options that we had and keeps it default.

Here is the option's documentation: https://github.com/safe-global/safe-apps-sdk/tree/main/packages/safe-apps-sdk#documentation

test app link: https://ifram-any.vercel.app

Thanks to @austintgriffith for suggestion 🙌


austintgriffith:
love it!

carletex (approved):

PR #665, opened 2023-12-15
JsDoc cleanup

technophile-04:
## Description

4th TODO from https://github.com/scaffold-eth/scaffold-eth-2/issues/617#issue-2003940907, also fixes #617

PR #668, opened 2023-12-19
Typos

carletex:
Noticed a couple of typos that were repeating on #665  (after merging :sweat:): `corrsponding` and `scaffold.congfig.ts`, so decided to create this PR with other typos that have been bugging me for a while.

Also added some new lines on long comments (trying to stay close to the 120 line length). This is just a convention, and some people have soft-wraps enabled anyway... but I think it's a good convention.

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useAutoConnect.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useAutoConnect.ts
@@ -6,7 +6,7 @@ import scaffoldConfig from "~~/scaffold.config";
 import { burnerWalletId } from "~~/services/web3/wagmi-burner/BurnerConnector";
 import { getTargetNetworks } from "~~/utils/scaffold-eth";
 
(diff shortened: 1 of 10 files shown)


technophile-04 (approved):
Tysm !!! Looks great to me !

> Also added some new lines on long comments (trying to stay close to the 120 line length). This is just a convention, and some people have soft-wraps enabled anyway... but I think it's a good convention.

Yup 💯! Lol while writing those comments I was enabling soft wrap manually should have wrapped them directly in actual file instead. 

It would have been great if prettier had handled this properly since we have `printWidth` set but it has problem with wrapping comments https://github.com/prettier/prettier/issues/265 :( 

I am also keeping close eye on [Biome](https://biomejs.dev/) for future 😂 , maybe [this](https://biomejs.dev/reference/configuration/#javascriptformatterlinewidth) from biome can solve it

PR #647, opened 2023-12-06
feat: add external flag to external contracts

sverps:
Fixes #636 

During the contracts deepMerge we can add `{ external: true }` to contracts that are declared in the `externalContracts.ts` file.

This can be used to make UI modifications based on this flag (UI part is WIP)

PR #675, opened 2023-12-29
remove console log from contractsData

technophile-04:
## Description

It by mistake went into #647, also tried finding some other unwanted console.logs but others seems good to me


damianmarti (approved):
:-)

PR #637, opened 2023-12-01
Update hardhat package plugins and add hardhat-verify

metoinside:
>  EDIT : Please just to https://github.com/scaffold-eth/scaffold-eth-2/pull/637#issuecomment-1879179247 for actual changes

## Description
As recorded with https://github.com/scaffold-eth/scaffold-eth-2/issues/314, we're unable to verify in some networks.  
It was returning an error as like below _scrollSepolia_, _arbitrumSepolia_, _polygonZkEvmTestnet_
image

That's because of `hardhat-etherscan` [plugin](https://github.com/NomicFoundation/hardhat/tree/main/packages/hardhat-etherscan) which we're actively using and it is depreciated.
Official docs are suggesting to migrate into `hardhat-verify` and this PR is trying to do it.

As we migrated to the new library, we had to upgrade some dependencies and I tried to have a minimal impact.
For example, we can now deploy `sepolia` right away but we need to add custom configuration and the keys for the `scrollSepolia`.
It means after this PR is approved, we need to update the [documentation](https://docs.scaffoldeth.io/deploying/deploy-smart-contracts#4-verify-your-smart-contract) on how to use it.

## Additional Information

- [x] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [x] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

_Closes #314_

After this PR, you can now verify a smart contract with
`yarn verify --network [networkOfYourChoice] <address> [constructorParameters]`

We have verified these contracts
- [sepolia.scrollscan.com/address/0xfdd6...](https://sepolia.scrollscan.com/address/0xfdd6c1a247ab7b4ea06fb9dc3fc00ea724a6131c#code)
- [testnet-zkevm.polygonscan.com/address/0xdd85...](https://testnet-zkevm.polygonscan.com/address/0xdd85aca5dc954f0dffe5575ade47bc193fd18bfe#code)

This is going to be my first PR, therefore please feel free to open a discussion or give feedback to improve

Your ENS/address: 0x63191383EDbd8531C442868d87555f1EA2f6ad4E

PR #685, opened 2024-01-14
Weekly CLI backmerge

technophile-04:
## Description

- App router migration (#535)
- Update hardhat package plugins and add hardhat-verify (#637)

Tried to peacefully resolve all merge conflicts and also tested it nicely, works great for me 🙌 

but would love if others too could just give it nice test and skim through file changed once 🙌

NOTE: We shouldn't squash & merge it instead do a normal merge

PR #692, opened 2024-01-16
Ethers v6 migration in hardhat

technophile-04:
## Description

Recently we updated all the hardhat related packages in #637, but it seems that if you uncomment : 

https://github.com/scaffold-eth/scaffold-eth-2/blob/e4c70071ae804f88d4d73662d8e1b52ea7e833bd/packages/hardhat/deploy/00_deploy_your_contract.ts#L35

You get an error, which I think is caused due to `hardhat-deploy-ethers` and `hardhat-ethers`  digging few discussion it seems that updating ethers to v6 does seems to solve this issue due to some internal api changes

### DIFF
--- a/packages/hardhat/deploy/00_deploy_your_contract.ts
+++ b/packages/hardhat/deploy/00_deploy_your_contract.ts
@@ -1,5 +1,6 @@
 import { HardhatRuntimeEnvironment } from "hardhat/types";
 import { DeployFunction } from "hardhat-deploy/types";
+import { Contract } from "ethers";
 
 /**
  * Deploys a contract named "YourContract" using the deployer account and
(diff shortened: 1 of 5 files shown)


technophile-04:
Tysm all !! I think after #692 we could merge this 🙌

rin-st:
Works great!

> Tysm all !! I think after https://github.com/scaffold-eth/scaffold-eth-2/pull/692 we could merge this 🙌

recursion 😄

rin-st (approved):

carletex:
> recursion 😄

lol 

Great job @technophile-04. Add uncommented example code as we talked on TG.

Let's also add this to the weekly CLI merge.

technophile-04:
> recursion 😄 

😂 lol my bad wanted to add that comment at #685

PR #697, opened 2024-01-22
Base Sepolia - Foundry

Hotmanics:
## Description

Adds Base Sepolia as a supported network when configuring for Foundry projects.

Your ENS/address: jacobhomanics.eth


technophile-04 (approved):
Thanks @Hotmanics 🙌

Issue #698, opened 2024-01-22
bug: lack of "use client" breaks the AddressInput component


derrekcoleman:
### Is there an existing issue for this?

- [X] I have looked through the [existing issues](https://github.com/scaffold-eth/scaffold-eth-2/issues)

### Which method was used to setup Scaffold-ETH 2 ?

git clone

### Current Behavior

Clicking an address in the left sidebar on the Debug page throws this error.
<img width="974" alt="Screenshot 2024-01-22 at 3 22 57 PM" src="https://github.com/scaffold-eth/scaffold-eth-2/assets/80121818/cf525e0e-78dd-4caa-b2dc-82f200d67154">

### Expected Behavior

Expect clicking the address to take me to http://localhost:3000/blockexplorer/address/[0x...], the blockexplorer page for that address.

### Steps To Reproduce

Fresh git clone install, start chain and frontend, go to Debug tab, click the owner
<img width="1188" alt="Screenshot 2024-01-22 at 3 26 41 PM" src="https://github.com/scaffold-eth/scaffold-eth-2/assets/80121818/20c0ab6d-0a9a-40bd-b55a-8d07bbe1226c">
 address.

### Anything else?

Clearly a server component issue. Here's the terminal output. 
<img width="1285" alt="Screenshot 2024-01-22 at 3 28 09 PM" src="https://github.com/scaffold-eth/scaffold-eth-2/assets/80121818/0724db09-56f8-4aa6-9a20-516ffdc0bff0">

derrekcoleman:
Adding "use client" to the top of these files got the page to load. Other input components probably need the same.
AddressInput.tsx, EtherInput.tsx, IntegerInput.tsx

Since I'm new to the party, I'd love some guidance. **Would it be useful for me to make a PR that adds "use client" to these Input components?** If so, I'd follow the process outline in CONTRIBUTING.md.

Or maybe I'm missing something! I'm new to the app router, so maybe I've got something fundamentally off on my side that doesn't require a change to SE-2.

technophile-04:
Hey @derrekcoleman tysm for raising the issue ! Yeah actually this is caused because of #688 which we merged few hours ago. Adding "use client" on top of `AddressInput` gives the warning checkout mentioned in #686 

A simple solution is adding "use client" at the top of `packages/nextjs/components/scaffold-eth/Input/index.ts` 

@derrekcoleman could you please create a PR with the above solution, Thanks 🙌

[...]

PR #688, opened 2024-01-15
Remove 'use client' from EtherInput, IntegerInput, and AddressInput to Resolve Serialization Lint Warning

KcPele:
## Description
Removed "use client" from EtherInput IntergerInput and AddessInput due to lint-warning: Props must be serializable for components
There is no need for a child component to have "use client" when the parent contains "use client" already https://github.com/vercel/next.js/discussions/46795
_Concise description of proposed changes, We recommend using screenshots and videos for better description_

## Additional Information

- [ ] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [ ] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

_Closes #{issue number}_

_Note: If your changes are small and straightforward, you may skip the creation of an issue beforehand and remove this section. However, for medium-to-large changes, it is recommended to have an open issue for discussion and approval prior to submitting a pull request._

Your ENS/address:

Issue #686, opened 2024-01-15
lint-warning:  Props must be serializable for components in the "use client" entry file, "onChange" is invalid


technophile-04:
### Description : 

Noticed this warning when you go to `EtherInput` or `IntegerInput` / other files which has `use client` on top and uses `onChange` as prop example : 

![Screenshot 2024-01-15 at 11 58 25 AM](https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/556dba8f-ad4f-427e-aeed-f3d985078487)

Related discussion : 

https://github.com/vercel/next.js/discussions/46795

KcPele:
The solution is to remove "use client" from the child which is in EtherInpute, IntergerInput and AddressInput.tsx

KcPele:
i am currently removing it

technophile-04:
completed at #688

PR #699, opened 2024-01-23
fix broken blockexplorer address page

technophile-04:
### Context : 

This was introduced in #688 which we recently merged. We removed "use client" from components like `AddressInput`, `EtherInput` etc in #688 because there was lint warning in them as mentioned in #686. 

### Reason why #698 happened : 

Screenshot 2024-01-23 at 8 03 18 PM

If you look at the error it occurred at `blockexplorer/address/[address]/page.tsx` and the cause of the error is `AddressInput`, but we are not using `AddressInput` anywhere instead we are using `Address` component which comes from `~~/components/scaffold-eth/index.ts` barrel file and since this barrel file also exports inputs barrel we are indirectly importing `AddressInput` in address page (although we are not using it). 

### Solution : 

I would suggest going through this comment discussion https://github.com/vercel/next.js/discussions/46795#discussioncomment-5248407 once its really nice. 

For now, I added a quick fix which is completely fine. An alternate solution is we can add "use client" to `~~/components/scaffold-eth/index.ts` since all the SE-2 components are reactive and expected to be used on the client. 

But yeah I think a future-proof solution might be removing barrel files at least from the components  folder

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Input/index.ts
+++ b/packages/nextjs/components/scaffold-eth/Input/index.ts
@@ -1,3 +1,5 @@
+"use client";
+
 export * from "./AddressInput";
 export * from "./Bytes32Input";
 export * from "./BytesInput";


carletex:
Thanks for this Shiv! Loved the discussion over the Vercel repo. 

Let's merge this for now and we can think about a more «clean» solution at some point.

carletex (approved):

PR #721, opened 2024-02-14
add basic example to show connected address

technophile-04:
## Description

Context: If you directly use `useAccount` / any other hooks on home you get the below error from Next which is not at all helpful and hard to debug 
<details>
<summary>
Not so helpful error: 
</summary>

Screenshot 2024-02-14 at 12 55 00 PM

</details>

The solution to this is adding "use client" at the beginning of the file (this might be not all obvious for beginners) 

This PR adds a default example of showing the connected address (this way we have "use client" at the top of file already and might save some users from pitfalls) 

Please feel free to suggest any other place where we could add currently added it below SE-2 text : 

<details>
<summary>
Demo Video
</summary>

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/addd844a-378f-443c-be63-46801393b6dd

</details>


packages/nextjs/app/page.tsx
```diff
@@ -1,16 +1,26 @@
...
   return (
     <>
       <div className="flex items-center flex-col flex-grow pt-10">
         <div className="px-5">
-          <h1 className="text-center mb-8">
+          <h1 className="text-center">
             <span className="block text-2xl mb-2">Welcome to</span>
             <span className="block text-4xl font-bold">Scaffold-ETH 2</span>
           </h1>
+          <div className="flex justify-center items-center space-x-2">
```
technophile-04, line 20: NITPICK:  Should we add a line above this ? :

```
{/* Can be moved to a separate client component instead of using "use client" for whole page */}

```

carletex: I won't personally add it. Seems to specific to NextJS!

Pabl0cks (approved):
Nice addition! I personally suffered from that kind of not so helpful error, and I think it can save time to beginners, and prevent some ragequit.

Another place where could be added is below the Tinker/Explore boxes.

<details>
<summary>Something like this</summary>

image
</details>

rin-st (approved):
👍

damianmarti:
@technophile-04 I have code in se-2 projects that uses **useAccount** without adding ```"use client"``` and useAccount is working fine. What I'm missing?

carletex:
> @technophile-04 I have code in se-2 projects that uses useAccount without adding "use client" and useAccount is working fine. What I'm missing?

@damianmarti This only happens after migrating to the [Next App Router](https://nextjs.org/docs/app) (we did it a couple of weeks ago in https://github.com/scaffold-eth/scaffold-eth-2/pull/535)

packages/nextjs/app/page.tsx
```diff
@@ -1,16 +1,26 @@
+"use client";
```
damianmarti, line 1: Maybe adding a comment here to know that this line is to be able to use useAccount

rin-st: it needed not only for `useAccount`, but for any react hook

rin-st:
> @technophile-04 I have code in se-2 projects that uses useAccount without adding "use client" and useAccount is working fine. What I'm missing?

Or you have 'use client' in parent components/page

technophile-04:
Tysm all merging it 🙌

PR #738, opened 2024-02-23
Fix cursor stealing & display loading for AddressInput

### LINKED ISSUE #710: Slow typing on <AddressInput> causes cursor stealing
image

If you are writing your ENS on AddressInput (slowly after the dot), it "steals" the cursor... and if you try to keep typing, the cursor is not there :D

Notice this testing ABI Ninja with Austin. But it's also happening here.

Let me know if it's clear. If not I can record a video.

technophile-04:
#### How AddressInput internally works:

- We have debounce time of 500ms for the input field.
- Actually we have isEns util which when true (carletex.e is valid ens), the hook useEnsAddress makes a request for corresponding ENS address
- We are disabling the input filed when the request is in progress.
- Input focus looses when its is disabled (it does not regain focus when enabled again)

#### Cause of the issue:

1. Since we have 500ms debounce time, as soon as you write carletex.e (valid ens), the request is made and input is disabled.So if there is delay >= 500ms b/w "e" & "t" of "eth" then input looses focus(and our key strokes are not registered).

### Solution:

1. Increase the debounce time to 900ms (So we give user a bit more window to type in "eth").
2. Incase the user is slow typer, We refocus the input after trying to resolve the ens address.

#### Some minor tweak to imporve UX:

1. Show "Resolving..." indicator when request is in progress(To just tell user the keystrokes won't be registered and some background process is going on)
2. Re focusing the input if there was error, success etc, So that user can continue typing if there was typo or wants to update to new value.

### Demo: 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/4ad2f1cb-146c-4be5-b465-477d06eab169

Closes #710

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
+++ b/packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
@@ -11,7 +11,7 @@ import { CommonInputProps, InputBase, isENS } from "~~/components/scaffold-eth";
 export const AddressInput = ({ value, name, placeholder, onChange, disabled }: CommonInputProps<Address | string>) => {
   // Debounce the input to keep clean RPC calls when resolving ENS names
   // If the input is an address, we don't need to debounce it
-  const [_debouncedValue] = useDebounceValue(value, 900);
+  const [_debouncedValue] = useDebounceValue(value, 500);
   const debouncedValue = isAddress(value) ? value : _debouncedValue;
   const isDebouncedValueLive = debouncedValue === value;
 
--- a/packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
+++ b/packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
@@ -1,6 +1,6 @@
 import { useCallback, useEffect, useState } from "react";
 import { blo } from "blo";
-import { useDebounce } from "usehooks-ts";
+import { useDebounceValue } from "usehooks-ts";
 import { Address, isAddress } from "viem";
 import { useEnsAddress, useEnsAvatar, useEnsName } from "wagmi";
 import { CommonInputProps, InputBase, isENS } from "~~/components/scaffold-eth";
@@ -11,7 +11,7 @@ import { CommonInputProps, InputBase, isENS } from "~~/components/scaffold-eth";
 export const AddressInput = ({ value, name, placeholder, onChange, disabled }: CommonInputProps<Address | string>) => {
   // Debounce the input to keep clean RPC calls when resolving ENS names
   // If the input is an address, we don't need to debounce it
-  const _debouncedValue = useDebounce(value, 900);
+  const [_debouncedValue] = useDebounceValue(value, 900);
   const debouncedValue = isAddress(value) ? value : _debouncedValue;
   const isDebouncedValueLive = debouncedValue === value;
 
--- a/packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
+++ b/packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
@@ -11,22 +11,30 @@ import { CommonInputProps, InputBase, isENS } from "~~/components/scaffold-eth";
 export const AddressInput = ({ value, name, placeholder, onChange, disabled }: CommonInputProps<Address | string>) => {
   // Debounce the input to keep clean RPC calls when resolving ENS names
   // If the input is an address, we don't need to debounce it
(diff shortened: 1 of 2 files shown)


packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
```diff
@@ -11,22 +11,32 @@ import { CommonInputProps, InputBase, isENS } from "~~/components/scaffold-eth";
 export const AddressInput = ({ value, name, placeholder, onChange, disabled }: CommonInputProps<Address | string>) => {
   // Debounce the input to keep clean RPC calls when resolving ENS names
   // If the input is an address, we don't need to debounce it
-  const _debouncedValue = useDebounce(value, 500);
+  const [_debouncedValue] = useDebounceValue(value, 900);
```
technophile-04, line 14: Lol `900ms` came from just brute force trying and seems like a good spot

1. fast typer wont need to wait too much before it starts resolving adderss
2. Gives a a good decent amount of window for slow typers to full complete domain 

But lol it is kind of personal preference, so please feel free to suggest any better number 🙌

packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
```diff
@@ -78,6 +97,18 @@ export const AddressInput = ({ value, name, placeholder, onChange, disabled }: C
...
             <span className="text-accent px-2">{enteredEnsName ?? ensName}</span>
           </div>
+        ) : (
+          (isEnsAddressLoading || isEnsNameLoading) && (
+            <div className="flex bg-base-300 rounded-l-full items-center">
+              <div className="w-[35px]">
+                <div className="h-[35px] w-[35px] rounded-full bg-gray-500 animate-pulse"></div>
+              </div>
+              <div className="text-accent px-2 flex items-end space-x-1">
+                <span>Resolving</span>
```
carletex, line 107: This is a bit too strong I think, Maybe we could:

- Leave the loading dots only
OR
- Use daisy skeleton: https://daisyui.com/components/skeleton/#skeleton---circle-with-content

The 2 option with the circle + bar will like great I think

carletex:
Love this @technophile-04 Thank you!!

Left a couple a comments

Pabl0cks (approved):
GJ @technophile-04, feels like magic!! Works pretty well in mobile too.

I feel a bit old saying that, but 900 seems ok to me 😅 Approving but will test again if you tweak the visual feedback while resolving.

packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
rin-st: voting for 500 since I think there are much less slow typers than fast typers between our users

packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
rin-st: voting for loading dots only

technophile-04:
Thanks all for the test's and review !! 

> voting for 500 since I think there are much less slow typers than fast typers between our users
> The main problem was the input not focusing after resolving (for slow typers), and this is fixed in this PR.

Yup I agree reverted it back to 500 at https://github.com/scaffold-eth/scaffold-eth-2/pull/738/commits/15cfdbfe309aa4192e7d770c17133719bc794b79

> Found one bug and it worked for every address I entered first time 

Thanks ! Fixed at https://github.com/scaffold-eth/scaffold-eth-2/pull/738/commits/c2f1cf3640d6cf6599f96d70a7c42e0e68955916

> The 2nd option with the circle + bar will look great I think

Going with this for showing loading, reason : This option really nicely handles layout shifting as compared to other option. 


https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/27166499-6a6a-40b0-849a-658bc1db9886


Also the Ens Avatar loading until now it used to come, go, and then pop up again with the image

packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
technophile-04: fixed at https://github.com/scaffold-eth/scaffold-eth-2/pull/738/commits/6efd33101fbdc0b34e1921e3e48e7ea9f61a4a2e

Pabl0cks (approved):
Looking good to me! Awesome job @technophile-04 🙌

technophile-04:
Merging this 🙌 Tysm All !

PR #734, opened 2024-02-19
Fixes issue #733

tjayrush:
## Description

Closes the issue by make both displays of the address 0x1234...1234

## Additional Information

- [ x ] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [ x ] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

_Closes #733

_Note: If your changes are small and straightforward, you may skip the creation of an issue beforehand and remove this section. However, for medium-to-large changes, it is recommended to have an open issue for discussion and approval prior to submitting a pull request._

Your ENS/address: trueblocks.eth

PR #740, opened 2024-02-26
Cli backmerge

technophile-04:
## Description

1. basic example to show connected address (#721)
2. Standardize displaying of address and follow ERC-55 (#734)
3. fix contract balance hot reload balance issue (#739)
4. Fix cursor stealing & display loading for AddressInput (#738)
5. Fix blockexplorer code tab (#741)
6. Match link name with actual tab name for Debug Contracts (#743)


technophile-04:
Tested it nicely and everything seems to work 🙌, but would love if someone could go through it once too 🙌

carletex (approved):
Tested and works great!

Not related to the back merge, but something to thing about:
image

Maybe when the wallet is not connected we should show a different thing.

PR #745, opened 2024-02-28
yarn flatten

carletex:
`yarn flatten`: Allow to flatten contracts into a single file.

I think we could do the same in Foundry: https://book.getfoundry.sh/reference/forge/forge-flatten

Open to suggestions!

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "hardhat-verify": "yarn workspace @se-2/hardhat hardhat-verify",
     "compile": "yarn workspace @se-2/hardhat compile",
     "generate": "yarn workspace @se-2/hardhat generate",
+    "flatten": "yarn workspace @se-2/hardhat flatten",
     "hardhat:lint": "yarn workspace @se-2/hardhat lint",
     "hardhat:lint-staged": "yarn workspace @se-2/hardhat lint-staged",
     "hardhat:test": "yarn workspace @se-2/hardhat test",
--- a/packages/hardhat/package.json
+++ b/packages/hardhat/package.json
@@ -8,7 +8,7 @@
     "deploy": "hardhat deploy",
     "fork": "MAINNET_FORKING_ENABLED=true hardhat node --network hardhat --no-deploy",
     "generate": "hardhat run scripts/generateAccount.ts",
-    "flatten": "hardhat flatten > FlattenedContracts.sol && echo '📄 Flattened contracts exported to packages/hardhat/FlattenedContracts.sol'",
+    "flatten": "hardhat flatten",
     "lint": "eslint --config ./.eslintrc.json --ignore-path ./.eslintignore ./*.ts ./deploy/**/*.ts ./scripts/**/*.ts ./test/**/*.ts",
     "lint-staged": "eslint --config ./.eslintrc.json --ignore-path ./.eslintignore",
     "test": "REPORT_GAS=true hardhat test --network hardhat",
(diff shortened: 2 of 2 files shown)


technophile-04:
Thanks Carlos 🙌

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/33b65c64-9c57-43fe-b8ed-9cf7f0e85a8d



For some reason when I run just `yarn flatten` , the `FlattendContracts.sol` only contains output till max line 258 rest output is eaten up. 

When I `cd package/hardhat` and manually run `yarn hardhat flatten ...` it works just fine, not sure if  its just with my system. 

If the contract is small (eg. I removed import `console.log` from `YourContract.sol`) It works nicely not sure if its due some limit on no. of line redirect (">") into file when from package.json script. 

cc @Pabl0cks could you please give it a try, it might be something to do with my machine

carletex:
Oh weird! @technophile-04 

What happens if you remove the redirect in the hardhat package.json? Do you get the full output in the console?

carletex:
no clue what might be:man_shrugging: 

> yeah I do get full output in console if I remove redirect from package.json script

With this removed in the package.json, does this work? `yarn flatten > test.sol`

technophile-04:
> yarn flatten > test.sol

Yup it creates `test.sol` at root of monorepo but works nicely ! 

Also if I put this in root package.json  `"flatten": "yarn workspace @se-2/hardhat flatten > test.sol",` it again breaks, so seems like an issue with redirect. 

Does it properly work for you? Lets maybe work for @Pabl0cks to test, I will also try to debug till then 🙌

carletex:
> it again breaks, so seems like an issue with redirect.

I guess it's an issue with redirect inside the package.json, since the redirection works fine in your system (`yarn flatten > test.sol`)

> Does it properly work for you? Lets maybe work for @Pabl0cks to test, I will also try to debug till then 🙌

Yep, it works for me as it's.

AN alternative could be to skip the redirection and just print out on stdout... and people can redirect themselves.

Pabl0cks:
> Does it properly work for you? Lets maybe work for @Pabl0cks to test, I will also try to debug till then 🙌

GJ Carlos and Shiv!! It's working fine to me, I get the same output with `yarn flatten` and with `yarn hardhat flatten > flattened.sol´ => 1648 lines and seems complete

Pabl0cks (approved):
Is working great to me, flattening all direct imports and imports into those imports

technophile-04 (approved):
> I guess it's an issue with redirect inside the package.json

💯💯

Hmm after some digging turns out something to do with yarn version (might be how I have globally configured `yarn`) 

When I do `yarn set version 1.22.19` It works nicely for me but for some reason it sets that version at project level :( 

<details>
<summary>
Demo Video 
</summary>


https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/7cd366f8-c624-4c18-9672-4cda46b5d12c

</details>


> AN alternative could be to skip the redirection and just print out on stdout

+1 for this, I think people can redirect themselves.... just to be safe, eg Autin intuitively doing [here](https://www.youtube.com/live/lmdj_QeqcdY?si=xGd1LItKh1rwegbZ&t=788)  

But please free to merge as it ! Tysm Carlos, and Pablo for test 🙌

Will create a PR adding this into foundry 🙌

carletex:
> eg Autin intuitively doing [here](https://www.youtube.com/live/lmdj_QeqcdY?si=xGd1LItKh1rwegbZ&t=788)

Autocompletion for SE-1 haha where it was printed in stdout

Decided to remove the redirect in the script, in case someone else faces the same issue as Shiv

technophile-04:
Tysm merge this 🙌

PR #746, opened 2024-02-29
add foundry flatten script

technophile-04:
## Description

There is catch, you need to pass in path for contract file eg: 

```bash
yarn flatten contracts/YourContract.sol
```

### DIFF
--- a/.changeset/wet-carrots-laugh.md
+++ b/.changeset/wet-carrots-laugh.md
@@ -1,0 +1,5 @@
+---
+"create-eth": patch
+---
+
+forge flatten for foundy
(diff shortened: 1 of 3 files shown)


rin-st:
Works great, thanks!

Not related to this pr and I think you already know it, but noticed that since base `README.md` is copied for foundry project, there is only hardhat information, not foundry

rin-st (approved):

technophile-04:
> but noticed that since base README.md is copied for foundry project, there is only hardhat information, not foundry

Ohh yeah ! Created #755 🙌

Merging this for now, Thanks !

Issue #753, opened 2024-03-06
useScaffoldContractWrite does not throw errors


corwinthill:
### Is there an existing issue for this?

- [X] I have looked through the [existing issues](https://github.com/scaffold-eth/scaffold-eth-2/issues)

### Which method was used to setup Scaffold-ETH 2 ?

git clone, npx create-eth@latest

### Current Behavior

useScaffoldContractWrite does not throw errors, but instead eats up the errors and displays it using the notification util. When calling a contract function using useScaffoldContractWrite within a try/catch block, regardless of whether the function call succeeds or fails, no exception is caught and the try block is fully executed. The current behavior thus allows for an error notification and a success notification to simultaneously appear.

### Expected Behavior

From my understanding, the intended functionality should be that useScaffoldContractWrite should throw an error so that functions utilizing it can be used within try/catch blocks. useScaffoldContractWrite relies on useTransactor, so an error caught by useTransactor should be thrown to useScaffoldContractWrite, and the error should then be thrown forward again to the contract function that is utilizing useScaffoldContractWrite. There should also only be one instance of the error notification being displayed.

### Steps To Reproduce

The issue comes from a setup similar to the code below.
```tsx
const { writeAsync: exampleFunction } = useScaffoldContractWrite({
  contractName: "exampleContract",
  functionName: "exampleFunction",
});

const handleExampleFunction = async () => {
  const loadingNotification = notification.loading("Loading ...");
  try {
    await exampleFunction();
    notification.remove(loadingNotification);
    notification.success("Success!");
  } catch (error) {
    notification.remove(loadingNotification);
    console.error(error);
  }
};
```
You'll want something on the frontend to trigger the function call (like a button). Then make sure that the function call will revert so that an error is thrown. A simple one to setup is not having any funds for gas. Once the function is called, then both the Success notification, and an error notification will be displayed.

### Anything else?

I'm not sure if its the best solution, but I have resolved this issue with the help of Shiv Bhonde. A few lines of code need to be changed on useScaffoldContractWrite.ts as well as useTransactor.tsx.
First, I added the following code after the line referenced in the link to useScaffoldContractWrite.ts
```ts
// Insert after line 87
throw new Error(e);
```
https://github.com/scaffold-eth/scaffold-eth-2/blob/cdd321e6b87477a557e3b295bad7592c234acdef/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts#L87

[...]

PR #758, opened 2024-03-07
Fix useScaffoldContractWrite so it properly throws errors

corwinthill:
## Description

- These changes reflect what is summarized at the bottom of issue #753
- Added line 88 to useScaffoldContractWrite.ts and changed line 98 in useTransactor.tsx
- These changes throw errors forward to a function utilizing useScaffoldContractWrite, and removes one notification error to prevent duplicate error notifications displaying.

## Additional Information

- I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

_Closes #753_

Your ENS/address:  zurriken.eth

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
@@ -3,7 +3,7 @@ import { useTargetNetwork } from "./useTargetNetwork";
 import { Abi, ExtractAbiFunctionNames } from "abitype";
 import { useContractWrite, useNetwork } from "wagmi";
 import { useDeployedContractInfo, useTransactor } from "~~/hooks/scaffold-eth";
-import { getParsedError, notification } from "~~/utils/scaffold-eth";
+import { notification } from "~~/utils/scaffold-eth";
 import { ContractAbi, ContractName, UseScaffoldWriteConfig } from "~~/utils/scaffold-eth/contract";
 
 type UpdatedArgs = Parameters<ReturnType<typeof useContractWrite<Abi, string, undefined>>["writeAsync"]>[0];
@@ -83,9 +83,7 @@ export const useScaffoldContractWrite = <
 
         return writeTxResult;
       } catch (e: any) {
-        const message = getParsedError(e);
-        notification.error(message);
-        throw new Error(e);
+        throw e;
       } finally {
         setIsMining(false);
       }
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldContractWrite.ts
@@ -85,6 +85,7 @@ export const useScaffoldContractWrite = <
       } catch (e: any) {
         const message = getParsedError(e);
         notification.error(message);
+        throw new Error(e);
       } finally {
         setIsMining(false);
       }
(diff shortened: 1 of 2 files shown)


technophile-04 (approved):
Hey thanks @corwinthill 🙌, just pushed few tweaks : 

### Updated  changes: 

1. Just update `useTransactor` to show notification error + throw whole error forward

2. `useScaffoldWriteContract`'s `writeContractAsync` will just throw error forward without showing any notifications.

### Reason for above changes:

#### useTransactor:

The main aim of this hook is to execute given txn + **show opinionated/general UI feedback**, so it makes sense here to show error notification. Hence https://github.com/scaffold-eth/scaffold-eth-2/pull/758/commits/b036b23513c5dedb23975fdb61cf50bcf1562408

#### useScaffoldWriteContract:

Since we are internally using `useTransactor` it should the show UI feedback needed but we don't need it to explicitly show error notification again and eat up the error.Similar to how wagmi's `writeAsync` works it should just throw error forward. Hence https://github.com/scaffold-eth/scaffold-eth-2/pull/758/commits/2aace01f702e5d53a372dcb02c204d0a63870920

cc @carletex @rin-st for thoughts / any other suggestions 

---

Since we are using `useTransactor` internally for `useScaffoldWriteContract` it makes UI feedback for `useScaffoldWriteContract` opinionated. Which kind of makes sense for beginners / intermediate users.

If advanced users want to show custom UI feedback they maybe shouldn't use `useScaffoldWriteContract` and use wagmi's `useWriteContract` and handle themselves.

But maybe in future we could make `useScaffoldWriteContract` less opinionated and let user handle the UI feedback themselves.

damianmarti:
@corwinthill @technophile-04 This looks great!! 

One related comment. Sometimes I ended up adding ifs to getParsedError, because the error thrown is not good enough, and you can get a better message checking something inside the error. Maybe adding a way to use a custom getParsedError is a nice way to improve that. What do you think? Did you have this issue before? We can move to a discussion if you think this is something we should implement...

rin-st (approved):
Lgtm!

rin-st:
> One related comment. Sometimes I ended up adding ifs to getParsedError, because the error thrown is not good enough, and you can get a better message checking something inside the error. Maybe adding a way to use a custom getParsedError is a nice way to improve that. What do you think? Did you have this issue before? We can move to a discussion if you think this is something we should implement...

Good point. I think we need to create a new issue for it, and your updates of `getParsedError` can be helpful. If you still have these projects somewhere of course :)

technophile-04:
>  I think we need to create a new issue for it, and your updates of getParsedError can be helpful. If you still have these projects somewhere of course :)

++, Tysm Damu ! and also Rinat for reveiw! 🙌

Merging this

damianmarti:
> > One related comment. Sometimes I ended up adding ifs to getParsedError, because the error thrown is not good enough, and you can get a better message checking something inside the error. Maybe adding a way to use a custom getParsedError is a nice way to improve that. What do you think? Did you have this issue before? We can move to a discussion if you think this is something we should implement...
> 
> Good point. I think we need to create a new issue for it, and your updates of `getParsedError` can be helpful. If you still have these projects somewhere of course :)

These are particular for each project/contract, you can take a look at https://github.com/damianmarti/fruit-market2/blob/main/packages/nextjs/components/scaffold-eth/Contract/utilsContract.tsx#L27

PR #760, opened 2024-03-07
Add TokenAmount component

luloxi:
## Description

Added TokenAmount component to nextjs package folder. It allows to display a token amount with a custom amount of decimals. If it's marked as ETH, it allows for showing its USD value with a button push, similar to Balance component.

## Additional Information

Your ENS/address: lulox.eth


rin-st:
Hey @luloxi ! Thanks for you pr!

As I understand, we can update Balance component with props like `ethPrecision` and `usdPrecision` (or `ethDecimals/usdDecimals`), and then use `Balance` component for Eth?

rin-st:
I think you can take this part of component

packages/nextjs/components/scaffold-eth/TokenAmount.tsx
```diff
@@ -0,0 +1,64 @@
...
+                      <span>{parseFloat(formatEther(BigInt(amount * price))).toFixed(2)}</span>
+                    </>
+                  )}
+                </div>
+              </button>
+            </div>
+          )}
+        </>
+      ) : (
+        <div className="w-full flex items-center justify-center">
```
rin-st, line 58: so if everything before this can be implemented with `Balance` component, in `TokenAmount` component will be just this `div`. I think you can add currency symbol here with options to add it to left or to right

rin-st:
After https://github.com/scaffold-eth/scaffold-eth-2/pull/760#discussion_r1518638192 this part of `Balance` component
```
{displayUsdMode ? (
          <>
            <span className="text-[0.8em] font-bold mr-1">$</span>
            <span>{(balance * price).toFixed(2)}</span>
          </>
        ) : (
          <>
            <span>{balance?.toFixed(4)}</span>
            <span className="text-[0.8em] font-bold ml-1">{targetNetwork.nativeCurrency.symbol}</span>
          </>
        )}
```
can be implemented like
```
{displayUsdMode ? (
          <TokenAmount
             amount={balance * price}
             precision={2}
             currencyPosition='left'
             currency={$}
          /> ) : (
          <TokenAmount
             amount={balance}
             precision={4}
             currencyPosition='left'
             currency={targetNetwork.nativeCurrency.symbol}
          />)
}
          
```

luloxi:
> Hey @luloxi ! Thanks for you pr!
> 
> As I understand, we can update Balance component with props like `ethPrecision` and `usdPrecision` (or `ethDecimals/usdDecimals`), and then use `Balance` component for Eth?

Yes, but Balance only reads the Balance of a contract, and it's not useful for simply displaying an amount of a token when reading from a contract read function.

packages/nextjs/components/scaffold-eth/TokenAmount.tsx
luloxi: Yes, it should include an option to specify a token symbol, and I've also noticed that I should separate precision from the decimals, which would affect how a token amount is read, so maybe I should use formatUnits instead maybe

luloxi:
> After [#760 (comment)](https://github.com/scaffold-eth/scaffold-eth-2/pull/760#discussion_r1518638192) this part of `Balance` component
> 
> ```
> {displayUsdMode ? (
>           <>
>             <span className="text-[0.8em] font-bold mr-1">$</span>
>             <span>{(balance * price).toFixed(2)}</span>
>           </>
>         ) : (
>           <>
>             <span>{balance?.toFixed(4)}</span>
>             <span className="text-[0.8em] font-bold ml-1">{targetNetwork.nativeCurrency.symbol}</span>
>           </>
>         )}
> ```
> 
> can be implemented like
> 
> ```
> {displayUsdMode ? (
>           <TokenAmount
>              amount={balance * price}
>              precision={2}
>              currencyPosition='left'
>              currency={$}
>           /> ) : (
>           <TokenAmount
>              amount={balance}
>              precision={4}
>              currencyPosition='right'
>              currency={targetNetwork.nativeCurrency.symbol}
>           />)
> }
>           
> ```
> 
> Please lmk if I understood something wrong. Thanks!

Naming "decimals" as "precision" is way more appropiate, so I'm all for it. And allowing to specify where the currency symbol would be displayed is also nice. It'd be nice to include also another parameter that allows to specify a string that would be used as the symbol too maybe.

packages/nextjs/components/scaffold-eth/TokenAmount.tsx
luloxi: What do you mean that it will be just the div you quoted?

luloxi:
> After [#760 (comment)](https://github.com/scaffold-eth/scaffold-eth-2/pull/760#discussion_r1518638192) this part of `Balance` component
> 
> ```
> {displayUsdMode ? (
>           <>
>             <span className="text-[0.8em] font-bold mr-1">$</span>
>             <span>{(balance * price).toFixed(2)}</span>
>           </>
>         ) : (
>           <>
>             <span>{balance?.toFixed(4)}</span>
>             <span className="text-[0.8em] font-bold ml-1">{targetNetwork.nativeCurrency.symbol}</span>
>           </>
>         )}
> ```
> 
> can be implemented like
> 
> ```
> {displayUsdMode ? (
>           <TokenAmount
>              amount={balance * price}
>              precision={2}
>              currencyPosition='left'
>              currency={$}
>           /> ) : (
>           <TokenAmount
>              amount={balance}
>              precision={4}
>              currencyPosition='right'
>              currency={targetNetwork.nativeCurrency.symbol}
>           />)
> }
>           
> ```
> 
> Please lmk if I understood something wrong. Thanks!

The intention of the component is to be able to be just that code you quoted. But it would also allow a custom source other than the balance of an address

rin-st:
> Yes, but Balance only reads the Balance of a contract, and it's not useful for simply displaying an amount of a token when reading from a contract read function.

Not only contracts, it can be regular addresses too.

packages/nextjs/components/scaffold-eth/TokenAmount.tsx
rin-st: > What do you mean that it will be just the div you quoted?

I was wrong a little bit, thought first part could be implemented just with `Balance` component

packages/nextjs/components/scaffold-eth/TokenAmount.tsx
```diff
@@ -0,0 +1,64 @@
...
+ */
+export const TokenAmount = ({ amount, decimals = 4, isEth = false }: TokenAmountProps) => {
+  const { targetNetwork } = useTargetNetwork();
+  const { price, isError, isLoading, onToggleBalance, isEthBalance } = useAccountBalance();
+
+  return (
+    <>
+      {isEth ? (
+        <>
+          {isLoading ? (
```
rin-st, line 23: `isLoading` and `isError` are about `balance` part of `useAccountBalance`, which you don't use. So as I understand you don't need these parts in the component

rin-st:
> The intention of the component is to be able to be just that code you quoted. But it would also allow a custom source other than the balance of an address

Oh, I understood, thank you

I'm proposing to start just with this props without toggling balance
```
<TokenAmount
             amount={balance * price}
             precision={2}
             currencyPosition='left'
             currency='$'
          />
```
Probably, with possibility of this
> It'd be nice to include also another parameter that allows to specify a string that would be used as the symbol too maybe.

Second iteration (next pull request).
Ability to toggle balance of every token (not only Eth) to $. New props - `price`. Additional logic `toggleBalance` and `isTokenBalance` inside `TokenAmount` component. 

cc @technophile-04 @carletex what do you think?

packages/nextjs/components/scaffold-eth/TokenAmount.tsx
```diff
@@ -0,0 +1,64 @@
+import { formatEther } from "viem";
+import { useAccountBalance } from "~~/hooks/scaffold-eth";
+import { useTargetNetwork } from "~~/hooks/scaffold-eth/useTargetNetwork";
+
+type TokenAmountProps = {
+  amount: number;
```
rin-st, line 6: also, when I pass `Number.MAX_SAFE_INTEGER` it's only 0,009 Eth. So you can't pass number greater. I think it's better to use `BigInt`

packages/nextjs/components/scaffold-eth/TokenAmount.tsx
technophile-04: ++, I think we could remove `useAccountBalance` completely from here, we could get `price` from here :  `const price = useGlobalState(state => state.nativeCurrencyPrice);` 

The use `useAccountBalance` initially confused me alot while reviewing the code

packages/nextjs/components/scaffold-eth/TokenAmount.tsx
technophile-04: Yup also another reason to use `BigInt` is while reading from contract you will mostly be getting as bigint and we don't want people to convert it `number` before passing it to component

technophile-04:
100% agree with https://github.com/scaffold-eth/scaffold-eth-2/pull/760#issuecomment-1987374250, with just a small change of renaming `currencyPosition` => `tokenSymbolPosition` & `currency` => `tokenSymbol` for first iteration. And also for second iteration we shouldn't use "balance" in names. But completely agree with the suggestions.

lol also I am kind of not completely sold on this component 😅, but it might be good for beginners.

luloxi:
I read your comments and simplified the TokenAmount component, integrating it into Balance component without visual or logic changes to it. Now the TokenAmount component doesn't have the switch between usd and eth display.

luloxi:
But it's broke outside it's Balance way of working, it's displaying a lot of zeros
image

rin-st:
> I read your comments and simplified the TokenAmount component, integrating it into Balance component without visual or logic changes to it. Now the TokenAmount component doesn't have the switch between usd and eth display.

Thanks @luloxi , great job!
I added some changes - simplified a bit `TokenAmount` component and removed `useAccountBalance` from `Balance` component since I think we'll remove that hook.

Regarding bigint for `amount` prop - it's trickier than I thought. 
- let's say if we have bigint and need to multiply it to float price, there is no simple universal way to do it. Also it can cause issues with precision
- if `amount` is bigint, we need to pass bigint even when we pass USD with symbol "$" 

So for now yes, let's pass a number and just show it with needed precision.

> But it's broke outside it's Balance way of working, it's displaying a lot of zeros

not sure hot is it even work. What have you passed as props?

rin-st:
> 100% agree with https://github.com/scaffold-eth/scaffold-eth-2/pull/760#issuecomment-1987374250, with just a small change of renaming currencyPosition => tokenSymbolPosition & currency => tokenSymbol for first iteration

I think `currency` is ok, since it's just a string and allows to pass let's say "USD" or "United States Dollars" if needed

packages/nextjs/components/scaffold-eth/TokenAmount.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const currencySymbol = <span className="text-[0.8em] font-bold">{currency}</span>;
+
+  return (
+    <div className="flex items-center justify-center gap-1">
+      {currencyPosition === "left" && currencySymbol}
+      {displayAmount}
+      {currencyPosition === "right" && currencySymbol}
+    </div>
+  );
+};
```
technophile-04, line 22: Umm not sure if this would be a great add or not :( 

I think the final version of this component will be something like clicking the component it will toggle and display `displayAmount * price` (`price` will be coming from prop) 

I think the toggling behavior can be easily achieved by https://daisyui.com/components/swap/

Also accepting `amount` as number not sure if its best way because usually while reading token amount you would get bigInt maybe people love to use `parseEther` / `parseUnits` to show it properly ?

Maybe I think it's better we let developer control all the styling too since its minimal, we can create a recipe in docs.scaffoldeth.io to help beginners if we see a lot of people struggling on this 🙌

technophile-04 (commented):
Just added my views https://github.com/scaffold-eth/scaffold-eth-2/pull/760#discussion_r1538101515

But would love to hear other views @rin-st , @carletex , @damianmarti 🙌, completely fine with merging this or maybe we could create an issue / discussion for exploring some new helpfulll components and then create PR's with those brainstormed components 🙌

Also thanks so much @luloxi for taking and tackling it, really appreciate it !

luloxi:
> > I read your comments and simplified the TokenAmount component, integrating it into Balance component without visual or logic changes to it. Now the TokenAmount component doesn't have the switch between usd and eth display.
> 
> Thanks @luloxi , great job! I added some changes - simplified a bit `TokenAmount` component and removed `useAccountBalance` from `Balance` component since I think we'll remove that hook.
> 
> Regarding bigint for `amount` prop - it's trickier than I thought.
> 
> * let's say if we have bigint and need to multiply it to float price, there is no simple universal way to do it. Also it can cause issues with precision
> * if `amount` is bigint, we need to pass bigint even when we pass USD with symbol "$"
> 
> So for now yes, let's pass a number and just show it with needed precision.
> 
> > But it's broke outside it's Balance way of working, it's displaying a lot of zeros
> 
> not sure how is it even work. What have you passed as props?

I have passed a BigInt that is being read from an event [here](https://github.com/luloxi/Easy2Pay/blob/main/packages/nextjs/pages/paymentRequests.tsx) 

It's previously converted like this to be used as input for TokenAmount: 
<TokenAmount amount={request.amount ? parseInt(request.amount.toString()) : 0} />

rin-st:
> I think the final version of this component will be something like clicking the component it will toggle and display displayAmount * price (price will be coming from prop)
...
Also accepting amount as number not sure if its best way because usually while reading token amount you would get bigInt maybe people love to use parseEther / parseUnits to show it properly ?

I noticed that its not so easy to multiply bigint and number, so I thought probably number is ok if TokenAmount is a dumb view component which knows nothing about value passed. But yes, we need to do multiplying anyway so better to do it inside component.

note about multiplying: I mean something like [some bigint] * 1.23456. How to multiply it? One way is [some bigint] * 123456n / 10_000n. It works for big numbers, but if bigint is small results will not be precise. Not sure we care about small bigint numbers though like 1 or 100 wei. So it seems we can just omit that problem and use bigint 🤷‍♂️ 

> Maybe I think it's better we let developer control all the styling too since its minimal, we can create a recipe in docs.scaffoldeth.io to help beginners if we see a lot of people struggling on this 🙌

> I think that the best option is to open a discussion about this component to decide if it's a useful component to include in the codebase and with what features.

Yes, let's discuss it!

> I have passed a BigInt that is being read from an event [here](https://github.com/luloxi/Easy2Pay/blob/main/packages/nextjs/pages/paymentRequests.tsx)
It's previously converted like this to be used as input for TokenAmount:
<TokenAmount amount={request.amount ? parseInt(request.amount.toString()) : 0} />

So you passed bigint instead of number, and you see zeros because of that. I think Typescript should prevent you to do that

technophile-04:
Closing this for now, context https://github.com/scaffold-eth/scaffold-eth-2/issues/786#issuecomment-2024963470 , https://github.com/scaffold-eth/scaffold-eth-2/issues/786#issuecomment-2025167483

But thanks @luloxi for taking time and contributing 🙌, feel bad for closing this :( but also maybe I think its really better if we first create and discussion / issue so that we don't encounter situations like this

PR #768, opened 2024-03-11
add .vercelignore

### LINKED ISSUE #767: Ignore hardhat / foundry deployements on vercel
### Description : 

Screenshot 2024-03-11 at 10 00 59 AM

Files changed : 

Screenshot 2024-03-11 at 10 06 23 AM

We could use https://vercel.com/docs/deployments/vercel-ignore

This was caused due to #757

technophile-04:
## Description

It's sad that Vercel don't respect `.gitgnore` inside it each package of monorepo, it was even uploading `.env` of hardhat :( : 

Screenshot 2024-03-11 at 10 47 00 AM

--- 

Also vercel doesn't respect `.vercelignore` inside each package, so had to copy all the things from each package `.gitignore` and append `packages/{packageName}` in `.vercelignore` at the root

Not sure what would be the better approach 

Here are docs : 
https://vercel.com/docs/deployments/vercel-ignore

Fixes #767

### DIFF
--- a/.vercelignore
+++ b/.vercelignore
@@ -1,0 +1,58 @@
+#Hardhat files
+
+packages/hardhat/node_modules
+packages/hardhat/.env
+packages/hardhat/coverage
+packages/hardhat/coverage.json
+packages/hardhat/typechain
+packages/hardhat/typechain-types
+packages/hardhat/temp
+
+packages/hardhat/cache
+packages/hardhat/artifacts
+
+packages/hardhat/artifacts-zk
+packages/hardhat/cache-zk
+
+packages/hardhat/deployments
+
+# -------------------------
+
+# Next.js files
+
+# dependencies
+packages/nextjs/node_modules
+packages/nextjs/.pnp
+packages/nextjs/.pnp.js
+
+# testing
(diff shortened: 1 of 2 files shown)


damianmarti (approved):
LGTM! Thanks!!

rin-st:
> Also vercel doesn't respect .vercelignore inside each package, so had to copy all the things from each package .gitignore and append packages/{packageName} in .vercelignore at the root

Regarding the [docs](https://vercel.com/docs/deployments/vercel-ignore#monorepos) it .vercelignore file should work from the root of every monorepo project. I tried it but it doesn't work for me too. Looks like a bug from their side

> Not sure what would be the better approach

For now I don't see other solutions than yours. Thank you!

rin-st (approved):

technophile-04:
> Regarding the [docs](https://vercel.com/docs/deployments/vercel-ignore#monorepos) .vercelignore file should work from the root of every monorepo project. I tried it but it doesn't work for me too.

Yeah :( 

Merging this for now but we should update it once we find a better a solution, Tysm guys for testing 🙌

PR #757, opened 2024-03-06
Fix vercel deployment mismatch

technophile-04:
## Description

My 99th try of trying to fix this. 

context #419 and all referenced PR inside it. 

Obviously there is a catch, had to add `next` as devDependency of monorepo because of https://github.com/orgs/vercel/discussions/6047 but now atleast it uploads whole monorepo (mainly `yarn.lock`) which fixes all the issues

Work flow doesn't change at all you just need to choose all defualt options and add vercel project : 

https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/32a76d20-c6e2-49fc-9072-bdac9f7a947b

PR #780, opened 2024-03-17
fix: vercel deployment linking from github

technophile-04:
## Description

After we merged #757 , it seems to broke the vercel deployment when you connect it through github example : 
github.com/technophile-04/test-vercel-deployments

Screenshot 2024-03-17 at 1 31 32 PM

This is because we are mentioning `outputDir` in `vercel.json` which is different when you connect through github


rin-st:
How to test it properly?
- Cloned main repo (add third party repo option), added it from vercel dashboard - deployment works already from main branch
- `npx create-eth@latest` hardhat - same error on main branch, and same error when I add your changes.

technophile-04:
> added it from vercel dashboard - deployment works already from main branch

This should actually fail, If you try doing : 

Screenshot 2024-03-18 at 12 51 56 PM


And then connecting the repo to vercel it gives the above error. For example I tried doing same [here](https://github.com/technophile-04/test-vercel-deployments)

> npx create-eth@latest hardhat - same error on main branch, and same error when I add your changes.

I just did `npx create-eth@latest hardhat` and added this changes to it and it worked for [example here](https://github.com/technophile-04/npx-vercel-deployment)

--- 

### Testing these changes : 

Maybe you could "Create a new template" and then push this PR changes to your repo of `main` branch and connect to github

rin-st:
Thank you! Tried both cases and everything works like you said.

> Cloned main repo (add third party repo option), added it from vercel dashboard - deployment works already from main branch

But after adding se-2 as a third party repo from vercel, deployment still works for main. I think it's not a problem though

Gj!

rin-st (approved):

damianmarti:
@technophile-04 I was able to deploy to Vercel with no issues!

Pabl0cks:
Tested after merge and is working great for me, gj @technophile-04! 🙌

PR #782, opened 2024-03-21
Create specific README files for foundry and hardhat options

### LINKED ISSUE #755: Update CLI scaffolded README
### Description : 

Currently, after scaffolding project through CLI the monorepo  README always shows hardhat instruction due to : 

https://github.com/scaffold-eth/scaffold-eth-2/blob/cli/templates/base/README.md

We can tackle this with two approaches : 

Simple : 
- Update [`templates/base/README.md`](https://github.com/scaffold-eth/scaffold-eth-2/blob/cli/templates/base/README.md) such that it contains both hardhat and foundry instruction checkout https://docs.scaffoldeth.io/quick-start/environment

Better approach: 
- We can create separate README's for both one in [`templates/extensions/hardhat`](https://github.com/scaffold-eth/scaffold-eth-2/tree/cli/templates/extensions/hardhat) for hardhat and second inside [`templates/extensions/foundry`](https://github.com/scaffold-eth/scaffold-eth-2/tree/cli/templates/extensions/foundry) with their respective instruction. 

Second approach make more sense to me since when

Pabl0cks:
## Description

Delete base template README and create 2 different README files, with the specific commands and file routes for Hardhat or Foundry. Add each file to their corresponding template folder.

## Additional Information

## Related Issues

Closes #755 

Your ENS/address: pabl0cks.eth


templates/extensions/foundry/README.md
```diff
@@ -0,0 +1,80 @@
+# 🏗 Scaffold-ETH 2
```
damianmarti, line 1: Maybe the common parts between both READMEs can be in another file, and when the extension is installed, the README is merged. But maybe it's too much and we can duplicate these texts and that is.

rin-st: Good point! But not sure it's easy to do with .md files and with multiple params

rin-st (approved):

technophile-04:
> Maybe the common parts between both READMEs can be in another file, and when the extension is installed, the README is merged.

> But not sure it's easy to do with .md files and with multiple params

Just templated the READEME's because if you choose "none" for "What solidity framework do you
want to use?" option you were not getting any README (lol but I agree that  "none" option mostly no one uses) 

For now templated the whole `QuickStart` section because it had differences based on option's selected. 

Also if anyone want's to test quickly you could select no for install dependencies and copy changes from #784

rin-st:
> Just templated the READEME's because if you choose "none" for "What solidity framework do you
want to use?" option you were not getting any README (lol but I agree that "none" option mostly no one uses)

Great!

> For now templated the whole QuickStart section because it had differences based on option's selected.

Noticed that half of quick start still copypasted. Maybe divide it to smaller but specific templates, like `runLocalNetworkDescription`, `deployContract` and `runTests`?

templates/base/README.md.template.mjs
```diff
@@ -25,56 +33,39 @@ Before you begin, you need to install the following tools:
...
-- Edit your deployment scripts in `packages/hardhat/deploy`
+export default withDefaults(contents, {
+  solidityFrameWork: "",
+  quickStart: `## Quickstart
 
-## Documentation
+To get started with Scaffold-ETH 2, follow the steps below:
 
-Visit our [docs](https://docs.scaffoldeth.io) to learn how to start building with Scaffold-ETH 2.
+1. Clone this repo & install dependencies
```
rin-st, line 56: also, why we're using this text instead of https://github.com/scaffold-eth/scaffold-eth-2/blob/cli/README.md?plain=1#L35

technophile-04: Ohhh nice catch! Updated it at https://github.com/scaffold-eth/scaffold-eth-2/pull/782/commits/33a3c654fa74d228c6107a80b2b395ebf56520f7 🙌

technophile-04:
> Maybe divide it to smaller but specific templates, like runLocalNetworkDescription, deployContract and runTests?

Ohh yes! I had the same thought but when we add more plugins like the subgraph etc I thought maybe its better to templatise the whole `quickstart` section. 

But yeah 100% agree that we could make it more granular 🙌 For this PR keeping it as it but in future we should update it 


Thanks all ! Mering this 🙌

Pabl0cks:
Thanks for the reviews! ♥ Nice catch and great improvements @technophile-04 🙌

PR #788, opened 2024-03-26
Removed useAccountBalance

Hotmanics:
## Description

- Removed `useAccountBalance` hook.
- Updated `Balance.tsx` to replace useAccountBalance by grabbing price from global state. Grabs the balance using `useBalance` passing in only address property and setting watch to true. Additionally, assigns a variable to the formatted balance and converts the result to a `Number`, then uses the variable during rendering.
- Updated `FaucetButton.tsx` to replace useAccountBalance by grabbing the balance using `useBalance` passing in the address obtained from `useAccount`. 
- Updated `~~/hooks/scaffold-eth/index.ts` to remove reference to `useAccountBalance`.

## Additional Information

## Related Issues

_Closes #687_

Your ENS/address: JacobHomanics.eth/0xc689c800a7121b186208ea3b182fAb2671B337E7


Hotmanics:
Not sure if we need to force passing in the chainId via `targetNetwork.id` when calling `useBalance`. I've omitted it for now.

packages/nextjs/components/scaffold-eth/FaucetButton.tsx
```diff
@@ -21,7 +22,10 @@ const localWalletClient = createWalletClient({
  */
 export const FaucetButton = () => {
   const { address } = useAccount();
-  const { balance } = useAccountBalance(address);
+
+  const { data: balance } = useBalance({
+    address,
```
technophile-04, line 27: I think we forget `watch:true` here

packages/nextjs/components/scaffold-eth/FaucetButton.tsx
```diff
@@ -21,7 +22,10 @@ const localWalletClient = createWalletClient({
  */
 export const FaucetButton = () => {
   const { address } = useAccount();
-  const { balance } = useAccountBalance(address);
+
+  const { data: balance } = useBalance({
```
technophile-04, line 26: Since `balance` here is object, the `className={balance ? 'ml-1' : '...'}` at line 65 will always be true and we will never show the tooltip

Lets maybe use a variable to track it maybe something like : 

```ts
const isBalanceZero = balance && balance.value === 0n;
```

So if `isBalanceZero` is true we show the tooltip

packages/nextjs/components/scaffold-eth/Balance.tsx
```diff
@@ -44,6 +55,8 @@ export const Balance = ({ address, className = "", usdMode }: BalanceProps) => {
     );
   }
 
+  const formattedBalance = Number(balance?.formatted);
```
rin-st, line 58: Number(undefined) is `NaN`, so it will break further if there is no balance. Better to check if balance exists `balance ? Number(balance.formatted) : 0`

technophile-04, line 58: If in case if the balance is `undefined` then `Number(undefined) => Nan` and this break all the 

And this will break further all multiplications with price, maybe its better we do : 

```ts
const formattedBalance = balance ? Number(balance.formatted) : 0; 
```

rin-st: don't copy my comment! 🤣

packages/nextjs/components/scaffold-eth/FaucetButton.tsx
Hotmanics: I think that isn't required, as that is to watch an update to the wallet's balance right?

The purpose of obtaining the balance in FaucetButton is to check if balance is undefined or not, and as a result, assigns different classes to a div element. 

Could we even remove useBalance and use useAccount instead? Then basically checking on if the wallet is connected or not (Seems like that is what is going on, but useAccount might be a more pure solution).

I may be missing some context surrounding this so apologies if so.

packages/nextjs/components/scaffold-eth/FaucetButton.tsx
Hotmanics: Here is the context I was missing. Makes sense.

The suggested change is now reflected in this commit:
[86841363057bdf0e5d1f2091a43e9ef3dc9e2699](https://github.com/scaffold-eth/scaffold-eth-2/pull/788/commits/86841363057bdf0e5d1f2091a43e9ef3dc9e2699)

packages/nextjs/components/scaffold-eth/FaucetButton.tsx
Hotmanics: I was missing the context based on another comment of yours!

However, I still did not need to add watch: true for the tooltip to automatically go away. 

Regardless, I've added it as an extra safeguard: [8c7250ac79dcef157d737ede2d968c39c51eac7b](https://github.com/scaffold-eth/scaffold-eth-2/pull/788/commits/8c7250ac79dcef157d737ede2d968c39c51eac7b)

packages/nextjs/components/scaffold-eth/Balance.tsx
Hotmanics: good catch! 
commit with the suggested change: [9bbc4c2f5386d9385fb404bf86b031636fa5bfa6](https://github.com/scaffold-eth/scaffold-eth-2/pull/788/commits/9bbc4c2f5386d9385fb404bf86b031636fa5bfa6)
@technophile-04 s/o to you for also recognizing the error!

technophile-04:
Thanks @Hotmanics !! Looking great !!

technophile-04 (approved):

PR #790, opened 2024-03-27
Templatise readme quickstart section

rin-st:
## Description

Templated quickstart section of cli readme, so we will change it in one place when needed

mentioned here: https://github.com/scaffold-eth/scaffold-eth-2/pull/782#issuecomment-2016526326

## Additional Information


templates/base/README.md.template.mjs
```diff
@@ -1,6 +1,68 @@
...
+
+1. Install dependencies if it was skipped in CLI:
+
+\`\`\`
+cd my-dapp-example
+yarn install
+\`\`\`
+
+${
+  Boolean(solidityFramework[0])
```
damianmarti, line 21: why solidityFramework (and all the other params) are an array?

rin-st: it's created that way 🤷‍♂️ 
https://github.com/scaffold-eth/scaffold-eth-2/blob/cli/src/tasks/copy-template-files.ts#L233-L251

damianmarti: ok, maybe change the names to plural, like **solidityFrameworks** so it's more obvious it's an array, what do you think?

rin-st: array is created by collecting params with the same names, so if we want to make it plural, it should be also plural in `args.mjs` files. For example here https://github.com/scaffold-eth/scaffold-eth-2/pull/790/files#diff-037ce7e84c771f6b64d461da1fcc2a3c5fc1eb5758f35c3dcf27e881714ad492R1 , and it's even worse.
Probably we need to rewrite part responsible for creating arrays later, but it's not related to this pr 🙂

technophile-04: > why solidityFramework (and all the other params) are an array?

Each of params is an array because, lets suppose in future we have two plugins `TheGraph` and `ReactQuery`. 

Now as a user you can choose both at same time ThGraph + react-query or only of them according to your needs. 

Since there is case where "n" plugins can be selected at same time hence we have it as an array, in this example case maybe we would have an `contextProvider` array. 

Now coming to our original solidityFrameWork case since solidity frame are mutually exclusive(they both can't happen at same time) lol in this case it might be counterintuitive...but maybe it will makes sense when we have plugins that are not mutually exclusive. 

> like solidityFrameworks so it's more obvious it's an array, 

> so if we want to make it plural, it should be also plural in args.mjs files.

Lol yeah it's weird, but we have to use the same param name as in `args` file, yeah we could refactor naming a bit, but maybe it kind of makes sense consider whole system flow but yeah lol not sure

technophile-04 (approved):
Thanks @rin-st 🙌, just reverted the formatting changes to `template/utils.js` at https://github.com/scaffold-eth/scaffold-eth-2/pull/790/commits/9dfd6503b9c378c7196e9a8feaa99fb5c22dc844 ... lol I think maybe we should configure prettier in CLI branch, will create an issue for it 🙌

Issue #797, opened 2024-04-04
Add `.env` to `.gitignore` in nextjs package


technophile-04:
#### Description : 

![Screenshot 2024-04-04 at 9 36 05 PM](https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/7b711e14-bce4-49fe-9e28-55b878474581)
Although Next [docs](https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#default-environment-variables) has this and makes complete sense. But sometimes people tinkering with both `hardhat` and `nextjs` package might loose the context and add secrets to `.env` in `nextjs` assuming it is gitignored similar to `hardhat` pacakge

technophile-04:
Completed at #798

PR #798, opened 2024-04-04
 Add .env to .gitignore in nextjs package #797

nathan-websculpt:
Add .env to .gitignore in nextjs package #797 - Although Next [docs](https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#default-environment-variables) has this and makes complete sense. But sometimes people tinkering with both hardhat and nextjs package might loose the context and add secrets to .env in nextjs assuming it is gitignored similar to hardhat pacakge
 
 Discussed with @technophile-04 on telegram

## Related Issues

_Closes #797 _

Your ENS/address: websculpt.eth

### DIFF
--- a/packages/nextjs/.gitignore
+++ b/packages/nextjs/.gitignore
@@ -26,6 +26,7 @@ yarn-error.log*
 .pnpm-debug.log*
 
 # local env files
+.env
 .env.local
 .env.development.local
 .env.test.local


carletex:
Thanks @nathan-websculpt ! 100% agree.

carletex (approved):

PR #801, opened 2024-04-05
Cli solidity framework option

rin-st:
## Description

Solidity framework option from CLI. It could be a discussion but I think in this case it's easier to discuss when you already see the code. Feel free to add suggestions 🙂 

usage examples (dev):
```
yarn cli --solidity-framework hardhat
yarn cli --solidity-framework=foundry
yarn cli -f none
```
In case of several solidity framework flags, last will be taken

Other options I thought about.
1. `--hardhat`, `--foundry` and `--skip-solidity-framework`. Looks bad. Need to check if two or more flags chosen and add additional `if`s to fix/throw error
2. `--extensions` flag. It's not clear how to add let's say `none`, and how to mix it with other extensions non-related to solidity frameworks

Not sure how it will work with nested extensions but I think it's not a problem to update recurring logic if needed.

Also in this pr I added error if both `--install` and `--skip-install` chosen

Thanks!

## Additional Information

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,7 +4,7 @@ const config: Config = {
   questions: [
     typedQuestion({
       type: "single-select",
-      name: "solidity-framework",
+      name: "solidityFramework",
       message: "What solidity framework do you want to use?",
       extensions: ["hardhat", "foundry", null],
       default: "hardhat",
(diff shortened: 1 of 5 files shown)


damianmarti:
I like it! 

I think it's good to be able to set the solidity framework to use directly from the command line.

technophile-04:
@rin-st could you please port this PR to create-eth repo? This is will also open the doors for https://github.com/scaffold-eth/create-eth/issues/23

PR #810, opened 2024-04-12
Tailwind dark variant working

carletex:
Tailwind offers the `dark` variant (same as `hover:`, etc) https://tailwindcss.com/docs/dark-mode, but it didn't work because we didn't have it properly configured. Since it's super easy to enable, I think we should have it in!

I updated TW to the latest version since it wasn't working with the current one:

image
----

To test, you can just copy paste this piece into the homepage (page.tsx). Note the `dark:bg-red-500`

```tsx
<code className="italic bg-base-300 dark:bg-red-500 text-base font-bold max-w-full break-words break-all inline-block">
  packages/nextjs/app/page.tsx
</code>
```

image
image

---

I had to restart the nextjs server (yarn start) to make it work after (yarn install)

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -49,7 +49,7 @@
     "eslint-plugin-prettier": "^4.2.1",
     "postcss": "^8.4.16",
     "prettier": "^2.8.4",
-    "tailwindcss": "^3.3.3",
+    "tailwindcss": "^3.4.3",
     "type-fest": "^4.6.0",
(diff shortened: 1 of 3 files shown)


technophile-04 (commented):
Looks great !! I think lot of people who don't use daisyUI and only use tailwind has this habit of using `dark:` 

Makes a lot of sense to have it default in SE-2 too ! 

Tested it nicely and works great https://tw-dark-test.vercel.app/ ! Thanke Carlos 🙌

Issue #807, opened 2024-04-09
Vercel CLI gives gated link instead of public URL


technophile-04:
### Description : 

When you deploy the frontend through CLI, vercel seems to output an gated link in CLI, example : 

![Screenshot 2024-04-09 at 3 26 46 PM](https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/a94798bd-086e-4f9e-8e9b-1045aa675ad6)

https://wagmiv2-test-cm4cx4xpi-technophile04s-projects.vercel.app <= If you try to open this link, it will tell you to login to vercel / unauthorized. 

Although vercel outputs gated link in CLI, one can guess the public URL since it mostly always follow this convention : 

`{projectName}.vercel.app` in my case I named the project as `wagmiv2-test` so https://wagmiv2-test.vercel.app

Not sure what's the best way to handle this, maybe mentioning this about doc or somewhere. 

--- 

https://www.youtube.com/live/uhTkEq70lls?si=5Cc6qd6wgB5K55MI&t=1766

^ Even in above vide when Austin clicked the generated link it showed "Authenticating"

rin-st:
I found that deployment protection can be changed/disabled only [per project](https://vercel.com/docs/security/deployment-protection#configuring-deployment-protection) , so one of the (bad) options is to go to project settings on vercel and disable it manually for every project

> Although vercel outputs gated link in CLI, one can guess the public URL since it mostly always follow this convention :
> {projectName}.vercel.app in my case I named the project as wagmiv2-test so https://wagmiv2-test.vercel.app/

it works only if no one used that `projectName`. For example I tried `wagmiv2-test` too and my URL is `https://wagmiv2-test-five.vercel.app/`.

So I think best way is to go to vercel dashboard using inspect link from CLI and get public URL here
<img width="1105" alt="Снимок экрана 2024-04-09 в 14 04 56" src="https://github.com/scaffold-eth/scaffold-eth-2/assets/25638585/87522ce3-90b9-49b6-b85a-576d7e40c6f3">

technophile-04:
> So I think best way is to go to vercel dashboard using inspect link from CLI and get public URL here

Yup agree, maybe we should mention this in docs and maybe we could even add "echo" a note after vercel cli is done logging the output

damianmarti:
> > So I think best way is to go to vercel dashboard using inspect link from CLI and get public URL here
> 
[...]

PR #811, opened 2024-04-12
Add metas for a default Farcaster Frame

carletex:
I was thinking about adding a default "dummy" [Farcaster Frame](https://docs.farcaster.xyz/learn/what-is-farcaster/frames) for SE-2. I say "dummy" because you usually add a backend route to handle different actions within the frame (but that would be too much)

To test locally, you can use: https://github.com/coinbase/onchainkit
-> clone, go to the framegear folder, npm install, npm run dev... and you get this:

frames

---

A few things:
- Is it worth it to add this?
- About NextJS Metadata. `layout.tsx` and `getMetadata.ts` feel a bit duplicated. Is there a way around it? Feels weird to have to add the frame meta to both places.
- VERCEL_URL: We use it for the base URL, and now Vercel returns the "protected deployment route" (instead of the production one. Related to https://github.com/scaffold-eth/scaffold-eth-2/issues/807 I think).. which make this a bit useless:

```ts
const baseUrl = process.env.VERCEL_URL
  ? `https://${process.env.VERCEL_URL}`
  : `http://localhost:${process.env.PORT || 3000}`;
```

Any ideas? (we can discuss in another issue/PR, sorry haha)

### DIFF
--- a/packages/nextjs/utils/scaffold-eth/getMetadata.ts
+++ b/packages/nextjs/utils/scaffold-eth/getMetadata.ts
@@ -30,5 +30,12 @@ export const getMetadata = ({
       description: description,
       images: [imageUrl],
     },
+    other: {
+      "fc:frame": "vNext",
+      "fc:frame:image": imageUrl,
+      "fc:frame:button:1": "Open my 🏗️ Scaffold-ETH 2 dApp",
+      "fc:frame:button:1:action": "link",
+      "fc:frame:button:1:target": baseUrl,
+    },
   };
 };
--- a/packages/nextjs/app/layout.tsx
+++ b/packages/nextjs/app/layout.tsx
@@ -44,6 +44,13 @@ export const metadata: Metadata = {
   icons: {
     icon: [{ url: "/favicon.png", sizes: "32x32", type: "image/png" }],
   },
+  other: {
+    "fc:frame": "vNext",
+    "fc:frame:image": imageUrl,
+    "fc:frame:button:1": "Open my 🏗️ Scaffold-ETH 2 dApp",
+    "fc:frame:button:1:action": "link",
+    "fc:frame:button:1:target": baseUrl,
+  },
 };
 
 const ScaffoldEthApp = ({ children }: { children: React.ReactNode }) => {


technophile-04 (approved):
Nicceee, Thanks @carletex !! Looks great and I think its nice to have! 


--- 

Things which we can handle in maybe different PR : 

> About NextJS Metadata. layout.tsx and getMetadata.ts feel a bit duplicated. Is there a way around it? Feels weird to have to add the frame meta to both places.

Yup agree, also with vercel giving gated URL its PITA to that `baseUrl` in both files 

I like the way how vercel in [their docs](https://arc.net/l/quote/wmenfehq) have created `shared-metadata.ts` file, maybe we could have that for all the common things in metadata ? 

<details>
<summary>
Docs Image : 
</summary>

Screenshot 2024-04-15 at 8 45 35 PM


</details>


> VERCEL_URL: We use it for the base URL, and now Vercel returns the "protected deployment route" (instead of the production one. Related to https://github.com/scaffold-eth/scaffold-eth-2/issues/807 I think).. which make this a bit useless:

:( , Just created https://github.com/vercel/vercel/discussions/11438 and also asked question on discord

packages/nextjs/utils/scaffold-eth/getMetadata.ts
```diff
@@ -30,5 +30,12 @@ export const getMetadata = ({
       description: description,
       images: [imageUrl],
     },
+    other: {
+      "fc:frame": "vNext",
+      "fc:frame:image": imageUrl,
+      "fc:frame:button:1": "Open my 🏗️ Scaffold-ETH 2 dApp",
+      "fc:frame:button:1:action": "link",
+      "fc:frame:button:1:target": baseUrl,
```
technophile-04, line 38: NITPICK: Should we accept this args in function and maybe have default value ? 

So that people can have respective button / URL for their route 

EG: I have `/builders` page and I am using `getMetadata` function in it

I can maybe have `getMetadata({framRelativeRoute: `/builders`, frameBtnLabel: "Open Builders page"})` ? 

Usefull when I share `https://frames-se2.vercel.app/builders`  lol but not sure if it makes that much sense

packages/nextjs/utils/scaffold-eth/getMetadata.ts
```diff
@@ -30,5 +30,12 @@ export const getMetadata = ({
       description: description,
       images: [imageUrl],
     },
+    other: {
+      "fc:frame": "vNext",
+      "fc:frame:image": imageUrl,
+      "fc:frame:button:1": "Open my 🏗️ Scaffold-ETH 2 dApp",
```
damianmarti, line 36: Should we use the **title** var here or add another element to the hash param?

packages/nextjs/utils/scaffold-eth/getMetadata.ts
damianmarti: I think it makes sense, and having a default **frameBtnLabel**.

carletex: > EG: I have /builders page and I am using getMetadata function in it
> I can maybe have getMetadata({framRelativeRoute: "/builders", frameBtnLabel: "Open Builders page"}) ?
> Usefull when I share https://frames-se2.vercel.app/builders lol but not sure if it makes that much sense

Do you think this is a use case for Frames? From what I've seen, people create a _single_ frame for an app (typically more complex than what we are doing here). Abstracting "framRelativeRoute" + "frameBtnLabel"  seems a bit too much.

If they want to build a real frame, they'll need to create a backend route with the dynamic steps, so these abstractions won't serve much.

My initial idea was to make a simple/silly "frame" so people could see it out of the box, and tweak as needed. But merging this PR might not be a good idea after all :D

packages/nextjs/utils/scaffold-eth/getMetadata.ts
carletex: yep, maybe even something like `Go to ${title}`. {title} should be fine tho. What do you think?

carletex:
(Sorry for deviating from the original PR haha)

> About NextJS Metadata. layout.tsx and getMetadata.ts feel a bit duplicated. Is there a way around it? Feels weird to have to add the frame meta to both places.

Seems like the obvious way to fix it. What am I missing here?

 https://github.com/scaffold-eth/scaffold-eth-2/pull/811/commits/cb47ae904651fa68de1a9aa3e24153368aeee7e8

Besides that, we could even add baseUrl to `scaffold.config.ts`.

packages/nextjs/utils/scaffold-eth/getMetadata.ts
technophile-04: > maybe even something like Go to ${title}

Umm this will be misleading right ? If people share https://frames-se2.vercel.app/debug and if they have this as title : 

https://github.com/scaffold-eth/scaffold-eth-2/blob/699b259fe176f720bd9da9946ff7a02b14bd8f36/packages/nextjs/app/debug/page.tsx#L6

The btn would say `Go to Debug Contracts` but instead the button will take you to homepage 

😂 sorry for thinking too much on this lol its just abi.ninja routes coming to mind and lol 100% agree its not that  too common people sharing random routes of their website instead they mostly share home page 

But I think would prefer "Open my 🏗️ Scaffold-ETH 2 dApp" over `Go to ${title}`, people can always tweak it if they want

packages/nextjs/utils/scaffold-eth/getMetadata.ts
carletex: Yes, you have a point.

We could:

a) leave it as it's 
b) as it's, but only add it for the homepage as a example
c) make it configurable as you were suggesting

I always try to push on things like (c) to keep it minimal (especially in things that doesn't seem too important). And yes, it's a tiny thing to add... but tiny things multiply over time and makes DX worse.

In any case, open to to add it!

My vote will be for (b), and then (a)

Thanks Shiv.

packages/nextjs/utils/scaffold-eth/getMetadata.ts
carletex: lol yeah, I was thinking about this while posting the other comment 1 min ago.

Option (b) will fix it, and we won't need to have a dynamic title.

I personally enjoy these discussions, makes me really think what is best & simpler for what we are looking for.

technophile-04:
> Seems like the obvious way to fix it. What am I missing here?

> https://github.com/scaffold-eth/scaffold-eth-2/commit/cb47ae904651fa68de1a9aa3e24153368aeee7e8

Niiiceee !!! 

Screenshot 2024-04-16 at 6 41 04 PM

I thought having `templates` inside `getMedata` would render home page metadata title as "Scaffold-ETH 2 App | Scaffold-ETH 2" but it doesn't and works nicely because of [this](https://arc.net/l/quote/zgbnkgar) and it also nicely renders other pages with template ! 

> we could even add baseUrl to scaffold.config.ts

+1 for this, looking at Next's discord people have been using / suggesting  to hardcode the domain URL directly and there's no way to get domain from vercel's env var  

So makes sense to have it in scaffold.config.ts 

Tysm @carletex !!

packages/nextjs/utils/scaffold-eth/getMetadata.ts
damianmarti: Yes, a or b is good enough for now.

rin-st: also voting for b (for me b > c > a)

rin-st:
> So makes sense to have it in scaffold.config.ts

+, tried to find a better way to set baseUrl too, but no luck 

---

Thanks for pr and interesting discussion!

rin-st (approved):

carletex:
Thanks all for the comments!!

I'm gonna do some tinkering and propose an initial solution.

Also trying this now: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#metadatabase

packages/nextjs/utils/scaffold-eth/getMetadata.ts
carletex: There is a problem with **b**

Currently, the homepage has "use client" (because of the account hook), so we can't export the metadata from there (only available for server-side pre-rendered pages).

I don't think is worth it to add a sub-component in the homepage (to move the `use client` there) just for the sake of this PR.

Alternatives:
- Leave the `other` in the getMetadata function
- Add the `other` param in the getMetadata function, and add the frame in `layout.tsx` (passing all the frame data in the other param)


What do you guys think?

carletex:
The metadataBase doesn't seem to work for me: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#metadatabase

image

This would be nice too.... but we have the VERCEL_URL issue, where it points to a usually protected deployment route (instead of the PROD URL)

If we don't make it work, maybe we just can leave it as it's in the PR (at least we are deduplicating the layout / getMetadata code)

packages/nextjs/utils/scaffold-eth/getMetadata.ts
damianmarti: I think we can leave getMetadata as is, it is not a big deal.

technophile-04 (approved):
> The metadataBase doesn't seem to work for me: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#metadatabase

:( , But I really love now have we have centralised / one `getMetadata` function ! Really easy to edit !! 

Thanks Carlos !!!

packages/nextjs/utils/scaffold-eth/getMetadata.ts
carletex: > Umm, I would vote to not add this frame thing at all 😅. Keeps the code simple and predictable.

Yeah, I understand why you feel this. Inconsistent behavior in the different routes.

So opted to remove it from this PR for now. We can discuss later on.

carletex:
So for now, decided to remove the Farcaster frame stuff from this PR. See this convo for context: https://github.com/scaffold-eth/scaffold-eth-2/pull/811#discussion_r1565987983

So at the end this PR just remove some duplication with the metadata.

plus

https://twitter.com/rauchg/status/1787664435097538776

This happened finally!! :fire: :fire:

technophile-04 (approved):
> The metadataBase doesn't seem to work for me: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#metadatabase
> 
> image
> 
> This would be nice too.... but we have the VERCEL_URL issue, where it points to a usually protected deployment route (instead of the PROD URL)
> 
> If we don't make it work, maybe we just can leave it as it's in the PR (at least we are deduplicating the layout / getMetadata code)

Maybe this will solved in latest Next version not completely sure though ? https://x.com/WillemJaap_/status/1787738993964048560


>  realized only works if you share the main domain from the project, not the deployment URL.

Yes, actually this problem is kind of related to #807....the deployment URL are unique and only accessible to team / authenticated users and hence when you share that URL in whatsapp / any other app, that app won't be able access / fetch anything from that URL 

If you try to open https://new-metadata-47koiot3h-tokodevs-projects.vercel.app in incognito it won't work there too. 

Merging this thanks all !!

PR #814, opened 2024-04-16
adds yarn setup commands

Hotmanics:
## Description

Simplifies the setup process when cloning a new project by adding `yarn setup` as a command from the root directory. As a result, it copies .env.example to .env. Without doing this command, you will receive an error when you try to deploy to localhost (atleast on projects using Foundry).

## Additional Information

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     ]
   },
   "scripts": {
+    "setup": "yarn workspace @se-2/hardhat setup",
     "account": "yarn workspace @se-2/hardhat account",
     "chain": "yarn workspace @se-2/hardhat chain",
     "fork": "yarn workspace @se-2/hardhat fork",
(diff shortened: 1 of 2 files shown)


technophile-04:
When using foundry with `npx create-eth@latest`  it creates `.env` for you already. So `yarn deploy` and `yarn chain` should work out of the box

Hotmanics:
> When using foundry with `npx create-eth@latest` it creates `.env` for you already. So `yarn deploy` and `yarn chain` should work out of the box

If you follow that workflow then yes, that works wonderfully!

However when dealing with forks using git clone, they do not follow that, and encounter errors during deployment (and possibly testing?). 

Currently making part of the workflow when cloning any new fork to be:

'cd packages/foundry`

`cp .env.example .env`

`cd ../../`

when it can be as simple as 

`yarn setup`. 

Maybe a problem the forks need to handle? IMO, SE2 should play as nicely with forks as possible.

Issue #786, opened 2024-03-25
Make components more customizable


technophile-04:
### Description : 

Currently, we have the following user-facing components at SE-2 https://docs.scaffoldeth.io/components.

The idea is to make them more customizable but also they follow some standardization so it's intuitive while using any SE-2 component.

technophile-04:
My thoughts:

I think the best part of SE-2 is we have components which are not abstracted in any library and they are part to repo directly.

So if people need to do some hardcore customization specific to their need they can always go to `Address.tsx` file and add the necessary styles directly or accepts more props if needed.

But yup I agree that having some basic props to more customize the components would be great for beginners and easy to use.

The approach we followed at #733 kind of adding `{element}ClassName` not sure if it's the best at current moment since it might get non intuitive for users to know which `{elementClassName}` goes to which element and they might need to look at the code of that Component to know which `className` goes to which element.

I think `wrapperClassName` kind of make sense but getting more granular than that might get non intuitive.

Props like `size` seems more intuitive to me and maybe we can think of some more on this line. Like maybe `borderRadius` to input components etc.

---

Maybe in case in future if we plan to abstract components in library then we might need to go with approach of `{elementClassName}` but I was looking for alternatives and love how [`shadcn/ui`](https://ui.shadcn.com/docs/components/alert) is doing it.

So instead of having only `Alert` component and then passing `{element}ClassName` to it as props, they ask you to do something like this :

```tsx
<Alert>
  <AlertTitle>Heads up!</AlertTitle>
  <AlertDescription>
    You can add components to your app using the cli.
  </AlertDescription>
</Alert>
```

This way you can pass `className` directly to `AlertTitle` & `AlertDescription` components and you know which `className` goes to which element.


Lol but yeah not sure if its best suited for our components but might make sense once we have lot of big components

Hotmanics:
This is a good comparison to look at:
The iPhone is elegant and intuitive which makes it very easy to use and understand. However, there is almost no room for customization; thus everything is forced to look and feel really cookie cutter.
Meanwhile, Android phones provide the user with many customizable options giving them flexibility and freedom, albeit requires more work on the user to do the setup and have a better understanding of the underlying technology.

[...]

PR #822, opened 2024-04-20
Add TokenSelector component and useToken

ByteAtATime:
## Description

I have added the TokenSelector component, a direct port of the ChooseTokenModal component from [scaffold-eth-svelte](https://github.com/ByteAtATime/scaffold-eth-svelte). It allows the user to choose a token (from the list provided by Uniswap) or input an address, in which case the token data will be read and, if the user chooses so, stored in localStorage. It also allows you to "suggest" common tokens, which are displayed as chips on the top.

I have also added a utility hook, `useToken`, that simply reads the `name`, `symbol`, and `decimals` of an ERC20 contract.

Here is an image of the TokenSelector:

TokenSelector

## Additional Information

## Related Issues

I have not created an issue for this, as @austintgriffith has expressed interest in getting the component ported from my Svelte implementation here.

The corresponding pull request in the docs repository is scaffold-eth/se2-docs#65.

### DIFF
--- a/packages/nextjs/components/scaffold-eth/TokenSelector.tsx
+++ b/packages/nextjs/components/scaffold-eth/TokenSelector.tsx
@@ -32,6 +32,7 @@ const TokenListing = ({
       onClick={() => onClick(token)}
     >
       {token.logoURI ? (
+        // eslint-disable-next-line @next/next/no-img-element
         <img src={token.logoURI} alt={token.name} className="h-6 w-6 rounded-full" />
       ) : (
         <div className="avatar placeholder">
@@ -224,6 +225,7 @@ export const TokenSelector = ({
                   onClick={() => selectToken(token)}
                 >
                   {token.logoURI ? (
+                    // eslint-disable-next-line @next/next/no-img-element
                     <img src={token.logoURI} alt={token.name} className="h-6 w-6 rounded-full" />
                   ) : (
                     <div className="avatar placeholder">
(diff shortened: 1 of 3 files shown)


ByteAtATime:
May I please ask why this has been marked as a draft?

technophile-04:
Hey thanks @ByteAtATime, but it seems a bit too much considering we are planning to keep base SE-2 clean, lean, and simple more on this [here](https://github.com/scaffold-eth/scaffold-eth-2/issues/786#issuecomment-2024963470). 

This component would have been a lot simpler and lot of its pieces could have been cut down depending on individual developer use case. Trying to create a generic component will make it complex, bloat the codebase, hard to customize for different people's use cases and maintain. 

I think it makes sense to invert the control to the developer, we provide them with very essential primitives and then they can build up from that (having customization as they want, add things that only they require). 

Maybe components like this could be a 3rd party extension to SE-2 CLI in the future. 


> I have also added a utility hook, useToken, that simply reads the name, symbol, and decimals of an ERC20 contract.

This makes sense ! Since wagmi is depreciating their implementation of `useToken` as [mentioned here](https://wagmi.sh/react/guides/migrate-from-v1-to-v2#deprecated-usetoken) maybe base SE-2 can provide that primitive to developers. 

So could you please make a PR for just `useToken` itself ? Also, instead of using multiple useContractRead you could use `useContractReads` for reading multiple functions at once as suggested [here](https://wagmi.sh/react/guides/migrate-from-v1-to-v2#deprecated-usetoken), Thanks  🙌

ByteAtATime:
> Maybe components like this could be a 3rd party extension to SE-2 CLI in the future.

A little off-topic, but do you think it would be possible to create a library like [svelte-add](https://github.com/svelte-add/svelte-add) or [astro integrations](https://docs.astro.build/en/guides/integrations-guide/)? Something like `scaffold-eth add TokenSelector` seems pretty handy. I'd be down to create something!

> So could you please make a PR for just `useToken` itself

I will do that, thanks. I'm really not sure why I had so many `useReadContract`'s, I don't think I got enough sleep 🤦

PR #823, opened 2024-04-22
chore: best practice - helping to avoid compatibility issues 

wolfcito:
## Description

Added an .nvmrc file to specify the version of Node.js to be used in the project. This ensures consistency in the development environment, especially when working in teams with different Node.js versions installed.

## Additional Information

Your ENS/address: wolfcito.eth


technophile-04:
Hey thanks @wolfcito, Not all people use `nvm` as their node package manager and we are not restricted to a specific version of node, anything above node 18 should work nicely, Also having this file won't hint people what version of node is preferred directly when they clone the repo / do install dependencies.

Maybe a good add would be adding [engines](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) fields in `package.json` to hint people directly what version of node is preferred when they install dependencies. 

Could you maybe create a PR adding `engines` field instead  ? Thanks !

wolfcito:
> Hey thanks @wolfcito, Not all people use `nvm` as their node package manager and we are not restricted to a specific version of node, anything above node 18 should work nicely, Also having this file won't hint people what version of node is preferred directly when they clone the repo / do install dependencies.
> 
> Maybe a good add would be adding [engines](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) fields in `package.json` to hint people directly what version of node is preferred when they install dependencies.
> 
> Could you maybe create a PR adding `engines` field instead ? Thanks !

@technophile-04 , for sure !

Actually, I suggest it because this message was shown on my terminal. That's why I have include to have specifically 18.17.0 as a minimun version
image


I've sent the pull request with the engines field. you can review it here: https://github.com/scaffold-eth/scaffold-eth-2/pull/833

Issue #785, opened 2024-03-23
Static Build generateStaticParams Error


tonynacumoto:
### Is there an existing issue for this?

- [X] I have looked through the [existing issues](https://github.com/scaffold-eth/scaffold-eth-2/issues)

### Which method was used to setup Scaffold-ETH 2 ?

git clone

### Current Behavior

I'm attempting to build a static site from the nextjs app.  When I add `output: 'export',` to my next.config.js file I get the error `Error: Page "/blockexplorer/transaction/[txHash]" is missing "generateStaticParams()" so it cannot be used with "output: export" config.`

I've looked around, and apparently this is an issue with combining dynamic routes with "use state". 

### Expected Behavior

I would hope that out of the box I could export the nextjs app without having to re-write any files.  Would be great to host on IPFS or other static site hosting platforms, which there probably is a way other than how I'm trying to do it -- will look deeper into the documentation.

### Steps To Reproduce

clone, add `output: 'export',` to next.config.js file, move into the `packages/nextjs` repo then `yarn install && yarn build`

### Anything else?

Nope, awesome repo guys :)

technophile-04:
Hey thanks for raising it 🙌, I think for now the only solution is to mark `blockexplorer/address` and `blockexplorer/transaction` as [private](https://nextjs.org/docs/app/building-your-application/routing/colocation#private-folders) by appending `_` before their name, example : 

![Screenshot 2024-03-24 at 10 37 32 PM](https://github.com/scaffold-eth/scaffold-eth-2/assets/80153681/fe01a86f-db1c-49dc-be1b-0b39838e14c4)

(Instead of marking both files as private we could mark whole blockexplorer dir as private before deploying, but due to import alias the build fails, something which can fixed on SE-2 side) 

But any which ways I people need to manually mark the dir's as private before deploying. 

--- 

Ideally this should have been solved if we have used `useParams` in `blockexplorer/address/[address]/page.ts` and same for `[txHash]/page.ts` instead of relying on Next Page props

But NextJs has the issue with `output: 'export',` checkout : https://github.com/vercel/next.js/issues/54393, they said [it will solved soon](https://github.com/vercel/next.js/issues/54393#issuecomment-1871629201) but they haven't yet :( 

An hacky solution which people are suggesting is to use `useSearchParams` but it seems too hacky and not sure how well it will play :(

tonynacumoto:
worked - thnxs

technophile-04:
Completed at #825

PR #824, opened 2024-04-23
use relative import in blockexplorer/[address] page

technophile-04:
## Description

This will allow people to use static build, by just renaming `blockexplorer` => `_blockexplorer`. 

Currently if you add `output: "export"` in `next.config.js` then do `yarn next:build` it fails checkout #785 for more details. 

A workaround is to not build any page in `blockexplorer` so by appending `_` before its name we mark it as private dir and nextjs ignores it. 

But currently in `main` branch if you add `output: "export"` in next.config.js and rename to `_blockexplorer` it fails because the import inside `blockexplorer/[address]/page.tsx` is messed up. 

Using relative import solves this problem, also I think it makes sense to have this relative import so people can just easily ignore `blockexplorer` dir while doing normal build too. 

Not the efficient / best solution, so please feel free to close this PR if it doesn't makes sense !

### DIFF
--- a/packages/nextjs/app/blockexplorer/address/[address]/page.tsx
+++ b/packages/nextjs/app/blockexplorer/address/[address]/page.tsx
@@ -1,7 +1,7 @@
+import { AddressComponent } from "../../_components/AddressComponent";
 import fs from "fs";
 import path from "path";
 import { hardhat } from "viem/chains";
-import { AddressComponent } from "~~/app/blockexplorer/_components/AddressComponent";
 import deployedContracts from "~~/contracts/deployedContracts";
 import { GenericContractsDeclaration } from "~~/utils/scaffold-eth/contract";
 


carletex:
Hey Shiv, thanks for this! Let me see if I understand:

So if people want to generate a static build of the NextJS app (e.g. they want to deploy to surge or any static site service) they will have to:
1. Add `output: "export"` in next.config.js
2. Rename `blockexplorer` folder to `_blockexplorer`
3. Use relative imports (not anymore if we merge this PR!)

If so, this is definitely a good step forward, but the process will still be manual (I usually delete the folder, lol). We could also document this somewhere.

It'd be super nice if we could find an automated way of doing this.
- https://github.com/vercel/next.js/discussions/11113. This would be ideal, but doesn't seem to have a working solution.

----

+1 on merging this PR tho!

technophile-04:
>  but the process will still be manual

Yeah, even I hate that part :( 

> It'd be super nice if we could find an automated way of doing this.

> https://github.com/vercel/next.js/discussions/11113. This would be ideal, but doesn't seem to have a working solution.

Also, even [here](https://github.com/vercel/next.js/discussions/29514) there is no proper answer

PR #825, opened 2024-04-23
add generateStaticParams to blockexplorer address and txHash

technophile-04:
## Description

An alternate solution to #824 (which doesn't require manual work) 

As the error mentioned in #785 : 

> Error: Page "/blockexplorer/transaction/[txHash]" is missing "generateStaticParams()" so it cannot be used with "output: export" config.

Therefore used [`generateStaticParams`](https://nextjs.org/docs/app/api-reference/functions/generate-static-params) from NextJs. 

```ts
export function generateStaticParams() {
  return [{ address: "0x0000000000000000000000000000000000000000" }];
}
```

having the above code in blockexplorers `/[address]` & `/[txHash]`, when NextJs build the prod version it will generate 2 static dummy pages and won't fail. 

--- 

Does having  `generateStaticParams` break the blockexplorer when tinkering around on localnetwork ? 

No, since [`dynamicPramas`](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamicparams) are enabled by default it should generate that page on demand. 

So TLDR; nothing breaks (at least In my testing) there is no side effect to current version of SE-2 and also having `output:export` in next.config.ts works out of the box.

The downside of the above thing, lol we have to add very hacky code in our codebase. 

---

<details>
<summary>

### Side quest : 

</summary>

Also an difference which I noticed on vercel deployment: 

| Main | This Branch | 
| ----- | ------------ | 
| Screenshot 2024-04-23 at 11 59 53 PM | Screenshot 2024-04-24 at 12 01 01 AM | 

Notice how on `main` branch is explicitly mentioned it spun up serverless functions for handling `/blockexplorer` page but didn't mention about it on this branch deployment. 

Since `dynamicParams` are enabled are by default it ideally should have shown right ? 

If I go to logs of request I made, it shows their that its using serverless functions : 

<details>
<summary>
Demo image: 
</summary>

Screenshot 2024-04-24 at 12 01 33 AM

</details>

</details>

### DIFF
--- a/packages/nextjs/app/blockexplorer/address/[address]/page.tsx
+++ b/packages/nextjs/app/blockexplorer/address/[address]/page.tsx
@@ -82,6 +82,8 @@ export function generateStaticParams() {
 
 const AddressPage = async ({ params }: PageProps) => {
   const address = params?.address as string;
+  if (address === "0x0000000000000000000000000000000000000000") return null;
+
   const contractData: { bytecode: string; assembly: string } | null = await getContractData(address);
   return <AddressComponent address={address} contractData={contractData} />;
 };
--- a/packages/nextjs/app/blockexplorer/address/[address]/page.tsx
+++ b/packages/nextjs/app/blockexplorer/address/[address]/page.tsx
@@ -76,6 +76,10 @@ const getContractData = async (address: string) => {
   return { bytecode, assembly };
 };
 
+export function generateStaticParams() {
+  return [{ address: "0x0000000000000000000000000000000000000000" }];
+}
+
 const AddressPage = async ({ params }: PageProps) => {
   const address = params?.address as string;
   const contractData: { bytecode: string; assembly: string } | null = await getContractData(address);
--- a/packages/nextjs/app/blockexplorer/address/[address]/page.tsx
+++ b/packages/nextjs/app/blockexplorer/address/[address]/page.tsx
@@ -3,6 +3,7 @@ import path from "path";
 import { hardhat } from "viem/chains";
 import { AddressComponent } from "~~/app/blockexplorer/_components/AddressComponent";
 import deployedContracts from "~~/contracts/deployedContracts";
+import { isZeroAddress } from "~~/utils/scaffold-eth/common";
 import { GenericContractsDeclaration } from "~~/utils/scaffold-eth/contract";
 
 type PageProps = {
@@ -77,13 +78,14 @@ const getContractData = async (address: string) => {
 };
 
 export function generateStaticParams() {
-  // An workaround to enable static exports in Next.js
+  // An workaround to enable static exports in Next.js, generating single dummy page.
   return [{ address: "0x0000000000000000000000000000000000000000" }];
 }
 
 const AddressPage = async ({ params }: PageProps) => {
   const address = params?.address as string;
-  if (address === "0x0000000000000000000000000000000000000000") return null;
+
+  if (isZeroAddress(address)) return null;
 
   const contractData: { bytecode: string; assembly: string } | null = await getContractData(address);
   return <AddressComponent address={address} contractData={contractData} />;
--- a/packages/nextjs/app/blockexplorer/address/[address]/page.tsx
+++ b/packages/nextjs/app/blockexplorer/address/[address]/page.tsx
@@ -77,6 +77,7 @@ const getContractData = async (address: string) => {
 };
 
 export function generateStaticParams() {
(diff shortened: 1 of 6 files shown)


packages/nextjs/app/blockexplorer/transaction/[txHash]/page.tsx
```diff
@@ -1,153 +1,19 @@
...
-
-        const functionCalled = transactionWithDecodedData.input.substring(0, 10);
-        setFunctionCalled(functionCalled);
-      };
-
-      fetchTransaction();
-    }
-  }, [client, txHash]);
 
-  return (
```
Comment (technophile-04) [inline packages/nextjs/app/blockexplorer/transaction/[txHash]/page.tsx:45]: Had to move all this in seprate component because `generateStaticParams` and "use client" cannot be present in same file

technophile-04:
Just asking for a review to know what you guys think  🙌, also we can completely close this if it feels too hacky ! 

Need to update component name, and maybe having other default txHash instead of using address as txHash? Will update it later

packages/nextjs/app/blockexplorer/address/[address]/page.tsx
```diff
@@ -76,8 +76,14 @@ const getContractData = async (address: string) => {
   return { bytecode, assembly };
 };
 
+export function generateStaticParams() {
+  return [{ address: "0x0000000000000000000000000000000000000000" }];
+}
+
 const AddressPage = async ({ params }: PageProps) => {
   const address = params?.address as string;
+  if (address === "0x0000000000000000000000000000000000000000") return null;
```
Comment (rin-st) [inline packages/nextjs/app/blockexplorer/address/[address]/page.tsx:85]: and `if (!address)` here. And the same for transactions

Tested and everything works fine.

Not very good too but looks a bit better. Will continue to try to find something better today, but not sure it's possible

rin-st:
btw we have `next:build` script in package.json but don't have `next:serve`. Probably it's worth to add it

rin-st:
> Side quest :
> Also an difference which I noticed on vercel deployment:

As I understand it's moved to [ISR](https://nextjs.org/docs/pages/building-your-application/data-fetching/incremental-static-regeneration) block
image

packages/nextjs/app/blockexplorer/address/[address]/page.tsx
```diff
@@ -76,8 +76,14 @@ const getContractData = async (address: string) => {
   return { bytecode, assembly };
 };
 
+export function generateStaticParams() {
+  return [{ address: "0x0000000000000000000000000000000000000000" }];
```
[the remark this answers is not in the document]
Comment (technophile-04) [reply packages/nextjs/app/blockexplorer/address/[address]/page.tsx:80]: Yeah, but when you have `output: "export"` and run `yarn next:build` it fails

<details>

<summary>
Example : 
</summary>

Screenshot 2024-04-24 at 10 44 14 AM


</details>


I think you should have valid values inside `generateStaticParam` while doing static export. 

But It nicely builds when you don't have `output: "export"`.

packages/nextjs/app/blockexplorer/address/[address]/page.tsx
Comment (technophile-04) [reply packages/nextjs/app/blockexplorer/address/[address]/page.tsx:85]: > and if (!address) here. And the same for transactions

If are not able to do https://github.com/scaffold-eth/scaffold-eth-2/pull/825#discussion_r1576845520, then this suggested if check won't work right ? since we need to have some kind of valid string passed from `generateStaticParams`

Maybe I can extract that zero address in util, since sometimes it handy for other stuff too

technophile-04:
> btw we have next:build script in package.json but don't have next:serve. Probably it's worth to add it

Yup, makes sense! updated it

packages/nextjs/app/blockexplorer/address/[address]/page.tsx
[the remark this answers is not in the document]
Comment (rin-st) [reply packages/nextjs/app/blockexplorer/address/[address]/page.tsx:81]: > Yeah, but when you have output: "export" and run yarn next:build it fails

oh my bad, forgot to write it again when switching between branches

rin-st (approved):

packages/nextjs/app/blockexplorer/address/[address]/page.tsx
Comment (carletex) [reply packages/nextjs/app/blockexplorer/address/[address]/page.tsx:86]: > Maybe I can extract that zero address in util, since sometimes it handy for other stuff too

I like this too!

It sounds like something viem could implement

carletex (approved):
I like this approach too. Yes, a bit hacky, but the hacky code is on the blockexplorer, a component that people are not usually going to modify.

technophile-04:
Merging this! Thanks all for reviews and suggestion !

PR #827, opened 2024-04-25
Update useScaffoldEventHistory hook

rin-st:
## Description

1. Moved `readEvents` to `useCallback`, so it's clear which dependencies it has
2. Since `fromBlock` is required prop, removed if's like `if (fromBlock)`
3. Combined two useEffects with the same functionality `readEvents();` and
```
if (!deployedContractLoading) {
    readEvents();
}
```
The one without `if (!deployedContractLoading)` was buggy - on load it threw errors to the console 
4. Changed default state of events to `[]`, so we don't guess is it array or undefined

## Additional Information

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
@@ -111,9 +111,11 @@ export const useScaffoldEventHistory = <
           setError(undefined);
         }
       } catch (e: any) {
-        console.error(e);
-        setEvents([]);
+        if (events.length > 0) {
+          setEvents([]);
+        }
         setError(e);
+        console.error(e);
       } finally {
         setIsLoading(false);
       }
@@ -153,7 +155,7 @@ export const useScaffoldEventHistory = <
         readEvents();
       }
     },
-    watch ? (targetNetwork.id !== chains.hardhat.id ? scaffoldConfig.pollingInterval : 4_000) : null,
+    watch && enabled ? (targetNetwork.id !== chains.hardhat.id ? scaffoldConfig.pollingInterval : 4_000) : null,
   );
 
   const eventHistoryData = useMemo(
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
@@ -135,6 +135,12 @@ export const useScaffoldEventHistory = <
   );
 
   useEffect(() => {
+    if (!deployedContractLoading) {
+      readEvents();
+    }
+  }, [readEvents, deployedContractLoading]);
+
+  useEffect(() => {
     // Reset the internal state when target network or fromBlock changed
     setEvents([]);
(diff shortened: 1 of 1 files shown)


packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -45,7 +45,7 @@ export const useScaffoldEventHistory = <
   watch,
   enabled = true,
 }: UseScaffoldEventHistoryConfig<TContractName, TEventName, TBlockData, TTransactionData, TReceiptData>) => {
-  const [events, setEvents] = useState<any[]>();
+  const [events, setEvents] = useState<any[]>([]);
   const [isLoading, setIsLoading] = useState(false);
```
damianmarti, line 49: should we change the default value to true?

damianmarti (approved):
Works well and the code looks good to me, I only left a small comment.

Thanks!!

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -56,95 +56,89 @@ export const useScaffoldEventHistory = <
...
+      JSON.stringify(filters, replacer),
+      fromBlockUpdated,
+      publicClient,
+      receiptData,
+      transactionData,
+    ],
+  );
 
   useEffect(() => {
     if (!deployedContractLoading) {
```
Pabl0cks, line 138: With the updated version, when there is an error, it keeps spamming the `readEvents` function, and we keep getting more and more errors in loop.

Should we add `&& !error` to this if condition?

Screenshot of the error spam:
image

Pabl0cks (commented):
Good job @rin-st!!

Tested it nicely and only found a problem when you pass `enabled: false`.

- In the old version you would get 1 `Error: Hook disabled` everytime it fires the hook (`watch: true`)
- In the new version, it keeps spamming the error in the console.

Rest worked great! 🙌

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
rin-st: I think no, so we can distinguish when there is contract loading `deployedContractLoading`, and when we're reading events `isLoading`

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
rin-st: Thank you Pablo, good catch! Fixed [25b4295](https://github.com/scaffold-eth/scaffold-eth-2/pull/827/commits/25b42955b14260514ce7d68d0630559cbf814750). There was a problem that in `catch` block events was set to empty array even it was already empty array and `useEffect` restarted

Pabl0cks: Working great now, TYSM!

Pabl0cks (approved):
Fix working great! Thanks for the quick fix! 🙌

rin-st:
Thanks! Merging this

PR #700, opened 2024-01-24
Wagmi v2 migration

technophile-04:
### Current noticed bugs: 
- [x] Rainbowkit modal not showing burnerwallet icon 
- [x] When args are type address, then you need to assert `0x${string}` for example [checkout this](https://github.com/scaffold-eth/scaffold-eth-2/blob/8279382c418d66940fd06dbde79ec3af3020c3c2/packages/nextjs/app/page.tsx#L49) (fixed at https://github.com/scaffold-eth/scaffold-eth-2/pull/700/commits/23179fabdb7fb49a5fddaafca533451502970d5b) 

### 📝 Notes: 
1. [Wagmi migration guide](https://wagmi.sh/react/guides/migrate-from-v1-to-v2)
2. [Viem migration guide](https://viem.sh/docs/migration-guide#_2-5-x-breaking-changes)
3. [Rainbowkit migration guide](https://www.rainbowkit.com/docs/migration-guide)

<details>
<summary>

### Misc Notes: 

</summary>

1. In case we get `Chain not configured`  error make sure we pass `chainId` to hooks (eg https://github.com/scaffold-eth/scaffold-eth-2/pull/700/commits/b6f406aab2c51cb9a0105f8ac983c1891bc90ba4)
2. Alchemy Networks mapping: https://github.com/alchemyplatform/alchemy-sdk-js/blob/master/src/types/types.ts#L81
3. why callbacks like `onError` were removed from `useQuery` (https://github.com/TanStack/query/discussions/5279) 
4. React query defaults: https://tanstack.com/query/latest/docs/framework/react/guides/important-defaults

<details>
<summary>Helpful resources for burnerWallet : </summary>

- Discussion created on [viem repo](https://github.com/wevm/viem/discussions/1988) and [discord dicussion](https://discord.com/channels/1156791276818157609/1208163716718133308)
 
- Wagmi creating Connector guide : https://wagmi.sh/dev/creating-connectors

- Wagmi mock connector: https://github.com/wevm/wagmi/blob/main/packages/core/src/connectors/mock.ts#L15

- Raibow kit WalletConnector implementation : https://github.com/rainbow-me/rainbowkit/blob/main/packages/rainbowkit/src/wallets/getWalletConnectConnector.ts

- Raibow kit Custom Wallet section : https://www.rainbowkit.com/docs/custom-wallets

- Viem http Transport : https://github.com/wevm/viem/blob/main/src/utils/rpc/http.ts#L24

[...]

PR #829, opened 2024-04-29
Return useAccountBalance hook

rin-st:
## Description

Basically `useAccountBalance` is `useBalance` with added `watch: boolean` property, so it's not the same  as it was [before](https://github.com/scaffold-eth/scaffold-eth-2/pull/788).

`useAccountBalance` was removed less than a month ago, but I think we should return it. Reason - `watch` property was [removed](https://wagmi.sh/react/guides/migrate-from-v1-to-v2#removed-watch-property) from `useBalance` hook. So whenever we need to use watched balance outside of `Balance` component, we need to dance with all those query/block number hooks etc. It adds complexity to the code and is also not so obvious to newcomers how to configure it properly.

notes:
- we can probably use another hook name to not confuse with the previous version
- default `watch` value is `true` since I think it is needed in most cases

## Additional Information

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useWatchBalance.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useWatchBalance.ts
@@ -3,12 +3,10 @@ import { useTargetNetwork } from "./useTargetNetwork";
 import { useQueryClient } from "@tanstack/react-query";
 import { UseBalanceParameters, useBalance, useBlockNumber } from "wagmi";
 
-type UseWatchBalanceParameters = UseBalanceParameters;
-
 /**
  * Wrapper around wagmi's useBalance hook. Updates data on every block change.
  */
-export const useWatchBalance = (useBalanceParameters: UseWatchBalanceParameters) => {
+export const useWatchBalance = (useBalanceParameters: UseBalanceParameters) => {
   const { targetNetwork } = useTargetNetwork();
   const queryClient = useQueryClient();
   const { data: blockNumber } = useBlockNumber({ watch: true, chainId: targetNetwork.id });


damianmarti (approved):
This looks very great to me!!

It makes the Balance and FauceButton components a lot simpler and cleaner.

Maybe call this new component with an explicit name like useWatchedBalance? 

And maybe remove the watch parameter, because if you don't want it watched you can just use the regular Wagmi useBalance component, right?

rin-st:
> Maybe call this new component with an explicit name like useWatchedBalance?

Thought about something like that, but with `watch` param it's not always "watched" :)

> And maybe remove the watch parameter, because if you don't want it watched you can just use the regular Wagmi useBalance component, right?

For me looks strange to use two different hooks for two slightly different cases, and hence for me is better to use `watch` param. But let's wait what others think, probably I'm wrong.

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
```diff
@@ -0,0 +1,20 @@
...
+export const useAccountBalance = ({ watch = true, ...useBalanceParameters }: UseAccountBalanceParameters) => {
+  const { targetNetwork } = useTargetNetwork();
+  const queryClient = useQueryClient();
+  const { data: blockNumber } = useBlockNumber({ watch, chainId: targetNetwork.id });
+  const { queryKey, ...restUseBalanceReturn } = useBalance(useBalanceParameters);
+
+  useEffect(() => {
+    queryClient.invalidateQueries({ queryKey });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [blockNumber, queryClient]);
```
technophile-04, line 17: We should remove `queryClient` from this dependency since it's an object this `useEffect` will be triggered unnecessarily a lot of time, That's why I didn't kept it as a dependency when we had this logic inlined in FaucetButton / Balance because it was triggering for very small user interactions

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
```diff
@@ -0,0 +1,20 @@
...
+type UseAccountBalanceParameters = UseBalanceParameters & { watch?: boolean };
+
+export const useAccountBalance = ({ watch = true, ...useBalanceParameters }: UseAccountBalanceParameters) => {
+  const { targetNetwork } = useTargetNetwork();
+  const queryClient = useQueryClient();
+  const { data: blockNumber } = useBlockNumber({ watch, chainId: targetNetwork.id });
+  const { queryKey, ...restUseBalanceReturn } = useBalance(useBalanceParameters);
+
+  useEffect(() => {
+    queryClient.invalidateQueries({ queryKey });
```
technophile-04, line 15: maybe add a condition above this line and add `watch` as dependency to `useEffect` : 

```ts
useEffect(() => {
 if(!watch) return;
 queryClient.invalidateQueries({ queryKey });

 }, [blockNumber, watch]);
```

This is allow people to keep `watch` property dynamic(derived from a state maybe) in case they want. 

Should we take `enabled` in consideration too before calling `queryClient.invalidateQueries({ queryKey });`  ? Lol just thinking out loud without trying

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from "react";
+import { useTargetNetwork } from "./useTargetNetwork";
+import { useQueryClient } from "@tanstack/react-query";
+import { UseBalanceParameters, useBalance, useBlockNumber } from "wagmi";
+
+type UseAccountBalanceParameters = UseBalanceParameters & { watch?: boolean };
+
```
technophile-04, line 7: Maybe we should add a comment, something like a tiny wrapper around wagmi's `useBalance` hooks which lets you pass watch property or something on that line, similar

technophile-04 (commented):
Thanks @rin-st !! I think this makes sense ! 

For naming, lol maybe we should call it `useScaffoldBalalnce` ? (denoting tiny wrapper around wagmi hook as our other hooks are) ? 

Lol but yeah `useAccountBalance` also makes sense, lets hear from @carletex  too on naming 🙌

technophile-04:
> 🥇 useWatchBalance (or useBalanceWatch useWatchedBalance). This will make a lot of sense if we remove the watch argument as Damu said (if you don't want to watch you can use wagmi's useBalance directly)

Yup makes sense and less confusion, so let's remove acceptance of `watch` property from hook and call it `useWatchBalance` .

rin-st:
> Maybe call this new component with an explicit name like useWatchedBalance?

> 🥇 useWatchBalance (or useBalanceWatch useWatchedBalance). This will make a lot of sense if we remove the watch argument as Damu said (if you don't want to watch you can use wagmi's useBalance directly)

> Yup makes sense and less confusion, so let's remove acceptance of watch property from hook and call it useWatchBalance 

I'd vote for `useScaffoldBalance`, but if you think it's better to `useWatchBalance` let's do it! I'll update pr soon

rin-st:
Renamed to `useWatchBalance`, removed `watch` param

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
rin-st: it's an object but it's memoized, so it doesn't add additional rerenders. At least for me, probably you saw other cases. But removed it anyway

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
rin-st: That `useEffect` was triggered only when `blockNumber` change, and `blockNumber` changed only `watch` was `true`, so looks like it's unnecessary.

But I removed watch param, so again we don't need that `if`

> Should we take enabled in consideration too before calling queryClient.invalidateQueries({ queryKey }); ? Lol just thinking out loud without trying

I don't know but tend to say no until we will need it often

Thanks for suggestions!

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
rin-st: Good point! [05b67af](https://github.com/scaffold-eth/scaffold-eth-2/pull/829/commits/05b67af858bab47646810cfd95850562268a4dac)

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
technophile-04: > But I removed watch param, so again we don't need that if

Ohh yes yes now it is not needed at all 🙌

packages/nextjs/hooks/scaffold-eth/useAccountBalance.ts
technophile-04: > it's an object but it's memoized, so it doesn't add additional rerenders.

Ohh yeah yeah my bad, Actually I had `queryKey` as dependency to `useEffect` which was causing it to unnecessarily re-render that's why I removed it at https://github.com/scaffold-eth/scaffold-eth-2/pull/700/commits/c42a9e3b64ae640c76655914e6ad634094d0e242, but you already didn't had it as dependency so it was fine, thanks  🙌

technophile-04 (approved):
Thanks @rin-st  !!!

rin-st:
Thanks Shiv!

Regarding [ee10269](https://github.com/scaffold-eth/scaffold-eth-2/pull/829/commits/ee1026955450c9a2b97ef16fc0e02ed2559be1e1) but I intentionally added new type. I can't find a proper article to explain why I prefer that way, so I just copy-paste ChatGpt answer. I don't say it's the right way, just something to consider

question:
why to add new type here
import SomeType from 'some-lib';
type ComponentPropsType = SomeType;

answer:
The line `type ComponentPropsType = SomeType;` in your TypeScript code is a way to define a new type alias named `ComponentPropsType` that directly references `SomeType`, which is imported from 'some-lib'. There are several reasons why you might want to define such a type alias:

1. **Simplicity and Clarity**: Defining `ComponentPropsType` can make your code more readable and maintainable. If `SomeType` is used as a component's props type across various parts of your application, renaming it to something more descriptive like `ComponentPropsType` can make it clearer what the role of this type is in the context of your application.

2. **Abstraction**: By aliasing `SomeType` as `ComponentPropsType`, you abstract away the specific details of the type implementation which is defined in 'some-lib'. This abstraction can be useful if you later decide to change `SomeType` to another type. You would only need to change the alias definition in one place rather than throughout your codebase.

3. **Extensibility**: Even if the alias initially is a direct reference (as in your example), setting up an alias like this can make it easier to extend or modify the type later. For instance, you might want to add additional properties or methods to `ComponentPropsType` by extending `SomeType` or combining it with other types.

4. **Consistency**: Using a type alias helps maintain consistency in naming conventions and type usage across your codebase, particularly if `SomeType` is a generic or less intuitively named type. This makes your code easier to understand and work with for other developers.

5. **Type Refinement**: You may want to refine or restrict the imported type in certain ways that are specific to your application. While your current line of code doesn't modify `SomeType`, it sets the stage for such potential modifications.

In essence, adding a type alias like `ComponentPropsType` can provide a foundation for better code structure, future scalability, and clearer communication in your codebase.

technophile-04:
> Regarding [ee10269](https://github.com/scaffold-eth/scaffold-eth-2/pull/829/commits/ee1026955450c9a2b97ef16fc0e02ed2559be1e1) but I intentionally added new type.

Ohh, I thought 95% you just removed `& { watch : true }` and forgot to inline the type and 5% it was intentional. So just removed it since it was just a small hook with very straightforward functionality (and not planning to extend it in the future) 

But yeah my bad should have asked before removing, and also the points make sense!!  Thanks for sharing!!

PR #831, opened 2024-04-30
use burner-connector package

### LINKED ISSUE #828: Extract Burner Connector to a different repo / package
We can create it inside the SE org.

technophile-04:
## Description

Fixes #828 

Code :  https://github.com/technophile-04/burner-connector 

Npm package : https://www.npmjs.com/package/burner-connector

--- 

Test app : https://burner-test.vercel.app/ (burner enabled on sepolia)


rin-st (approved):
Lgtm!

Tested both locally and on sepolia, works great!

damianmarti (approved):
Tested on Sepolia and working great!!

The documentation from the package looks good too!

technophile-04:
Thanks all for testing !! Merging this 🙌

--- 

Also Carlos and I yesterday tried renaming the package to `@scaffold-eth/burner-connector` so that its [scoped](https://docs.npmjs.com/cli/v10/using-npm/scope). But when we tried registering scaffold-eth "organization" on npm it was saying the name was unavailable :( . Not completely sure if its saying unavailable because there is already registered "package" [scaffold-eth](https://www.npmjs.com/package/scaffold-eth) by someone. Carlos tried raising a ticket to npm regarding name dispute but not sure when they will reply back. 

So for this package we are keeping the name as it is and for future packages we will try to keep it scoped under scaffold-eth, hopefully we will get the org name soon

PR #836, opened 2024-05-03
Adjust QR Code modal spacing to fix copy icon size on bigger fonts

### LINKED ISSUE #835: Adjust `AddressQRCodeModal` spacing to fix icon size problem with some fonts
While testing se-2-challenges we realized that with some addresses, the QR code modal was not showing copy icon properly. [See discussion](https://github.com/scaffold-eth/se-2-challenges/pull/171#issuecomment-2092721888)

If people change default font they should review overall UI and fix their specific needs, but since this is happening in our challenges, we could adjust spacing a bit to cover Challlenges case and give a bit more space for other font cases.

image

Pabl0cks:
## Description

Fixes #835 

Feel free to close this PR if we should not handle issues on different fonts 🙌. Thought it would good to create because was happening in SRE challenges, but maybe not!

For address `0x04AAa2B551eaAb54568a5546be5d61407cA9D70c` is looking like this in Challeneges after fix:

image


rin-st (approved):
lgtm!

technophile-04 (approved):
Makes sense! Thanks @Pabl0cks 🙌

PR #840, opened 2024-05-07
Adds export from `useWatchBalance` to hooks index file

rin-st:
## Description

Adds export from `useWatchBalance` to hooks index file. (I forgot to add when created it 🤷‍♂️ )
Plus sorted index file alphabetically

## Additional Information


rin-st:
> Ohh nice catch ! Thanks @rin-st , also could you also add useTargetNetwork too to index.ts I think its a nice helpful hook and was also mentioned in https://github.com/scaffold-eth/scaffold-eth-2/discussions/789#discussion-6421005

Thanks! Added

> Nicee! , I don't think our prettier is configured to do that so lol I think it again is going to get messed up in the future but yeah makes sense for now!

Temporary order is better than nothing 😄 . We can add it later if needed, but not in this pr

Pabl0cks (approved):
Nice catch! If the refactor of all the imports needs to be done manually, I can help with that 🙌

PR #846, opened 2024-05-15
update useTransaction parameters

technophile-04:
## Description

When trying to send raw transaction using `useTransactor` hook we get below type error : 
Screenshot 2024-05-15 at 5 00 31 PM


carletex (approved):
Thanks Shiv! Thanks for the pairing :P

PR #851, opened 2024-05-20
Add default favicon

### LINKED ISSUE #850: Missing a default favicon
After https://github.com/scaffold-eth/scaffold-eth-2/commit/3ff50f2f076cc8eb91c68c4aafa9c88e4e782b8d meta handling optimizations, we lost our default icon from layout.

Currently is showing default browser favicon:

image

Pabl0cks:
## Description

Fixes #850 

In order to get our default favicon back, we could follow different approaches (let me know if I missed a better approach!):

1. Add it explicitly to `getMetadata`.
2. Add it as optional prop with default value to `getMetadata`, similar to how we handling unfurl image with `imageRelativePath` prop.
3. Moving `favicon.png` from /public to /app and rename it to `icon.png` or change the format to `favicon.ico`. See [this](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/app-icons)

I went for 1 just because it felt simple, and if I had to change the favicon for my app, I'd probably search favicon/icon in the code. 

I don't have a strong opinion about this, would love to know what do you think the best practice would be. TYSM ♥

### DIFF
--- a/packages/nextjs/utils/scaffold-eth/getMetadata.ts
+++ b/packages/nextjs/utils/scaffold-eth/getMetadata.ts
@@ -42,5 +42,8 @@ export const getMetadata = ({
       description: description,
       images: [imageUrl],
     },
+    icons: {
+      icon: [{ url: "/favicon.png", sizes: "32x32", type: "image/png" }],
+    },
   };
 };


technophile-04 (approved):
Tysm @Pabl0cks !! Yeah I think 1. makes more sense to me too, since very less users are gonna have different favicon for different page so they can update `getMetaData`  function to take in extra favicon arg or maybe don't use `getMetaData` function at all have different metadata for each page in page itself 

So yup I think this looks great !!

carletex:
Thanks Pablo!

PR #853, opened 2024-05-27
Better transaction result formatting in debug page

ByteAtATime:
## Description

This PR improves the formatting of the transaction result in the debug page, especially for numbers, tuples, and structs.

Here is a table comparing the changes (click on images to enlarge):

| Before | After | Notes |
|---|---|---|
| old-int | new-int new-int-2 | To prevent mis-identifications of numbers and to allow the user to copy the value, they can now toggle the number/ether views. |
| old-tuple | new-tuple | The tuple now recursively displays each element, with some added quality-of-life details (most notably the array index). |
| old-struct | new-struct | Like the tuple, this recursively displays each element, with the key displayed at left. |
| old-complex-struct | new-complex-struct |  Example of a more complicated item. Notice how the `arr` field is nicely indented. |

The contract used in these examples can be found [here](https://gist.github.com/ByteAtATime/7e5463948302768ac400f96ab474221d).

## Additional Information

_I would consider this to be a "straightforward" change, so have not filed an issue._


carletex:
Hey @ByteAtATime thanks for this, it looks cool!

A few open questions/comments before reviewing the code:

- The icon might be too big. e.g.
image

- Is `/` the best character to represent the change to ETH? Are we "toggling"? if so, we might want a different icon to to toggle back to wei.

Let's see what others think too!

Thanks again sir!

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -24,33 +24,87 @@ export const displayTxResult = (
   if (typeof displayContent === "bigint") {
     try {
       const asNumber = Number(displayContent);
-      if (asNumber <= Number.MAX_SAFE_INTEGER && asNumber >= Number.MIN_SAFE_INTEGER) {
-        return asNumber;
+      if (asText) {
```
rin-st, line 27: Looks like all the logic inside this `if` is already in `NumberDisplay` component, so probably no need to duplicate it here?

ByteAtATime:
> * The icon might be too big. e.g.

Honestly, it probably is, just wasn't sure what the best practice is considering this is already `btn-sm` from daisyUI. I'm a little hesitant to override styles manually, but I think that might be the best way?

> * Is `/` the best character to represent the change to ETH? Are we "toggling"? if so, we might want a different icon to to toggle back to wei.

Hmm, I'm not completely sure where I got that from. I suppose it's to "divide" (by 1e18), but then I guess we would have to replace it with "*" for "multiply by 1e18". I agree with your "toggle", would the "ArrowsRightLeft" icon work better?

image

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -24,33 +24,87 @@ export const displayTxResult = (
...
 
   return JSON.stringify(displayContent, replacer, 2);
 };
 
-const displayTxResultAsText = (displayContent: DisplayContent) => displayTxResult(displayContent, true);
+const NumberDisplay = ({ value }: { value: bigint }) => {
+  const [isEther, setIsEther] = useState(false);
+
+  const asNumber = Number(value);
+  if (asNumber <= Number.MAX_SAFE_INTEGER && asNumber >= Number.MIN_SAFE_INTEGER) {
```
rin-st, line 66: I'm also not sure about this restriction. With this `if` we can't convert let's say `Number.MAX_SAFE_INTEGER` = `9007199254740991` to `0.009...`. But can convert `9007199254740992`.

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -24,33 +24,87 @@ export const displayTxResult = (
...
+    return String(value);
+  }
+
+  return (
+    <span>
+      {isEther ? "Ξ" + formatEther(value) : String(value)}
```
rin-st, line 72: I think we need add `.toFixed(4)` like it is in the `Balance` component. Without it result can be too long

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
ByteAtATime: This behavior is in line with previous behavior. I believe this is to ensure that things like lengths or counts, usually small numbers, aren't mistaken for wei.

I think the limit can be changed, however. What would be a good limit? I'm thinking somewhere around 1e9 (one billion) might make sense, but at the same time would be formatted as 0.000000001, which seems like too small a number.

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
ByteAtATime: The main difference is in that `Balance` is used for display, while the debug page is used for, well, debugging. I personally believe this precision should be allowed (if you don't care about the precise value, just ignore the rest of the decimal points).

I don't think "too long" is a problem; no matter the output, there should be exactly one value per line (this holds even for tuples/arrays and structs). Therefore, the worst that can happen is that very specific values need horizontal scrolling, which again is great for people that need the precision, and should be fine for people that don't.

I'd welcome differing opinions though, this is just my personal view.

rin-st: good point, you're right, thanks

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
technophile-04: Ohh yeah, we could safely this condition and also `asText` param from `displayContent` 

It was only used to handle the elements inside array's,  but since now we are nicely rendering each of elements with component we could remove this. 

So yup lets remove this

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
technophile-04: > With this if we can't convert let's say Number.MAX_SAFE_INTEGER = 9007199254740991 to 0.009.... But can convert 9007199254740992

Yeah agree, but not sure either whats the best restrictions would be,  I think current restriction looks fine to me for now since previously also it has been this same case

rin-st: ok, let's leave it as is

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -24,33 +24,87 @@ export const displayTxResult = (
...
 
-  if (typeof displayContent === "string" && isAddress(displayContent)) {
-    return asText ? displayContent : <Address address={displayContent} />;
+  if (typeof displayContent === "string") {
+    if (isAddress(displayContent)) {
+      return asText ? displayContent : <Address address={displayContent} />;
+    }
+
+    if (isHex(displayContent)) {
+      return displayContent; // don't add quotes
```
technophile-04, line 47: I think we could remove the comment

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -24,33 +24,87 @@ export const displayTxResult = (
...
+        data-tip={isEther ? "Format as number" : "Format as ether"}
+        onClick={() => setIsEther(!isEther)}
+      >
+        /
+      </button>
+    </span>
+  );
+};
+
+const TupleDisplay = ({ value }: { value: DisplayContent[] }) => {
```
technophile-04, line 84: Nitpick: Maybe call this as `ArrayDisplay` because in abi representation actually structs are called as tuple 

<details>

<summary>
Example : 
</summary>

Screenshot 2024-05-29 at 2 41 38 PM


</details>

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -24,33 +24,87 @@ export const displayTxResult = (
...
-
-    return asText ? (
-      displayable
-    ) : (
-      <span style={{ overflowWrap: "break-word", width: "100%" }}>{displayable.replaceAll(",", ",\n")}</span>
-    );
+    return asText ? JSON.stringify(displayContent, replacer) : <TupleDisplay value={displayContent} />;
+  }
+
+  if (typeof displayContent === "object") {
```
technophile-04, line 55: Notice the two structs : 
Screenshot 2024-05-29 at 2 37 14 PM


While doing a write transaction we get a receipt, which is also an object more precisely nested object :(

<details>

<summary>
Example : 
</summary>

Screenshot 2024-05-29 at 2 53 08 PM


</details>

Maybe here : 

https://github.com/scaffold-eth/scaffold-eth-2/blob/c670c1f0d841b4a4e7ddf74d67e62c154d099d81/packages/nextjs/app/debug/_components/contract/TxReceipt.tsx#L43

We check if txResult?.txResult exist then we pass it ?

technophile-04:
Thanks @ByteAtATime !! looking great ! Added a couple of comments and just pushed a small commit at https://github.com/scaffold-eth/scaffold-eth-2/pull/853/commits/37df0ec817d1a5ab53cc44353cd0597f269a0abb since there was small bug at `TxReceipt` component more context [here](https://github.com/scaffold-eth/scaffold-eth-2/pull/853#discussion_r1618581486) 

---

In DisplayContractVarible part, the nested items inside struct and array are highlighted more as compared to normal variable : 

<details>

<summary>
Example : 
</summary>

Screenshot 2024-05-29 at 3 24 51 PM

</details>


<details>

<summary>
Test code : 
</summary>

```sol
// SPDX-License-Identifier: MIT
pragma solidity >=0.8.0 <0.9.0;

contract YourContract {
	struct SimpleStruct {
		uint256 val;
		uint256 secondVal;
		bool flag;
		address addr;
	}

	struct ComplexStruct {
		uint256 val;
		bytes data;
		uint256[5] arr;
	}

	uint256[10] public arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];

	address public owner = 0x5B38Da6a701c568545dCfcB03FcB875f56beddC4;

	address public owner2 = 0x67cE05a68887E1a5fD4F208d3f7F7eC66aCc8534;

	function simpleArray() public view returns (uint256[10] memory) {
		return arr;
	}

	function updateOwner(address _owner) public {
		owner = _owner;
	}

	function getAddr(address _dummyAddress) public pure returns (address) {
		return _dummyAddress;
	}

	function simpleTuple(uint256 _val) public pure returns (uint256, uint256) {
		return (_val, _val * 2);
	}

	function OnlyDiplaySimpleStruct()
		public
		pure
		returns (SimpleStruct memory)
	{
		return
			SimpleStruct(
				5,
				5 * 2,
				true,
				0x5B38Da6a701c568545dCfcB03FcB875f56beddC4
			);
	}

	function simpleStruct(
		uint256 _val
	) public pure returns (SimpleStruct memory) {
		return
			SimpleStruct(
				_val,
				_val * 2,
				true,
				0x5B38Da6a701c568545dCfcB03FcB875f56beddC4
			);
	}

	function simpleInt(uint256 _val) public pure returns (uint256) {
		return _val;
	}

	function simpleUint256Array(
		uint256 _val
	) public pure returns (uint256[5] memory) {
		return [_val, _val * 2 ether, _val * 3 ether, _val * 4, _val * 5];
	}

	function complexStruct(
		uint256 _val
	) public pure returns (ComplexStruct memory) {
		return
			ComplexStruct(
				_val,
				abi.encodePacked(_val),
				[_val, _val * 2, _val * 3, _val * 4, _val * 5]
			);
	}
}

```

</details>


if its get too tedious to handle this we could also have another PR which focuses on styling and maybe also have accordian to display complex variables

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
technophile-04, line 55: While writing to contract, since we are doing a geneic object check here, the TxReceipt is also showing "struct"  keyword : 

<details>

<summary>
Example : 
</summary>

Screenshot 2024-05-29 at 3 30 13 PM


</details>

ByteAtATime: Would adding an `isReceipt` variable be fine? We could then refactor out a component, say, `ObjectFieldDisplay` that displays a field (`[key]: [val]`) and we could use that in `TupleDisplay` (where `[key]` is the index), in `StructDisplay` (where `[key]` is the field name), and also in a new `ReceiptDisplay` (where there is no title).

Not sure if that made sense, feel free to ask for clarification.

ByteAtATime:
> Now we're showing the int value, with the "Format as ether" tooltip. Some might think we're converting it to ETH (Ξ symbol on result helps with confusion).
>
> Calculating the symbol of the ERC20 to show it on the tooltip + result may be a good solution, but probably would make the code messy.
Well, the thing is that the contract is not passed into the function, and even if it were we would also need to load the `decimals` and everything.

> If so, maybe changing the tooltip to "Divide by 1e18" helps a bit?
I think this might be the best approach, I will add a commit. If anyone else has any ideas, I'd be happy to hear them!

> Or maybe setting the default result to the old one? And letting user multiply.
The problem with this approach is that it might confuse the user that they're receiving a decimal for a `uint256` function. (Not sure about this, but I think it might be a problem for new users).

> Just thinking out loud! Maybe it's just me that felt it a bit confusing
Nope, it's definitely a problem we need to fix! Never thought of testing it with other contracts haha, great testing!

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
ByteAtATime: Should be updated in commit https://github.com/ByteAtATime/scaffold-eth-2/commit/946900a6747e508d46a9d614450618c56c9a7c9f, I'd be grateful if someone can check this

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
ByteAtATime: Personally, I would keep it, because it seems weird to just return the raw value in this specific condition.

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
ByteAtATime: Fixed in https://github.com/ByteAtATime/scaffold-eth-2/commit/4a25c0d09d5971bce21bb826ae5d73c09bca7827

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
technophile-04: >  say, ObjectFieldDisplay that displays a field ([key]: [val]) and we could use that in TupleDisplay (where [key] is the index), in StructDisplay (where [key] is the field name), and also in a new ReceiptDisplay (where there is no title).

Yeah I like this ! Let's create `ObjectFieldDisplay` component use that in ArrayDisplay and StuctDisplay. 

Regarding `isReceipt`, Lets not complicate `displayTxResult` more, keep as it is. 

I think after having `ObjectFieldDisplay` component, we should be directly use it in `TxReceipt`, so here : 

https://github.com/scaffold-eth/scaffold-eth-2/blob/5df118e637e6d33dc113756ab319052e606de604/packages/nextjs/app/debug/_components/contract/TxReceipt.tsx#L43

```diff
- <pre className="text-xs pt-4">{displayTxResult(txResult)}</pre> 
+ <ObjectFiledDisplay  data={txResult} />

// OR 

<pre className="text-xs pt-4"> 
+ <ObjectFiledDisplay  data={txResult} />
</pre> 

```

Which ever looks good

technophile-04:
> , with the "Format as ether" tooltip. Some might think we're converting it to ETH (Ξ symbol on result helps with confusion).

Wooah lol didn't thought it but makes sense !! 

> Calculating the symbol of the ERC20 to show it on the tooltip + result may be a good solution, but probably would make the code messy. If so, maybe changing the tooltip to "Divide by 1e18" helps a bit?

Umm ERC20 symbol will be adding complexity and I think its not worth it that much (sound a great feature for Abi.Ninja though). Like "Divide by 1e18" idea! I think for now it is best approach. 

> Or maybe setting the default result to the old one? And letting user multiply.

> The problem with this approach is that it might confuse the user that they're receiving a decimal for a uint256 function. (Not sure about this, but I think it might be a problem for new users).

Yeah I agree with this also keeping things a bit raw can make it more intuitive  for developer too

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -24,33 +24,87 @@ export const displayTxResult = (
...
     } catch (e) {
       return "Ξ" + formatEther(displayContent);
     }
   }
 
-  if (typeof displayContent === "string" && isAddress(displayContent)) {
-    return asText ? displayContent : <Address address={displayContent} />;
+  if (typeof displayContent === "string") {
+    if (isAddress(displayContent)) {
+      return asText ? displayContent : <Address address={displayContent} />;
```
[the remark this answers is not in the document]
ByteAtATime: Fixed in https://github.com/ByteAtATime/scaffold-eth-2/commit/b66e87dd8da38a29aa08077a1bf7193eb7ea2076

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
ByteAtATime: Hmm, just realized the problem with this: `ObjectFieldDisplay` recursively calls `displayTxResult`, so while the first level works, the nested items still show tuple/struct:
image

Should I just leave it like the above picture?

technophile-04: I think the picture looks fine, lol bug turned into feature ? 

I think it's not too bad, and also handling it will add more complexity 

So yeah for this PR lets have it similar to pic without "struct" word just below "Transaction Receipt"

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -13,44 +14,99 @@ type DisplayContent =
   | undefined
   | unknown;
 
+type ResultFontSize = "sm" | "base" | "xs" | "lg" | "xl" | "2xl" | "3xl";
+
 export const displayTxResult = (
   displayContent: DisplayContent | DisplayContent[],
-  asText = false,
+  fontSize: ResultFontSize = "sm",
```
technophile-04, line 21: lets have the default `fontSize` as "base" (Which is currently followed in `main` branch `DisaplayVariable` component) 

And lets explicitly pass "sm" for all the nested things

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
ByteAtATime: Alright, committed in https://github.com/ByteAtATime/scaffold-eth-2/commit/2183474b83e46167046e464a467a37be02a66811

I forgot about that hehe, updated in https://github.com/ByteAtATime/scaffold-eth-2/commit/35adf37fbf8a85078d00b40de8c14f56fc3d4d50

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
ByteAtATime: 🫡 https://github.com/ByteAtATime/scaffold-eth-2/commit/6d54d6763a69d19ac5f8568fec5bc6de809a2e69

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -13,44 +14,101 @@ type DisplayContent =
...
+}: {
+  name: string;
+  value: DisplayContent;
+  size: ResultFontSize;
+  leftPad?: boolean;
+}) => {
+  return (
+    <div className={`flex flex-row ${leftPad && "ml-4"}`}>
+      <span className="text-gray-500 dark:text-gray-400 mr-2">{name}:</span>
+      <span className="text-base-content">{displayTxResult(value, size)}</span>
```
technophile-04, line 89: Since we want struct / object values to always to be small in size 

Lets directly pass "sm" to `displayTxResult`  and remove accepting "size" prop

Same will apply for Array nd Struct Display component

ByteAtATime: I believe the reason I added this was because the transaction receipts were `xs`

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -13,44 +14,101 @@ type DisplayContent =
...
+    </div>
+  );
+};
+
+const ArrayDisplay = ({ value, size }: { value: DisplayContent[]; size: ResultFontSize }) => {
+  return (
+    <div className="flex flex-col gap-y-1">
+      array
```
technophile-04, line 97: Ohh just realised, If array is empty (example condition will be when contract write function does not emit any event) it shows just array : 

Screenshot 2024-06-01 at 10 29 23 PM

Might confuse people, that we are not able to render whole array properly 

Myabe just add condition  : 

```ts
{value.length ? "array" : "[]"}
```

also nitpick: should we call it "values" with "s"  lol just wired while writting value.length, maybe do same for "StructDisplay" and calls its prop as `struct` instead of "value" ?  But we can keep it as it is too !

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
technophile-04: Ohh I see, then let's keep it as it is thanks !

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -13,44 +14,101 @@ type DisplayContent =
...
+  size,
+  leftPad = true,
+}: {
+  name: string;
+  value: DisplayContent;
+  size: ResultFontSize;
+  leftPad?: boolean;
+}) => {
+  return (
+    <div className={`flex flex-row ${leftPad && "ml-4"}`}>
```
technophile-04, line 87: Ohh we should use proper ternary here, and leftPad is false the keep it empty "" string 

If use && then it adds "false" as className : 
Screenshot 2024-06-01 at 10 44 05 PM

technophile-04:
Tysm @ByteAtATime !! I think we are almost there just added last couple of comments but its looking great !

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -13,44 +14,101 @@ type DisplayContent =
...
+
+  return (
+    <span>
+      {isEther ? "Ξ" + formatEther(value) : String(value)}
+      <span
+        className="tooltip tooltip-secondary font-sans ml-2"
+        data-tip={isEther ? "Multiply by 1e18" : "Divide by 1e18"}
+      >
+        <button className="btn btn-primary btn-circle btn-xs" onClick={() => setIsEther(!isEther)}>
+          <ArrowsRightLeftIcon className="h-3 w-3" />
```
technophile-04, line 68: Ohh I think we decided to change this to "/" and "*" and also tooltip text right ? 

https://github.com/scaffold-eth/scaffold-eth-2/pull/853#pullrequestreview-2085318361

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
ByteAtATime: Hmm, I did that because I don't think there's anything that would clash. Fixed it anyway: https://github.com/ByteAtATime/scaffold-eth-2/commit/19ef4c7d7cb5a6f9a1107f2ad70817d140f93714

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
ByteAtATime: Done: https://github.com/ByteAtATime/scaffold-eth-2/commit/a89b901652c27d63530f89a9afdb463463d56f32

Renamed: https://github.com/ByteAtATime/scaffold-eth-2/commit/c59c443ccf91887bb47ac50f6ee2b381d059c95d

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
ByteAtATime: Oh, I changed the tooltip and the icon is now ArrowsRightLeft as I commented: https://github.com/scaffold-eth/scaffold-eth-2/pull/853#issuecomment-2135369866

Do you think it would be better to change it to "/" and "*"?

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
technophile-04: Yup but lol you never know if a dev using SE-2 adds that as a class

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
technophile-04: Umm I think it's better be explicit ? 

But cc @carletex and @rin-st once

rin-st: For me both variants are good but yes probably "/" and "*" are better.

rin-st:
Снимок экрана 2024-06-02 в 11 56 23

Noticed that
1. Color of tooltip in light mode is the same as color of tx result block
2. Because of `overflow: hidden` sometimes tooltip looks clipped

packages/nextjs/app/debug/_components/contract/ReadOnlyFunctionForm.tsx
```diff
@@ -77,11 +77,11 @@ export const ReadOnlyFunctionForm = ({
       </p>
       {inputElements}
       <div className="flex justify-between gap-2 flex-wrap">
-        <div className="flex-grow w-4/5">
+        <div className="flex-grow max-w-[80%]">
```
Pabl0cks, line 80: Checking in mobile I noticed a couple of nitpicks on Read results:

- The copy button in `<address />` component it's very small for some resolutions.
- The wei/ether format change button jumps to 2nd line.

| | |
|-|-|
| receipt-mobile-1 | receipt-mobile-2 |

Tried with `<div className="flex-grow max-w-[90%] sm:max-w-[80%]">` but it's not perfect either.

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -13,44 +14,101 @@ type DisplayContent =
...
+  if (asNumber <= Number.MAX_SAFE_INTEGER && asNumber >= Number.MIN_SAFE_INTEGER) {
+    return String(value);
+  }
+
+  return (
+    <span>
+      {isEther ? "Ξ" + formatEther(value) : String(value)}
+      <span
+        className="tooltip tooltip-secondary font-sans ml-2"
+        data-tip={isEther ? "Multiply by 1e18" : "Divide by 1e18"}
```
carletex, line 65: image

This is how we show the "*" in the uint fields. For consistency, we should show 10^18 here or 1e18 in the input. I think I prefer the latter, but I like both options if we are consistent

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
carletex: Same, not feeling strong, but might lean into `/` and `*`

----

image

We are still giving the button too much importance:
- It's using the primary color. (meant for main actions)
- It's too tall: It's misaligning the key and the value too (e.g. [0] and 230000..)

I think it should be waaaay more subtle. This feature is nice to have, but the UI for this shouldn't draw attention. 

Tried without a background: maybe it's too subtle? haha

image

Another option would be to click the number it self (with a tooltip when hovering). I don't like this one too much because i use double click for selecting the text/word and copy.

We can also ask Andrea, so she could give us a designer's take.

technophile-04: > We are still giving the button too much importance:

yeah makes sense, updated it to be btn-ghost and decreased the opacity to 65, this is how it looks now : 

| normal | on hover | 
| ------- | ----- | 
| Screenshot 2024-06-10 at 8 32 45 PM | Screenshot 2024-06-10 at 8 32 52 PM | 

What do you think ? But yeah we can ask Andrea to find better opotion

---

Umm also  kept the same ArrowLeftRightIcon instead of `/` and `*` since they  were not looking that great : 

| `/` | `*` | 
| ------- | ----- | 
| Screenshot 2024-06-10 at 8 08 08 PM | Screenshot 2024-06-10 at 8 08 30 PM |

carletex: That looks better I think! The only thing is that the line is not aligned (because of the button height?)

image

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
technophile-04: Ohh yeah updated IntergerInput to follow 1e18 at https://github.com/scaffold-eth/scaffold-eth-2/pull/853/commits/25bd2c1f4a8720926cbfcedf7371c4547c3d9c52

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
technophile-04: Ohh yes yes just fixed it at https://github.com/scaffold-eth/scaffold-eth-2/pull/853/commits/6d168a236bf41a57ca79c88662b65dac7aa53059: 

| Before | After | 
| ------- | ----- | 
| Screenshot 2024-06-10 at 9 26 01 PM | Screenshot 2024-06-10 at 9 25 46 PM |

ByteAtATime:
Oh my god... I'd forgotten about this PR!

Thanks @technophile-04 for all the commits and fixes!

What are the main issues that need to be fixed? I see that the `<NumberDisplay />` has weird issues with the button, is that the only issue that we have right now?

technophile-04 (approved):
> I see that the <NumberDisplay /> has weird issues with the button,

You mean to say the alignment? already fixed at https://github.com/scaffold-eth/scaffold-eth-2/pull/853/commits/6d168a236bf41a57ca79c88662b65dac7aa53059

Maybe we can do some more touchup's to UI in separate PR 

Thanks @ByteAtATime for the PR and other for review and suggestions, Merging this 🙌

packages/nextjs/app/debug/_components/contract/ReadOnlyFunctionForm.tsx
technophile-04: Create an issue for Address component https://github.com/scaffold-eth/scaffold-eth-2/issues/863

PR #854, opened 2024-05-28
Add zeros to native currency price when price is integer

rin-st:
## Description
Add zeros to native currency price when price is integer

## Additional Information

Before:
Снимок экрана 2024-05-28 в 12 38 48

After:
Снимок экрана 2024-05-28 в 12 39 27


damianmarti (approved):
Looks good! 

It doesn't matter too much for ETH because of the high value, but fixed to 2 decimals is better when the value is lower!

Thanks!!

rin-st:
The main point is that it is price, so it's easier to read and understand than 3917 or 3917.3

Merging this, thanks!

PR #856, opened 2024-05-29
Add use effect on Balance component for the price

carletex:
The `usdMode` flag wasn't working on our `Balance` component, since the price takes some time to load and we were always setting `displayUsdMode` to false on the first render (price > 0 condition)

You can test in `page.tsx` 
```tsx
/* This will always show the Balance in ETH mode, even if we say usdMode */
<Balance address={connectedAddress} usdMode /> 
```

This PR adds the missing `useEffect` (as in EtherInput).

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Balance.tsx
+++ b/packages/nextjs/components/scaffold-eth/Balance.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { Address, formatEther } from "viem";
 import { useTargetNetwork } from "~~/hooks/scaffold-eth/useTargetNetwork";
 import { useWatchBalance } from "~~/hooks/scaffold-eth/useWatchBalance";
@@ -27,6 +27,10 @@ export const Balance = ({ address, className = "", usdMode }: BalanceProps) => {
   });
 
   const [displayUsdMode, setDisplayUsdMode] = useState(price > 0 ? Boolean(usdMode) : false);
+
+  useEffect(() => {
+    setDisplayUsdMode(price > 0 ? Boolean(usdMode) : false);
+  }, [usdMode, price]);
 
   const toggleBalanceMode = () => {
     if (price > 0) {


packages/nextjs/components/scaffold-eth/Balance.tsx
```diff
@@ -28,6 +28,10 @@ export const Balance = ({ address, className = "", usdMode }: BalanceProps) => {
 
   const [displayUsdMode, setDisplayUsdMode] = useState(price > 0 ? Boolean(usdMode) : false);
```
damianmarti, line 29: If the price is not loaded in the first render, maybe you can just use here _useState(false)_

rin-st:
Fix works great! 

But I think we need to consider rewriting a bit the logic of the `Balance` and `EtherInput` components (in another pull request). Because `usdMode` prop is basically doesn't mean real mode inside the component because of internal toggling logic. I'm thinking about two options
1. Rename `usdMode` to `defaultUsdMode` and use that value as default in `useState`. Remove `useEffect`. (I prefer this)
2. Remove inner usd/eth logic. Use only `usdMode` prop (or `mode`)
Yes, with that options we can't change mode of let's say group of `EtherInput`s, but not sure we need it.

Also, who remembers why we added that price > 0 condition? Looks like it's redundant

packages/nextjs/components/scaffold-eth/Balance.tsx
```diff
@@ -28,6 +28,10 @@ export const Balance = ({ address, className = "", usdMode }: BalanceProps) => {
...
 
+  useEffect(() => {
+    setDisplayUsdMode(price > 0 ? Boolean(usdMode) : false);
+  }, [usdMode, price]);
```
rin-st, line 33: Also, in every `price` change `displayUsdMode` will change to `usdMode` (unexpected behavior/bug)

A bit more details

`usdMode` is `true`. `displayUsdMode` is `true`. Price is `3000`.  -> toggle `displayUsdMode` to `false` -> price changes to `3001` -> `useEffect` triggers. `displayUsdMode` changes to `true` because `usdMode` is `true`.

packages/nextjs/components/scaffold-eth/Balance.tsx
carletex: The price is not loaded in the first component render **+ first-page render**, but the price is already loaded on the store when you navigate through pages... so I think it's useful to keep it to avoid flickering in some situations.

carletex:
Thanks for the review @rin-st !

> who remembers why we added that price > 0 condition?  Looks like it's redundant

The USD mode only makes sense when we have a price for the native currency. If it's loading (or if the fetch fails) it doesn't make sense to show the USD mode... since you won't be able to convert to the native currency (which is what we want for any onchain operation / show balance, etc). So I think that was the reason.

> Rename usdMode to defaultUsdMode and use that value as default in useState. Remove useEffect. (I prefer this)

This will only work if we ignore my previous comment (the price condition). I agree that `defaultUsdMode` might be a better name.

> Remove inner usd/eth logic. Use only usdMode prop (or mode)

I don't like this one since it'd require people to do their logic for the basic toggle USD / ETH.

---

I'm not sure if these components need a refactor, but like you said, it should probably happen in another PR.

I just created this one as a hotfix, to tackle a bug that is currently happening on main (and tried to be consistent with other components)

packages/nextjs/components/scaffold-eth/Balance.tsx
carletex: This is currently not happening out of the box since enablePolling is set to false in `useNativeCurrencyPrice`, but it'll happen if we enable it.

I guess we could fix it with a ref? like `const prevPriceRef = useRef(price);` and only calling setDisplayUsdMode if `prevPriceRef.current` was 0 and now is > 0.

rin-st:
Thank you for explanation!

> I just created this one as a hotfix, to tackle a bug that is currently happening on main (and tried to be consistent with other components)

Agree, lgtm!

packages/nextjs/components/scaffold-eth/Balance.tsx
rin-st: Probably, let's fix it in another pr

technophile-04 (approved):
Thanks @carletex for the quick fix !! Merging this 

--- 

For another PR : 

Agree on naming that it should be called `defaultUdMode` 

Regarding working, wouldn't it be more intuitive if we had this flow, just thinking out loud : 
1. Initial state of `price` (global store is `undefined`) [denoting we haven't made any request yet for price]
2. The initial state of `displayUsdMode` should be only derived from `defaultUdMode` and not the price. 
   - `const [displayUsdMode, setDisplayUsdMode] = useState(defaultUsdMode);`
 3. We should have an extra global loading state (`isFetchingPrice`) which denotes price is being fetched
 4. The loading skeleton will be shown if `isFetchingPrice || isBalanceInETHLoading ...` are true
 5. An `useEffect` which checks if `price === 0 &&  defaultUsdMode` then `setDisplayUsdMode(false)` [With maybe a console.error("Error getting price") ] 
 
I think `isFetchingPrice` is good to have in global store since `price` is derived from async action and people would love to have it exposed to workaround the flickering / layout shift

PR #859, opened 2024-06-03
feat: useDisplayUsdMode hook

rin-st:
## Description

Adds `useDisplayUsdMode` hook which helps to work with nativeCurrency/USD values.

Hook takes as parameter `defaultUsdMode` which sets an initial value, and returns `{ displayUsdMode, toggleDisplayUsdMode }`. `toggleDisplayUsdMode` allows to change inner `displayUsdMode`, and `displayUsdMode` basically returns current USD mode.

`defaultUsdMode` and `displayUsdMode` variables are booleans. `true` means `USD`, `false` means Native Currency.

See discussion [here](https://github.com/scaffold-eth/scaffold-eth-2/pull/856)

Note:
- I didn't set default state to `undefined` since with it we need additional conditions like `if (typeof nativeCurrencyPrice !== "undefined" && nativeCurrencyPrice > 0)` etc. 
- I assigned default `isFetching` value for price to `true`, so it's clear that if `!isFetching && price === 0` then fetch failed.

## Additional Information

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Input/EtherInput.tsx
+++ b/packages/nextjs/components/scaffold-eth/Input/EtherInput.tsx
@@ -1,6 +1,7 @@
-import { useEffect, useMemo, useState } from "react";
+import { useMemo, useState } from "react";
 import { ArrowsRightLeftIcon } from "@heroicons/react/24/outline";
 import { CommonInputProps, InputBase, SIGNED_NUMBER_REGEX } from "~~/components/scaffold-eth";
+import { useDisplayUsdMode } from "~~/hooks/scaffold-eth/useDisplayUsdMode";
 import { useGlobalState } from "~~/services/store/store";
 
 const MAX_DECIMALS_USD = 2;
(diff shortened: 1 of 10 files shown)


packages/nextjs/hooks/scaffold-eth/useDisplayUsdMode.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const useDisplayUsdMode = ({ defaultUsdMode = false }: { defaultUsdMode?: boolean }) => {
+  const nativeCurrencyPrice = useGlobalState(state => state.nativeCurrency.price);
+  const isPriceFetched = nativeCurrencyPrice > 0;
+  const predefinedUsdMode = isPriceFetched ? Boolean(defaultUsdMode) : false;
+  const [displayUsdMode, setDisplayUsdMode] = useState(predefinedUsdMode);
+
+  useEffect(() => {
+    setDisplayUsdMode(predefinedUsdMode);
+  }, [predefinedUsdMode]);
```
rin-st, line 12: Also I changed this `useEffect` dependency, so it does not depend on price and not trigger on every price change, and hence work with `pollingInterval`. See https://github.com/scaffold-eth/scaffold-eth-2/pull/856#discussion_r1619316376

packages/nextjs/hooks/scaffold-eth/useNativeCurrencyPrice.ts
```diff
@@ -11,24 +11,23 @@ const enablePolling = false;
  */
 export const useNativeCurrencyPrice = () => {
   const { targetNetwork } = useTargetNetwork();
+  const [isFetching, setIsFetching] = useState(true);
   const [nativeCurrencyPrice, setNativeCurrencyPrice] = useState(0);
```
technophile-04, line 15: Realised that this state are same as global state. 

`useNativeCurrencyPrice` is only used in `ScaffoldEthAppWithProvider`

Maybe rename this  to `useInitializeNativeCurrencyPrice` ? And instead of using `useState` maybe use states from store and set them here ? 

Doing this also clean up `ScaffoldEthAppWithProvider` because there we again don't need to set the global state , we just call this hook at top. 

What do you think ? M I missing something ?

packages/nextjs/hooks/scaffold-eth/useDisplayUsdMode.ts
technophile-04: Niccee !!! Just to be sure I tested by adding this to `useNaticeCurrencyPrice` hook : 

```ts
  useEffect(() => {
    // Increase the price by 1000 every 2 seconds
    const interval = setInterval(() => {
      setNativeCurrencyPrice(price => price + 1000);
    }, 2000);

    return () => clearInterval(interval);
  }, [nativeCurrencyPrice]);
```

And it doesn't trigger this hook

packages/nextjs/components/scaffold-eth/Input/EtherInput.tsx
```diff
@@ -122,8 +114,8 @@ export const EtherInput = ({
         >
```
technophile-04, line 114: Nitpick: 

Above this where there is data-tip maybe use : 

```ts
 data-tip={isNativeCurrenPriceFetching ? "Fetching price" : "Unable to fetch price"}
```



Because if you put this in `page.tsx` : 



```ts
const [etherInputValue, setEtherInputValue] = useState("2");
<EtherInput usdMode value={etherInputValue} onChange={setEtherInputValue} />
```
Lol too much attention to detail, but since we have access to `isNativeCurrenPriceFetching` maybe its worth it ?

technophile-04 (approved):
Tysm @rin-st !! Tested it and works great ! 

Lol initally I was bit hesitant to have this hook in `hooks/scaffold-eth` since seem too specific to SE-2 internal use but then realised that we have hooks like `useAnimateConfig` etc too in their, not sure what the best place to keep this internal hooks but I think its fine for now !

packages/nextjs/hooks/scaffold-eth/useNativeCurrencyPrice.ts
rin-st: You're right, thanks! I thought about it but forgot/for some reason left it. Updated! [1460c4d](https://github.com/scaffold-eth/scaffold-eth-2/pull/859/commits/1460c4d68d7cc514d4c1ca9e3ee9b2c44eebc9a3)

rin-st:
> Lol initally I was bit hesitant to have this hook in hooks/scaffold-eth since seem too specific to SE-2 internal use but then realised that we have hooks like useAnimateConfig etc too in their, not sure what the best place to keep this internal hooks but I think its fine for now !

I agree that the hooks folder is a little bit mess. Maybe in future create two folders like core (main hooks from docs) and internal/helpers/utils(all other hooks)?

technophile-04:
Niceee tysm @rin-st !! 

--- 

>  Maybe in future create two folders like core (main hooks from docs) and internal/helpers/utils(all other hooks)?

Yup makes sense !

PR #871, opened 2024-06-24
lock typescript and abitype version

### LINKED ISSUE #870: bug: Type error: Type 'string' is not assignable to type '`0x${string}`'.

git clone

### Current Behavior

Failed to compile.
./app/blockexplorer/_components/AddressComponent.tsx:22:26
Type error: Type 'string' is not assignable to type '`0x${string}`'.
> 22 |                 <Address address={address} format="long" />
     |                          ^
Error: Command "yarn run build" exited with 1

### Expected Behavior

That String could be assignable to Address

### Steps To Reproduce

Deploying to vercel using git clone.

### Anything else?

_No response_

technophile-04:
### Description: 

Seems there is some problem in new typescript version, it not registering AddressType properly. So for now locked it's version. 

Also locked the version of abitype (just to be future proof, since we have also locked viem and wagmi version). 

Fixes #870 

### To test: 
Run on main branch and it fails : 

```shell
yarn vercel
```

Switch to this pr branch and run `yarn vercel` and it should work

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -43,7 +43,7 @@
     "@types/react": "^18.0.9",
     "@types/react-copy-to-clipboard": "^5.0.4",
     "@typescript-eslint/eslint-plugin": "^5.39.0",
-    "abitype": "^1.0.2",
+    "abitype": "1.0.2",
     "autoprefixer": "^10.4.12",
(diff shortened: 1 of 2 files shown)


carletex:
Thanks Shiv!

carletex (approved):

PR #878, opened 2024-07-01
Use keystores in Scaffold-Eth

jrcarlos2000:
## Description

Adds keystore for `yarn generate` and also for all scripts run inside `script/deploy.s.sol`.
keystore is saved to '~/.foundry/keystores/scaffold-eth-default'

```bash
yarn account:generate // generates a random wallet and saves to keystore
```

```bash
yarn account:import // uses prompt to ask for private key and password
```
## Additional Information


packages/foundry/package.json
```diff
@@ -4,18 +4,20 @@
...
+    "deploy": "yarn check-keystore && forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --account scaffold-eth-default --rpc-url ${1:-default_network} --broadcast --legacy && node scripts-js/generateTsAbis.js",
+    "deploy:verify": "yarn check-keystore && forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --account scaffold-eth-default --rpc-url ${1:-default_network} --broadcast --legacy --verify ; node scripts-js/generateTsAbis.js",
     "flatten": "forge flatten",
     "fork": "anvil --fork-url ${0:-mainnet} --chain-id 31337 --config-out localhost.json",
     "format": "forge fmt && prettier --write ./script/**/*.js",
-    "generate": "node scripts-js/generateAccount.js",
+    "account:generate": "shx rm ~/.foundry/keystores/scaffold-eth-default 2>/dev/null ; cast wallet import scaffold-eth-default --private-key $(cast wallet new | grep 'Private key:' | awk '{print $3}')",
+    "account:import" : "shx rm ~/.foundry/keystores/scaffold-eth-default 2>/dev/null ; cast wallet import scaffold-eth-default --interactive",
```
rin-st, line 15: redundant space after `"account:import"`

rin-st:
Hey @jrcarlos2000 ! Overall working great!

But I noticed some things we possibly need to change:
1. We need to add information to readme that after `yarn:deploy` we need to enter empty password
2. I think it's better to use another keystore for `account:generate`. Let's say I want to deploy my contract to sepolia. I run `account:generate` and then deploy it. Then I want to change contract, so I work locally again, and when `yarn chain` my generated sepolia account is gone with all my sepolia eth. So I need to save private key of sepolia account somewhere or generate a new one. But as I understand if we will use another keystore this problem disappears.
3. `yarn account` doesn't work
4. I'm getting this warning all the time, even when deploying to sepolia
```
Warning (9302): Return value of low-level calls not used.
  --> script/DeployHelpers.s.sol:33:9:
   |
33 |         deployer.call{ value: SCAFFOLD_BASE_BALANCE }("");
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[⠆] Compiling...
```
Can we get rid of it?

jrcarlos2000:
thanks @rin-st , yes still testing out what would be the best Dev Ex here before coming up with an updated README.
> I think it's better to use another keystore for account:generate. Let's say I want to deploy my contract to sepolia. I run account:generate and then deploy it. Then I want to change contract, so I work locally again, and when yarn chain my generated sepolia account is gone with all my sepolia eth. So I need to save private key of sepolia account somewhere or generate a new one. But as I understand if we will use another keystore this problem disappears.

so for this, we can do a pre-check, only generates a wallet after `yarn chain`, when there is no existing keystore.

>  I'm getting this warning all the time on deploy, even when deploying to sepolia

yes, will get rid of this

jrcarlos2000:
update : i have made a change:

when you do `yarn account:generate` or `yarn account:import` your keystore will be named `scaffold-eth-custom`, users will be able to set this on `.env` , please check the `.env.example`. so they can either used the anvil account ( scaffold-eth-default) or the imported / generated account ( at scaffold-eth-custom ) 

we also allow them to give it a name if provided. 

`yarn account:generate --name KEYSTORE_NAME ` , will need to update on `.env`

`yarn account:import --name KEYSTORE_NAME ` , will need to update on `.env`

KEYSTORE_NAME defaults to `scaffold-eth-custom`

rin-st:
Thanks for the quick fixes! 🔥 

Regarding `yarn account`. Could you take address from `scaffold-eth-custom` keystore, not as parameter? It worked without parameters before and also we don't need to save address before passing to `yarn account`

packages/foundry/package.json
```diff
@@ -3,19 +3,22 @@
   "version": "0.0.1",
   "type": "module",
   "scripts": {
-    "account": "node scripts-js/ListAccount.js",
-    "chain": "anvil --config-out localhost.json",
+    "verify-keystore": "cast wallet address 2>/dev/null ||  echo 'Keystore verified' && exit 1",
```
rin-st, line 6: if there's an error in `verify-keystore` script, it echoes 'Keystore verified', so 'Keystore' is `process.argv[2]`. And terminal shows this

```
yarn account                                           
🔍 Checking account: Keystore
🚫️ Invalid account, please run `yarn account:generate` or `yarn account:import`
```

Maybe allow it to just throw an error? Or show `🔍 Checking account` log only if account is valid

packages/foundry/script/DeployHelpers.s.sol
```diff
@@ -15,20 +15,38 @@ contract ScaffoldETHDeploy is Script {
...
+        if (!success) {
+          revert();
+        }
+        vm.stopBroadcast();
+      }
+      vm.startBroadcast(deployer);
     } else {
-      return vm.envUint("DEPLOYER_PRIVATE_KEY");
+      vm.startBroadcast();
+      (, deployer,) = vm.readCallers();
```
rin-st, line 42: same lines at start of `if` and `else`. Can be moved upper `if/else`

packages/foundry/package.json
jrcarlos2000: yes, if it is successfull it will go through, the problem is that this is an interactive setup, for now is the best solution i could get. it will show you an error if it fails and a good logging telling you what to do `yarn account:generate` . if the echo is not used, the next command wont run

jrcarlos2000:
> Thanks for the quick fixes! 🔥
> 
> Regarding `yarn account`. Could you take address from `scaffold-eth-custom` keystore, not as parameter? It worked without parameters before and also we don't need to save address before passing to `yarn account`

yes this will be taken from the .env , it will be set to `scaffold-eth-default` first and `scaffold-eth-custom` when user has generated a wallet. However you still need to unlock it. we still need the password to decrypt the keystore.

cc: @technophile-04

packages/foundry/package.json
jrcarlos2000: i have updated this script, it should be better now @rin-st

rin-st: Thanks! Now it doesn't log `🔍 Checking account` line on error

technophile-04:
After running `yarn chain` and then `yarn deploy` it gives: 

image

packages/foundry/script/DeployHelpers.s.sol
```diff
@@ -15,20 +15,37 @@ contract ScaffoldETHDeploy is Script {
...
-      string memory mnemonic = abi.decode(mnemonicBytes, (string));
-      return vm.deriveKey(mnemonic, 0);
-    } else {
-      return vm.envUint("DEPLOYER_PRIVATE_KEY");
+      // starts a broadcast and reads the caller
+      vm.stopBroadcast();
+      // check balance of first anvil account
+      // if balance is not 0, is the first run
+      uint256 balance = ANVIL_FIRST_ACCOUNT.balance;
+      if (balance > 1 ether && deployer != vm.addr(ANVIL_FIRST_PK)) {
```
technophile-04, line 33: Maybe we could update this condition as 

`if(balance == 0 && vm.addr(ANVIL_FIRST_PK))`

and while sending eth maybe just send 500 ETH?, Also lets declare 500 ETH as constant above so if anyone want to change he can update it

packages/foundry/script/DeployHelpers.s.sol
```diff
@@ -15,20 +15,37 @@ contract ScaffoldETHDeploy is Script {
   string root;
   string path;
   Deployment[] public deployments;
+  uint256 constant ANVIL_FIRST_PK =
+    0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80;
+  address constant ANVIL_FIRST_ACCOUNT =
+    0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266;
```
technophile-04, line 21: Maye instead of using anvils first account to send funds maybe using anvil's last account make more sense? 

Because in SE-2 frontend on localhost anvil first account behaves as faucet to send fund to burner wallet. 

So I think maybe we should use last account just to be safe

packages/foundry/script/DeployHelpers.s.sol
jrcarlos2000: i have simplified the logic by using only `balance == 0` because if your account is the `LAST_ANVIL_ACCOUNT` , it is supposed to have funds, otherwise everything else ( even the local faucet on script would fail )

technophile-04:
Tysm Carlos for this!!

As I understand `ETH_KEYSTORE_ACCOUNT` basically denotes which keystore to use.

SE-2 basically will be interacting with two keystore named:

1. scaffold-eth-default => This keystore uses anvil's 0th account private key, password is set to empty string `''`
2. scaffold-eth-custom => This will be generated when you run `yarn account:generate`, users sets the password.

I kinda like how we are getting account of keystore from `.env`. This means if I have created an keystore account "my-account", I could just add it in `.env#ETH_KEYSTORE_ACCOUNT=my-account` and everything should work in context of `my-account` address as deployer / sending transaction and also `yarn account` will list details of `my-account` address.

^ So if above is true(it seems to be, on my testing) then this makes `yarn account:import` kind of redundant? If people want to use their preferred account with PK then they can just update `.env#ETH_KEYSTORE_ACCOUNT` with their account?

### Difference b/w current SE-2 flow and keystore flow is:

1. Whenever you run `yarn deploy`(on localhost) you need to hit "Enter"
2. After you run `yarn account:generate`, you have to manually update `.env#ETH_KEYSTORE_ACCOUNT` with `scaffold-eth-custom` to use account generated by you to deploy contract.
3. If you do `yarn deploy`(on localhost) after pt.2, user has to enter `scaffold-eth-custom` password instead of hitting enter since now we will be using `scaffold-eth-custom` account always.

### Some DX improvements:

1. Considering point 2. and 3, I think we should show the user whats the current account/address being used at that time when he is entering the password(Because I was kind of confused initally,  should I hit "Enter" or use my generated account password without looking at `.env`)

2. (optional): Umm thinking out loud, with keystore it feels like we should have single deploy account for all SE-2 instance. Currently(in this branch) it removes the previous instance `scaffold-eth-custom` from keystore and generates a new one.
    - Imagninary flow:
        1. Users run `yarn generate` => we check if `scaffold-eth-custom` is already present in keystore
        2. If not, we create `scaffold-eth-cusom` with random address.
        3. If its present, then in CLI we directly show the "Using address/account 0x23423423", and then "Enter the password"
        4. There could be an another script like `yarn account:dangerous-re-generate` (this will remove `scaffold-eth-custom`) and generate new one


But yeah would love to know others thought as well!

packages/foundry/script/DeployHelpers.s.sol
```diff
@@ -15,20 +19,42 @@ contract ScaffoldETHDeploy is Script {
...
+      vm.startBroadcast(ANVIL_LAST_PK);
+      (bool success,) = deployer.call{ value: ANVIL_BASE_BALANCE / 2 }("");
+      if (!success) {
+        revert DeployerHasNoBalance();
+      } else {
+        try this.anvil_setBalance(ANVIL_LAST_ACCOUNT, ANVIL_BASE_BALANCE) {
+          emit AnvilSetBalance(ANVIL_LAST_ACCOUNT, ANVIL_BASE_BALANCE);
+        } catch {
+          revert FailedAnvilRequest();
+        }
```
technophile-04, line 45: question: Since we are setting the balance here directly, can't we just directly set the balance of `deployer` here? 

Instead of first sending fund from anvil last account to deployer then resetting the balance of anvil last account?

packages/foundry/script/DeployHelpers.s.sol
```diff
@@ -61,6 +87,30 @@ contract ScaffoldETHDeploy is Script {
...
+    inputs[2] = "POST";
+    inputs[3] = "http://localhost:8545";
+    inputs[4] = "-H";
+    inputs[5] = "Content-Type: application/json";
+    inputs[6] = "--data";
+    inputs[7] = requestPayload;
+
+    vm.ffi(inputs);
+  }
+
```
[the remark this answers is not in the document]
jrcarlos2000: this would work on testing, this is a script running on a network, which is localhost / anvil

packages/foundry/script/DeployHelpers.s.sol
jrcarlos2000: i tried doing this first and it seems that the deployer will get the balance only after the a transaction is run , i.e. if we do `yarn deploy` again. This is not ideal because we want to deploy on this transaction using the deployer account.

packages/foundry/script/DeployHelpers.s.sol
```diff
@@ -15,20 +19,42 @@ contract ScaffoldETHDeploy is Script {
...
+    0x2a871d0798f97d79848a013d4936a73bf4cc922c825d33c1cf7073dff6d409c6;
+  address constant ANVIL_LAST_ACCOUNT =
+    0xa0Ee7A142d267C1f36714E4a8F75612F20a79720;
+  uint256 constant ANVIL_BASE_BALANCE = 1000 ether;
+
+  function _startBroadcast() internal returns (address deployer) {
+    vm.startBroadcast();
+    (, deployer,) = vm.readCallers();
+
+    if (block.chainid == 31337 && deployer.balance) {
```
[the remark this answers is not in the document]
jrcarlos2000: some type went through, i have updated the `ANVIL_BASE_BALANCE` it was 10k eth

technophile-04:
Stressing on the point, in my thinking keystore kind of changes how we have been preaching to deploy contracts to testnet / mainnet.

Let's suppose I have 2 SE-2 instance (DAO-1 and NFT-proj-2).

Previously each of the project had their respective deployer key independent of each other. Example running `yarn generate` in `NFT-proj-2` won't affect DAO-1 project's pk at all.

With keystore since we have only one global namespace(`sacffold-eth-custom`), running `yarn generate` in `NFT-proj-2` will override the pk of `DAO-1` project(loosing the deployer account of `DAO-1` project)

So, Ideally I think `yarn generate` should tell the user do you want to override previous account or not?(if its second time). If its very first time we create an account.

Similarly to [here](https://arc.net/l/quote/nsbgdulu), we could mention if you want to use your personal keystore account, you could update the `.env#ETH_KEYSTORE_ACCOUNT` with your preffered account name.


> 2. the yarn account:import is a helper function for a non trivial operation
cast wallet import ${1:-scaffold-eth-custom} --interactive // here is worth mentioning that the user can select the name if they pass --name {ACCOUNT_NAME} and use it later on the .env
if i want to put my current pk into a keystore i would need to use this command anyways ( or something similar ) this is interactive which is good!

Umm I kind of confused and like 51% **not** to add `yarn account:import` and 49% to add it. kj

The reason : 
1. It add volume to scripts in package.json, if we didn't have it then the options could be just `yarn generate` (to generate an account), `yarn acount` to list account details (same as current SE-2)
2. Its just an abstractions over `cast wallet import ACCOUNT_NAME --interactive`, also foundry people seems to prefer running their own commands (for beginners we could have a line about in the docs) 
3. We don't have to do any magic, like removing `scaffold-eth-custm` etc. 

> 3. what we can add is setting the custom or "named" account to be the one that .env points to, after the user calls yarn generate or yarn account:import . I wouldnt prefer this imo

Lol confused on this too. Both option makes sense to me but yeah telling people to manually update `.env` with  `scaffold-eth-custom`/ their prefered account.  

Also @jrcarlos2000 can we achieve point 1. of "Some DX improvements" from https://github.com/scaffold-eth/scaffold-eth-2/pull/878#issuecomment-2220397549, I think it should be just adding `echo 'using $ETH_KEYSTORE_ACCOUNT'` before everything?

jrcarlos2000:
> Its just an abstractions over cast wallet import ACCOUNT_NAME --interactive, also foundry people seems to prefer running their own commands (for beginners we could have a line about in the docs)

I agree with you @technophile-04 , for DX 1, 2 we can achieve it. keeping only one single `scaffold-eth` keystore account

Lets see other thoughts !

rin-st:
> ### Some DX improvements:
> 1. Considering point 2. and 3, I think we should show the user whats the current account/address being used at that time when he is entering the password(Because I was kind of confused initally,  should I hit "Enter" or use my generated account password without looking at `.env`)
> 2. (optional): Umm thinking out loud, with keystore it feels like we should have single(testnet/mainnet) deploy account for all SE-2 instance. Currently(in this branch) it removes the previous instance `scaffold-eth-custom` from keystore and generates a new one.
>    
>    * Imagninary flow:
>      
>      1. Users run `yarn generate` => we check if `scaffold-eth-custom` is already present in keystore
>      2. If not, we create `scaffold-eth-cusom` with random address.
>      3. If its present, then in CLI we directly show the "Using address/account 0x23423423", and then "Enter the password"
>      4. There could be an another script like `yarn account:dangerous-re-generate` (this will remove `scaffold-eth-custom`) and generate new one
> 
> But yeah would love to know others thought as well!



Agree, good points!

> what we can add is setting the custom or "named" account to be the one that .env points to, after the user calls yarn generate or yarn account:import . I wouldnt prefer this IMO

I wouldn't prefer this too

> With keystore since we have only one global namespace

hm, I didn't take it into account!

> So, Ideally I think yarn generate should tell the user do you want to override previous account or not?(if its second time). If its very first time we create an account.
> Similarly to [here](https://arc.net/l/quote/nsbgdulu), we could mention if you want to use your personal keystore account, you could update the .env#ETH_KEYSTORE_ACCOUNT with your preffered account name.

+++

> Umm I kind of confused and like 51% not to add yarn account:import and 49% to add it.
> 2. Its just an abstractions over cast wallet import ACCOUNT_NAME --interactive, also foundry people seems to prefer running their own commands (for beginners we could have a line about in the docs)

Since I'm not a foundry pro, I'd prefer to leave it. It's easier to check `package.json` than find information in the docs. We can always remove it if not needed

> 3. We don't have to do any magic, like removing scaffold-eth-custm etc.

As an option we can make a name required for `account:import`, so it won't rewrite `scaffold-eth-custom` by default

portdeveloper:
Hey @jrcarlos2000 amazing work!

I have some questions. Please excuse my dum dumness if there is any.

So here is what it would look like for a normal user:
1. The user clones the repo, does a yarn install and foundry too
2.` yarn chain` generates a default keystore
3. the user then is able to run yarn deploy but she is asked a password that she needs to leave blank and click enter

My question here is:
- How hard is it to remove the empty pw submission? 
Maybe introduce a new command:
`yarn deploy:local` that would not ask a password and just yolo the contract into local.

Why I am wondering if this would work or not is that I believe that this empty pw thing will baffle lots of newbies. (It confused me at the beginning)

What I would suggest is to have a different command for local deployments. Since local deployments are entirely different from deployments on real chains, it makes sense to separate them.

`yarn deploy:local` would deploy contracts on the local anvil node.:

https://github.com/user-attachments/assets/81928915-4160-408e-b0e0-d3e151f830a4


`    "deploy:local": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${1:-default_network} --broadcast --legacy --ffi --password '' && node scripts-js/generateTsAbis.js",
`

The method above would ensure two things:
1. Ease of use for newbies: very verbose, `yarn deploy:local` deploys the contracts locally instead of `yarn deploy` deploying locally. 
2. Separation: now we have an entirely new command for local deployments, which is not possible to mix with the deploy we already have.

`yarn deploy` would still use the local node as a default value as a guard because there is no harm in deploying to a local node.

More on `yarn deploy`.

- Ask the user: chain to deploy
- Ask the user: keystore to use
- Show a confirmation (and maybe the deployer details) just before deployment

This would take a little bit of effort but it would be better in terms of UX

technophile-04:
@jrcarlos2000 is there a way we could in deploy script while doing `yarn deploy` that if the rpc url is `127:....` then we conditionally pass the `--password` with `''` string and if the resolved `${1:default_network}` is different we don't pass the `--pasword` flag?

jrcarlos2000:
Updates: 

- Updated from latest version of branch `foundry`
- Added a `--password` flag, that uses a default password when on localhost , namely `localhost`, `yarn chain` && `yarn account` are also updated to follow this pattern.
- Users will have to pass a password for other accounts they import
```bash
## Usage
yarn account --password PASSWORD ## defaults to localhost if not passed
yarn deploy --network NETWORK --password PASSWORD ## defaults to localhost if not passed
```
- Unable to log the account being used, however if the decryption fails, it will log the corresponding account , users should be aware that the account they are using is the account they have already pointed to on their `.env`
```bash
Failed to decrypt keystore "/Users/carlosssramosss/.foundry/keystores/scaffold-eth-custom"
```
- Added `yarn:account:force-import` to force override on `scaffold-eth-custom` account. 

It looks cleaner right now @technophile-04

escottalexander:
Testing this out and when I run `yarn deploy` I get this error:
```
[elliott@fedora foundry-deploy-test]$ yarn deploy
[⠢] Compiling...
No files changed, compilation skipped
error: the following required arguments were not provided:
  --keystore <PATHS>

Usage: forge script --keystore <PATHS> --fork-url <URL> --password <PASSWORDS> --broadcast --legacy --ffi --etherscan-api-key <KEY> <PATH> [ARGS]...

For more information, try '--help'.
```

Prior to running I `yarn installed` and `yarn account:generate`. Did I miss a step?

technophile-04:
Thanks @escottalexander for testing it out! 

Could you please see if you have `.env` present in `pacakges/foundry/.env` and it should have this init `ETH_KEYSTORE_ACCOUNT=scaffold-eth-default` . 

You don't need to create the `.env` it should have been automatically created when you first clone and `yarn install` if it didn't that it might be a bug which we need to solve (could you please tell the OS you are using?) 

If `.env` was already present could you try upgrading foundry version. You  could just run 

```shell
foundryup
```

This should make all the foundry related stuff like anvil etc upto date and then could you try running `yarn deploy` again?

escottalexander:
Thanks @technophile-04 
Perhaps I did `yarn install` before being in the correct branch because I did have a .env but it did not have the ETH_KEYSTORE_ACCOUNT in it. Deleting the .env and performing a `yarn install` set up a .env with that variable and the deploy was successful. So hopefully just a user error on my part and not a rel issue. Thanks!

portdeveloper:
Upon seeing @escottalexander 's reply here, I tested it with a clean install of SE-2 and I got the same error.

After `yarn install`, I tried to run `yarn deploy`. And below is what I get:
`
...
[⠊] Compiling...
[⠑] Compiling 31 files with Solc 0.8.26
[⠒] Solc 0.8.26 finished in 2.56s
Compiler run successful!
error: the following required arguments were not provided:
  --keystore <PATHS>

Usage: forge script --keystore <PATHS> --fork-url <URL> --password <PASSWORDS> --broadcast --legacy --ffi <PATH> [ARGS]...

For more information, try '--help'.
make: *** [Makefile:16: deploy] Error 2
`
image

As @technophile-04 mentioned, `yarn install` should create a .env file in the directory shown above. (packages/foundry)

I tried to look around a bit but I couldn't figure out where the code that creates the .env file 


Info about my system:

`PS C:\Users\port> wsl cat /proc/version
Linux version 5.15.153.1-microsoft-standard-WSL2 (root@941d701f84f1) (gcc (GCC) 11.2.0, GNU ld (GNU Binutils) 2.37) #1 SMP Fri Mar 29 23:14:13 UTC 2024`

I run wsl2 ubuntu on a win11 EDU machine.

I have run foundryup to get foundry to the latest version.

technophile-04:
Just merged the makefile PR, would love a last couple of tests. 

```shell
git clone https://github.com/scaffold-eth/scaffold-eth-2.git foundry-se-2 --branch feat/keystore
```

Then 

```
cd foundry-se-2 && yarn install
```

then try SE-2 magic command `yarn chain`, `yarn deploy`, `yarn start` they should work out of box hopefully. 

Before deploying to testnet: 

```shell
yarn account:generate

#enter some password
```

Then update `.env#ETH_KEYSTORE_ACCOUNT=scaffold-eth-custom` . 

```shell
yarn account

# enter same password as above to unlock and see the balance
```

Send some eth to generate account on sepolia and then: 

```shell
yarn deploy --network sepolia

# enter same password which you used early
```

This should deploy the contract to sepolia with generated account. 

and lastly : 

```shell
yarn verify --network sepolia
```

This should verify the contract

escottalexander:
@technophile-04 `yarn account:generate` is successful but `yarn account fails`. I did update the .env to point to scaffold-eth-custom and I verified that the keystore exists in /.foundry/keystores. Let me know if you want me to try something to diagnose further.

```bash
[elliott@fedora foundry-se-2]$ yarn account:generate
Enter password: 
`scaffold-eth-custom` keystore was saved successfully. Address: 0x822bda2a647b1dba555d002964dba61c21d2b84c
Please update .env file with ETH_KEYSTORE_ACCOUNT=scaffold-eth-custom
[elliott@fedora foundry-se-2]$ yarn account
/bin/sh: line 1: source: .env: file not found
make[1]: *** [Makefile:30: verify-keystore] Error 1
🚫️ Unable to access keystore account scaffold-eth-custom

Please run `yarn account:generate` to generate deployer keystore account and update `ETH_KEYSTORE_ACCOUNT=scaffold-eth-custom` in `.env` file
```

technophile-04:
Thanks @escottalexander for tests! Not sure what caused that problem but pushed a commit so that we don't do sourcing of `.env` in make file. Hopefully it should work now could you please try it once again?

escottalexander:
@technophile-04 I just repeated the steps in a fresh download of the repo (and removed my existing scaffold-eth-custom keystore) and this time it is asking for my password which is good but it fails after I give it my password. I tried twice (recreated the keystore) being very careful to enter my password correctly.

I am running Fedora Linux 40.

technophile-04:
> @technophile-04 I just repeated the steps in a fresh download of the repo (and removed my existing scaffold-eth-custom keystore) and this time it is asking for my password which is good but it fails after I give it my password. I tried twice (recreated the keystore) being very careful to enter my password correctly.

Strange, not sure whats happening did you get error as previous one after you enter password? Could you paste the error? 

Also after you do `yarn account:generate` could you try just running plain command: 

```shell
cast wallet address --account scaffold-eth-custom

# after entering the password it should log you wallet address
```

escottalexander:
@technophile-04 This is the error I get:
```bash
[elliott@fedora foundry-se-2]$ yarn account
Enter keystore password:

🚫️ Unable to access keystore account scaffold-eth-custom

💡 If you haven't generated a deployer keystore account yet, please run `yarn account:generate`. Then update the `.env` file with `ETH_KEYSTORE_ACCOUNT=scaffold-eth-custom`
```

The simple cast command works. It prints the wallet address after I give it my password.

technophile-04:
> I don't know much about C Makefiles but it appears when I run yarn account it tries to get my address by running cast wallet address(via make verify-keystore). This command seems to fail when I try it by itself:

Ohh actually since we have `.env` configured in `pacakges/foundry` with `ETH_KEYSTORE_ACCOUNT` foundry picks that variable without us needing to explicitly mention  `scaffold-eth-custom` it get gets that from `.env` : 

image


Also was debugging this with @portdeveloper and I think we finally fixed the issue, seems like executing commands from makefile seems to pass argv at different position for different OS. But it should be good now and hopefully should work 🙌

portdeveloper:
I have just tested it all except the verify, will test it when etherscan is out of maintenance, and everything works great!
hyped up to see these changes in prod!!

jrcarlos2000:
Have tested this, everything still works perfectly for me on Mac OS @technophile-04 😁

escottalexander:
@technophile-04 @carletex Works great! No issues this time. I was able to follow all the steps you posted including deploying to sepolia and verifying the contract. Nice work!

portdeveloper:
Tested it and it all works amazing!!!

Though I want to note one thing here:
`yarn deploy --network sepolia `
deploys and verifies the contract

So we do not need to run:
`yarn verify --network sepolia`

Wanted to point this out to ask if this is the intended behavior or not. 

Is it better to make it separate or keep it as is for ease of use?

technophile-04:
> Wanted to point this out to ask if this is the intended behavior or not.
> Is it better to make it separate or keep it as is for ease of use?

Actually they are separate already, `yarn deploy --network sepolia` should only deploy the contract without verifying it. 

I think maybe you felt this way because you got "contract already verified" when you deployed "YourContract.sol"? I this is because etherscan is being smart

portdeveloper:
> > Wanted to point this out to ask if this is the intended behavior or not.
> > Is it better to make it separate or keep it as is for ease of use?
> 
> Actually they are separate already, `yarn deploy --network sepolia` should only deploy the contract without verifying it.
> 
> I think maybe you felt this way because you got "contract already verified" when you deployed "YourContract.sol"? I think this is because etherscan is being smart marking contracts already verified which have same bytecode?

That is right haha I should've tested it with a separate contract before adding that comment!

Now after testing with another contract, I saw that everything works as it should!
lesgoo

technophile-04:
Thanks to @portdeveloper for noticing! Actually fork was not woking, just fixed it at https://github.com/scaffold-eth/scaffold-eth-2/pull/878/commits/cbc3ac9f3fc849cb43993415b06460b507eaaa10

So now : 
By default `yarn fork` will fork mainnet.

and `yarn fork --network ${networkName}` will fork mentioned network name. Eg: `yarn fork --network sepolia` will fork sepolia

technophile-04:
Merging this into `foundry` branch 🙌 Thanks @jrcarlos2000 and also others for feedback/help!!

PR #885, opened 2024-07-12
feat: bundler module resolution

rin-st:
## Description

Change ts `moduleResolution` to `Bundler`

Context: https://github.com/scaffold-eth/create-eth/pull/74#discussion_r1675316747

Also added paths for abitype/viem/wagmi since with `Bundler` `abi.d.ts` is ignored

Note: Could you check is everything ok with types? My Vscode getting mad, it can show or not show errors for same code. (as you see build works fine)

## Additional Information

### DIFF
--- a/packages/nextjs/tsconfig.json
+++ b/packages/nextjs/tsconfig.json
@@ -15,10 +15,7 @@
     "jsx": "preserve",
     "incremental": true,
     "paths": {
-      "~~/*": ["./*"],
-      "abitype": ["./node_modules/abitype"],
-      "viem/node_modules/abitype": ["./node_modules/viem/node_modules/abitype"],
-      "wagmi/node_modules/abitype": ["./node_modules/wagmi/node_modules/abitype"]
+      "~~/*": ["./*"]
     },
     "plugins": [
       {
--- a/packages/nextjs/tsconfig.json
+++ b/packages/nextjs/tsconfig.json
@@ -9,13 +9,16 @@
     "noEmit": true,
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "Bundler",
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
     "incremental": true,
     "paths": {
-      "~~/*": ["./*"]
+      "~~/*": ["./*"],
+      "abitype": ["./node_modules/abitype"],
+      "viem/node_modules/abitype": ["./node_modules/viem/node_modules/abitype"],
+      "wagmi/node_modules/abitype": ["./node_modules/wagmi/node_modules/abitype"]
     },
     "plugins": [
       {
--- a/packages/nextjs/types/abitype/abi.d.ts
+++ b/packages/nextjs/types/abitype/abi.d.ts
@@ -1,23 +1,7 @@
 import "abitype";
-import "viem";
-import "wagmi";
 
 type AddressType = string;
 
 declare module "abitype" {
-  export interface Register {
-    AddressType: AddressType;
-  }
-}
-
-declare module "viem/node_modules/abitype" {
-  export interface Register {
-    AddressType: AddressType;
-  }
-}
-
-declare module "wagmi/node_modules/abitype" {
-  export interface Register {
-    AddressType: AddressType;
-  }
+  export type Address = AddressType;
 }
--- a/packages/nextjs/types/abitype/abi.d.ts
+++ b/packages/nextjs/types/abitype/abi.d.ts
@@ -3,5 +3,7 @@ import "abitype";
 type AddressType = string;
 
 declare module "abitype" {
-  export type Address = AddressType;
+  export interface Register {
+    AddressType: AddressType;
+  }
 }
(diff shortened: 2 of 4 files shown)


technophile-04:
Thanks @rin-st ! 

> Note: Could you check is everything ok with types? My Vscode getting mad, it can show or not show errors for same code. (as you see build works fine)

Yup it seems to work fine for me in both vscode and neovim. 

But when I update the typescript to 5.5.3, it seems to break again :( 

To reproduce: 

1. Up typescript version

```shell
yarn workspace @se-2/nextjs up typescript
```

2. Check for types: 

```shell
yarn next:check-types
```

Umm, we could merge this already but I am bit skeptical because above thing and also on this branch when I hover over "viem/node_modules/abitype" in `abi.d.ts` it doesn't seem to resolve the path: 

| this branch | main branch | 
| ------- | ----- | 
| Screenshot 2024-07-15 at 11 51 26 AM | Screenshot 2024-07-15 at 11 52 29 AM | 

See how on `main` it nicely resolves the whole path

rin-st:
I don't know why, but extending `interface Register` doesn't work for me with `moduleResolution: "Bundler"`. 

Possible solution is to rewrite `Address` type directly, not through `Register`. It works, but am I missing something? Do we really need to use `Register`. Also updated TS version and looks like everything works fine. [1528a71](https://github.com/scaffold-eth/scaffold-eth-2/pull/885/commits/1528a711ff2fd6edbd4344f5250baf6ca2e38884) . Of course I'll revert it if something is wrong

As I remember we added `string` because it will be easier to work with our components to newcomers, am I right? Another solution is to remove `abi.d.ts` and add some `as `0x${string}`, but we need to think if it really helps or not

packages/nextjs/tsconfig.json
```diff
@@ -9,13 +9,16 @@
...
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
     "incremental": true,
     "paths": {
-      "~~/*": ["./*"]
+      "~~/*": ["./*"],
+      "abitype": ["./node_modules/abitype"],
+      "viem/node_modules/abitype": ["./node_modules/viem/node_modules/abitype"],
+      "wagmi/node_modules/abitype": ["./node_modules/wagmi/node_modules/abitype"]
```
technophile-04, line 21: I think we can remove viem and wagmi mapping, from here now?

technophile-04 (approved):
> Possible solution is to rewrite Address type directly, not through Register. It works, but am I missing something? Do we really need to use Register. Also updated TS version and looks like everything works fine.

Actually it was suggested in abitype docs:  https://abitype.dev/config#overview

also while digging I found this in their [release logs](https://github.com/wevm/abitype/releases): 
Screenshot 2024-07-16 at 1 43 45 PM

Maybe this was the reason `Register` were not working? 

--- 

But your current approach seems to works nicely, also tried deploying to vercel and works nicely too.

I think with this approach the problem will reappear again in future when viem / wagmi internal abitype version deviates from our `abitype` version? But since we have locked the version, I think we can only encounter it while manually trying to update the viem / wagmi. 

We can merge this though since currently everything works and its required for https://github.com/scaffold-eth/create-eth-extensions/pull/11. Just cc @carletex once.

rin-st:
> Maybe this was the reason Register were not working?

Thanks Shiv, yes! 🙏  After update `abitype` to `1.0.5` problem disappeared, so I switched to `Register` again.
There were no errors but for being more confident I also updated `wagmi`/`viem` to latest versions

packages/nextjs/types/abitype/abi.d.ts
```diff
@@ -7,15 +7,3 @@ declare module "abitype" {
     AddressType: AddressType;
   }
 }
-
-declare module "viem/node_modules/abitype" {
```
rin-st, line 11: removed this since it worked with different abitype versions (1.0.0 in viem, 1.0.5 in our package.json)

packages/nextjs/types/abitype/abi.d.ts
```diff
@@ -7,15 +7,3 @@ declare module "abitype" {
...
-  export interface Register {
-    AddressType: AddressType;
-  }
-}
-
-declare module "wagmi/node_moudles/abitype" {
```
rin-st, line 17: looks like we don't need to declare module for wagmi at all. No abitype in `wagmi`'s deps, and typo here (`node_moudles`) and everything worked fine

packages/nextjs/tsconfig.json
rin-st: Thanks! [6023ccb](https://github.com/scaffold-eth/scaffold-eth-2/pull/885/commits/6023ccb96d0371acc01ccb5945980992ca999441)

packages/nextjs/types/abitype/abi.d.ts
technophile-04: lol it was hiding all along until now!

technophile-04 (approved):
> Thanks Shiv, yes! 🙏 After update `abitype` to `1.0.5` problem disappeared, so I switched to `Register` again. There were no errors but for being more confident I also updated `wagmi`/`viem` to latest versions

Noicce! Yup also updating wagmi / viem make sense just to be sure.

All in all, I think everything looks great and works fine!! 

- We are using suggested way to update addressType using `Register`
- Deploys to vercel nicely : https://ts-address-type.vercel.app/
- Also we have switched to `moduleResolution: "Bundler"` 

Thanks Rinat!! Merging this!

PR #893, opened 2024-07-17
Fixes BigInt parsing losing precision in IntegerInput

iPaulPro:
## Description

When entering a large number into an `IntegerInput` component, it is parsed as a JSON string and loses precision.

For example, entering `85604566775863511922702956178915374775782418036268384638508224920020197708340` is parsed into `85604566775863516368369806302131844469805175632226226214501379502487210295296`.

This is fixed by first attempting to convert the `string` value to a `BigInt` and then returning that value if successful.

## Additional Information

Your ENS/address: paulburke.eth

### DIFF
--- a/packages/nextjs/app/debug/_components/contract/utilsContract.tsx
+++ b/packages/nextjs/app/debug/_components/contract/utilsContract.tsx
@@ -21,6 +21,11 @@ const isJsonString = (str: string) => {
 // Recursive function to deeply parse JSON strings, correctly handling nested arrays and encoded JSON strings
 const deepParseValues = (value: any): any => {
   if (typeof value === "string") {
+    try {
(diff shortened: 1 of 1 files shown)


technophile-04 (approved):
Hey thanks @iPaulPro for reporting and tackling it!! 

Just pushed a commit at https://github.com/scaffold-eth/scaffold-eth-2/pull/893/commits/6c601e2e04087aad6c7975ad95d09c0cbcc4a3dc, because earlier if we pass empty string they were parsed as 0n. 


https://github.com/user-attachments/assets/4ba015a9-3a84-427c-b705-44ab1e937194

iPaulPro:
@technophile-04 Great catch and thanks for the incredibly fast merge!

PR #894, opened 2024-07-17
fix: inputs parsing numbers instead of address

technophile-04:
### Description :

Actually this was caused due to #893. 

When people try to enter address is converted to BigInt. 

https://github.com/user-attachments/assets/5cda53c3-8e16-4cdf-b5e1-c23cf70f1e03

### DIFF
--- a/packages/nextjs/app/debug/_components/contract/utilsContract.tsx
+++ b/packages/nextjs/app/debug/_components/contract/utilsContract.tsx
@@ -19,7 +19,7 @@ const isJsonString = (str: string) => {
 };
 
 const isBigInt = (str: string) => {
-  if (str.trim().length === 0) return false;
+  if (str.trim().length === 0 || str.startsWith("0")) return false;
   try {
     BigInt(str);
     return true;


packages/nextjs/app/debug/_components/contract/utilsContract.tsx
```diff
@@ -19,7 +19,7 @@ const isJsonString = (str: string) => {
 };
 
 const isBigInt = (str: string) => {
-  if (str.trim().length === 0) return false;
+  if (str.trim().length === 0 || str.startsWith("0")) return false;
```
technophile-04, line 22: Umm not sure whats the best way, but this seems to work fine. 

The reason I am not doing `startsWith("0x")` and used only "0", because if you keep `startsWith("0x")` then this happens with string starting with 0, notice how it ignores 0 in input string


https://github.com/user-attachments/assets/958346f5-acda-475a-b15a-84917c437f57

rin-st (approved):
Lgtm!

technophile-04:
Thanks merging this for now as a patch, maybe we could find a more robust soultuion in future

PR #781, opened 2024-03-21
Deployment and verification abstraction

Hotmanics:
## Rationale

In a personal project, I faced a challenge and experienced confusion and errors when I needed more than just the `Deploy.s.sol` script, and made a `DeployDemo.s.sol` script. However there were a large number of changes to my project that I needed to do. 

# Original Process:
1. Copy and rename the yarn commands `deploy`, `deploy:verify`, and `verify` yarn commands to `deployDemo`, `deployDemo:verify`, and `verifyDeployDemo`.
2. Update the references in the the new yarn commands from `Deploy.s.sol` to `DeployDemo.s.sol`.
3. Duplicate and rename the file `generateTsAbis.ts` to `generateTsAbis_Demo.ts`.
4. Duplicate and rename the file `VerifyAll.s.sol` to `VerifyAll_Demo.s.sol`.
5. Update the references in `generateTsAbi_Demo.ts` from `Deploy.s.sol` to `DeployDemo.s.sol`.
6. Update the references in `VerifyAll_Demo.ts` from `Deploy.s.sol` to `DeployDemo.s.sol`.
7. Return to the new yarn commands and change all references of `generateTsAbis.ts` to `generateTsAbis_Demo.ts` and `VerifyAll.s.sol` to `VerifyAll_Demo.s.sol`. 

# New Process
1. Copy and rename the yarn commands `deploy`, `deploy:verify`, and `verify` yarn commands to `deployDemo`, `deployDemo:verify`, and `verifyDeployDemo`.
2. Update the references in the the new yarn commands from `Deploy.s.sol` to `DeployDemo.s.sol`.

## Description

Simplifies the process of creating and running additional deploy scripts and verifications. 

## Additional Information

- [x] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [x] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

Your ENS/address: JacobHomanics.eth

PR #898, opened 2024-07-23
foundry: separate out your contract deployments

technophile-04:
### Description: 

Separate your contract deploy script from so that we enforce people to follow a pattern. 

The reason we need to enforce this pattern is because `exportDeployments()` should be called only once and too after all the deployments are done. 

So basically `Deploy.s.sol` file is the main file which import different deploy script of different contracts and in the end it calls the `exportDeployments()` which creates all the necessary files for generateTsAbi.js fucntion. 

Checkout second part of this for more details : https://github.com/scaffold-eth/scaffold-eth-2/pull/781#issuecomment-2018642266

CC @jrcarlos2000 feel free suggest some cleanups / better practices in foundry. 

### Some notes: 

- Kindof similar patter followed in this twitter thread https://x.com/jj_ranalli/status/1656384664020434949
- Foundry discussion / issue having better deployment mangement flow : https://github.com/foundry-rs/foundry/issues/4732
- [forge-deploy](https://github.com/wighawag/forge-deploy) from the creator of `hardhat-deploy` (which we are currently using for better dev flow in hardhat) but forge-deploy seems to add extra steps / commands? (saying this without trying, and something we need to maybe tinker in future PR's)

### DIFF
--- a/packages/foundry/script/Deploy.s.sol
+++ b/packages/foundry/script/Deploy.s.sol
@@ -6,27 +6,30 @@ import "./DeployHelpers.s.sol";
 import { DeployYourContract } from "./00_deploy_your_contract.s.sol";
 
 contract DeployScript is ScaffoldETHDeploy {
+  uint256 deployerPrivateKey;
+
   error InvalidPrivateKey(string);
 
-  function run() external {
-    // check for the deployer account
-    uint256 deployerPrivateKey = setupLocalhostEnv();
+  constructor() {
+    deployerPrivateKey = setupLocalhostEnv();
+  }
+
+  function run() external checkDeployerPrivateKey deploymentExporter {
+    DeployYourContract deployYourContract = new DeployYourContract();
+    deployYourContract.run();
+  }
+
+  modifier checkDeployerPrivateKey() {
     if (deployerPrivateKey == 0) {
       revert InvalidPrivateKey(
         "You don't have a deployer account. Make sure you have set DEPLOYER_PRIVATE_KEY in .env or use `yarn generate` to generate a new random account"
       );
     }
-
-    DeployYourContract deployYourContract = new DeployYourContract();
-    deployYourContract.run();
-
-    /**
-     * This function generates the file containing the contracts Abi definitions.
-     * These definitions are used to derive the types needed in the custom scaffold-eth hooks, for example.
-     * This function should be called last.
-     */
-    exportDeployments();
+    _;
   }
 
-  function test() public { }
+  modifier deploymentExporter() {
+    _;
+    exportDeployments();
+  }
 }
--- a/packages/foundry/script/Deploy.s.sol
+++ b/packages/foundry/script/Deploy.s.sol
@@ -17,6 +17,10 @@ contract DeployScript is ScaffoldETHDeploy {
   function run() external ScaffoldEthDeployerRunner {
     DeployYourContract deployYourContract = new DeployYourContract();
     deployYourContract.run();
+
+    // deploy more contracts here
+    // DeployMyContract deployMyContract = new DeployMyContract();
+    // deployMyContract.run();
   }
 
   modifier ScaffoldEthDeployerRunner() {
--- a/packages/foundry/script/Deploy.s.sol
+++ b/packages/foundry/script/Deploy.s.sol
@@ -3,27 +3,22 @@ pragma solidity ^0.8.19;
 
 import "../contracts/YourContract.sol";
 import "./DeployHelpers.s.sol";
+import { DeployYourContract } from "./00_deploy_your_contract.s.sol";
 
 contract DeployScript is ScaffoldETHDeploy {
   error InvalidPrivateKey(string);
 
   function run() external {
+    // check for the deployer account
     uint256 deployerPrivateKey = setupLocalhostEnv();
     if (deployerPrivateKey == 0) {
       revert InvalidPrivateKey(
         "You don't have a deployer account. Make sure you have set DEPLOYER_PRIVATE_KEY in .env or use `yarn generate` to generate a new random account"
       );
     }
-    vm.startBroadcast(deployerPrivateKey);
 
-    YourContract yourContract = new YourContract(vm.addr(deployerPrivateKey));
-    console.logString(
-      string.concat(
-        "YourContract deployed at: ", vm.toString(address(yourContract))
-      )
-    );
-
-    vm.stopBroadcast();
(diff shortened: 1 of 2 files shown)


0xquantum3labs:
```solidity
//SPDX-License-Identifier: MIT
pragma solidity ^0.8.19;

import { YourContract } from "../contracts/YourContract.sol";
import { ScaffoldETHDeploy, console } from "./DeployHelpers.s.sol";

contract DeployScript is ScaffoldETHDeploy {
  error InvalidPrivateKey(string);

  modifier deploymentExporter() {
    _;
    exportDeployments();
  }

  function run() external deploymentExporter {
    address deployer = _startBroadcast();
    YourContract yourContract = new YourContract(deployer);
    console.logString(
      string.concat(
        "YourContract deployed at: ", vm.toString(address(yourContract))
      )
    );
    _stopBroadcast();
  }

  function test() public { }
}
```

we can avoid having users explicitly calling `exportDeployments()` by using this modifier which will execute after function `run()`

jrcarlos2000:
```solidity
//SPDX-License-Identifier: MIT
pragma solidity ^0.8.19;

import { YourContract } from "../contracts/YourContract.sol";
import { ScaffoldETHDeploy, console } from "./DeployHelpers.s.sol";

contract DeployScript is ScaffoldETHDeploy {
  error InvalidPrivateKey(string);

  modifier deploymentExporter() {
    _;
    exportDeployments();
  }

  function run() external deploymentExporter {
    address deployer = _startBroadcast();
    YourContract yourContract = new YourContract(deployer);
    console.logString(
      string.concat(
        "YourContract deployed at: ", vm.toString(address(yourContract))
      )
    );
    _stopBroadcast();
  }

  function test() public { }
}
```

we can avoid having users explicitly calling `exportDeployments()` by using this modifier which will execute after function `run()`

packages/foundry/script/Deploy.s.sol
```diff
@@ -3,35 +3,33 @@ pragma solidity ^0.8.19;
...
+
   error InvalidPrivateKey(string);
 
-  function run() external {
-    uint256 deployerPrivateKey = setupLocalhostEnv();
+  constructor() {
+    deployerPrivateKey = setupLocalhostEnv();
+  }
+
+  function run() external checkDeployerPrivateKey deploymentExporter {
```
0xquantum3labs, line 17: i think we can have one single modifier `ScaffoldEthDeployerRunner` that does :
```solidity
checkDeployerPrivateKey();
_;
exportDeployments();
```

instead of having two, this is less error prone, in case users forget to add one of them. @technophile-04

jrcarlos2000, line 17: we can have one single modifier `ScaffoldEthDeployerRunner` that does :
```solidity
modifier ScaffoldEthDeployerRunner() {
     checkDeployerPrivateKey(); // function here
     _;
     exportDeployments();
}
```

instead of having two, this is less error prone, in case users forget to add one of them. @technophile-04

jrcarlos2000:
another way could be having `abstract` contract for each deploy script.
we can share the `deployments` array for `named` contracts, for example when deploying multiple instances of a same contract so that we can export all of them with different names. This is currently available and is useful, hardhat used to have a similar functionality. 

- Deploy script 1

```solidity
//SPDX-License-Identifier: MIT
pragma solidity ^0.8.19;

import "../contracts/YourContract.sol";
import "./DeployHelpers.s.sol";

abstract contract DeployYourContract is ScaffoldETHDeploy {
  function run() public virtual {
    uint256 deployerPrivateKey = setupLocalhostEnv();
    vm.startBroadcast(deployerPrivateKey);

    YourContract yourContract = new YourContract(vm.addr(deployerPrivateKey));
    deployments.push(Deployment("YourContract0", address(yourContract)));
    console.logString(
      string.concat(
        "YourContract deployed at: ", vm.toString(address(yourContract))
      )
    );

    vm.stopBroadcast();
  }
}
```

- Deploy script 2
```solidity
//SPDX-License-Identifier: MIT
pragma solidity ^0.8.19;

import "../contracts/YourContract.sol";
import "./DeployHelpers.s.sol";

abstract contract DeployYourContract1 is ScaffoldETHDeploy {
  function run() public virtual{
    uint256 deployerPrivateKey = setupLocalhostEnv();
    vm.startBroadcast(deployerPrivateKey);

    YourContract yourContract = new YourContract(vm.addr(deployerPrivateKey));
    deployments.push(Deployment("YourContract1", address(yourContract)));
    console.logString(
      string.concat(
        "YourContract deployed at: ", vm.toString(address(yourContract))
      )
    );

    vm.stopBroadcast();
  }
}
```


- main deploy script 

```solidity
//SPDX-License-Identifier: MIT
pragma solidity ^0.8.19;

import { DeployYourContract } from "./00_deploy_your_contract.s.sol";
import { DeployYourContract1 } from "./01_deploy_your_contract.s.sol";

contract DeployScript is DeployYourContract, DeployYourContract1 {
  uint256 deployerPrivateKey;

  error InvalidPrivateKey(string);

  modifier scaffoldEthDeployer() {
    deployerPrivateKey = setupLocalhostEnv();
    if (deployerPrivateKey == 0) {
      revert InvalidPrivateKey(
        "You don't have a deployer account. Make sure you have set DEPLOYER_PRIVATE_KEY in .env or use `yarn generate` to generate a new random account"
      );
    }
    _;
    exportDeployments();
  }

  function run()
    public
    override(DeployYourContract, DeployYourContract1)
    scaffoldEthDeployer
  {
    DeployYourContract.run();
    DeployYourContract1.run();
  }
}

```

jrcarlos2000:
> another way could be having abstract contract for each deploy script.
we can share the deployments array for named contracts, for example when deploying multiple instances of a same contract so that we can export all of them with different names. This is currently available and is useful, hardhat used to have a similar functionality.

The current approach doesnt support this shared state of `deployments` for `named` contracts @technophile-04

technophile-04:
Ohh I see, so with current approach people are not able to same contract(name) multiple times  it gets override by most recent one? 

I think we kind of the same problem with current setup of hardhat-deploy(where you are not able to keep track of contract with same names)

Yeah your approach makes sense! The only thing which I don't like is people in their own contract deploy script has to remember to do : 
```solidity
deployments.push(Deployment("YourContract1", address(yourContract)));
```

but yeah I am not sure what's other way to handle this either, something not that important for now but we should def tackle this in future! Thanks @jrcarlos2000 for all the suggestions!!

Issue #901, opened 2024-07-27
bug: ESLint Error


MattPereira:
### Which method was used to setup Scaffold-ETH 2 ?

git clone, npx create-eth@latest

### Current Behavior

Linter not working so i dont get squigles warning me of unused variables, imports, ect.

![image](https://github.com/user-attachments/assets/0dffd4a5-c00d-431a-8091-c51d9ea48b88)

I do still get blocked from committing and get warnings explained when the lint script runs

![image](https://github.com/user-attachments/assets/830e6025-f605-4648-bf7d-cbb5e632452f)

### Expected Behavior

Here is what it looks like on older version of scaffold when Linter works without error

![image](https://github.com/user-attachments/assets/8df655d2-f66a-4359-b930-54d6cddc5a50)

### Steps To Reproduce

Happens instantly with fresh clone and fresh npx create-eth@latest

### Anything else?

The ESLint error logs

```
[Info  - 1:46:49 PM] ESLint server is starting.
[Info  - 1:46:49 PM] ESLint server running in node v20.9.0
[Info  - 1:46:49 PM] ESLint server is running.
[Info  - 1:46:50 PM] ESLint library loaded from: /home/matthu/Desktop/test/packages/nextjs/node_modules/eslint/lib/api.js
Pages directory cannot be found at /home/matthu/Desktop/test/pages or /home/matthu/Desktop/test/src/pages. If using a custom path, please configure with the `no-html-link-for-pages` rule in your eslint config file.
[Error - 1:46:52 PM] An unexpected error occurred:
[Error - 1:46:52 PM] Error: Cannot find package '@trivago/prettier-plugin-sort-imports' imported from /home/matthu/Desktop/test/noop.js
    at new NodeError (file:///home/matthu/Desktop/test/packages/nextjs/node_modules/prettier/index.mjs:16425:5)
    at packageResolve (file:///home/matthu/Desktop/test/packages/nextjs/node_modules/prettier/index.mjs:17366:9)
    at moduleResolve (file:///home/matthu/Desktop/test/packages/nextjs/node_modules/prettier/index.mjs:17406:18)
    at defaultResolve (file:///home/matthu/Desktop/test/packages/nextjs/node_modules/prettier/index.mjs:17497:16)
    at resolve2 (file:///home/matthu/Desktop/test/packages/nextjs/node_modules/prettier/index.mjs:17514:12)
[...]

PR #875, opened 2024-06-28
up prettier

technophile-04:
### Description: 

Updated prettier version to latest version, because of https://github.com/scaffold-eth/create-eth/pull/55#issuecomment-2195411531. 

Also thought maybe its nice chance to up other prettier related things.

PR #905, opened 2024-08-09
revert #875

technophile-04:
## Description

Reverts the changes from #875. 

The reason for #875 was https://github.com/scaffold-eth/create-eth/pull/55#issuecomment-2195411531. Since in `create-eth` we are no more using prettier from outside to format files and instead using SE-2 instance prettier to format files we don't require this changes anymore. 

The reason for reverting this is #901.  

To see if this solve #901 : 

Copy paste the below in `page.tsx` insde Home component: 

<details>

<summary>
Test snippet : 
</summary>

```ts
  const [count, setCount] = useState(0);
  const [name, setName] = useState("shiv");
  // should get an eslint error
  const x = "shiv";

  useEffect(() => {
    setCount(count + 1);
    console.log(count);
    setName("shiv");
    console.log(name);
    // should get an eslint warning 
  }, [count]);

```

</details>

And hopefully you should get eslint warning / error at mentioned places in the comment.


MattPereira:
@technophile-04 you are my hero :heart_eyes:

image

rin-st:
Works great to me! And I agree that this solution is better because it's more stable! Thank you!

rin-st (approved):

technophile-04:
Thanks all! Merging this!

PR #912, opened 2024-08-20
foundry keystore makefile

technophile-04:
### Description: 

Switched to makefile. Also tweaked a flow a bit in latest changes of #878 while deploying to live network you had to pass the raw password in cli: 

```shell
yarn deploy --network sepolia --password my_password

# had to do same for yarn account 
```

Now we are using makefiles and in the new flow : 

```shell
yarn deploy
```
^ By default deploying to locahost won't ask you for password (already tackled in latest #878)

But when you pass it an live network which is not locahost then it will give you an interactive cli to type password.

Same for `yarn account`

cc @jrcarlos2000 to check once if I did not miss anything.


packages/foundry/package.json
```diff
@@ -3,22 +3,20 @@
...
-    "deploy": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${1:-localhost} --password ${3:-localhost} --broadcast --legacy --ffi && node scripts-js/generateTsAbis.js",
-    "deploy:verify": "forge build --build-info --build-info-path out/build-info/ && forge script script/Deploy.s.sol --rpc-url ${1:-localhost} --password ${3:-localhost} --broadcast --legacy --ffi --verify ; node scripts-js/generateTsAbis.js",
-    "flatten": "forge flatten",
-    "fork": "anvil --fork-url ${0:-mainnet} --chain-id 31337",
-    "format": "forge fmt && prettier --write ./script/**/*.js",
-    "account:generate": "cast wallet import ${1:-scaffold-eth-custom} --private-key $(cast wallet new | grep 'Private key:' | awk '{print $3}')",
-    "account:import": "cast wallet import ${1:-scaffold-eth-custom} --interactive",
-    "account:force-import": "shx rm ~/.foundry/keystores/scaffold-eth-custom 2>/dev/null; cast wallet import ${1:-scaffold-eth-custom} --interactive",
-    "lint": "forge fmt --check && prettier --check ./script/**/*.js",
-    "postinstall": "shx cp -n .env.example .env",
```
jrcarlos2000, line 19: missed the `postinstall`

packages/foundry/Makefile
```diff
@@ -0,0 +1,84 @@
...
+
+# List account
+account:
+	@node scripts-js/ListAccount.js $$(make verify-keystore)
+
+# Generate a new account
+account-generate:
+	@cast wallet import $${1:-scaffold-eth-custom} --private-key $$(cast wallet new | grep 'Private key:' | awk '{print $$3}')
+	@echo "Please update .env file with ETH_KEYSTORE_ACCOUNT=$${1:-scaffold-eth-custom}"
+
```
jrcarlos2000, line 41: I think missed the ACCOUNT_NAME from `package.json`
```suggestion
	@cast wallet import $(ACCOUNT_NAME) --private-key $$(cast wallet new | grep 'Private key:' | awk '{print $$3}')
	@echo "Please update .env file with ETH_KEYSTORE_ACCOUNT=$(ACCOUNT_NAME)"
```

packages/foundry/Makefile
```diff
@@ -0,0 +1,84 @@
...
+# Import an existing account
+account-import:
+	@cast wallet import $${1:-scaffold-eth-custom} --interactive
+
+
```
jrcarlos2000, line 46:
```suggestion
	@cast wallet import $(ACCOUNT_NAME) --interactive
```

packages/foundry/Makefile
```diff
@@ -0,0 +1,84 @@
...
+	node scripts-js/generateTsAbis.js
+
+# Verify keystore
+verify-keystore:
+	@cast wallet address 
+
+# List account
+account:
+	@node scripts-js/ListAccount.js $$(make verify-keystore)
+
```
jrcarlos2000, line 36:
```suggestion
verify-keystore:
	@set -a && source .env && \
	if [ "$$ETH_KEYSTORE_ACCOUNT" = "scaffold-eth-default" ]; then \
		cast wallet address --password localhost; \
	else \
		cast wallet address; \
	fi

# List account
account:
	@address=$$(make -s verify-keystore | tail -n 1) && \
	node scripts-js/ListAccount.js $$address
```

jrcarlos2000: I think since we are using a make file we can also default the password for `scaffold-eth-default` account.

packages/foundry/Makefile
```diff
@@ -0,0 +1,84 @@
...
+lint:
+	forge fmt --check && prettier --check ./script/**/*.js
+
+# Run tests
+test:
+	forge test
+
+# Verify contracts
+verify:
+	forge build --build-info --build-info-path out/build-info/ && \
```
jrcarlos2000, line 82: i think we already have a `build` make script above

technophile-04:
Tysm @jrcarlos2000! updated with all the changes! Merging this 🙌

PR #916, opened 2024-08-25
fix: useScaffoldEventHistory caching

### LINKED ISSUE #903: useScaffoldEventHistory uses data from old fetch with different filters

git clone

### Current Behavior

I use useScaffoldEventHistory twice on a single page with different filters. It fetches right data. I run again and the old data does not change, it use overrides the data with first run of useScaffoldEventHistory (even the constants have a differnt name). Any ideas??

This:

```
<PostCardList key="-1" filters={{ _post_hash: postHash }} />
<PostCardList key="-2" filters={{ _parent_post_hash: postHash }} />
```
and PostCardList.tsx:

```
export const PostCardList = ({ filters }: PostCardListProps) => {

const { data: postCreatedEvents, isLoading } = useScaffoldEventHistory({
    contractName: "PostCreator",
    eventName: "PostCreated",
    fromBlock: 0n,
    watch: true,
    filters: filters 
  });
...
```

But the second One gets overridden with contents of first one?? Both solidity parameters are indexed. And I also put a key inside my looped list items (unique hash)

### Expected Behavior

_No response_

### Steps To Reproduce

Use useScaffoldEventHistory on same page with different filters

### Anything else?

_No response_

rin-st:
## Description

Fixes useScaffoldEventHistory caching

Fixes #903

## Additional Information

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useState } from "react";
+import { useEffect, useState } from "react";
 import { useTargetNetwork } from "./useTargetNetwork";
 import { useInfiniteQuery } from "@tanstack/react-query";
 import { Abi, AbiEvent, ExtractAbiEventNames } from "abitype";
@@ -11,10 +11,6 @@ import {
   UseScaffoldEventHistoryConfig,
   UseScaffoldEventHistoryData,
 } from "~~/utils/scaffold-eth/contract";
-
-const getRandomId = () => {
-  return Math.random().toString(36).slice(2, 11);
-};
 
 const getEvents = async (
   getLogsParams: GetLogsParameters<AbiEvent | undefined, AbiEvent[] | undefined, boolean, BlockNumber, BlockNumber>,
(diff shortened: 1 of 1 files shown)


technophile-04 (commented):
Ohh I think the main underlying problem was I missed adding `filters` in `queryKey` : 

https://github.com/scaffold-eth/scaffold-eth-2/blob/a132cce62c1064bacff8dc1722115bd9200153f7/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts#L100-L108

Since we didn't had `filters` in query key `react-query` was treating both queries as same and hence not respecting the filters provided.

technophile-04:
Just pushed commits to updated here is the demo fixing [mrr](https://github.com/phaitonican/scuffold-eth-2-event-history-test) mentioned [here](https://github.com/scaffold-eth/scaffold-eth-2/issues/903#issuecomment-2285574189)


https://github.com/user-attachments/assets/62a5b9fc-8b60-4664-9d9f-8253761db43b

packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
```diff
@@ -105,6 +106,7 @@ export const useScaffoldEventHistory = <
         eventName,
         fromBlock: fromBlock.toString(),
         chainId: targetNetwork.id,
+        filters: JSON.stringify(filters, replacer),
```
technophile-04, line 109: We could have passed the `filters` obj directly but sometimes there can be bigInt init it too. When react-query make a queryHash of all this items it uses JSON.stringify internally and not able to parse bigInt. 

So here we are making the react-query's life easier by handling the object with bigInt earlier so its easier for it to create query hash

rin-st:
Can't approve my own PR 😄 . Thank you for update, lgtm!

technophile-04 (approved):
Merging this 🙌

PR #926, opened 2024-09-05
Reset enteredEns on Address input when changing the value

carletex:
There is a bug in the AddressInput component that shows when changing the value from the outside without triggering `onChange`.

To test:

1. Enter `atg.eth` => it resolves and shows the blockie, ensName, and avatar
2. Click the Change address button. Everything works except the ensName which remains the same.

```tsx
"use client";

import { useState } from "react";
import { useAccount } from "wagmi";
import { AddressInput } from "~~/components/scaffold-eth";

  const [inputAddress, setInputAddress] = useState("");
  return (
    <>
      <div className="flex items-center flex-col flex-grow pt-10 gap-2">
        <AddressInput value={inputAddress} onChange={setInputAddress} />
        <button
          className="btn btn-primary"
          onClick={() => setInputAddress("0x60583563d5879c2e59973e5718c7de2147971807")}
        >
          Change Address
        </button>
      </div>
    </>
  );
};

export default Home;
```

----

We found this while working on https://github.com/BuidlGuidl/buidlguidl.com/pull/33

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
+++ b/packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from "react";
+import { useEffect, useState } from "react";
 import { blo } from "blo";
 import { useDebounceValue } from "usehooks-ts";
 import { Address, isAddress } from "viem";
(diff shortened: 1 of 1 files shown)


packages/nextjs/components/scaffold-eth/Input/AddressInput.tsx
```diff
@@ -66,6 +66,10 @@ export const AddressInput = ({ value, name, placeholder, onChange, disabled }: C
     onChange(ensAddress);
   }, [ensAddress, onChange, debouncedValue]);
 
+  useEffect(() => {
+    setEnteredEnsName(undefined);
```
rin-st, line 70: since you added this `useEffect`, `handleChange` function can be deleted and use `onChange={onChange}` for `InputBase` props

carletex: oh yes!! great catch https://github.com/scaffold-eth/scaffold-eth-2/pull/926/commits/a090b57b94d17ea431794005cebbc4d802522944

rin-st (approved):
lgtm!

PR #932, opened 2024-09-10
update README with keystore details

technophile-04:
## Description

Updated the README to mention some details about keystore accounts so people using it will get a gist. 

Added "Deploying to live networks" but in future we should delete this section from here and update `docs.scaffoldeth.io` with it.


carletex (approved):
Great stuff @technophile-04 Thanks!

We can iterate if we need, but this looks great.

PR #933, opened 2024-09-10
Up next react sept 24

rin-st:
## Description

Step 2 of https://github.com/scaffold-eth/scaffold-eth-2/pull/917#issuecomment-2320138630

Decided to combine next/react with CSS libs or we will have 15 pull requests just to update dependencies

Plus
- updated `react-copy-to-clipboard` types since it's broken with newest react
- updated `vercel` version
- updated `viem`, `wagmi` and `rainbowkit` last patches

## Additional Information

- [x] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [x] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

PR #937, opened 2024-09-15
Shadcn rainbowkit button + dropdown

rin-st:
## Description

Custom rainbowkit button + dropdown implemented with shadcn

See https://github.com/scaffold-eth/scaffold-eth-2/pull/937#discussion_r1760213449 and https://github.com/scaffold-eth/scaffold-eth-2/pull/937#discussion_r1760199979 first

Buttons functionality

https://github.com/user-attachments/assets/7d0ba4f7-7508-46c0-8b5f-ccc67f7000f2

---
With networks switch

https://github.com/user-attachments/assets/4d8693e2-df21-49b0-896e-3e270bea4c4e

---
Wrong network dropdown

https://github.com/user-attachments/assets/77ec7cb6-0074-40ee-ab7a-869b1a1aa74d

---
Dark theme

https://github.com/user-attachments/assets/2fd32a10-eb2f-4d4c-b8d5-8376eb83feeb

### DIFF
--- a/packages/nextjs/components/ui/button.tsx
+++ b/packages/nextjs/components/ui/button.tsx
@@ -4,7 +4,7 @@ import { type VariantProps, cva } from "class-variance-authority";
 import { cn } from "~~/lib/utils";
 
 const buttonVariants = cva(
-  "inline-flex items-center justify-center whitespace-nowrap rounded-full text-sm font-medium transition-colors focus-visible:outline-none focus-visible:ring-0 focus-visible:ring-ring disabled:pointer-events-none disabled:opacity-50",
+  "inline-flex items-center justify-center whitespace-nowrap rounded-full text-sm font-medium transition-colors focus-visible:outline-none focus-visible:ring-1 !ring-accent focus-visible:ring-ring disabled:pointer-events-none disabled:opacity-50",
   {
     variants: {
       variant: {
--- a/packages/nextjs/components/ui/button.tsx
+++ b/packages/nextjs/components/ui/button.tsx
@@ -12,7 +12,7 @@ const buttonVariants = cva(
         destructive: "bg-destructive text-destructive-foreground shadow-sm hover:bg-destructive/90",
         outline: "border border-input bg-background shadow-sm hover:bg-accent hover:text-accent-foreground",
         secondary: "bg-secondary text-secondary-foreground shadow-sm hover:bg-secondary/80",
-        ghost: "hover:!bg-muted",
+        ghost: "hover:bg-accent hover:text-accent-foreground",
         link: "text-primary underline-offset-4 hover:underline",
       },
       size: {
--- a/packages/nextjs/components/ui/button.tsx
+++ b/packages/nextjs/components/ui/button.tsx
@@ -4,7 +4,7 @@ import { type VariantProps, cva } from "class-variance-authority";
 import { cn } from "~~/lib/utils";
 
 const buttonVariants = cva(
-  "inline-flex items-center justify-center whitespace-nowrap rounded-full text-sm font-medium transition-colors focus-visible:outline-none focus-visible:ring-1 !ring-accent focus-visible:ring-ring disabled:pointer-events-none disabled:opacity-50",
+  "inline-flex items-center justify-center whitespace-nowrap rounded-md text-sm font-medium transition-colors focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring disabled:pointer-events-none disabled:opacity-50",
   {
     variants: {
       variant: {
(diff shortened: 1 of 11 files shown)


packages/nextjs/components/ui/button.tsx
```diff
@@ -4,21 +4,21 @@ import { type VariantProps, cva } from "class-variance-authority";
...
+  "inline-flex items-center justify-center whitespace-nowrap rounded-full text-sm font-medium transition-colors focus-visible:outline-none focus-visible:ring-0 focus-visible:ring-ring disabled:pointer-events-none disabled:opacity-50",
   {
     variants: {
       variant: {
         default: "bg-primary text-primary-foreground shadow hover:bg-primary/90",
         destructive: "bg-destructive text-destructive-foreground shadow-sm hover:bg-destructive/90",
         outline: "border border-input bg-background shadow-sm hover:bg-accent hover:text-accent-foreground",
         secondary: "bg-secondary text-secondary-foreground shadow-sm hover:bg-secondary/80",
-        ghost: "hover:bg-accent hover:text-accent-foreground",
+        ghost: "hover:bg-accent",
```
rin-st, line 15: Color is not changed intentionally. Scroll to ghost in docs https://ui.shadcn.com/docs/components/button, color is very organic, so when we'll update our theme I think it will be fixed. 

And I also think that when possible we don't need change colors in `ui` folder components. Again it should work great everywhere after configuring theme

packages/nextjs/components/ui/button.tsx
```diff
@@ -4,21 +4,21 @@ import { type VariantProps, cva } from "class-variance-authority";
 import { cn } from "~~/lib/utils";
 
 const buttonVariants = cva(
-  "inline-flex items-center justify-center whitespace-nowrap rounded-md text-sm font-medium transition-colors focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring disabled:pointer-events-none disabled:opacity-50",
+  "inline-flex items-center justify-center whitespace-nowrap rounded-full text-sm font-medium transition-colors focus-visible:outline-none focus-visible:ring-0 focus-visible:ring-ring disabled:pointer-events-none disabled:opacity-50",
```
rin-st, line 7: notice ring-0: With ring-1 behavior is [strange](https://github.com/radix-ui/primitives/issues/1164), inside dropdown it's focused even on hover and you cant move focus with `tab` (but can with `up` and `down` buttons). And after page change there's no focus on hover. Haven't met radixui bugs before though. So for now I set it to ring-0

https://github.com/user-attachments/assets/b1c9150a-6a1c-4bae-8723-905b21fc4ba2

Update: I think it's ok to make it the same color as button. Also found a small bug that our menu doesn't open on focus + click to enter. Will fix it tomorrow

rin-st: updated ring color to ring-accent for now since same color as button makes it invisible. Fixed menu and qrcode modal so they can be opened by `enter` press and closed by `esc`. Other menu buttons also works on `enter` except `Copy address` because `CopyToClipboard` component works only on click. Looks like it's better to write our custom component, since `react-copy-to-clipboard` is outdated, written as [class comonent](https://github.com/nkbt/react-copy-to-clipboard/blob/master/src/Component.js) and needs additional types https://github.com/scaffold-eth/scaffold-eth-2/pull/933#issue-2515987545. If you agree, I'll create an issue for it

technophile-04:
Hey @rin-st thanks for such detailed explanation love it! 

I feel like we shouldn't touch the components added from shadcn yet(unti will we are in dev phase) and let them be stock as it is. Umm I don't want us to mimic daisyUI SE-2 with shadcn and keep it as stock as possible also avoid passing of extra `className` while using them(until we are in development stage). Its completely fine if things breaks a bit / there very rough conner since things are not final yet. 

^ We can always polish stuff later and as you mentioned in https://github.com/scaffold-eth/scaffold-eth-2/pull/937#discussion_r1760199979 we are also not final on theme yet. When we update we can always again visit this things. Also the reason I say we keep things as close to shadcn without updating much is this will make adding shadcn stuff to SE-2 look more compatible when people start using it. 

For example if we update shadcn btn from rounded-md => rounded-full. Now if a developer using SE-2 plans to use Alert component from shadcn he has to manually update that component to bit more rounded. 

what do you say? Thanks @rin-st!! Also please fee free to push changes directly to that branch 🙌

rin-st:
> For example if we update shadcn btn from rounded-md => rounded-full. Now if a developer using SE-2 plans to use Alert component from shadcn he has to manually update that component to bit more rounded.

It completely depends of our default styles. Like now, our buttons are always rounded, so when dev uses `<button className='btn' />` its always rounded. So I just kept it as is now. 

> I feel like we shouldn't touch the components added from shadcn yet(unti will we are in dev phase) and let them be stock as it is. Umm I don't want us to mimic daisyUI SE-2 with shadcn and keep it as stock as possible also avoid passing of extra className while using them(until we are in development stage). Its completely fine if things breaks a bit / there very rough conner since things are not final yet.

But if we want to take default shadcn styles and change it later then ok, let's do it.

rin-st:
Reverted ui changes and passing classNames

But I'm not sure regarding the last. We will need to do this anyway, so looks like it's better to keep that classNames from the start, so we will not forget to change it in future. 

See, how it looks. With classnames:

https://github.com/user-attachments/assets/2021166f-513d-44d6-bf45-df013669dc88

Without classNames - trigger button padding, dropdown border, buttons width, disconnect color (we can make disconnect button type="destructive", but it will be with red bg, so not so good I think):

https://github.com/user-attachments/assets/92a531ae-1f38-4426-a717-bcbf58b22fd3

everything except buttons width is optional for now though

rin-st:
Returned full width for menu items for now. 
Also fixed a bug, menu items were clickable during closing animation. Fixed version: see where I click to close the menu. Before fixing I'd be disconnected

https://github.com/user-attachments/assets/11f54dd5-8c43-4fc8-b27f-242369e849eb

---

Merging this, I think current state is good enough to move forward

PR #938, opened 2024-09-16
make solidity complier as arrays

technophile-04:
### Description: 

I think its makes to have it as array especially really helpful for people building the extensions this way we have our `0.8.17` by default but also allow others building extension to share example of contracts which uses different solidity versions and settings. Original discussion https://github.com/scaffold-eth/create-eth/pull/122


rin-st (approved):
Lgtm!

PR #940, opened 2024-09-16
Fix rounding issue when converting to wei

lukaw3d:
## Description

Related to https://github.com/scaffold-eth/scaffold-eth-2/pull/894 https://github.com/scaffold-eth/scaffold-eth-2/pull/893

https://abi.ninja/0xde30da39c46104798bb5aa3fe8b9e0e1f348163f/1?methods=transfer
|  |  |
| --- | --- |
| abi ninja_0xde30da39c46104798bb5aa3fe8b9e0e1f348163f_1_methods=transfer | abi ninja_0xde30da39c46104798bb5aa3fe8b9e0e1f348163f_1_methods=transfer (1) |

## Additional Information

## Related Issues

_Closes #{issue number}_

Your ENS/address:


packages/nextjs/components/scaffold-eth/Input/IntegerInput.tsx
```diff
@@ -23,7 +24,7 @@ export const IntegerInput = ({
     if (typeof value === "bigint") {
       return onChange(value * 10n ** 18n);
     }
-    return onChange(BigInt(Math.round(Number(value) * 10 ** 18)));
+    return onChange(parseUnits(value, 18));
```
rin-st, line 27: Thank you for your pr @lukaw3d ! Good point!

I think instead of this block 
```
if (typeof value === "bigint") {
  return onChange(value * 10n ** 18n);
}
return onChange(BigInt(Math.round(Number(value) * 10 ** 18)));
```
we can use just
`return onChange(parseEther(value.toString())`

cc @carletex @technophile-04

technophile-04: > return onChange(parseEther(value.toString())

Yes! I tried going through all previous conversations #220, #101, #104, #36 and this seem safe and smoothes way to do it! And yes the current `IntegerInput` api is bad while using it you need to mention `useState<string | bigInt>`. I feel like input should have same api as plain html/react input onChange should work with string. 

So yeah let's go with `onChange(parseEther(value.toString())` and remove `useState<string | bigInt>`. Also  re-searched on why we were returning bigInt in some cases and the `if (typeof value === "bigint")` and there seems no need to do that, it just seems naive Shiv copy-pasted some code from SE-1 without understanding and we didn't removed it and kept handling that case. Checkout https://github.com/scaffold-eth/scaffold-eth-2/pull/36/files#r1764521420 

So yeah lets go with https://github.com/scaffold-eth/scaffold-eth-2/pull/940#discussion_r1763119411 what Rinat suggested there. 

<details>

<summary>
Pasting the test contract found in other PR : 
</summary>

```sol
import "hardhat/console.sol";

contract YourContract {
	event SetPurpose(address sender, string purpose);
	int256 public intVariable = 1;

	string public purpose = "Building Unstoppable Apps!!!";
	uint256 public totalSupply = 2.5 ether;
	mapping(address => uint256) public counter;
	bytes32 public myBytes;

	constructor() payable {
		// what should we do on deploy?
	}

	function setIntVariable(int256 _intVariable) public {
		intVariable = _intVariable;
	}

	function setPurpose(string memory newPurpose) public {
		purpose = newPurpose;
		console.log(msg.sender, "set purpose to", purpose);
		emit SetPurpose(msg.sender, purpose);
	}

	function withAddressAndAndValueIncrementCounter(
		address _add
	) public payable {
		require(msg.value > 0.0001 ether, "Not enough");
		counter[_add] += 1;
	}

	function withUint8AndBytes(uint8 _u, bytes32 _b) public {
		counter[msg.sender] = _u;
		myBytes = _b;
	}

	function counterPlusX(
		address _add,
		uint256 _x
	) public view returns (uint256) {
		return counter[_add] + _x;
	}

	// to support receiving ETH by default
	receive() external payable {}

	fallback() external payable {}
}

```

</details>

lukaw3d: Simplified, but
```diff
- return onChange(parseEther(value.toString())
+ return onChange(parseEther(value).toString())
```

isValidInteger can also stop supporting bigint now, but I don't know if external users rely on it
https://github.com/scaffold-eth/scaffold-eth-2/blob/bb1e49095bb99dfc132a56700e503f8246f21ece/packages/nextjs/components/scaffold-eth/Input/utils.ts#L79

technophile-04:
Thanks @lukaw3d for finding and fixing! Also @rin-st for suggestion! Merging this 🙌

technophile-04 (approved):

PR #909, opened 2024-08-14
Exclude external links from triggering progress bar

portdeveloper:
## Description

-> Modifies the nprogress bar so that it only forks for non external links.

-> Implements the changes from:
https://github.com/BuidlGuidl/extensions-hackathon/pull/23#event-13881221641

https://github.com/user-attachments/assets/0f191f3c-b02b-4782-981c-89192d958b21

## Additional Information

- [ ] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [ ] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

## Related Issues

Fixes #908 

_Note: If your changes are small and straightforward, you may skip the creation of an issue beforehand and remove this section. However, for medium-to-large changes, it is recommended to have an open issue for discussion and approval prior to submitting a pull request._

Your ENS/address:

PR #943, opened 2024-09-18
Remove nprogress + use next-nprogress-bar

carletex:
When we migrated to the App Router https://github.com/scaffold-eth/scaffold-eth-2/pull/535 we switched from https://www.npmjs.com/package/next-nprogress-bar to https://www.npmjs.com/package/nprogress + creating a custom integration with the App Router.

We already had an issue with external `_blank` links https://github.com/scaffold-eth/scaffold-eth-2/pull/909 and now I realized that we have the same problem with hash links (#section). The progress bar will start to load (it shouldn't do anything)

One solution will be to just parse the #links and fix it, but created this PR with another alternative: moving back to `next-nprogress-bar` which has been updated and supports the App Router nicely. Under the hood it does the same thing that we are doing, but it's handling some stuff a bit better. So it'll be moving custom logic to an external library.

Happy to close this, just wanted to hear your take on it.

Thanks!!


rin-st (approved):
Agree let's return to `next-progress-bar` since it's up to date. Working great for me!

technophile-04 (approved):
Makes sense! Also seems to nicely work with a tags(not showing the loader)

PR #955, opened 2024-10-09
depreciate default alchemy key

technophile-04:
## Description

Removed the default alchemy key and also added `fallback` to the transport. 

#### Frontend: 

We check if alchemy key is set via process.env if its present then we do [alchemyRpc, public] (this is the fallback order) , if alchemy key is not set we just use generalRpc. 

### Hardhat: 

Updated all the networks with public rpc URLs (got this public URL's comes from viem/chains) . For forking people will still need to add the alchemy api key.


packages/hardhat/hardhat.config.ts
```diff
@@ -12,7 +12,7 @@ import "hardhat-deploy-ethers";
 
 // If not set, it uses ours Alchemy's default API key.
 // You can get your own at https://dashboard.alchemyapi.io
-const providerApiKey = process.env.ALCHEMY_API_KEY || "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF";
+const providerApiKey = process.env.ALCHEMY_API_KEY || "";
```
carletex, line 15: Do we want to keep this? It seems it's only being used when mainnet forking. Maybe

packages/nextjs/scaffold.config.ts
```diff
@@ -20,7 +20,7 @@ const scaffoldConfig = {
   // You can get your own at https://dashboard.alchemyapi.io
   // It's recommended to store it in an env variable:
   // .env.local for local testing, and in the Vercel/system env config for live apps.
-  alchemyApiKey: process.env.NEXT_PUBLIC_ALCHEMY_API_KEY || "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF",
+  alchemyApiKey: process.env.NEXT_PUBLIC_ALCHEMY_API_KEY || "",
```
carletex, line 23: Remove the "this is ours Alchemy`s default API KEY"

Another thing: Should we hint something about this in the README? 

Here maybe:
image

- You can add your Alchemy API Key in `scaffold.config.ts` if you want more reliability in your RPC requests. 

(or something similar)

packages/hardhat/hardhat.config.ts
```diff
@@ -51,43 +51,43 @@ const config: HardhatUserConfig = {
       },
     },
     mainnet: {
-      url: `https://eth-mainnet.alchemyapi.io/v2/${providerApiKey}`,
+      url: `https://cloudflare-eth.com`,
```
rin-st, line 54: nit question: How to decide which one to use here? As I understand you took one of the stable ones, right?

packages/hardhat/hardhat.config.ts
technophile-04: Remove ALCHEMY_API_KEY from `.env.exmaple` itself and added FORKING_URL variable there so people can use any RPC url they want? Didn't set the default FORK_URL value to `https://cloudflare-eth.com` because if people run `yarn fork`  then `https://cloudflare-eth.com` will crash and also block that person's IP and it will affect frontend (since in frontend we are using same RPC URL for ENS)

packages/nextjs/scaffold.config.ts
technophile-04: Ohh yeah missed thanks updated!

packages/hardhat/hardhat.config.ts
technophile-04: Yess lol I just copied what viem/chains uses for their public rpc URLs (Assuming lol they did some research on it). 

Also was thiking in future, instead of hardcoding this URL's we could bring viem/chains use the url from their so that when viem/chain updates we get updated URL's too. But yeah it has it pro's and con's

rin-st: Make sense, thanks! Agree, it's better to use import that URLs later then

packages/hardhat/hardhat.config.ts
```diff
@@ -10,14 +10,13 @@ import "@nomicfoundation/hardhat-verify";
...
-// If not set, it uses ours Alchemy's default API key.
-// You can get your own at https://dashboard.alchemyapi.io
-const providerApiKey = process.env.ALCHEMY_API_KEY || "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF";
 // If not set, it uses the hardhat account 0 private key.
 const deployerPrivateKey =
   process.env.DEPLOYER_PRIVATE_KEY ?? "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
 // If not set, it uses ours Etherscan default API key.
 const etherscanApiKey = process.env.ETHERSCAN_API_KEY || "DNXJA8RX2Q3VZ4URQIWP7Z68CJXQZSC6AW";
+// forking rpc url
+const forkingURL = process.env.FORKING_URL || "";
```
carletex, line 19: won't the mainnet one work? `https://cloudflare-eth.com`

Or does it need to be a special RPC?

(Trying to understand why we are using an env var for this)

carletex (approved):
Great stuff @technophile-04 Thanks!!

Left a comment (more like a question haha) but feel free to merge if you want.

packages/hardhat/hardhat.config.ts
technophile-04: Screenshot 2024-10-10 at 9 05 07 AM

Public rpc are bad forking and most of the time doesn't work. The above SS is for `https://cloudflare-eth.com`. 

Screenshot 2024-10-10 at 8 59 34 AM

The above is `eth.llama.rpc` and it didn't work too :( 

And also mentioned here https://github.com/scaffold-eth/scaffold-eth-2/pull/955#discussion_r1793305774 it might  be the case that the `https://cloudflare-eth.com` block the IP and then frontend will not work too because it uses the same RPC URL :(

technophile-04:
Thanks all! Merging this for now 🙌 but we can update later.

PR #964, opened 2024-10-14
Use mainnet RPC for default forking URL

moltam89:
Hey guys,

with this change we can use **yarn fork** out of the box.

Cheers,
Tamas


moltam89:
Thanks @technophile-04 !

I noticed today that yarn fork works with foundry and I guessed that we want this for hardhat as well.

But probably it is the other way and foundry should not use https://cloudflare-eth.com for forking either.

technophile-04:
>  I noticed today that yarn fork works with foundry and I guessed that we want this for hardhat as well.

> But probably it is the other way and foundry should not use https://cloudflare-eth.com/ for forking either.

Ohh yes we could probably remove it foundry too. Because it will also block the frontend if users fork make a lot of request during fork.

PR #966, opened 2024-10-18
Add a SE-2 cursor project rule

portdeveloper:
This PR introduces a `.cursorrules` file to enhance the development experience with Scaffold-ETH 2. The file contains:

- Instructions for answering Scaffold-ETH 2 related questions
- Important notes on using correct hook names (useScaffoldReadContract, useScaffoldWriteContract, these were used wrongly by AI as useScaffoldContractRead and useScaffoldContractWrite respectively)
- Example Q&As for common scenarios

I have read the [documentation for prompt engineering on Claude.](https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/overview)

**And these were the steps:**

1. Prompt generator - is sth online
2. Be clear and direct - tried with using clear verbs
3. Use examples (multishot) - gave two eamples
4. Let Claude think (chain of thought) - used the usual "think step-by-step" trick
5. Use XML tags - used them whenever possible to tell ai about important stuff or just to categorize
6. Give Claude a role (system prompts) - an expert in scaffold-eth 2
7. Prefill Claude’s response - didn't use this as answers might be different?
8. Chain complex prompts - didn't use it
9. Long context tips - the user doesn't use a long context by default?

To be honest, prompt engineering does not overlap with what a .cursorrules file is 100%. But I believe that using a .cursorrrules file such as this one makes sense as it improves how a newbie might interact with SE-2.

Also, the user might want to index se-2 docs, to better benefit from Cursor and also from the .cursorrules file, as we tell the LLM to check the docs as well.:
image

I believe the .cursorrules file can be improved on, and I would love to hear your feedback on this.

### How to test
1. Clone the repo and switch to the branch.
2. Go to cursor settings and index se-2 docs
3. Open a chat
4. While asking your first question, please use CTRL+ENTER for windows and CMD+ENTER for macos for cursor to see all the files


damianmarti:
@portdeveloper this looks good, I will try it. Is there a way to add the documentation link on the cursorrules file, so the user do not need to add this manually?

portdeveloper:
@damianmarti Thanks!

I put it as a link there, but that way, the AI does not always read it. 
I also tried putting docs in a single file here, and then putting it into the cursorrules file,
https://raw.githubusercontent.com/portdeveloper/crawl-se-2-docs/refs/heads/main/flattened_docs.md

But since it is looong, the AI again does not read it well. The only way here is to index it using Cursor's Docs feature.

portdeveloper:
I have been playing around with this cursorrules file again and I think it might also make sense to ship this as an extension. 

A file is still a file and people might not want this weird file in their se-2 folders if they are not using cursor

portdeveloper:
Thanks for the feedback @damianmarti ! Could you please give it another go? I have made some changes to it and I would love to get feedback on them.

portdeveloper:
Updates:

Cursor now has project rules! And it works a lot better than a .cursorrules file! 
Read about it here: https://docs.cursor.com/context/rules-for-ai

I have included a very simple rules file. Please check if it works while getting the correct names.
Usually Cursor tends to use useScaffoldContractWrite and useScaffoldContractRead, which are the old names for hooks.

From my manual tests talking to Cursor, while we use this new rules file, the correct names are used!

Let me know if it works for you as well!

rin-st:
This is great Port!

`useScaffoldReadContract` works well! For `useScaffoldWriteContract` it uses wrong (or old) API by default, but can fix it after some chatting

.cursor/rules/scaffold-eth-hooks.mdc
technophile-04, line 1: Umm so is there a genereal prompt file too where we could mention other stuff like how the project is structured etc? 

Also whats the advantage of this way instead of supplying the `full-llm.txt` file by @ ing the docs?

technophile-04 (commented):
Tyms @portdeveloper! This worked for me 🙌 like initially it used the wrong name but then it say lininting errors nd then update the hooks with right name. 

Also would have been great if we could have told it to use SE-2 components, like while developing the UI it used custom things but then prompted it to use SE-2 components it used them correctly.  

Just made this with couple of prompts : https://coin-flip-rho-gray.vercel.app/

.cursor/rules/scaffold-eth-hooks.mdc
portdeveloper: What I noticed is that it does not even need the project rules. Can you test without the file as well? It just gets the correct names every time for me

damianmarti:
It's working great for me now!! No mistakes from the AI!

Thanks @portdeveloper for keep pushing it!

portdeveloper:
Interesting observation: No matter how you add the docs link to the rule file, Cursor cannot reference it. So the user still has to add the docs manually to the repo

packages/foundry/lib/forge-std
technophile-04, line 1: Let's remove this files

carletex:
Hey all! 

I spend the morning with Pablo & Shiv (live in Spain!) tinkering with this.

The conclusion: non-deterministic things sucks :D

We tested 3 versions of the rules files (aso Sonnet 3.5 and 3.7. 3.7 was usually better):

1. Port's version (you can rollback to this commit to test https://github.com/scaffold-eth/scaffold-eth-2/pull/966/commits/6033ca669cbc5cc1b2ee21569a03dfba99d6c064)

- This was fast, and got the hooks right (sometimes not in the first try, but it auto-fixed)
- Besides the read/write hooks, it didn't use the "scaffold-way" most of the times: Address component, yarn start (was doing yarn dev inside nextjs folder), etc.

2. Full LLM file pasted into the rules: https://docs.scaffoldeth.io/llms-full.txt

- This was a slower and results similar to Port's (but was slightly better on the scaffold-way)

3. The current version of this PR: https://github.com/scaffold-eth/scaffold-eth-2/pull/966/commits/a20fb77f8f29db55b6efcae121a64847749db293

We came to this version after a few tries and I think it has the best of both. Fast & scaffold-aware.


I have to say that most of the tries one-shotted (even if it had to auto correct some times)

----

This is what we did for testing:

- We start with cursor with a clean SE-2 install (no chain running, etc)
- First prompt:
```
I want to modify the exiting smart contract and add a delegate feature.

The owner of the contract will be able to set a delegate. And the the delegate will be able to witdraw from the contract (as the owner)

Then I want to create a custom UI (in the homepage) where the current delegate is display, and a input + button the update the delegate with a new Delegate Address.
```

(this should create a working version and should run chain / deploy /start)

- When is worklng, second prompt:`
```
Can you change the Owner to be "<WHATEVER_BURNER_ADDRESS>"?
```

(this should modify the script + run yarn deploy again)

- Last prompt:
```
How could I deploy this to Sepolia? Don't do it, just tell me the steps to follow.
```

(beside the basics, it should mention vercel and ipfs and hint to modify scaffold-config with the network).

----

Let's do a round of testing and see if we are happy with the results! Thanks all for working on this <3

rin-st:
> This is what we did for testing:

I tried to do the same

> First prompt:

Contract was a bit broken at the first try, it didn't handle `Ownable` properly but fixed at the second try. Also, it added frontend changes to /pages/index.tsx and I asked to change it to app router and then it was fixed

> When is worklng, second prompt:

This worked well

> Last prompt:

Also worked well but cursor didn't mention ipfs. Asked for it and then got full answer.

---

After it I asked about event hooks and its implementation and about se-2 components and everything worked well!

Great job guys, it's really awesome!

rin-st (approved):

damianmarti (approved):
This is working great for me, too!!

I had some issues trying to use an ENS in the deploy script, but overall, the AI got it right to implement code that interacts with the contract using the hooks and the components.

technophile-04 (approved):

carletex (approved):
Merging this! I think this is a good first iteration. Let's tweak as we go.

Thanks @portdeveloper for pushing for this and everything that helped too <3

PR #971, opened 2024-10-22
Up wagmi and viem

rin-st:
## Description

Up wagmi and viem versions


technophile-04 (approved):
Thanks @rin-st it works great!!

PR #972, opened 2024-10-24
Add shell script for prod deployments + modify Makefile accordingly

portdeveloper:
Hello! 

## This PR:
-> adds a shell script to give the user a menu for selecting a keystore when deploying to any network other than the local anvil node
-> modifies the makefile to pass in the RPC_URL from the foundry.toml file to the shell script
-> works in linux and macos, since foundry does not work on windows natively, windows is ignored. So we should test this on: linux + wsl + macos
-> removes the need to modify the .env file when changing keystores to use

## Possible To-Do:
-> ‘—keystore’ flag (thx to Shiv for the suggestion) I have some working code on this, but it accepts variables like this:
`yarn deploy --network sepolia my-keystore` 
instead of 
`yarn deploy --network sepolia --keystore my-keystore`
We can also introduce a new script as I couldn't find how to achieve this without doing so. Tried passing in another flag in the package.json, but did not work.


technophile-04:
Hey port thanks for this!! Can you please resolve the merge conflicts we did some tweaks in #1011 🙌

portdeveloper:
Hey ofc! Will work on this today!

portdeveloper:
Will do some more testing + UX improvements and ask for your review!

portdeveloper:
Alright now we get:
-> a direct deployment to the local chain when we run `yarn deploy`
-> a menu in the console when we do a `yarn deploy --network sepolia`
image
-> a direct deployment to the specified chain with the specified keystore when we do a `yarn deploy --network sepolia --keystore zort`
image


https://github.com/user-attachments/assets/653f9df5-b73a-48be-9ec6-db74e79fe143



Also the `deploy-verify` target in the Makefile is unreachable I think 🤔

technophile-04:
This is great @portdeveloper been testing it for a while and looks great!! 

Some questions / improvements: 

1. Shall we use `.js` script instead of shell script? So that it's easy for management and don't add extra `scripts` dir for shell script, and maybe import that function in `parseArgs.js` file directly instead of calling `spawnSync` ? 

2. Here can we make more clear maybe something like : `y/n or c to check balance again` 
    
    Screenshot 2025-02-15 at 1 19 44 PM

3. The check balance step is good but yeah we need to write the password two times while deploying it. If some person have long password which might be common to have...it might be PITA to enter it two times. 

^ Maybe to avoid after user selecting keystore we ask a question wheather he want's to check balance or not, if not we directly deploy (only 1 time password enter) if he wishes to check balance then enter the password twice. 

4. I think we should update the `.env.example` https://github.com/scaffold-eth/scaffold-eth-2/blob/d73d3ebf313864c36aef69bde87203cf0a9a7c57/packages/foundry/.env.example#L18 mentioning that this is only used while local development and can use any keystore name if they want.

portdeveloper:
Thanks for the review @technophile-04 ! I think your points make great sense! Let me see if I can do a js file!

portdeveloper:
Alright so now we can:

- [ ] Deploy to localhost normally
- [ ] Deploy to any chain by using --network chainname, it let's you select a keystore 

Things we cannot do:
- [ ]  --keystore flag does not work. Enabling the user to type in stuff to the terminal is definitely harder than it is with a shell script.

technophile-04:
Tysm Port! Played around with it, and looks nice! just some comments: 

1. I think we need to update `yarn account` behaviour too in similar way because currently it looks at `.env` but we could keep the flow similar to this where we prompt the list of keystore to check balance for (without the 0th option) 
2. Lets update the `yarn account:import` to also ask for keystore name instead of defaulting it to `scaffold-eth-custom`. 

####  Bug: 

Even though I have `.env#ETH_KEYSTORE_ACCOUNT=scaffold-eth-custom` on localhost it's always using `scaffold-eth-default` I think we need to allow people to use keystore also during localhost development if they want to test. 

---

> `--keystore` flag does not work. Enabling the user to type in stuff to the terminal is definitely harder than it is with a shell script.

> However, we can ask the user to pass in `--keystore` keystorename `--password passwordhere` instead

Yeah the CLI just freezes if I pass `--keystore`. Lets remove that flag for now. We don't want user to enter the password in plain text directly probably. We could take the same path as hardhat where we require user to ender password in masked format. But yeah let's tackle `--keystore` flag in different PR.

portdeveloper:
I have turned on the dolphin mode. Since I can't make it accept passwords with js, now we dive into js and resurface to makefile multiple times

portdeveloper:
I have identified the EVIL LINE.
Just importing that selectKeystore file (not even using it btw!) breaks the password input:
Here is the little problem kid, captured in LIVE FOOTAGE!!!
https://screen.studio/share/oOHVh5CO

portdeveloper:
Below is a quick overview of the current state of this PR:

We use an .env var named LOCALHOST_KEYSTORE_ACCOUNT to have a default keystore.

Commands and what they do:
### deploy related commands
`yarn deploy` -> Uses LOCALHOST_KEYSTORE_ACCOUNT from .env file and deploys to anvil
`yarn deploy --keystore hello` -> Uses hello keystore and deploys to anvil
`yarn deploy --network sepoila` -> Lets the user select a keystore interactively and deploys to sepolia OR create a new keystore
`yarn deploy --network sepolia --keystore i-love-se2` -> Uses i-love-se2 keystore and deploys to sepolia, skipping interactive keystore selection 
``

### other commands
`yarn account` -> Lets the user select a keystore interactively to display balances for it on all chains that we have in foundry.toml(much like the previous implementation, but with interactive selection of any keystore the user has)
`yarn account:import` -> Asks the user for a name and then a private key, uses those values to create a new keystore

packages/foundry/Makefile
```diff
@@ -39,26 +39,26 @@ deploy-and-generate-abis: deploy generate-abis
...
-	if grep -q "scaffold-eth-default" .env; then \
-		cast wallet address --password localhost; \
-	else \
-		cast wallet address; \
-	fi
-
 # List account
 account:
-	@node scripts-js/ListAccount.js $$(make verify-keystore)
+	@node scripts-js/ListAccount.js $(RPC_URL)
```
technophile-04, line 44: We can remove this now? Since in package.json we are directly calling `checkAccountBalnce.js`? Or maybe lets update this to do 

```
account: 
  @node scripts-js/checkAccountBalnce.js $(RPC_URL)
```

packages/foundry/scripts-js/getAddressBalance.js
technophile-04, line 1: Let's remvoe this also remove `chec-balance` from makefile I think we don't need this. 

We have `yarn account` which uses checkAccountBalnce right?

packages/foundry/scripts-js/ListAccount.js
technophile-04, line 1: rename this to ListKeystores

packages/foundry/scripts-js/selectKeystore.js
technophile-04, line 1: Rename this to SelectOrCreateKeystore?

technophile-04 (changes requested):
> Just importing that selectKeystore file (not even using it btw!) breaks the password input:

Wohhhhh nice catch, thanks for the research and explanation 🙏 

--- 

TYSM Port this is looking almost solid, just added a some last comments but it works nicely!  

also just pushed a small commit at https://github.com/scaffold-eth/scaffold-eth-2/pull/972/commits/c1216fa944613a79f49113959abcc69fe31e7706 to hide private key during account:import (behavior inline with hardhat account:import)

packages/foundry/Makefile
portdeveloper: Thanks! just updated it to yarn account and also removed RPC_URL from that command as it is not needed

packages/foundry/.env.example
```diff
@@ -16,4 +16,4 @@ ALCHEMY_API_KEY=oKxs-03sij-U_N0iOlrSsZFr29-IqbuF
 # Etherscan API key is used to verify the contract on etherscan.
 ETHERSCAN_API_KEY=DNXJA8RX2Q3VZ4URQIWP7Z68CJXQZSC6AW
 # Default account for localhost / use "scaffold-eth-custom" if you wish to use a generated account or imported account
-ETH_KEYSTORE_ACCOUNT=scaffold-eth-default
+LOCALHOST_KEYSTORE_ACCOUNT=scaffold-eth-default
```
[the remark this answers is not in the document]
portdeveloper: Or:
# Keystore account to be used for localhost deployments only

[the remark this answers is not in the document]
technophile-04: Yup updated at 76cf9874ae17 also removed the above lines on `DEPLOYER_PRIVATE_KEY `

technophile-04 (approved):
TYSM Port!! just pushed some commits this is looking solid now 🙌 

Will create a PR to update the SE-2 docs merging this now 🚀🚀

PR #974, opened 2024-10-28
Forking url for foundry

moltam89:
Hey guys,

This PR is the same as https://github.com/scaffold-eth/create-eth/pull/141/.

Cheers,
Tamas

### DIFF
--- a/packages/foundry/Makefile
+++ b/packages/foundry/Makefile
@@ -9,6 +9,11 @@ setup-anvil-wallet:
 chain: setup-anvil-wallet
 	anvil
 
+# include variables from .env file if it exists
+ifneq (,$(wildcard .env))
+    include .env
+endif
+
 # Start a fork
 fork: setup-anvil-wallet
 	anvil --fork-url ${FORK_URL} --chain-id 31337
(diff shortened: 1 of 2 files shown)


technophile-04:
Thanks @moltam89 !! Umm thinking more on this, I think since now the our default Alchemy started working maybe we could use it as previously or even use public default rpc url and use alchemy in frontned.  

The reason I say this because the devX is really bad, like after you run `yarn fork` it gives you this error: 

Screenshot 2024-10-30 at 12 46 31 PM


People don't have any idea where they should put fork url. They need to need search for command how its used package.json => packages/foundry/package.json => makefile => fork 

Gonna discuss this with SE-2 team! Keeping this on hold for now, but thanks for the PR!!

moltam89:
Thanks @technophile-04 !

Yeah, if we could run `yarn fork` out of the box, that would be the best!

**Note:**
Hardhat throws an ugly error message as well when we run yarn fork.
Screenshot 2024-10-30 at 08 13 28

technophile-04:
Thanks Tam! Closing this because we added alchemy key back 🙌 

https://github.com/scaffold-eth/scaffold-eth-2/pull/988

Issue #939, opened 2024-09-16
bug: Debug tabs behavior


rin-st:
### Is there an existing issue for this?

- [X] I have looked through the [existing issues](https://github.com/scaffold-eth/scaffold-eth-2/issues)

### Which method was used to setup Scaffold-ETH 2 ?

git clone

### Current Behavior

While working on https://github.com/scaffold-eth/scaffold-eth-2/pull/933 I found very strange behavior: whenever I click on tab with contract, it always change to `YourContract` tab. 

https://github.com/user-attachments/assets/0b67d24f-dcb2-4b53-b0f2-771051dedb43

It's not because of its index or alphabetical order, it's always `YourContract` (if you have that tab, of course). Clearing localstorage doesn't help

### Expected Behavior

Clicked contract tab should be chosen

### Steps To Reproduce

I debugged it and understood that it was because of my other browser tab, which had only `YourContract`.
So, steps:
- run default se-2 (with one contract - YourContract)
- stop your frontend app (but don't close browser tab [tab 1] )
- add another contract and deployment file for it
- `yarn deploy --reset` your contracts (or just `yarn deploy` if you re-run `yarn chain`)
- open new browser tab [tab 2]
- `yarn start` your app
- open it on new browser tab [tab 2]

### Anything else?

Not sure if we need to fix it for now since it's relatively rare case and not so easy to fix. But at least we know what causes this

rin-st:
The bug is not related to the mentioned PR, it also works on `main`

technophile-04:
Thanks @rin-st!! So the summary is that if this problem occurs people need to make sure they close all other browser TABs with `http://localhost:3000` and then refresh the latest one and it should start working. Just tried this and works nicely. 

Umm I think instead of relying on usehooks-ts `useLocalStorage` logic maybe we should write a our custom simple logic for DebugContracts. 

The reason I feel usehooks `useLocalstorage` have this problem is because they are listening to localstroage for all tabs and keeping them in sync : 

![Screenshot 2024-10-28 at 10 39 34 AM](https://github.com/user-attachments/assets/f8d91f22-1e95-4a88-a1cd-32604a50a26e)

But since the outdated tab doesn't have the new contract name, it updates the contractName to contractName[0] in outdated tab and latest tab listen to those changes and also updates it.

PR #975, opened 2024-10-28
Fix debug page contracts view

rin-st:
## Description

Fixes https://github.com/scaffold-eth/scaffold-eth-2/issues/939

One of the possible (and simple) solutions is to use sessionStorage instead of localStorage. As I understand it doesn't change anything except one case: 
User worked with debug page where he chose some non-first contract. User closes browser tab with his debug page and open debug page in another browser tab. With localStorage, last chosen contract will be selected. With sessionStorage first contract will be selected. I believe it's not a problem at all

Additionally, using sessionStorage adds possibility to view different contracts of debug page in different browser tabs

----

What I also tried to do:
- use simple hash function depending on contractNames
```
const [selectedContract, setSelectedContract] = useLocalStorage<ContractName>(
    ${selectedContractStorageKey}-${hash},
    contractNames[0],
    { initializeWithValue: false },
  );
```
it works great, but adds new entry to localstorage for every hash
- use one localStorage for `selectedContract` and one for `hash`. But in this case I needed to use `useEffects` to handle logic and again it didn't work for multiple tabs, same as https://github.com/scaffold-eth/scaffold-eth-2/issues/939

### DIFF
--- a/packages/nextjs/app/debug/_components/DebugContracts.tsx
+++ b/packages/nextjs/app/debug/_components/DebugContracts.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useEffect, useMemo } from "react";
-import { useLocalStorage } from "usehooks-ts";
+import { useSessionStorage } from "usehooks-ts";
 import { BarsArrowUpIcon } from "@heroicons/react/20/solid";
 import { ContractUI } from "~~/app/debug/_components/contract";
 import { ContractName, GenericContract } from "~~/utils/scaffold-eth/contract";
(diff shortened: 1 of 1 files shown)


carletex:
> With sessionStorage first contract will be selected. I believe it's not a problem at all

Agree.

The main reason we have this functionality is to avoid "losing the contract focus" if you refresh the page. And session storage does the trick too

carletex (approved):

technophile-04 (approved):
> With localStorage, last chosen contract will be selected. With sessionStorage first contract will be selected. I believe it's not a problem at all

Yess event I think its not a big deal at all! 

Just tested out nicely and #939 is fixed! Thanks @rin-st!

PR #978, opened 2024-11-02
Updated graph-cli and graph-ts versions

siddhant-k08:
## Description

Updated to graph-cli version to 0.87.0 and graph-ts version to 0.35.1

Your ENS/address: siddhantk08.eth

### DIFF
--- a/packages/subgraph/package.json
+++ b/packages/subgraph/package.json
@@ -17,8 +17,8 @@
         "clean-node": "rm -rf graph-node/data/"
     },
     "dependencies": {
-        "@graphprotocol/graph-cli": "^0.55.0",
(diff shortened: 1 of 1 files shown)


technophile-04:
Thanks @siddhant-k08! Maybe a good idea might be creating a PR to [`create-eth-extensions`](https://github.com/scaffold-eth/create-eth-extensions/blob/fee12c651cf887c15738d1c5bd6a122847c251a6/extension/packages/subgraph/package.json#L21) repo, since we are no longer maintaining this branch and `npx create-eth@latest` with subgraph gives latest changes.

siddhant-k08:
Hey @technophile-04 , sorry I found that out after already creating a PR. But I've already created a similar PR to the create-eth-extensions branch too

PR #991, opened 2024-11-20
Remove hardcoded colors

rin-st:
## Description

Removed hardcoded colors so colors will be better for extensions.

Changes:

general:
- copy button near address and transactions is the same color as default text
- changed input content and placeholder colors
- switch theme toggle color is now `text-base-content`, so it has the same color as switch icon, has better contrast and also is better for extensions, for example sre-challenges light theme
- changed balance error border color
- changed receipt object keys colors

debug page:
- display variable is now `text-base-content/80` and also normal font-weight so it has better contrast with var title

blockexplorer:
- changed color of searchbar placeholder
- changed hardcoded color of pagination button to `btn-disabled`

## Additional Information

### DIFF
--- a/packages/nextjs/components/SwitchTheme.tsx
+++ b/packages/nextjs/components/SwitchTheme.tsx
@@ -29,7 +29,7 @@ export const SwitchTheme = ({ className }: { className?: string }) => {
       <input
         id="theme-toggle"
         type="checkbox"
-        className="toggle toggle-primary bg-primary hover:bg-primary border-primary"
+        className="toggle bg-base-content focus:bg-base-content border-base-content"
         onChange={handleToggle}
         checked={isDarkMode}
       />
--- a/packages/nextjs/components/SwitchTheme.tsx
+++ b/packages/nextjs/components/SwitchTheme.tsx
@@ -29,7 +29,7 @@ export const SwitchTheme = ({ className }: { className?: string }) => {
       <input
         id="theme-toggle"
         type="checkbox"
-        className="toggle bg-base-content focus:bg-base-content border-base-content"
+        className="toggle toggle-primary bg-primary hover:bg-primary border-primary"
         onChange={handleToggle}
         checked={isDarkMode}
       />
--- a/packages/nextjs/components/scaffold-eth/Address/Address.tsx
+++ b/packages/nextjs/components/scaffold-eth/Address/Address.tsx
@@ -177,7 +177,7 @@ export const Address = ({
             </AddressLinkWrapper>
           </span>
           <AddressCopyIcon
-            className={`ml-1 text-base-content ${copyIconSizeMap[addressSize]} cursor-pointer`}
+            className={`ml-1 ${copyIconSizeMap[addressSize]} cursor-pointer`}
             address={checkSumAddress}
           />
         </div>
--- a/packages/nextjs/components/scaffold-eth/Address/Address.tsx
+++ b/packages/nextjs/components/scaffold-eth/Address/Address.tsx
@@ -177,7 +177,7 @@ export const Address = ({
             </AddressLinkWrapper>
           </span>
           <AddressCopyIcon
-            className={`ml-1 text-sky-600 ${copyIconSizeMap[addressSize]} cursor-pointer`}
+            className={`ml-1 text-base-content ${copyIconSizeMap[addressSize]} cursor-pointer`}
             address={checkSumAddress}
           />
         </div>
--- a/packages/nextjs/components/scaffold-eth/Input/InputBase.tsx
+++ b/packages/nextjs/components/scaffold-eth/Input/InputBase.tsx
@@ -50,7 +50,7 @@ export const InputBase = <T extends { toString: () => string } | undefined = str
     <div className={`flex border-2 border-base-300 bg-base-200 rounded-full text-accent ${modifier}`}>
       {prefix}
       <input
-        className="input input-ghost focus-within:border-transparent focus:outline-none focus:bg-transparent h-[2.2rem] min-h-[2.2rem] px-4 border w-full font-medium placeholder:text-accent/80 text-base-content/90 focus:text-base-content/90"
+        className="input input-ghost focus-within:border-transparent focus:outline-none focus:bg-transparent h-[2.2rem] min-h-[2.2rem] px-4 border w-full font-medium placeholder:text-accent/70 text-base-content/70 focus:text-base-content/70"
         placeholder={placeholder}
         name={name}
         value={value?.toString()}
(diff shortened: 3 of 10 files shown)


packages/nextjs/components/SwitchTheme.tsx
```diff
@@ -29,7 +29,7 @@ export const SwitchTheme = ({ className }: { className?: string }) => {
       <input
         id="theme-toggle"
         type="checkbox"
-        className="toggle toggle-primary bg-primary hover:bg-primary border-primary"
+        className="toggle bg-base-content focus:bg-base-content border-base-content"
```
portdeveloper, line 32: new:
image
old:
image

Maybe the switch button looks a little bit too pronounced with these changes?

portdeveloper (approved):
Just dropped one little nitpicky comment, other than that, it looks good to me!

packages/nextjs/components/SwitchTheme.tsx
rin-st: Yes I agree, but It will be difficult for some extensions themes to make all the colors look good in both light and dark theme. But if to make the color `base-content` we are sure that colors will be distinguishable. I can add for example 70-90 opacity, but not sure we need it here

packages/nextjs/components/scaffold-eth/Address/Address.tsx
```diff
@@ -177,7 +177,7 @@ export const Address = ({
             </AddressLinkWrapper>
           </span>
           <AddressCopyIcon
-            className={`ml-1 text-sky-600 ${copyIconSizeMap[addressSize]} cursor-pointer`}
+            className={`ml-1 text-base-content ${copyIconSizeMap[addressSize]} cursor-pointer`}
```
technophile-04, line 180: Maybe we completely remove text-color className from this. 

Then it will follow the color which people set in outer container, otherwise it will still remain hardcode 

image

packages/nextjs/components/scaffold-eth/Input/InputBase.tsx
```diff
@@ -50,7 +50,7 @@ export const InputBase = <T extends { toString: () => string } | undefined = str
     <div className={`flex border-2 border-base-300 bg-base-200 rounded-full text-accent ${modifier}`}>
       {prefix}
       <input
-        className="input input-ghost focus-within:border-transparent focus:outline-none focus:bg-transparent focus:text-gray-400 h-[2.2rem] min-h-[2.2rem] px-4 border w-full font-medium placeholder:text-accent/50 text-gray-400"
+        className="input input-ghost focus-within:border-transparent focus:outline-none focus:bg-transparent h-[2.2rem] min-h-[2.2rem] px-4 border w-full font-medium placeholder:text-accent/80 text-base-content/90 focus:text-base-content/90"
```
technophile-04, line 53: Maybe make this all `/70` or `/60` lol `/90` still pops out too much

packages/nextjs/components/SwitchTheme.tsx
technophile-04: > Maybe the switch button looks a little bit too pronounced with these changes?

I agree, I think maybe we should revert this changes? 

Its fine if colors don't go nicely in extensions but we make sure by default base se-2 looks nice. 

I just tried previous changes to SRE base template and it looks fine : 
Screenshot 2024-11-21 at 4 26 37 PM
Screenshot 2024-11-21 at 4 26 54 PM

Lol light mode doesn't look that great but I think for extensions its fine also non-deterministic what people gonna put in their tailwind.config.js template, its better we optimize for base SE-2?

technophile-04 (commented):
Thanks Rinat this is great!! Just added a couple of small comments

packages/nextjs/components/scaffold-eth/Address/Address.tsx
rin-st: good point, changed for other copy icons too
[155418f](https://github.com/scaffold-eth/scaffold-eth-2/pull/991/commits/155418fd7084cb85214c743b4615474e7be276ac)

packages/nextjs/components/scaffold-eth/Input/InputBase.tsx
rin-st: Changed all to `/70`, when less than 70 it becomes difficult to read the text, at least for me

Screenshot 2024-11-21 at 15 59 10

Screenshot 2024-11-21 at 15 59 29

Pabl0cks (approved):
Nice job @rin-st ! Looks good to me 🙌

packages/nextjs/components/scaffold-eth/Input/InputBase.tsx
Pabl0cks: Yeah for me 70 is ok, looking at pre-PR styles they could look like a disabled field.

packages/nextjs/components/SwitchTheme.tsx
rin-st: Reverted [2113370](https://github.com/scaffold-eth/scaffold-eth-2/pull/991/commits/21133704253ddec64e966c6f534bfe64bc96fb92), but I leave some thougths regarding it

> Its fine if colors don't go nicely in extensions but we make sure by default base se-2 looks nice.

> Lol light mode doesn't look that great but I think for extensions its fine also non-deterministic what people gonna put in their tailwind.config.js template, its better we optimize for base SE-2

The problem is that sre will be group of the most used (or one of the most used) extensions and sre users are important for us. Also, I think most of the users use default light theme. I think we need to make it as good as we can. 

For me, the changed version looks a bit worse than current, but still ok. Well, if not count that I broke hover part a bit :). Since toggle color in that version is the same as color of text, it will work by default with most of the extensions

---

As a note, we can still change a bit sre default colors so maybe we will make it better without changing toggle inside se-2

technophile-04: > As a note, we can still change a bit sre default colors so maybe we will make it better without changing toggle inside se-2

Ohh yes, I was thinking that people could always tweak their tailwind theme since its non-deterministic which we can't handle. 

> The problem is that sre will be group of the most used (or one of the most used) extensions. Also, I think most of the users use default light theme.
> For me, the changed version looks a bit worse than current, but still ok. Well, if not count that I broke hover part a bit :). Since toggle color in that version is the same as color of text, it will work fine by default with most of the extensions

Yup, I agree, maybe let's merge this if we don't find a way to tweak the challenge color and make it look better we can come back to this discussion on SE-2 and tweak SE-2 color palette a bit so that it suits all extensions as well 🙌 , Thanks Rinat!!

technophile-04 (approved):
Tysm @rin-st 🙏!

PR #993, opened 2024-11-21
Remove clone command + add npx install

carletex:
As we discussed yesterday, we are moving away from SE-2 clone method. The first step is to remove it from the docs / READMEs / etc.

Related https://github.com/scaffold-eth/scaffoldeth.io/pull/41

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ This command will install all the necessary packages and dependencies, so it mig
 yarn chain
 ```
 
-This command starts a local Ethereum network using Hardhat. The network runs on your local machine and can be used for testing and development. You can customize the network configuration in `hardhat.config.ts`.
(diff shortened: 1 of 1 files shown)


technophile-04 (commented):
Currently the below commands are hardhat specific 

 https://github.com/scaffold-eth/scaffold-eth-2/blob/e5a666b78c5f7c8dfc8f827643128612dd23c71f/README.md?plain=1#L46 
 
Umm maybe make description for those 3 magic commands a bit more generic and then point to docs directly? 
\

Pabl0cks:
> Umm maybe make description for those 3 magic commands a bit more generic and then point to docs directly?

Sounds good, I can tweak them if you want

Pabl0cks:
> Sounds good.

Tried to remove all the references to specific configuration of Hardhat, and pointed to the docs instead. Also pushed a couple of small tweaks:
- Add foundry to tech stack
- Add extensions hint in installation setup

Let me know if you prefer to remove any of that, or go more minimal in the Readme. Thanks! 🙌

technophile-04 (approved):
Thanks! I think this looks great for first iteration!

PR #999, opened 2024-11-30
[Feature] Project Detail Consolidation

JacobHomanics:
## Description

Consolidates many of the project details into `scaffold.config.ts`.

SE-2 is cloned and forked for many purposes to create many unique applications. These unique applications of course have unique titles, descriptions, images, metadata, etc. There are currently a myriad of places where SE-2 metadata is spread throughout the project, which can make it a pain to update the metadata once a fork/clone becomes mature enough for its own branding. Easily leading to the possibility of complete applications to still have SE-2 branding present, which ends up feeling out of place.

This PR aims to alleviate these issues and remove the developer's need to scour the project to ensure they changed every single piece of SE-2 branding as possible to their own application's branding.

### Give credit where credit is due
Of course, none of this is possible without SE-2. Thus, maybe we could introduce a new section that gives SE-2 the proper plug space that it deserves?

## Additional Information

### DIFF
--- a/packages/nextjs/app/layout.tsx
+++ b/packages/nextjs/app/layout.tsx
@@ -1,12 +1,13 @@
 import "@rainbow-me/rainbowkit/styles.css";
 import { ScaffoldEthAppWithProviders } from "~~/components/ScaffoldEthAppWithProviders";
 import { ThemeProvider } from "~~/components/ThemeProvider";
+import scaffoldConfig from "~~/scaffold.config";
 import "~~/styles/globals.css";
 import { getMetadata } from "~~/utils/scaffold-eth/getMetadata";
 
 export const metadata = getMetadata({
-  title: "Scaffold-ETH 2 App",
-  description: "Built with 🏗 Scaffold-ETH 2",
+  title: scaffoldConfig.projectName,
+  description: scaffoldConfig.projectDescription,
 });
 
 const ScaffoldEthApp = ({ children }: { children: React.ReactNode }) => {
--- a/packages/nextjs/app/layout.tsx
+++ b/packages/nextjs/app/layout.tsx
@@ -6,7 +6,7 @@ import "~~/styles/globals.css";
 import { getMetadata } from "~~/utils/scaffold-eth/getMetadata";
 
 export const metadata = getMetadata({
-  title: scaffoldConfig.projectName,
+  title: `${scaffoldConfig.projectName} App`,
   description: scaffoldConfig.projectDescription,
 });
 
--- a/packages/nextjs/app/layout.tsx
+++ b/packages/nextjs/app/layout.tsx
@@ -1,13 +1,13 @@
 import "@rainbow-me/rainbowkit/styles.css";
 import { ScaffoldEthAppWithProviders } from "~~/components/ScaffoldEthAppWithProviders";
 import { ThemeProvider } from "~~/components/ThemeProvider";
-import scaffoldConfig from "~~/scaffold.config";
+import projectConfig from "~~/project.config";
 import "~~/styles/globals.css";
 import { getMetadata } from "~~/utils/scaffold-eth/getMetadata";
 
 export const metadata = getMetadata({
-  title: `${scaffoldConfig.projectName} App`,
-  description: scaffoldConfig.projectDescription,
+  title: `${projectConfig.name} App`,
+  description: projectConfig.description,
 });
 
 const ScaffoldEthApp = ({ children }: { children: React.ReactNode }) => {
--- a/packages/nextjs/app/page.tsx
+++ b/packages/nextjs/app/page.tsx
@@ -5,7 +5,7 @@ import type { NextPage } from "next";
 import { useAccount } from "wagmi";
 import { BugAntIcon, MagnifyingGlassIcon } from "@heroicons/react/24/outline";
 import { Address } from "~~/components/scaffold-eth";
-import scaffoldConfig from "~~/scaffold.config";
+import projectConfig from "~~/project.config";
 
 const Home: NextPage = () => {
   const { address: connectedAddress } = useAccount();
@@ -16,7 +16,7 @@ const Home: NextPage = () => {
         <div className="px-5">
           <h1 className="text-center">
             <span className="block text-2xl mb-2">Welcome to</span>
-            <span className="block text-4xl font-bold">{scaffoldConfig.projectName}</span>
+            <span className="block text-4xl font-bold">{projectConfig.name}</span>
           </h1>
           <div className="flex justify-center items-center space-x-2 flex-col sm:flex-row">
             <p className="my-2 font-medium">Connected Address:</p>
--- a/packages/nextjs/app/page.tsx
+++ b/packages/nextjs/app/page.tsx
@@ -5,6 +5,7 @@ import type { NextPage } from "next";
 import { useAccount } from "wagmi";
 import { BugAntIcon, MagnifyingGlassIcon } from "@heroicons/react/24/outline";
 import { Address } from "~~/components/scaffold-eth";
+import scaffoldConfig from "~~/scaffold.config";
 
 const Home: NextPage = () => {
   const { address: connectedAddress } = useAccount();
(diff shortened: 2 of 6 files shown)


technophile-04 (commented):
Screenshot 2024-12-02 at 11 40 18 PM

Umm initially even I felt that we had SE-2 branding at lot's of place, but looking at the files changed I don't see that much value init and its bloating the `scaffold.config.ts` a lot. 

I would personally vote for not adding anything and we mention about `getMetadata` util in [docs.scaffoldeth.io ](https://docs.scaffoldeth.io/deploying/deploy-nextjs-app), since that's a powerful util and was actually added so that its easy for people to deal with metadata.  

Maybe `githuRepoLink` and `projectTitle` make a bit sense but again I feel that this values could be just hardhacode in the UI HTML abstracting it / hoisting it to scaffold.config don't make that much. 

The reason why present fields in scaffold.config make sense  to me is because they kind of orchestrate certain internal behaviour of SE-2 so that people don't need to go into weeds of SE-2 to "configure" those things. 

Configuring this UI vars from scaffold.config feels a bit weird to me. Since any which way people need to touch this files for updating Header links etc. 

But let's see what others have to say 🙌

JacobHomanics:
> Screenshot 2024-12-02 at 11 40 18 PM
> Umm initially even I felt that we had SE-2 branding at lot's of place, but looking at the files changed I don't see that much value init and its bloating the `scaffold.config.ts` a lot.
> 
> I would personally vote for not adding anything and we mention about `getMetadata` util in [docs.scaffoldeth.io ](https://docs.scaffoldeth.io/deploying/deploy-nextjs-app), since that's a powerful util and was actually added so that its easy for people to deal with metadata.
> 
> Maybe `githuRepoLink` and `projectTitle` make a bit sense but again I feel that this values could be just hardhacode in the UI HTML abstracting it / hoisting it to scaffold.config don't make that much.
> 
> The reason why present fields in scaffold.config make sense to me is because they kind of orchestrate certain internal behaviour of SE-2 so that people don't need to go into weeds of SE-2 to "configure" those things.
> 
> Configuring this UI vars from scaffold.config feels a bit weird to me. Since any which way people need to touch this files for updating Header links etc.
> 
> But let's see what others have to say 🙌

Pushed a new commit.

Moved the properties to its own `project.config.ts`. This removes the bloat from `scaffold.config.ts`.

The main benefits that I see from the PR's current implementation are:
1. The developer no longer needs to scour random files to change the project's properties. 
In reference to the Unity Game Engine Editor, we can see that they group the specific project's details in its own window (a `*.config.ts` equivalent). It doesn't send them to different areas of the Editor to change each property. Scaffold, currently, also "defines" these properties in specific component files. 
player-settings-window
2. In the latest commit of the PR, we can see that `project.config.ts`'s `project.name` is referenced by both `layout.tsx` and `page.tsx`. Fixing the issue that project details get changed somewhere, but not elsewhere.
a. In SE-2 template, it makes sense to have the `header.tsx`, `layout.tsx`, and `page.tsx` to have separate "names" (Scaffold-ETH 2 App, Scaffold-ETH 2, Scaffold-ETH) defined for the App. However, SE-2 is unusual, in which most other apps will probably consolidate each instance to simply use their app's name (I.E. `Dust Aggregator`). Since there is no need for something like `Dust Aggregator App`, `Dust Aggregator 2`, and `Dust Aggregator`. 

@technophile-04 eager to hear thoughts on the updated reasoning and file changes.

packages/nextjs/app/layout.tsx
```diff
@@ -1,12 +1,13 @@
...
 import { ThemeProvider } from "~~/components/ThemeProvider";
+import projectConfig from "~~/project.config";
 import "~~/styles/globals.css";
 import { getMetadata } from "~~/utils/scaffold-eth/getMetadata";
 
 export const metadata = getMetadata({
-  title: "Scaffold-ETH 2 App",
-  description: "Built with 🏗 Scaffold-ETH 2",
+  title: `${projectConfig.name} App`,
+  description: projectConfig.description,
```
rin-st, line 10: We have `getMetadata` also on `/blockexplorer` and `/debug` pages, and I don't think we need to change that. For me, the default description here is in line with other metadata descriptions, so I think we don't need to move description to variable

packages/nextjs/app/page.tsx
```diff
@@ -15,7 +16,7 @@ const Home: NextPage = () => {
         <div className="px-5">
           <h1 className="text-center">
             <span className="block text-2xl mb-2">Welcome to</span>
-            <span className="block text-4xl font-bold">Scaffold-ETH 2</span>
+            <span className="block text-4xl font-bold">{projectConfig.name}</span>
```
rin-st, line 19: This page is mostly rewritten for new apps so I think no need to use variable here

packages/nextjs/components/Footer.tsx
```diff
@@ -58,17 +58,17 @@ export const Footer = () => {
               </p>
               <a
                 className="flex justify-center items-center gap-1"
-                href="https://buidlguidl.com/"
+                href={projectConfig.creatorLink}
```
rin-st, line 61: see some lines above, we have `Built with <HeartIcon className="inline-block h-4 w-4" /> at` and it will work fine with creatorLink only if it a company or smth like that. For me, "https://buidlguidl.com/" is good and shouldn't change

packages/nextjs/components/Footer.tsx
```diff
@@ -58,17 +58,17 @@ export const Footer = () => {
...
                 target="_blank"
                 rel="noreferrer"
               >
-                <BuidlGuidlLogo className="w-3 h-5 pb-1" />
-                <span className="link">BuidlGuidl</span>
+                <projectConfig.creatorSvg className="w-3 h-5 pb-1" />
```
rin-st, line 65: Same for svg and creator. Also, svg could be different and wont work well with common styles

packages/nextjs/components/Header.tsx
```diff
@@ -90,11 +91,11 @@ export const Header = () => {
         </div>
         <Link href="/" passHref className="hidden lg:flex items-center gap-2 ml-4 mr-6 shrink-0">
           <div className="flex relative w-10 h-10">
-            <Image alt="SE2 logo" className="cursor-pointer" fill src="/logo.svg" />
+            <Image alt="SE2 logo" className="cursor-pointer" fill src={projectConfig.iconPathHeader} />
```
rin-st, line 94: I don't see a reason to change logo path. Why just not remove default `logo.svg` and add a new one? It also helps to not forget about removing old one

packages/nextjs/components/Footer.tsx
```diff
@@ -58,17 +58,17 @@ export const Footer = () => {
...
+                <span className="link">{projectConfig.creator}</span>
               </a>
             </div>
             <span>·</span>
             <div className="text-center">
-              <a href="https://t.me/joinchat/KByvmRe5wkR-8F_zz6AjpA" target="_blank" rel="noreferrer" className="link">
+              <a href={projectConfig.supportLink} target="_blank" rel="noreferrer" className="link">
```
rin-st, line 71: Not sure lot's of projects has tg chats. Probably we need to remove or at least rename that link. But discuss about it first

rin-st (commented):
Thank you for PR @JacobHomanics !

Good idea, but I'm also thinking it's not worth it.

For me can make sense only these variables, see my other comments to understand why

```
  name: string;
  repoLink: string;
  nameHeader: string; // logoName?
  descriptionHeader: string; // logoDescription?
```
and I'm still doubt about logo vars, it's not a very big problem to change it in place. So as a result we have too small amount of vars we can move to config and hence looks like it's easier to just change them in place

technophile-04:
So I think the summary would be: 

1. We remove the Support link from footer we don't need it in every instance. 
2. We just put `description`, `name`  variable in `scaffold.config.ts`. 
   - We use that `name` here   `%s | ${scaffoldConfig.name}`
   https://github.com/scaffold-eth/scaffold-eth-2/blob/7d373d2c725ddb32afe41b0f0571c5f7c89d5083/packages/nextjs/utils/scaffold-eth/getMetadata.ts#L6
   - And in `layout` for  `getMetaData` too
   - we use `name` in Header logo 
   
3.  And let's revert all other changes 🙌

PR #1005, opened 2024-12-03
move 99_deployScript as hardhat task

technophile-04:
### Description: 

This is something which we did at CTF repo. Doing this no more requires `99_deployScript` to be present and also avoids the problem people deleting it by mistake

### DIFF
--- a/packages/hardhat/hardhat.config.ts
+++ b/packages/hardhat/hardhat.config.ts
@@ -9,6 +9,8 @@ import "solidity-coverage";
 import "@nomicfoundation/hardhat-verify";
 import "hardhat-deploy";
 import "hardhat-deploy-ethers";
+import { task } from "hardhat/config";
+import generateTsAbis from "./scripts/generateTsAbis";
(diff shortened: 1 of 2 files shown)


carletex (approved):
Smart tweak :)

rin-st (approved):

technophile-04:
Thanks guys! Merging this 🙌

damianmarti:
@technophile-04 nice!! 

So the generateTsAbis script is run always you deploy now and you don't have to add the generateTsAbis tag anymore (it does not exist anymore), right? The abis are always generated after any deployment...

technophile-04:
> you don't have to add the generateTsAbis tag anymore (it does not exist anymore), right? 

yeah that's the main reason 🙌 So now if people want to run a particular deploy script then can just do `yarn deploy --tags solution2` earlier they had to mention `generateTsAbis` tag too like `yarn deploy --tags solution2, generateTsAbis`

PR #1008, opened 2024-12-05
Encrypt deployer PK on .env file (when using hardhat)

carletex:
TODO:

- [x] `yarn account:import`

**Update 1**: Check https://github.com/scaffold-eth/scaffold-eth-2/pull/1008#issuecomment-2523846288. 

OG comment:
---

We have been talking about this for a while, and I'm drafting this PR to start the discussion and tinker. I think Hardhat v3 will come with a proper keystore (as foundry) but for now let's see if we can come with an easy solution.

I've been trying a bunch of things, but I've been finding roadblocks (maybe it's a good excuse to do our own simple package :D)

- Tried https://github.com/smartcontractkit/env-enc/, but I don't really like the workflow of it
- Tried https://github.com/tka85/dotenvenc, but required a uncrypted file to encrypt. I don't want to write to disk (even if it's for a moment) the unencrypted data... tried with their [encryptValuesOnly](https://github.com/tka85/dotenvenc/blob/master/src/index.ts#L164) but then it was creating a "readable" file (which is not totally safe). We could do "in-memory files" but wanted to avoid the extra complexity. Happy to reconsider.
- Tried https://dotenvx.com/, but it has a file with the "keys" to encrypt so it defeats the purpose.

Let's be open about other options too... but I'm here presenting a simpler approach (which is not working yet, lol)

## Goal

My goal is to have a simple password encryption system, so I went for the easies option: keep using the same `.env` file but store the encrypted account.

- :heavy_check_mark:  `yarn generate` prompts for a password* (+ repeat to validate), generates a random account as before, and saves the encrypted data** on `.env`
```.env
DEPLOYER_PRIVATE_KEY={"address":"634fe4716de4cdce1b0f0f1538af9385305268ba","id":"a5193ccb-f08e-45ea-a45b-c929519c166c","version":3,"Crypto":{"cipher":"aes-128-ctr","cipherparams":{"iv":"5d6a4a86f7630419c93922332a65157f"},"ciphertext":"32bc2935a5339d62f6adddca1f94c7542b46ac9e779de7ad30922e7de2aea661","kdf":"scrypt","kdfparams":{"salt":"6ba9acac9f3427b404e160ed9b8e91c02594f51dafca357c9964b226003f2f22","n":131072,"dklen":32,"p":1,"r":8},"mac":"4870c238eeda8d5efd2a40e60b66f97df5fcc1af3c36b14d9d4207e596a0c9ca"},"x-ethers":{"client":"ethers/6.13.2","gethFilename":"UTC--2024-12-05T18-03-35.0Z--634fe4716de4cdce1b0f0f1538af9385305268ba","path":"m/44'/60'/0'/0/0","locale":"en","mnemonicCounter":"befb5d31b92467e64f4835cb59934db8","mnemonicCiphertext":"4f3bc210655d7f2b8407c159df9d0b10","version":"0.1"}}
```
image

- :heavy_check_mark: `yarn account` prompts for a password and shows the account info as before
image

- :red_circle: `hardhat.config.ts` needs to be able to prompt for the password (at least when --network is not hardhat) so it's able to parse the PK. This is not working at the moment (so I removed `process.env.DEPLOYER_PRIVATE_KEY for now, so we can test the other commands). I'll continue to tinker, but open to ideas!! (so far, unsuccessful haha)

`*` I'm using `@inquirer/password` since it's really small
`**` I'm using wallet encrypt from ethers since we are already using ethers.

We'd also need a `yarn account:import` in case you want to import your PK (prompts for your PK and password)

What do you guys think?


damianmarti:
We can encrypt the PK using AES from the crypto library and save the encrypted PK. Then, when the user runs a yarn account or yarn deploy, we must ask for the password first, decrypt the PK, and set the DEPLOYER_PRIVATE_KEY env var with the decrypted PK. Am I missing something?

carletex:
> We can encrypt the PK using AES from the crypto library and save the encrypted PK. Then, when the user runs a yarn account or yarn deploy, we must ask for the password first, decrypt the PK, and set the DEPLOYER_PRIVATE_KEY env var with the decrypted PK. Am I missing something?

I think if you read carefully what I wrote and the current code, you'll see that what are you saying is the same thing that I'm trying.

Most of the stuff is already working (password encryption/decryption, yarn generate, yarn account), but there is an issue with how hardhat config works. You can check it out if you try it locally. Feel free to suggest any code if you make it work!

I'm currently trying an extra script to be run before any hardhat command to make it work (not the most fancy way) and load the env var.

carletex:
> Yes, this was what I was thinking about. I think this is the easier way because it should be transparent to hardhat what we are doing.

I think this might be the only way... And maybe that's why https://github.com/smartcontractkit/env-enc/ has a weird workflow (they couldn't find a way around hardhat config)

I'll update in a bit. Thanks Damu!!

carletex:
Update: pushed https://github.com/scaffold-eth/scaffold-eth-2/pull/1008/commits/3ae16c4407f77479fe889f2b7ad537992b9c7ec4 and it seems to be working.

This is the simplest way that I found to load the decrypted PK in memory and run everything as usual. I think `deploy` (when a network != hardhat/localhost) is the only place where we actually need the PK. (well maybe for hardhat scripts too).

So everything runs as always, but when you deploy it check if the network is specified. If so, decrypt the PK, load it into the env, spaw a process and run the deploy command as usual. 

Added a ToDo on the top PR comment (yarn account:import)

Please test, and let me know if you find a better way or if I'm missing something. Also @Pabl0cks , could you check on Windows?

packages/hardhat/scripts/runHardhatDeployWithPK.ts
```diff
@@ -0,0 +1,54 @@
...
+import password from "@inquirer/password";
+import { spawn } from "child_process";
+
+/**
+ * Unencrypts the private key and runs the hardhat deploy command
+ */
+async function main() {
+  const isNetworkProvided = process.argv.slice(2).includes("--network");
+
+  if (!isNetworkProvided) {
```
carletex, line 13: Maybe we want to check that if network is specified, it's different that "hardhat"

damianmarti: Yes, it will be a safe check.

rin-st:
Tested it, works great to me!

I'll play with it next days but for now I think it's a good solution!

damianmarti:
> Update: pushed [3ae16c4](https://github.com/scaffold-eth/scaffold-eth-2/commit/3ae16c4407f77479fe889f2b7ad537992b9c7ec4) and it seems to be working.
> 
> This is the simplest way that I found to load the decrypted PK in memory and run everything as usual. I think `deploy` (when a network != hardhat/localhost) is the only place where we actually need the PK. (well maybe for hardhat scripts too).
> 
> So everything runs as always, but when you deploy it check if the network is specified. If so, decrypt the PK, load it into the env, spaw a process and run the deploy command as usual.
> 
> Added a ToDo on the top PR comment (yarn account:import)
> 
> Please test, and let me know if you find a better way or if I'm missing something. Also @Pabl0cks , could you check on Windows?

Tested and it's working great!!

This will force to use the encrypted PK, right? If I want to use a wallet from a PK I won't be able to do it, right?

carletex:
> This will force to use the encrypted PK, right? If I want to use a wallet from a PK I won't be able to do it, right?

Not yet! But I plan to add a `yarn account:import`command, which will let you add your PK (will ask for PK + password and will encrypt it)

I'll tackle it (+ little tweaks) next week!

Also open to any other feedback. Thanks!!

damianmarti:
> We'd also need a `yarn account:import` in case you want to import your PK (prompts for your PK and password)

Yeah, sorry again, it was there on the PR text from the beginning. I think I didn't sleep well last night :-)

packages/hardhat/hardhat.config.ts
```diff
@@ -13,8 +13,9 @@ import { task } from "hardhat/config";
 import generateTsAbis from "./scripts/generateTsAbis";
 
 // If not set, it uses the hardhat account 0 private key.
+// You can generate a random account with `yarn generate` or `yarn account:import` to import your existing PK
 const deployerPrivateKey =
-  process.env.DEPLOYER_PRIVATE_KEY ?? "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
+  process.env.__RUNTIME_DEPLOYER_PRIVATE_KEY ?? "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
```
carletex, line 18: using this internal/weird looking env var so people won't use it

packages/hardhat/scripts/listAccount.ts
```diff
@@ -3,17 +3,26 @@ dotenv.config();
 import { ethers, Wallet } from "ethers";
 import QRCode from "qrcode";
 import { config } from "hardhat";
+import password from "@inquirer/password";
 
 async function main() {
-  const privateKey = process.env.DEPLOYER_PRIVATE_KEY;
+  const encryptedKey = process.env.DEPLOYER_PRIVATE_KEY_ENCRYPTED;
```
carletex, line 9: We could actually list the account without decrypting, since in generate JSON from ethers (where the PK is encrypted) you can get the Ethereum address.

But decided you keep it like this... in case someone forgets their password and send ETH directly.

packages/hardhat/scripts/generateAccount.ts
carletex, line 1: There is some shared functionality in all of these account-related scripts. I started to create a util file.... but it seemed unnecessary so I reverted back. But happy to reconsider.

carletex:
Ok, this is ready to review! 

Pushed the `yarn account:import` script + some tweaks on the copy/comments/env var names.

Please test and let me know if I'm missing something. Left some comments in the code too

A couple of things that were on my mind.
1. Harhdat scripts won't work out of the box (because you have to decrypt the PK before, as we do in `runHardhatDeployWithPK`. I think it's not a big deal (since people doing hardhat scripts are not beginners)
2. Give a warning if we detect `DEPLOYER_PRIVATE_KEY` env var. I think it's not worth it...but was in my mind :D

packages/hardhat/scripts/generateAccount.ts
rin-st: voting for moving it to util file, so it will be easier to update later

packages/hardhat/scripts/listAccount.ts
rin-st: > But decided you keep it like this... in case someone forgets their password and send ETH directly

I don't get what you mean here. If someone forgets the password, can he see the balances?

packages/hardhat/scripts/generateAccount.ts
```diff
@@ -1,42 +1,55 @@
...
 
-/**
- * Generate a new random private key and write it to the .env file
- */
-const setNewEnvConfig = (existingEnvConfig = {}) => {
-  console.log("👛 Generating new Wallet");
+const getValidatedPassword = async () => {
+  while (true) {
+    const pass = await password({ message: "Enter a password to encrypt your private key:", mask: true });
+    const confirmation = await password({ message: "Confirm password:", mask: true });
```
rin-st, line 11: not sure we need we need a mask for passwords. If someone sees for example that my password is 4 symbols it's relatively easy to find it

packages/hardhat/package.json
```diff
@@ -3,12 +3,14 @@
   "version": "0.0.1",
   "scripts": {
     "account": "hardhat run scripts/listAccount.ts",
+    "account:import": "hardhat run scripts/importAccount.ts",
+    "account:generate": "hardhat run scripts/generateAccount.ts",
```
rin-st, line 7: intuitively at the first try I entered `yarn account:generate <my_pk>` I think we need to add info how to use it to the docs. Since if people will use this `yarn account:generate <my_pk>` too, they will have their private key in terminal history. Also, we can mention that they need to clear clipboard after copying their PK to `account:generate`

rin-st (approved):
Works great for me!

Added some nipticks/comments, but can be ignored or changed later. Thanks!

package.json
```diff
@@ -10,13 +10,15 @@
   },
   "scripts": {
     "account": "yarn workspace @se-2/hardhat account",
+    "account:import": "yarn workspace @se-2/hardhat account:import",
+    "account:generate": "yarn workspace @se-2/hardhat account:generate",
```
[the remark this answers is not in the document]
damianmarti: Yes, but we will need to change README and docs...

packages/hardhat/scripts/generateAccount.ts
damianmarti: The advantage of using a mask is that the user gets better feedback while typing, right?

package.json
```diff
@@ -10,13 +10,15 @@
...
     "chain": "yarn workspace @se-2/hardhat chain",
     "fork": "yarn workspace @se-2/hardhat fork",
     "deploy": "yarn workspace @se-2/hardhat deploy",
     "verify": "yarn workspace @se-2/hardhat verify",
     "hardhat-verify": "yarn workspace @se-2/hardhat hardhat-verify",
     "compile": "yarn workspace @se-2/hardhat compile",
-    "generate": "yarn workspace @se-2/hardhat generate",
+    "generate": "yarn account:generate",
```
damianmarti, line 21: Oh, it works with generate too. Nice.

package.json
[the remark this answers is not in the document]
damianmarti: Nevermind.

damianmarti (approved):
It's working great!! And the code looks good to me.

packages/hardhat/scripts/generateAccount.ts
carletex: > that my password is 4 symbols it's relatively easy to find it

yep, but they also need to steel your .env file :D

> The advantage of using a mask is that the user gets better feedback

Yes, this is why I added it.

But happy to reconsider

packages/hardhat/scripts/listAccount.ts
carletex: If you forget your password:

- you run `yarn account` and it'll ask for your password. So you won't send anything
- If we don't require a password, you might send it... and then (e.g. when you deploy) you'd realize that you don't remember the password and the funds might be lost.

That was the only reason I did it like this.

Happy to reconsider.

packages/hardhat/package.json
carletex: Agree, we need to update the docs!

Pabl0cks (commented):
Awesome job @carletex !! This looks great!

On windows, as usual, there is some weird stuff, but I think it's easily fixed!

- `yarn account` works as expected. My git bash is a bit weird because it shows the raw password when you type it, but I will try on my laptop (the console on my desktop it's always been a bit buggy). On Windows Powershell the password is always hided 👌
- `yarn deploy` throws error, but applied this [solution ](https://stackoverflow.com/questions/37459717/error-spawn-enoent-on-windows) and works great.

The error was: 
```
Error: spawn hardhat ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:286:19)
    at onErrorNT (node:internal/child_process:484:16)
    at processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawn hardhat',
  path: 'hardhat',
  spawnargs: [ 'deploy' ]
}
```
Adding `shell` to both `spawn`  in `runHardhatDeployWithPK.ts` when it's windows, fixed it, and hopefully doesn't break the rest of OS behaviours.

```
stdio: "inherit",
env: process.env,
shell: process.platform === 'win32'

packages/hardhat/scripts/listAccount.ts
technophile-04: > But decided you keep it like this... in case someone forgets their password and send ETH directly.

Yeah makes sense! and in foundry also we have same worksflow

packages/hardhat/scripts/generateAccount.ts
technophile-04: Umm not sure if its worth it. Lol don't want create an extra dir in `hardhat` package for our internal usage. 

Maybe we can `export` some functions from `generateAccount` file and use them inside `importAccount` instead of creating a new util file?

packages/hardhat/scripts/generateAccount.ts
technophile-04: I think maybe we hide it? 

similar to foundry workflow and also because of https://github.com/scaffold-eth/scaffold-eth-2/pull/1008#pullrequestreview-2491802349

well we don't need to follow foundry, but also due Pablos facing that issue its fine we hide it

technophile-04 (approved):
Tysm @carletex this is great! And works nicely, tested with different cases like `--tags`, import pk + generating and deployments in general everything seems to work smooth! 

Hopefully it works nicely on windows and then update the docs 🙌

packages/hardhat/scripts/generateAccount.ts
rin-st: It's not a big deal for now and all 3 options are ok, we can always change it later if it will become more complex. Yes, let's export for now (but I'd prefer new file :D)

packages/hardhat/scripts/listAccount.ts
rin-st: > That was the only reason I did it like this.

Got it, thanks!

packages/hardhat/package.json
rin-st: created an issue https://github.com/scaffold-eth/se-2-docs/issues/103

packages/hardhat/scripts/generateAccount.ts
carletex: yeah, I think it makes sense to follow foundry and also avoiding the error on Windows.

Removed the mask for all the inputs (PK & password) d976c0a2c58d

packages/hardhat/package.json
carletex: Perfect, thanks!

packages/hardhat/scripts/generateAccount.ts
carletex: I just started exporting from generateAccount... but I didn't like where I was going. I prefer having the code duped (but easy to understand) than confusing. We are only duplicating the `getValidatedPassword` (and we could make `setNewEnvConfig` generic too)

I think we leave it as it's... or...

we could create an `account` util with all the functions (`getValidatedPassword`, `getWalletFromPrivateKey`, and maybe `setNewEnvConfig`) and we use them from each of the main scripts.

But since this is just a refactor optimization... maybe in another PR?

carletex:
I knew it was going to fail on Windows! haha

Pushed https://github.com/scaffold-eth/scaffold-eth-2/pull/1008/commits/c8b08a797d955fbc276eb9e2af2eee634d3c764e. Thanks @Pabl0cks! It works for me. Let's see if it works for the MacOS guys too :)

There are a couple of open things:
- Update docs: https://github.com/scaffold-eth/se-2-docs/issues/103. 
- Refactor thingy: https://github.com/scaffold-eth/scaffold-eth-2/pull/1008#discussion_r1876362537. If we want to do this I think we could save it for another PR.

packages/hardhat/scripts/generateAccount.ts
rin-st: > But since this is just a refactor optimization... maybe in another PR?

Yes, I agree. Or leave it as is if we can't find a way everyone agree with

PR #1010, opened 2024-12-08
reverse rpc fallback order

technophile-04:
## Description

I am not sure what's the best solution for this is, but realized that public RPC URL's are not reliable. Especially recognized this with sepolia public rpc url which wagmi has `https://rpc2.sepolia.org`. 

It seems to read the wrong state example balance and also not able to fetch read variables. 

<details>

<summary>
    Demo video:     
</summary>

https://github.com/user-attachments/assets/05f787a6-3ad2-4047-85eb-9359187b080a

</details>

## To test the problem: 

1. Switch to main branch.

2. Copy this inside `externalContracts.ts`

<details>

<summary>
    externalContracts.ts
</summary>

```ts

/**
 * @example
 * const externalContracts = {
 *   1: {
 *     DAI: {
 *       address: "0x...",
 *       abi: [...],
 *     },
 *   },
 * } as const;
 */
const externalContracts = {
  11155111: {
    YourContract: {
      address: "0x4E439fde29737bD2B8a89CC2C2dDf9926c624194",
      abi: [
        {
          inputs: [
            {
              internalType: "address",
              name: "_owner",
              type: "address",
            },
          ],
          type: "constructor",
        },
        {
          anonymous: false,
          inputs: [
            {
              indexed: true,
              internalType: "address",
              name: "greetingSetter",
              type: "address",
            },
            {
              indexed: false,
              internalType: "string",
              name: "newGreeting",
              type: "string",
            },
            {
              indexed: false,
              internalType: "bool",
              name: "premium",
              type: "bool",
            },
            {
              indexed: false,
              internalType: "uint256",
              name: "value",
              type: "uint256",
            },
          ],
          name: "GreetingChange",
          type: "event",
        },
        {
          inputs: [],
          name: "greeting",
          outputs: [
            {
              internalType: "string",
              name: "",
              type: "string",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "owner",
          outputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "premium",
          outputs: [
            {
              internalType: "bool",
              name: "",
              type: "bool",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "string",
              name: "_newGreeting",
              type: "string",
            },
          ],
          name: "setGreeting",
          outputs: [],
          stateMutability: "payable",
          type: "function",
        },
        {
          inputs: [],
          name: "totalCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          name: "userGreetingCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "withdraw",
          outputs: [],
          type: "function",
        },
        {
          stateMutability: "payable",
          type: "receive",
        },
      ],
      inheritedFunctions: {},
    },
  },
} as const;

```

</details>

3. Go to http://localhost:3000/debug

### Solution: 

We use alchmey URL first and then fallback to public RPC. Well I mean this is gonna burst the scaffold alchemy rpc but not sure what's the best solution here is.

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -31,7 +31,7 @@
     "react-dom": "^18.3.1",
     "react-hot-toast": "^2.4.0",
     "usehooks-ts": "^3.1.0",
-    "viem": "^2.21.54",
+    "viem": "2.21.54",
     "wagmi": "2.12.23",
     "zustand": "^5.0.0"
   },
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -32,7 +32,7 @@
     "react-hot-toast": "^2.4.0",
     "usehooks-ts": "^3.1.0",
     "viem": "2.21.54",
-    "wagmi": "2.12.23",
+    "wagmi": "2.13.4",
     "zustand": "^5.0.0"
   },
   "devDependencies": {
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -31,7 +31,7 @@
     "react-dom": "^18.3.1",
     "react-hot-toast": "^2.4.0",
     "usehooks-ts": "^3.1.0",
-    "viem": "2.21.32",
+    "viem": "^2.21.54",
     "wagmi": "2.12.23",
     "zustand": "^5.0.0"
   },
--- a/packages/nextjs/scaffold.config.ts
+++ b/packages/nextjs/scaffold.config.ts
@@ -8,6 +8,8 @@ export type ScaffoldConfig = {
   onlyLocalBurnerWallet: boolean;
 };
 
+export const DEFAULT_ALCHEMY_API_KEY = "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF";
+
 const scaffoldConfig = {
   // The networks on which your DApp is live
   targetNetworks: [chains.hardhat],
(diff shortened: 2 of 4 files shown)


carletex:
> but not sure what's the best solution here is.

Not sure either, but sharing some thoughts here as well. 

I think the Sepolia one is the most problematic (noticed this when trying to deploy from hardhat when we had the public RPCs... it was always failing giving weird errors). The problem is that the RPC is not _failing_ (where the fallback would work)... it's just returning incorrect stuff.

I guess **wevm** people noticed it too because they changed it on `viem` last month https://github.com/wevm/viem/commit/79a183046b92f90d3e8ce707fe1e28b4523d7e5e. I guess it'd come with the newer versions of wagmi/view.

A couple of solution could be:
1. If this is mainly Sepolia, let's just update and leave the RPC order as it was.
   - PROS: 
     - Our alchemy key won't suffer
   - CONS: 
       - public RPCs tend to fail
       - If people change to use their alchemy key... it'd still fail because it's the fallback (A weird hack would be to detect if a different API is being used... if so, make alchemy the default)
2. If it's a general thing, let's tweak the order. 
   - PROS:
     - Works most of the time (unless limits are hit)
   - CONS:
     - People killing our RPC

technophile-04:
> If people change to use their alchemy key... it'd still fail because it's the fallback (A weird hack would be to detect if a different API is being used... if so, make alchemy the default)

Love this! Makes sense becoz if people add their API key they would accept more reliable result and SE-2 was still using public RPC URL first (which might give wrong data) 

Also updated the viem and wagmi so that we the default rpc from viem works nicely out of the box

carletex (approved):
I think this is a great solution. Thanks Shiv!!

rin-st:
Working and looking great to me, thanks! Just added a micro change, moved default alchemy API key to the variable so it could be changed from one place

rin-st (approved):

technophile-04:
> Working and looking great to me, thanks! Just added a micro change, moved default alchemy API key to the variable so it could be changed from one place

Niceee and makes sense tysm!! Merging this 🙌

PR #1012, opened 2024-12-14
Disable vercel telemetry

rin-st:
Context: https://github.com/scaffold-eth/se-2-challenges/pull/245#pullrequestreview-2503416298

I think we need to disable telemetry by default, because
- vercel cli asks to do additional actions to approve telemetry before deploy
- I am not sure people want to vercel to collect their data

Vercel telemetry [docs](https://vercel.com/docs/cli/about-telemetry)


technophile-04:
Thankss! Yeah I think it's fine and makes sense to diable it 🙌

technophile-04 (approved):

Issue #1013, opened 2024-12-15
bug: useScaffoldEventHistory duplicates events data


rin-st:
### Is there an existing issue for this?

- [X] I have looked through the [existing issues](https://github.com/scaffold-eth/scaffold-eth-2/issues)

### Which method was used to setup Scaffold-ETH 2 ?

git clone

### Current Behavior

Part of the data could be duplicated when using `useScaffoldEventHistory`

Reason: 
we're using `blockNumber` from `useBlockNumber` inside `getNextPageParam` of `useScaffoldEventHistory`. But `blockNumber` is not cached and refreshes if you use new instance `useScaffoldEventHistory`, for example from another page. 

So for example you're setting events from first page and use `useScaffoldEventHistory` from that page `pageParams` could be for example `0n, 2n, 3n` and everything works as expected. But if you switch to another page with `useScaffoldEventHistory` for the same events, your page `pageParams` will be something like `0n, 3n, 3n`

### Expected Behavior

Events data is not duplicated

### Steps To Reproduce

- change `app/page.tsx` code to

<details><summary>Details</summary>
<p>

```
"use client";

import { useState } from "react";
import Link from "next/link";
import type { NextPage } from "next";
import { useAccount } from "wagmi";
import { BugAntIcon, MagnifyingGlassIcon } from "@heroicons/react/24/outline";
import { Address } from "~~/components/scaffold-eth";
import { useScaffoldEventHistory, useScaffoldWriteContract } from "~~/hooks/scaffold-eth";

const Home: NextPage = () => {
  const { address: connectedAddress } = useAccount();

  const [newGreeting, setNewGreeting] = useState("");

  const { writeContractAsync: writeSetGreeting } = useScaffoldWriteContract("YourContract");

  const handleSetGreeting = async () => {
    try {
      await writeSetGreeting({
        functionName: "setGreeting",
        args: [newGreeting],
        value: 0n,
      });
      setNewGreeting(""); // Clear input after successful transaction
    } catch (error) {
      console.error("Error setting greeting:", error);
    }
  };

  const { data: greetingChangeEvents } = useScaffoldEventHistory({
    contractName: "YourContract",
    eventName: "GreetingChange",
    fromBlock: 0n,
    watch: true,
  });

  console.log(greetingChangeEvents);

  return (
    <>
      <div className="flex items-center flex-col flex-grow pt-10">
        <div className="px-5">
          <h1 className="text-center">
[...]

PR #1014, opened 2024-12-15
Fix useScaffoldEventHistory duplicated events

rin-st:
See description and steps to reproduce here https://github.com/scaffold-eth/scaffold-eth-2/issues/1013

Fixes https://github.com/scaffold-eth/scaffold-eth-2/issues/1013


rin-st:
+ fixed types since we don't have logs in our events

see https://github.com/scaffold-eth/scaffold-eth-2/blob/74d2baa59c01de2eebf6a6ef2466bbef7d16ec8c/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts#L35-L37

technophile-04 (approved):
This is great! Thanks Rinat for repro example and fix! works nicely and makes sense!!

PR #931, opened 2024-09-09
Optional chain parameter in Scaffold hooks

JacobHomanics:
## Description

Provides the option to pass a different network into the scaffold hooks than the one that the user's wallet  is currently connected to. This allows for applications that can show data from different blockchains than that of which the user's wallet is connected to. The hooks still abide by the rules that the contract instance's need to be within the `deployedContracts` file and `targetNetworks` needs to have the expected chain(s) present.

live example: https://scaffold-eth-optional-chains-example-nextjs.vercel.app/
live example repo: https://github.com/Hotmanics/scaffold-eth-optionalChainsExample
## Additional Information

- [x] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [x] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

Your ENS/address: jacobhomanics.eth

PR #1015, opened 2024-12-17
make `useScaffoldWriteContract` & `useDeployedContractInfo` backward compatible

technophile-04:
### Description: 

This allows developers to consume `useScaffoldWriteContract` and `useDeployedContractInfo` hook in both ways until we make breaking changes release : 

Example of `useScaffoldWriteContract`

Old: 
```ts
  const { writeContractAsync: depeRecatedWriteAsync } = useScaffoldWriteContract("YourContract");
```

When using the Old way the IDE should give the warning : 

Screenshot 2024-12-17 at 2 12 08 PM

New: 
```ts
  const { writeContractAsync: newWriteAsync } = useScaffoldWriteContract({
    contractName: "YourContract",
    chainId: 31337,
  });
```

### To test: 

<details>

<summary>
    Copy this in page.tsx
</summary>

```tsx
"use client";

import { useAccount } from "wagmi";
import { Address } from "~~/components/scaffold-eth";
import { useDeployedContractInfo, useScaffoldReadContract, useScaffoldWriteContract } from "~~/hooks/scaffold-eth";

  const { address: connectedAddress } = useAccount();
  const { data: depecatedContractInfo } = useDeployedContractInfo("YourContract");
  const { data: newContractInfo } = useDeployedContractInfo({ contractName: "YourContract", chainId: 31337 });

  console.log({ depecatedContractInfo, newContractInfo });

  const { writeContractAsync: depeRecatedWriteAsync } = useScaffoldWriteContract("YourContract");

  const { writeContractAsync: newWriteAsync } = useScaffoldWriteContract({
    contractName: "YourContract",
    chainId: 31337,
  });

  const { data: currGreeting } = useScaffoldReadContract({
    contractName: "YourContract",
    functionName: "greeting",
  });

  const handleWriteWithDeperecated = async () => {
    try {
      await depeRecatedWriteAsync({
        functionName: "setGreeting",
        args: ["With deprecated write!"],
      });
    } catch {}
  };

  const handleNewWrite = async () => {
    try {
      await newWriteAsync({
        functionName: "setGreeting",
        args: ["With new write!"],
      });
    } catch {}
  };

  return (
    <>
        <div className="px-5">
          <div className="flex justify-center items-center space-x-2 flex-col sm:flex-row">
            <p className="my-2 font-medium">Connected Address:</p>
          </div>
        </div>
        <p>{currGreeting}</p>
        <div className="flex flex-col space-y-3">
          <button onClick={handleWriteWithDeperecated} className="btn btn-primary">
            Write with Deperecated
          </button>

          <button onClick={handleNewWrite} className="btn btn-primary">
            Write with New
          </button>
        </div>
      </div>
    </>
  );
};

export default Home;
```

</details>

### Question: 

Shall we also `console.warn("Warning: Using useScaffoldWriteContract with a string parameter is deprecated. Please use the object parameter version instead")` log this warning in browser? if people use string param?

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldWriteContract.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldWriteContract.ts
@@ -1,7 +1,7 @@
 import { useState } from "react";
 import { MutateOptions } from "@tanstack/react-query";
 import { Abi, ExtractAbiFunctionNames } from "abitype";
-import { Config, useAccount, useWriteContract } from "wagmi";
+import { Config, UseWriteContractParameters, useAccount, useWriteContract } from "wagmi";
 import { WriteContractErrorType, WriteContractReturnType } from "wagmi/actions";
 import { WriteContractVariables } from "wagmi/query";
 import { useSelectedNetwork } from "~~/hooks/scaffold-eth";
@@ -15,6 +15,34 @@ import {
   UseScaffoldWriteConfig,
 } from "~~/utils/scaffold-eth/contract";
 
+type ScaffoldWriteContractReturnType<TContractName extends ContractName> = Omit<
+  ReturnType<typeof useWriteContract>,
+  "writeContract" | "writeContractAsync"
+> & {
+  isMining: boolean;
+  writeContractAsync: <
+    TFunctionName extends ExtractAbiFunctionNames<ContractAbi<TContractName>, "nonpayable" | "payable">,
+  >(
+    variables: ScaffoldWriteContractVariables<TContractName, TFunctionName>,
+    options?: ScaffoldWriteContractOptions,
+  ) => Promise<WriteContractReturnType | undefined>;
+  writeContract: <TFunctionName extends ExtractAbiFunctionNames<ContractAbi<TContractName>, "nonpayable" | "payable">>(
+    variables: ScaffoldWriteContractVariables<TContractName, TFunctionName>,
+    options?: Omit<ScaffoldWriteContractOptions, "onBlockConfirmation" | "blockConfirmations">,
+  ) => void;
+};
+
+export function useScaffoldWriteContract<TContractName extends ContractName>(
+  config: UseScaffoldWriteConfig<TContractName>,
+): ScaffoldWriteContractReturnType<TContractName>;
+/**
+ * @deprecated Use object parameter version instead: useScaffoldWriteContract({ contractName: "YourContract" })
+ */
+export function useScaffoldWriteContract<TContractName extends ContractName>(
+  contractName: TContractName,
+  writeContractParams?: UseWriteContractParameters,
+): ScaffoldWriteContractReturnType<TContractName>;
+
 /**
  * Wrapper around wagmi's useWriteContract hook which automatically loads (by name) the contract ABI and address from
  * the contracts present in deployedContracts.ts & externalContracts.ts corresponding to targetNetworks configured in scaffold.config.ts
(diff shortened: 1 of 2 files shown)


packages/nextjs/hooks/scaffold-eth/useScaffoldWriteContract.ts
```diff
@@ -15,23 +15,56 @@ import {
...
-  writeContractParams,
-}: UseScaffoldWriteConfig<TContractName>) => {
+export function useScaffoldWriteContract<TContractName extends ContractName>(
+  configOrName: UseScaffoldWriteConfig<TContractName> | TContractName,
+  writeContractParams?: UseWriteContractParameters,
+): ScaffoldWriteContractReturnType<TContractName> {
+  const finalConfig: UseScaffoldWriteConfig<TContractName> =
+    typeof configOrName === "string"
+      ? { contractName: configOrName, writeContractParams, chainId: undefined }
+      : (configOrName as any);
```
carletex, line 60: Do we need to specify the type here? It seems that it can be inferred.

If not, should it be `UseScaffoldWriteConfig<TContractName>`?

carletex (commented):
Good stuff @technophile-04 this is great!

Left a tiny comment.

> Shall we also console.warn("Warning: Using useScaffoldWriteContract with a string parameter is deprecated. Please use the object parameter version instead") log this warning in browser? if people use string param?

No strong opinion here. Up to you all.

P.S. I was trying to understand why we needed the explicit return type (introduced in #931) and not sure if we needed it there, but we definitely need it here because of the overloads (requires a function declaration syntax). So all good!

packages/nextjs/hooks/scaffold-eth/useScaffoldWriteContract.ts
technophile-04: If we remove the its being inffered as `any` lol it's fine but in next step when I was destructuring I wanted that autcompletions with keys. 

```suggestion
  const finalConfig =
    typeof configOrName === "string"
      ? { contractName: configOrName, writeContractParams, chainId: undefined }
      : configOrName;
``` 

Actually earlier while playing I had this. 

Inferring  + not asserting `any` at `configOrName`. 

But the problem is when `deployedContracts.ts` is empty. the type `TContractName` i.e `export type ContractName = keyof Contracts;` is inferred as `string | number` instead of `ContractName` being only `string`  [weird how TS works](https://stackoverflow.com/questions/57185972/typescript-infers-object-key-to-number-instead-of-string). 

Now since it is inferred  as `string | number` in the next step I wasn't able to destructure  it becoz it was giving error that `contractName, writeContractParams` doesn't exist on type `string | number`.

carletex: Damn, types.

Thanks for the explanation!! <3

technophile-04: > If not, should it be UseScaffoldWriteConfig<TContractName>?

Yeah just asserted the `configOrName` with it and now its being inferred at https://github.com/scaffold-eth/scaffold-eth-2/pull/1015/commits/3bef6823cc89805fce3b67e974584fa52006a28f

lol forgot to click "add comment" here

rin-st:
Lgtm! For vscode/cursor it even cross out deprecated variants!
image

> Shall we also console.warn("Warning: Using useScaffoldWriteContract with a string parameter is deprecated. Please use the object parameter version instead") log this warning in browser? if people use string param?

No strong opinion too, but agree with Damu

technophile-04:
Added the warning log feel free to suggest a better message!

damianmarti (approved):
@technophile-04 The console warn message looks good to me. Thanks!!

carletex (approved):

PR #1016, opened 2024-12-17
Move warnings to useEffect

rin-st:
Without this we show warnings on every render so all the console is yellow, try example from https://github.com/scaffold-eth/scaffold-eth-2/pull/1015


Pabl0cks (commented):
Thanks for tackling this Rinat! It was a bit annoying hah

For the write hook, maybe it would be better to show a warning everytime we execute the writeContractAsync?

rin-st:
> For the write hook, maybe it would be better to show a warning everytime we execute the `writeContractAsync`?

I'm not sure. Yes, if we want to make it very annoying

But I'm tending not to change it since devs already see one console message for every hook. And the amount of executions is not correlating with amount of places to change.

technophile-04 (approved):
Thanks @rin-st!!

PR #1022, opened 2024-12-20
Keystore Randomization

JacobHomanics:
## Description

We can improve keystore security by generating a unique one in every SE-2 application. 

When `yarn-generate` is called, it will create a unique name and private key for the user.

I will try to improve the documentation but would appreciate help with updating it!

## Additional Information

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,12 +71,12 @@ Visit your app on: `http://localhost:3000`. You can interact with your smart con
 yarn generate
 ```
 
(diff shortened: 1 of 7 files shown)


technophile-04:
https://github.com/scaffold-eth/scaffold-eth-2/blob/25f23f62e260ffa580887b2f2557b1e5db2dfa8b/.github/pull_request_template.md?plain=1#L14

Hey Jacob, I think its really important that you make discussions/issue first rather than jumping directly into PR's also looking into different PR's, issues and discussions before creating your own discussion might also give you and us more context on things. 

Closing this for now, not rejecting PR but maybe first we should discuss/brainstorm a bit.

PR #1027, opened 2025-01-06
Add vercel login script

### LINKED ISSUE #1025: Vercel login
I tried to deploy to Vercel from the cli in my new computer:

I got this (which is expected (and stated in the [docs](https://docs.scaffoldeth.io/deploying/deploy-nextjs-app))

But then how do I login?

With the current commands (in `package.json`) doesn't seem possible. If I install Vercel cli globally, then it works after login.

Should we add `yarn vercel:login`? Or am I missing something?

damianmarti:
Add Vercel login script to allow to login to Vercel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "test": "yarn hardhat:test",
     "vercel": "yarn workspace @se-2/nextjs vercel",
     "vercel:yolo": "yarn workspace @se-2/nextjs vercel:yolo",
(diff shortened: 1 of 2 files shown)


carletex (approved):
Let's merge this for now (so people can log in, especially when doing the challenges), but let's iterate with a more generic solutions.

Thanks Damu!

See #1025

PR #1042, opened 2025-01-27
Add forge clean to run it with yarn

phipsae:
## Description

This little addition let you run forge clean from the monorepo root with `yarn foundry:clean`. If you run the command, it removes all the build artifacts and cached files in the foundry directory.


technophile-04:
Tysm @phipsae!!

technophile-04 (approved):

PR #1043, opened 2025-01-27
Add hardhat clean to run it with yarn

phipsae:
## Description

This little addition let you run hardhat clean from the monorepo root with yarn hardhat:clean. If you run the command, it removes all the build artifacts and cached files in the foundry directory.


technophile-04 (approved):
Thanks @phipsae! Makes sense!!

PR #1049, opened 2025-02-10
Up wagmi, viem and rainbowkit

technophile-04:
### Description: 

Upgraded wagmi, viem, rainbowkit to latest version. 

Since wagmi/viem now by default removed the simulation of contract write we have to handle it now, check out https://github.com/wevm/wagmi/issues/4516

Initially, I thought I would add simulation logic to `useTransactor` but it seems there is no simulation hook / util for normal transaction in wagmi/viem and its only available for contract interaction.  So it made sense to add the logic in `useScaffoldWriteContract` itself. 

Also added a new option to `useScaffoldWriteContract` and `disbaleSimulate` if anyone wants to disable simulation. 

Please feel free to be nitpicky on naming / approach! 

--- 

### To test: 

<details>

<summary>
    Replace this in `externalContracts.ts`
</summary>

```ts

/**
 * @example
 * const externalContracts = {
 *   1: {
 *     DAI: {
 *       address: "0x...",
 *       abi: [...],
 *     },
 *   },
 * } as const;
 */
const externalContracts = {
  11155111: {
    YourContract: {
      address: "0x3A897AaAcFCdd13b6A5F2b9fb85F36AFFa49442c",
      abi: [
        {
          inputs: [
            {
              internalType: "address",
              name: "_owner",
              type: "address",
            },
          ],
          type: "constructor",
        },
        {
          anonymous: false,
          inputs: [
            {
              indexed: true,
              internalType: "address",
              name: "greetingSetter",
              type: "address",
            },
            {
              indexed: false,
              internalType: "string",
              name: "newGreeting",
              type: "string",
            },
            {
              indexed: false,
              internalType: "bool",
              name: "premium",
              type: "bool",
            },
            {
              indexed: false,
              internalType: "uint256",
              name: "value",
              type: "uint256",
            },
          ],
          name: "GreetingChange",
          type: "event",
        },
        {
          inputs: [],
          name: "greeting",
          outputs: [
            {
              internalType: "string",
              name: "",
              type: "string",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "owner",
          outputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "premium",
          outputs: [
            {
              internalType: "bool",
              name: "",
              type: "bool",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "string",
              name: "_newGreeting",
              type: "string",
            },
          ],
          name: "setGreeting",
          outputs: [],
          stateMutability: "payable",
          type: "function",
        },
        {
          inputs: [],
          name: "totalCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [
            {
              internalType: "address",
              name: "",
              type: "address",
            },
          ],
          name: "userGreetingCounter",
          outputs: [
            {
              internalType: "uint256",
              name: "",
              type: "uint256",
            },
          ],
          stateMutability: "view",
          type: "function",
        },
        {
          inputs: [],
          name: "withdraw",
          outputs: [],
          type: "function",
        },
        {
          stateMutability: "payable",
          type: "receive",
        },
      ],
      inheritedFunctions: {},
    },
  },
} as const;

```

</details>

<details>

<summary>
    Replace `page.tsx` with 
</summary>

```ts
"use client";

import { useScaffoldWriteContract } from "~~/hooks/scaffold-eth";

  const { writeContractAsync } = useScaffoldWriteContract({
    contractName: "YourContract",
    // disableSimulate: true,
  });

  const handleWithdraw = async () => {
    try {
      await writeContractAsync({
        functionName: "withdraw",
      });
    } catch (error) {
      console.error("Withdrawal failed", error);
    }
  };

  return (
    <>
        <button className="btn btn-primary" onClick={handleWithdraw}>
          Withdraw
        </button>
      </div>
    </>
  );
};

export default Home;
```

</details>

Updated `scaffold.config.ts` with `chains.sepolia`. 

The deployed contract is basic `YourContract` from SE-2 with owner being my account. So ideally withdraw function will fail saying "Not Owner" and won't give wallet pop-up at all. 

In `main` branch if you `cd packages/nextjs && yarn up wagmi viem` and try to withdraw from above branch it will give metamask pop up even when trascation will revert after sending.

### DIFF
--- a/packages/nextjs/app/debug/_components/contract/WriteOnlyFunctionForm.tsx
+++ b/packages/nextjs/app/debug/_components/contract/WriteOnlyFunctionForm.tsx
@@ -16,7 +16,7 @@ import {
 import { IntegerInput } from "~~/components/scaffold-eth";
 import { useTransactor } from "~~/hooks/scaffold-eth";
 import { useTargetNetwork } from "~~/hooks/scaffold-eth/useTargetNetwork";
-import { simulateContractWriteAndNotifyError } from "~~/utils/scaffold-eth/common";
+import { simulateContractWriteAndNotifyError } from "~~/utils/scaffold-eth/contract";
 
 type WriteOnlyFunctionFormProps = {
   abi: Abi;
--- a/packages/nextjs/app/debug/_components/contract/WriteOnlyFunctionForm.tsx
+++ b/packages/nextjs/app/debug/_components/contract/WriteOnlyFunctionForm.tsx
@@ -4,7 +4,9 @@ import { useEffect, useState } from "react";
 import { InheritanceTooltip } from "./InheritanceTooltip";
 import { Abi, AbiFunction } from "abitype";
 import { Address, TransactionReceipt } from "viem";
-import { useAccount, useWaitForTransactionReceipt, useWriteContract } from "wagmi";
+import { Config, useAccount, useConfig, useWaitForTransactionReceipt, useWriteContract } from "wagmi";
+import { simulateContract } from "wagmi/actions";
+import { WriteContractVariables } from "wagmi/query";
 import {
   ContractInput,
   TxReceipt,
@@ -16,6 +18,7 @@ import {
 import { IntegerInput } from "~~/components/scaffold-eth";
 import { useTransactor } from "~~/hooks/scaffold-eth";
 import { useTargetNetwork } from "~~/hooks/scaffold-eth/useTargetNetwork";
+import { getParsedError, notification } from "~~/utils/scaffold-eth";
 
 type WriteOnlyFunctionFormProps = {
   abi: Abi;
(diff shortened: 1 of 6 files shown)


packages/nextjs/hooks/scaffold-eth/useScaffoldWriteContract.ts
```diff
@@ -81,6 +83,16 @@ export function useScaffoldWriteContract<TContractName extends ContractName>(
...
 
+  const simulateAndNotifyError = async (params: WriteContractVariables<Abi, string, any[], Config, number>) => {
+    try {
+      await simulateContract(wagmiConfig, params);
+    } catch (error) {
+      const parsedError = getParsedError(error);
+      notification.error(parsedError);
+      throw error;
+    }
+  };
```
technophile-04, line 94: We could have inlilned this in `sendContractWriteAsyncTx` function itself but notice the extra try catch here in this function. 

Couldn't inline it in `sendContractWriteAsyncTx` because if we show `notification.error` in `sendContractWriteAsyncTx` catch block then it might show error notification two times one when simulation failed other time incase the `useTransactor` fails. And even third time when user handles the error in his app.

packages/nextjs/app/debug/_components/contract/WriteOnlyFunctionForm.tsx
```diff
@@ -41,17 +44,31 @@ export const WriteOnlyFunctionForm = ({
 
   const { data: result, isPending, writeContractAsync } = useWriteContract();
 
+  const wagmiConfig = useConfig();
+
+  const simulateAndNotifyError = async (params: WriteContractVariables<Abi, string, any[], Config, number>) => {
```
rin-st, line 49: maybe add `wagmiConfig` as param an move to utils or keep in one file, since the function is duplicated?

technophile-04: Ohh yeah initially I thought since its being used in just two files so not worth it, but yeah makes sense to have it in util 🙌

Just moved it under `utils/scaffold-eth/contract.ts` lol but not 100% sure if it's the right place since that file contains all types but also makes sense to have that function in that file since its util function is related to contract interaction.

rin-st: Thanks! Probably we need to move types to some separated types file(s) later

technophile-04: Ohh yes 💯!! Has been in my mind always!

technophile-04:
Thanks Rinat mergin this 🙌

PR #1039, opened 2025-01-25
Use bgipfs for upload

azf20:
## Description

To upload a site to BuidlGuidl's IPFS node, in the nextjs package:

```
yarn ipfs:init
yarn ipfs
```

These commands wrap initialisation and upload commands from [bgipfs](https://www.npmjs.com/package/bgipfs). This uses a default proxy endpoint (no auth required at this time)

PR #833, opened 2024-05-01
chore: adding engines fields in package.json to hint people

wolfcito:
## Description

I have added 'engines' to the package.json file. This practice ensures consistency in the development environment, which is crucial when collaborating with colleagues who may have different versions of Node.js installed

## Additional Information

- [x] I have read the [contributing docs](/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md) (if this is your first contribution)
- [x] This is not a duplicate of any [existing pull request](https://github.com/scaffold-eth/scaffold-eth-2/pulls)

Your ENS/address: wolfcito.eth

PR #1052, opened 2025-02-21
Update node version requirements on Readme

Pabl0cks:
We just merged #1039 to add `yarn ipfs` using the new `bgipfs` package, but Windows users will have problems with SE-2 `yarn install` if they don't update to >20.18.1.

Changed required version to 20.18.3 on the Readme (closest lts version) so no users have problems with `bgipfs` package installation.


rin-st:
Hey Pablo, could you please also update main `package.json` engines?

technophile-04:
Ohh I think it would be nice to add the node version check in create-eth too like the CLI will exit directly if it finds the node version is less than our required. 

Like the engines field already does that but does not work for everyone as I remember it was not respected for Damu [here](https://github.com/scaffold-eth/scaffold-eth-2/pull/833#issuecomment-2100792421)

technophile-04 (approved):

carletex:
> Ohh I think it would be nice to add the node version check in create-eth too like the CLI will exit directly if it finds the node version is less than our required.

Yes! We talked about this before: https://github.com/scaffold-eth/create-eth/issues/162

PR #1055, opened 2025-02-24
[Draft] Custom address input

rin-st:
Custom address input with shadcn from `AddressInput`. Only UI components changed and used with hooks from https://github.com/scaffold-eth/scaffold-eth-2/pull/1054 . Basically, user could replace them in place but I created new component

To test you can open faucet modal and see three inputs here. First is our current `AddressInput`. Second is the same `AddressInput` with abstracted hooks. Third is this custom shadcn address input

note: Ignore first commit, it's mostly shadcn config. And some styles/placeholders are broken because of that config


rin-st:

PR #1056, opened 2025-02-24
[Draft] Tinkering with AddressInput v2

rin-st:
Second version of `AddressInput`. See v1 here: https://github.com/scaffold-eth/scaffold-eth-2/pull/1054

Usage: 
```tsx
<AddressInput2
    placeholder="Destination Address"
    value={inputAddress ?? ""}
    onChange={value => setInputAddress(value as AddressType)}
    InputComponent={InputBase2}
    InputPrefixComponent={AddressInputPrefix}
    InputSuffixComponent={AddressInputSuffix}
/>
```

Main change: UI components moved to `AddressInput2` props. That UI components should have certain props. 

Comparing to v1
Pros:
1. `AddressInput` could be uploaded to NPM without default component props. In that case all the UI components (PropComponents) should be implemented from scratch by user. We could have examples at the docs though, or keep them in near folder but see [3]
2. Different `PropComponent`'s could be used for different `AddressInput`s. Not sure if it good or bad. On the one side it gives composability but on the other side it could add complexity to the project

Cons: 
3. If we use default components props, then that components should be accessible from `AddressInput` and hence distributed somehow together or just from the all components lib. So, component will become very similar to v1
4. A bit more complex

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Input/AddressInput2/index.tsx
+++ b/packages/nextjs/components/scaffold-eth/Input/AddressInput2/index.tsx
@@ -1,3 +1,4 @@
+import { ComponentType, ReactNode } from "react";
 import { AddressInputPrefix } from "./_components/AddressInputPrefix";
 import { AddressInputSuffix } from "./_components/AddressInputSuffix";
 import { InputBase2 } from "./_components/InputBase2";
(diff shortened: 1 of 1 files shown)


packages/nextjs/components/scaffold-eth/Input/AddressInput2/index.tsx
```diff
@@ -1,22 +1,58 @@
...
  * Address input with ENS name resolution
  */
-export const AddressInput2 = ({ value, name, placeholder, onChange, disabled }: CommonInputProps<Address | string>) => {
+export const AddressInput2 = ({
+  value,
+  name,
+  placeholder,
+  onChange,
+  disabled,
+  InputComponent = InputBase2,
```
rin-st, line 44: see [1] and [3] in the initial comment regarding default values

rin-st:

PR #1064, opened 2025-03-10
Small fixes to homepage layout

JacobHomanics:
## Description

The layout wasn't working as expected due to incorrect flex-col/flex-row setup, especially on different screen sizes. The proposed changes provide a clean layout for all screen sizes.

## Additional Information

Tailwind follows a mobile-first approach, meanwhile the SE-2 homepage seemed to have been following a desktop-first approach: https://tailwindcss.com/docs/responsive-design#working-mobile-first which led to unexpected behaviour.

Because of this, the example: `flex-col sm:flex-row` does not work as expected. On small screens, it actually defaults to `flex-col`. Since when tailwind sees `flex-col`, it actually interprets it as `sm:flex-col`.

## Proposed Changes

The `Connected Addresses` section now simply is a `flex-col`, since with the (now removed) `sm:flex-row`, mobile screens would have content displaying horizontally off-screen.

The `Call To Action` section inhibits the mobile first approach by defaulting to a `flex-col` and on screens medium or larger it turns into a `flex-row`.

Your ENS/address: jacobhomanics.eth

### DIFF
--- a/packages/nextjs/app/page.tsx
+++ b/packages/nextjs/app/page.tsx
@@ -17,7 +17,7 @@ const Home: NextPage = () => {
             <span className="block text-2xl mb-2">Welcome to</span>
             <span className="block text-4xl font-bold">Scaffold-ETH 2</span>
           </h1>
-          <div className="flex justify-center items-center space-x-2 flex-col sm:flex-row">
+          <div className="flex justify-center items-center space-x-2 flex-col">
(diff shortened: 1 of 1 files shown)


technophile-04 (approved):
Thanks @JacobHomanics! yes we had small horizontal scroll on `main` branch in mobile screens... this fixes it too 🙌

PR #1073, opened 2025-03-24
Forward user to creating a keystore if none exist

### LINKED ISSUE #1072: bug: If a user has no keystores yet, yarn deploy --network network_name fails

- [x] I have looked through the [existing issues](https://github.com/scaffold-eth/scaffold-eth-2/issues)

npx create-eth@latest

### Current Behavior

Currently when a user does a `yarn deploy --network sepolia`, they get:

Image

### Expected Behavior

It should either open yarn account:generate or maybe automatically select 0 on the selection menu

### Steps To Reproduce

1. Clone or do a npx create-eth@latest with foundry 
2. Make sure that you have no existing keystores
3. Try to deploy to a network!

### Anything else?

_No response_

portdeveloper:
When the user has no keystores yet, the `yarn deploy --network network_name` just fails.
This PR prevents that by showing the users the menu instead of giving the user an error message. 
I think we can definitely improve this but this should be a good start!

Before:
image

After:
image

Fixes #1072

### DIFF
--- a/packages/foundry/scripts-js/selectOrCreateKeystore.js
+++ b/packages/foundry/scripts-js/selectOrCreateKeystore.js
@@ -1,35 +1,36 @@
-import { readdirSync, existsSync } from "fs";
+import { readdirSync } from "fs";
 import { join } from "path";
 import { spawnSync, spawn } from "child_process";
 import readline from "readline";
 import { fileURLToPath } from "url";
 
 async function selectOrCreateKeystore() {
+  // Create readline interface only when function is called
   const rl = readline.createInterface({
     input: process.stdin,
     output: process.stdout,
   });
 
   const keystorePath = join(process.env.HOME, ".foundry", "keystores");
-  let keystores = [];
 
   try {
-    if (existsSync(keystorePath)) {
-      keystores = readdirSync(keystorePath).filter(
-        (keystore) => keystore !== "scaffold-eth-default"
-      );
-    }
+    const keystores = readdirSync(keystorePath).filter(
+      (keystore) => keystore !== "scaffold-eth-default"
+    );
 
     if (keystores.length === 0) {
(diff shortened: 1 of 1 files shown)


technophile-04 (commented):
Umm port can't we just have take simple route and here : 

https://github.com/scaffold-eth/scaffold-eth-2/blob/aa17fa4aff90c9aaeaa59e590ab59ecf3c70ea1d/packages/foundry/scripts-js/selectOrCreateKeystore.js#L17-L19 

Just updated it to be : 

```js
    const keystores = existsSync(keystorePath)
      ? readdirSync(keystorePath).filter(
          (keystore) => keystore !== "scaffold-eth-default"
        )
      : [];
``` 

Because I think [selectOrCreateKeystore](https://github.com/scaffold-eth/scaffold-eth-2/blob/aa17fa4aff90c9aaeaa59e590ab59ecf3c70ea1d/packages/foundry/scripts-js/selectOrCreateKeystore.js) we already have the logic for showing user that keystore is not present and please use 0 to create a new keystore? It's just that this logic didn't work earlier because we were assuming atleast user will have `scaffold-eth-default` present.  

I mean this case is rare that user won't be having anything inside `~/.foundry/keystores` because when you run `yarn chain` create that keystore.  

But yeah it's good to handle that edge case too, but I think we should take a simpler approach 🙌

portdeveloper:
The way I got this error was when I tried to directly deploy to prod without running yarn chain or yarn start while I am helping someone see how easy it is to go prod with SE-2.

About your suggestion I think it is a lot simpler but I thought maybe the wording would be better if we handled it as something separate.
Since the one below has select a keystore etc. By making the code more complex we show a better msg + path to the user but we overcomplicate the code maybe.
image

But I say why not! If it feels too wrong we can just change it to this more complex implementation in the future.

technophile-04 (approved):

PR #1075, opened 2025-03-24
up next version

technophile-04:
## Description

Next recently had a critical security bug in middleware which they fixed in this version checkout : 

https://nextjs.org/blog/cve-2025-29927


rin-st (approved):
Thanks Shiv!

PR #1077, opened 2025-03-24
Migrate to Eslint v9

### LINKED ISSUE #1076: Migrate to Eslint v9
Since Nextjs 15 [supports](https://nextjs.org/blog/next-15-rc2#eslint-9-support) eslint v9 we can finally move from v8

rin-st:
## Description

Migration of nextjs and hardhat packages to eslint v9

Migrator: https://www.npmjs.com/package/@eslint/migrate-config

Also updated typescript, prettier and eslint related packages

---

Notes regarding `Nextjs` part

`eslint-plugin-prettier`: Lets you enforce Prettier formatting through ESLint's reporting mechanism 
```
plugins: {
  prettier: prettierPlugin,
},
...
rules: {
      ...,
      "prettier/prettier": [
        "warn",
        {
          endOfLine: "auto",
        },
      ],
    },
```
`eslint-config-prettier`:
- Turns off all ESLint rules that might conflict with Prettier's formatting
- Doesn't run Prettier at all
- Simply prevents double-reporting and conflicts
```
extends: compat.extends("next/core-web-vitals", "next/typescript", "prettier"),
```

Fixes #1076


packages/nextjs/eslint.config.mjs
```diff
@@ -0,0 +1,43 @@
...
+
+      "prettier/prettier": [
+        "warn",
+        {
+          endOfLine: "auto",
+        },
+      ],
+    },
+  },
+]);
```
technophile-04, line 43: Can't we do something like this: 


```ts
import { FlatCompat } from "@eslint/eslintrc";
import prettier from "eslint-plugin-prettier/recommended";
import path from "node:path";
import { fileURLToPath } from "node:url";

const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);
const compat = new FlatCompat({
  baseDirectory: __dirname,
});

const eslintConfig = [
  ...compat.config({
    extends: ["next/core-web-vitals", "next/typescript", "prettier"],
    plugins: {
      prettier,
    },
    rules: {
      "@typescript-eslint/no-explicit-any": "off",
      "@typescript-eslint/ban-ts-comment": "off",
      "prettier/prettier": [
        "warn",
        {
          endOfLine: "auto",
        },
      ],
    },
  }),
];

export default eslintConfig;
```

Because I looked next js docs and they aren't using the `defineConfig`  instead they just spread the compat.config in an array: 

https://nextjs.org/docs/app/api-reference/config/eslint#with-typescript

Also, I think the `globalIgnores` is redundant.

Also when you bootstrap a new next project it gives like this: 

```ts
import { dirname } from "path";
import { fileURLToPath } from "url";
import { FlatCompat } from "@eslint/eslintrc";

const __filename = fileURLToPath(import.meta.url);
const __dirname = dirname(__filename);

const compat = new FlatCompat({
  baseDirectory: __dirname,
});

const eslintConfig = [...compat.extends("next/core-web-vitals", "next/typescript", "prettier")];

export default eslintConfig;

```

technophile-04 (commented):
Thanks @rin-st, just tested it out and seems to work great! 

Just added a small comment if we make that file more simple following how the new `create-next-app` follows the path

packages/nextjs/eslint.config.mjs
rin-st: > Because I looked next js docs and they aren't using the defineConfig instead they just spread the compat.config in an array:
> https://nextjs.org/docs/app/api-reference/config/eslint#with-typescript

Migrator added it and it's used in most of [examples](https://eslint.org/docs/latest/use/configure/configuration-files). 

Your config doesn't work for me - no underlines in code, and errors when trying `yarn next:lint`. Is it work for you? I'm not sure at all how it works in example from nextjs docs since if you look ton the definition of `defineConfig`, it handles `extends` keys in config objects array. Looks like without it `extends` shouldn't even work.

Screenshot 2025-04-01 at 22 03 45


Screenshot 2025-04-01 at 22 04 04

> Also, I think the globalIgnores is redundant.

Thanks, removed! [486845d](https://github.com/scaffold-eth/scaffold-eth-2/pull/1077/commits/486845d89cec8fca9548341331e21f77fca1bb5c)

rin-st: hmm, as I understand it should work with `{ FlatCompat } from "@eslint/eslintrc"`. I'll take a closer look tomorrow

technophile-04 (approved):
Tysm @rin-st this is working great 🙌

also tried out : 

https://github.com/scaffold-eth/scaffold-eth-2/issues/901 

And seems to work nicely too 🙌

PR #1084, opened 2025-04-16
Fix Address component copying address to clipboard 

damianmarti:
When the Address component is inside a dialog (like the one created when using a DaisyUI modal), the copy to the clipboard for the address is not working.

You can try it at https://clipboard-broken.vercel.app/ , open the modal, and try to copy the address. The address will not be copied to the clipboard.

This PR does not use react-copy-to-clipboard library to copy to the clipboard. It uses just navigator.clipboard.writeText

https://clipboard-test-sable.vercel.app/

In the previous link, you can test the Address component with the new copy to clipboard implementation. Open the modal and try to copy the address. The address should be copied to the clipboard now. No permission should be asked by the browser.

There is another button "**read from clipboard**" added just for testing the use of navigator.clipboard.readText. This will ask the user for permission to read the clipboard content.

This was tested on Chrome, Firefox, and Edge, on Ubuntu and Windows.

If this works, I think the react-copy-to-clipboard library is not needed anymore since the major advantage was to give browser compatibility, but modern browsers should work using navigator.clipboard.writeText (https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText#browser_compatibility).

We can remove this library dependency if we change 3 components to use navigator.clipboard.writeText: **TransactionHash**, **TxReceipt**, and **AddressInfoDropdown**.


technophile-04 (commented):
Thanks @damianmarti for this and all the research! Yup after research for a while, I think it safe to react-copy-clipboard and use the `navigator.clipboard.writeText` I checked the code of some famous website's and docs too they also are using `navigator.clipboard.writeText` only. 

Also lol I remember having `types` issue with `react-copy-clipboard` because of the way it has it's internal dependencies with `*` so it was also causing problems when installed via `npx create-eth@latest` few months ago, so lets remove it and use `navigator.clipboard.writeText`, Thanksss 🙌

rin-st:
Working great, thank you @damianmarti !

> We can remove this library dependency if we change 3 components to use navigator.clipboard.writeText: TransactionHash, TxReceipt, and AddressInfoDropdown.

It would be great, could you change them too?

> Also lol I remember having types issue with react-copy-clipboard because of the way it has it's internal dependencies with * so it was also causing problems when installed via npx create-eth@latest few months ago, so lets remove it and use navigator.clipboard.writeText, Thanksss 🙌

Yes, and we added type definitions to `/types/react-copy-to-clipboard`. Remove that folder too, please

damianmarti:
Thanks @technophile-04 and @rin-st for the review!

All the components are using the Clipboard API to copy content now, and the react-copy-to-clipboard library was removed!

rin-st:
Thanks Damu! Tested everything and it's working great!

One more question: is it possible to move logic to `useCopyToClipboard` hook?

So we can use it like

```
const {copyToClipboard, isCopiedToClipboard} = useCopyToClipboard;

...
return <>
  <button onClick={() => copyToClipboard(text)} /> 
  {isCopiedToClipboard ? <CheckIcon /> : <CopyIcon />}
</>
```

damianmarti:
> Thanks Damu! Tested everything and it's working great!
> 
> One more question: is it possible to move logic to `useCopyToClipboard` hook?
> 
> So we can use it like
> 
> ```
> const {copyToClipboard, isCopiedToClipboard} = useCopyToClipboard();
> 
> ...
> return <>
>   <button onClick={() => copyToClipboard(text)} /> 
>   {isCopiedToClipboard ? <CheckIcon /> : <CopyIcon />}
> </>
> ```

@rin-st thanks for testing!

Great idea! I will implement this!

packages/nextjs/app/blockexplorer/_components/TransactionHash.tsx
```diff
@@ -2,12 +2,20 @@
...
   const [addressCopied, setAddressCopied] = useState(false);
 
+  const handleCopy = () => {
+    navigator.clipboard.writeText(hash).then(() => {
+      setAddressCopied(true);
+      setTimeout(() => {
+        setAddressCopied(false);
+      }, 800);
+    });
+  };
```
technophile-04, line 17: It would be nice to use async / await instead of .then

technophile-04:
Thanks @damianmarti!! Along with @rin-st suggestions on using hook can we also use asyn / await with try / catch and just log the error. Thanks 🙌

packages/nextjs/app/blockexplorer/_components/TransactionHash.tsx
damianmarti: Implemented in new hook.

damianmarti:
@rin-st @technophile-04 Thanks for the review and testing! 

Implemented the new userCopyToClipboard hook!

technophile-04:
Tysm @damianmarti!! Just pushed some nitpick on renmaings similar to how they were named earlier in particular components 

Merging this 🙌

technophile-04 (approved):

PR #1090, opened 2025-05-06
example of ComposedProviders 

technophile-04:


rin-st:
I think we can close this after https://github.com/scaffold-eth/create-eth/pull/302?

technophile-04:
Yesss! Closing this

PR #1091, opened 2025-05-06
yarn account:reveal-pk

### LINKED ISSUE #1088: Allow revealing PK when using hardhat
Since we merged #1008 we are protecting the PK  using an encrypted json. 

There is no straightforward way to reveal the PK (in case you want to export it and use it somewhere else).

In foundry, you can do `cast wallet decrypt-keystore`.

Maybe we can do `yarn account:reveal-pk` or similar?

carletex:
Fixes #1088

### DIFF
--- a/packages/hardhat/scripts/revealPK.ts
+++ b/packages/hardhat/scripts/revealPK.ts
@@ -17,8 +17,7 @@ async function main() {
   let wallet: Wallet;
   try {
     wallet = (await Wallet.fromEncryptedJson(encryptedKey, pass)) as Wallet;
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  } catch (e) {
+  } catch {
     console.log("❌ Failed to decrypt private key. Wrong password?");
     return;
   }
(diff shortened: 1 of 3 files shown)


packages/hardhat/scripts/revealPK.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  console.log("👀 This will reveal your private key on the console.\n");
+
+  const pass = await password({ message: "Enter your password to decrypt the private key:" });
+  let wallet: Wallet;
+  try {
+    wallet = (await Wallet.fromEncryptedJson(encryptedKey, pass)) as Wallet;
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  } catch (e) {
```
damianmarti, line 21: You can do a catch with no argument if you will not use it, and you avoid the eslint-disable:

```
try {
...
} catch {
...
}
```

damianmarti (approved):
Working good! Nice feature!

packages/hardhat/scripts/revealPK.ts
carletex: Oh, should we remove it from all the scripts, then? cc @technophile-04 

It's that, or printing the error from `e`

Issue #1092, opened 2025-05-08
bug: `useNetworkColor` is always returning `DEFAULT_NETWORK_COLOR`


Pabl0cks:
### Is there an existing issue for this?

- [x] I have looked through the [existing issues](https://github.com/scaffold-eth/scaffold-eth-2/issues)

### Which method was used to setup Scaffold-ETH 2 ?

git clone

### Current Behavior

`useSelectedNetwork` hook returns network objects directly from `scaffoldConfig.targetNetworks`, which are plain `viem/chains` objects and don't have the extra `color` metadata from `NETWORKS_EXTRA_DATA`.  

As a result, hooks like `useNetworkColor` fall back to the default color (`#666666`) instead of the intended network-specific color. 

### Expected Behavior

`useSelectedNetwork` should return network objects with all extra metadata (e.g. `color`) so that UI components display the correct network color and other attributes.

### Steps To Reproduce

On a clean SE-2 build, can try any chain from `scaffold.config.ts` and will show `#666666` color in the RainbowKit button instead of their specific color.

rin-st:
#1093

PR #1093, opened 2025-05-08
Add chain metadata to `useSelectedNetwork` hook

Pabl0cks:
## Description

Adds chain metadata to  `useSelectedNetwork` response, so other hooks like `useNetworkColor` have the color metadata from the chain to work as intended.

## Additional Information

## Related Issues

_Closes #1092

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useSelectedNetwork.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useSelectedNetwork.ts
@@ -3,6 +3,10 @@ import { useGlobalState } from "~~/services/store/store";
 import { AllowedChainIds } from "~~/utils/scaffold-eth";
 import { NETWORKS_EXTRA_DATA } from "~~/utils/scaffold-eth/networks";
 
+/**
+ * Given a chainId, retrives the network object from `scaffold.config`,
+ * if not found default to network set by `useTargetNetwork` hook
+ */
 export function useSelectedNetwork(chainId?: AllowedChainIds) {
   const globalTargetNetwork = useGlobalState(({ targetNetwork }) => targetNetwork);
   const targetNetwork =
--- a/packages/nextjs/hooks/scaffold-eth/useSelectedNetwork.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useSelectedNetwork.ts
@@ -1,16 +1,19 @@
 import scaffoldConfig from "~~/scaffold.config";
 import { useGlobalState } from "~~/services/store/store";
 import { AllowedChainIds } from "~~/utils/scaffold-eth";
-import { NETWORKS_EXTRA_DATA } from "~~/utils/scaffold-eth/networks";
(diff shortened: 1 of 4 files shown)


packages/nextjs/hooks/scaffold-eth/useSelectedNetwork.ts
```diff
@@ -1,8 +1,12 @@
...
+import { NETWORKS_EXTRA_DATA } from "~~/utils/scaffold-eth/networks";
 
 export function useSelectedNetwork(chainId?: AllowedChainIds) {
-  const targetNetwork = useGlobalState(({ targetNetwork }) => targetNetwork);
-  return scaffoldConfig.targetNetworks.find(targetNetwork => targetNetwork.id === chainId) ?? targetNetwork;
+  const globalTargetNetwork = useGlobalState(({ targetNetwork }) => targetNetwork);
+  const targetNetwork =
+    scaffoldConfig.targetNetworks.find(targetNetwork => targetNetwork.id === chainId) ?? globalTargetNetwork;
+
+  return targetNetwork ? { ...targetNetwork, ...NETWORKS_EXTRA_DATA[targetNetwork.id] } : targetNetwork;
```
rin-st, line 11: Thanks Pablo, working great!

Tried to play with it and I think it's easier to use colors data just in `getNetworkColor` function. So we don't need to add color to network object and hence we can remove usage of `NETWORKS_EXTRA_DATA` from this and `useTargetNetworks` hook. (And also move`DEFAULT_NETWORK_COLOR` to `NETWORKS_EXTRA_DATA.default`). What do you think?

image

technophile-04: > So we don't need to add color to network object and hence we can remove usage of NETWORKS_EXTRA_DATA from this and useTargetNetworks hook.

Umm, I think its nice to have `NETWORKS_EXTRA_DATA` returned by `useTargetNework` hook since note that `NETWORKS_EXTRA_DATA` doesn't only contain colors but also `nativeCurrencyTokenAddress` for certain chains. And other people can extend it too like we did it in [abi.ninja](https://github.com/BuidlGuidl/abi.ninja/blob/2c694b711d33764b2d0de53b0e73de30e09f5e11/packages/nextjs/utils/scaffold-eth/networks.ts#L56) 

> And also moveDEFAULT_NETWORK_COLOR to NETWORKS_EXTRA_DATA.default

Again like if we just treat `NETWORKS_EXTRA_DATA` as an object for color then it makes sense, but if we look at other possible attributes it could expose then it doesn't make that much sense?

rin-st: Yes, you right, I saw on the first lines of the object and didn't even noticed `nativeCurrencyTokenAddress` fields )

rin-st (approved):
Thanks for updates and explanation, lgtm!

PR #1095, opened 2025-05-11
Remove daisy UI

### LINKED ISSUE #1079: Remove DaisyUI
This is something that has been on the table for some time: removing DaisyUI and mimicking the same functionality (only the things we use) with Tailwind classes/utilities (@apply, etc)

 There are some PROS:
- Cleaner codebase just with Tailwind
- Allow easier integrations with other libs (e.g. shadcn)
- Extra: if we want to keep going with `scaffold-ui`, this is also a good thing.

Also, some of the _useful_ functionality doesn't work super smoothly (modals, tooltip). I think they could be easily replicated.

On the other hand, I like themes, color handling, semantic classes, etc. We'd need to find a (TW native) replacement for it.

Let's analyze what we are using from DaisyUI and figure out what we'd need to create the functionality we want. Then, let's decide if it is worth it.

rin-st:
[In progress]

Mostly completed
- modal component
- dropdown component
- tooltip component
- table, fix styles

Todo
- input
- textarea
- toggle
- button
- theme

...

Fixes #1079


carletex:
Thanks for exploring this, Rinat! Great job!!

Took a quick high-level look as you suggested. I think it's good to discuss before moving forward.

> What do you think about that approach at all? Is it fit for us? Maybe some other ideas?

In general, I like the approach, and I think this could make a lot of sense. I agree with the PROS / CONS. 

----

Something that we could discuss: where should we draw the line between just tailwind classes & components.

=> It's clear to me that modals (and similar) should be a Component (because they have some custom styling + "JS" functionality). 

=>  On the other side, it's not that clear for things like buttons, since it's _only_ styling. I guess the alternative would be to create some extra utility classes (using `@apply`?) like btn-primary, btn-sm, etc. This applies to things like badges, links, (tooltips?), etc.

I wonder if there would be a nice dynamic way to create variations with themes/colors/sizes for all of these utility classes.

Pushing back on myself, if we are expecting people to use either raw TW or Shadcn, maybe the approach of having `button`, etc as components is not a bad idea.

But yes, for me, not relying on custom components for simpler things seems better. I feel that it makes SE-2 less opinionated and closer to HTML / CSS. People can always create their own components (custom, shadcn, etc)

Anyway, just opening the discussion!

Thanks again!!!

rin-st:
> => On the other side, it's not that clear for things like buttons, since it's only styling. I guess the alternative would be to create some extra utility classes (using @apply?) like btn-primary, btn-sm, etc. This applies to things like badges, links, (tooltips?), etc.

I'll try to add more pros/cons. 

**How I see it with utility classes:**
Pros:
- less components and js as already mentioned
- > But yes, for me, not relying on custom components for simpler things seems better. I feel that it makes SE-2 less opinionated and closer to HTML / CSS. 

Cons:
- it allows to make mistakes like writing `className="btn-sm btn-lg"` for the same button
- we need to add information to the docs why we use utility for buttons and components for modals and if I would be newcomer it would be weird to me
> People can always create their own components (custom, shadcn, etc)
- (less important) if someone wants to keep part of our button when creating custom component, he will need to migrate that part to component from utility class 
- (less important) if we keep daisyui-like classNames (`btn`, `btn-sm` etc), our users could think that daisy still works and won't understand, why `btn` classes work but `modal` is not since they will expect daisyui to work.

**With components:**
Pros:
- it's much more familiar to react devs and hence it's much easier to configure and update for specific project needs
- typechecks: you couldn't set two sizes at the same time (as `btn-sm btn-lg`)
- as Carlos mentioned, it's easier to update component to use Shadcn/whatever

Cons:
- it could be specific style cases which is not fit any props and/or conflict with them. Sometimes it could be easier to fix that cases with utility classes. For example in the PR you could see that inside `Dropdown` I use regular button, not `Button` component. But it could be fixed (or maybe not))

Non of these are critical but I'm still voting for components

---

> I wonder if there would be a nice dynamic way to create variations with themes/colors/sizes for all of these utility classes.

If I undersood you right, It could be implemented the same way as in the daisy, relying on CSS variables. Example:

```
.btn {
  height: var(--size);
}

.btn-lg {
  --size: calc(var(--size-field, 0.25rem) * 12);
}

.btn-xl {
  --size: calc(var(--size-field, 0.25rem) * 14);
}
```

carletex:
This is great stuff!!!

I'll sleep on it a bit and will reply with some thoughts too.

rin-st:
> So actually we could go with shadCN way of doing things (following its theme handling etc) and replace radix with our own HTML CSS things.

Could you add more details, please? As I understand you're proposing to do something like this, but not sure

- use default shadcn config and vars
- create our own components based on that vars/classnames etc so it will be similar to and compatible with other shadcn and shadcn registry components. Other than that vars there will be no dependencies like radix UI, and we can build components using whatever we want

something like https://www.jollyui.dev/docs ?

technophile-04:
So here are the path which we have: 

### Doing things our custom way: 

Creating our own custom css variables nd components where required.

Pros: 
1. Zero dependency 
2. Ability for us to have valid html/css components

Cons: 
1. We need to handle themeing (other people building on top will have inconsistent style)
2. Allowing people to intuitively customize stuff (merging classes etc)
3. educating people how everything works

### Doing things shadcn way: 

ShadCN mainly provides themeing, utilites to merge class names and have variants, cli. 

By default on shadCN website example components are build using RadixUI but we can swap that or not use any library at all.

We now have two paths in shadCN:

1. Use components mentioned on ShadCN directly (i.e using RadixUI)
2. Don't use shadCN website examples instead build our own components.

| 1                                                                                                                                                                                                                                                                                                  | 2                                                                                                                                                                |     |     |
| -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------- | --- | --- |
| PROS: <br><br>- We don't need  to think of JS logic, and having docs on how to use modal instead we point to default shadCN website<br><br>- Not just our components but people can bring other useful components from shadCN website and will go nicely with whole SE-2<br><br>- Devs familiarity | PROS:<br><br>- We have valid html / css components.<br><br>- Since we will be building components we can easily fix things not thinking of bug in other library. |     |     |
|                                                                                                                                                                                                                                                                                                    |                                                                                                                                                                  |     |     |
|                                                                                                                                                                                                                                                                                                    |                                                                                                                                                                  |     |     |
 
 (lol sorry for extra rows / cols my weird Obsidian note app had it)
 
 Also, another thing which Carlos mentioned was if we go with 1. We have to make sure that if people use just <button> instead <Button> we could have some default styles to such directly using <button> looks similar to <Button>.

rin-st:
Closing this since we migrated se-2 to scaffold-ui

rin-st:
I'll create new PR to update the places where daisy is still in use

rin-st:
Continuing discussion here, hope you're getting notifications from closed PRs

Now we have scaffold-ui so some our previous opinions could change
My opinion about current paths

0. Don't change anything, just use daisy.
Since we already have scaffold-ui working, and default create-eth app with daisy ui working well with scaffold-ui, probably we don't need to change anything at all. But there is one trouble. If I want to implement my own components with scaffold-ui and some other library incompatible with daisy-ui, then I need to
- remove daisy-ui. Every daisyui component will break
- rewrite all the daisy-ui related components
which is difficult. This leads to next option

1. Use react components with tailwind classNames, something like this PR.
pros
- Daisy-ui removed, no need to do it yourself when needed like in [0]. Almost every ui lib is compatible with tailwind, so rewriting components in case of using other lib will be easier one by one
cons
- not just html/css components

Other pros/cons https://github.com/scaffold-eth/scaffold-eth-2/pull/1095#issuecomment-2879601949 https://github.com/scaffold-eth/scaffold-eth-2/pull/1095#issuecomment-2883369304

Other paths Shiv mentioned:

> So here are the path which we have:

> ### Doing things our custom way:
> Creating our own custom css variables nd components where required.
> 
> Pros:
> 
> 1. Zero dependency
> 2. Ability for us to have valid html/css components
> 
> Cons:
> 
> 1. We need to handle themeing (other people building on top will have inconsistent style)
> 2. Allowing people to intuitively customize stuff (merging classes etc)
> 3. educating people how everything works

2. Our own components/styles. I agree with all the points but I mostly don't like [Cons 3]. It feels very weird to make se-2 users remember new classNames for buttons inputs etc
3. Shadcn [registry](https://ui.shadcn.com/docs/registry) allows to move most of our frontend components there if needed and also it makes it easy to install that components to any shadcn compatible app. We also can move scaffold-ui part to shadcn registry [directory](https://ui.shadcn.com/docs/directory)  which could add visibility of scaffold-ui to directories users and AI. I think it could be good since I dont see web3 registries there for now. 

One more thing, I think in cases 2.1 and 2.2 we need to rewrite scaffold-ui components to shadcn to make it similar to default components.

Regarding the same problem as in [0], if someone wants to remove Shadcn and wants to use his own library, it also possible to change components one by one. But if he wants to use daisy it could be not so easy, but I'm not sure here

> 1. Use components mentioned on ShadCN directly (i.e using RadixUI)

3.1.  I think this is an easy and reliable way. All the components are high quality and well tested. But there's no just html/css components and radix could be a bit difficult for newbies. But I think it's not a big problem since lots of people already famillar with default shadcn components. And hopefully people won't need to often touch radix ui part

> 2 Don't use shadCN website examples instead build our own components.

3.2. To make our components compatible with other shadcn components apart from other work we should preferably change classnames to the ones that shadcn recommends, like bg-background text-foreground https://ui.shadcn.com/docs/theming. Without it, to change the styles, users will need to add our custom variables --color-primary, --color-secondary etc and change them, or change every component styles or use other methods

 So, we need to
 - to write components compatible with shadcn from scratch or take partially from this PR.
 - find and fix all the bugs by yourself
 - maintain them long term
 
 I think every step above will require much more efforts than just use shadcn radixui components 

---

To sum up:

Options from better to worse (Imho):
1 > 3.1 > 0 > 2 > 3.2

What do you think?

technophile-04:
Thanks Rinat! Just thinking out loud, I feel in this AI era, does the UI framework even really matter anymore? Also considering the Scaffold-ETH 2 something in b/w prod framework and dev playground  I don't think developers deeply care about which UI framework we are using. When they're going to production or have a proper engineering team, that's when they might think about this stuff. Also, I feel the current setup seems to work great with the latest AI models. So I don't see a strong advantage in spending our energy on this migration right now. I think it's fine to keep things as they are and let the AI handle the UI styling changes as they see fit 

But yup, the tailwind with default components makes more sense to me again as compared to other approaches incase we plan to go about tackline around it, I feels thats the best one!

PR #1096, opened 2025-05-12
fix dark mode variant

technophile-04:
## Description

Actually after Tw v4 and DaisyUI v5 updated we broke the dark variant and this fixes it. 

In `page.tsx` try setting the span with : 

```html
<span
className="block text-2xl mb-2 text-blue-400 dark:text-red-500">
Welcome to
</span>
```

Thanks to Philip and Gill for bringing this up 🙌


rin-st (approved):
Lgtm!

PR #1099, opened 2025-05-20
Up viem, wagmi and rainbowkit

rin-st:
Will fix https://github.com/scaffold-eth/create-eth/issues/259 after backmerging to `create-eth` 

(waits for `burner-connector` update)

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "@heroicons/react": "^2.1.5",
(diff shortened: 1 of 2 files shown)


rin-st:
typecheck is broken, but as I understand it will be fixed after https://github.com/scaffold-eth/burner-connector/pull/38

technophile-04 (approved):
Tysm @rin-st! Just tested it out both locally and deploying contract to sepolia and everything seems to work nicely 🙌

PR #1100, opened 2025-05-20
Update Viem Package 2.23.0 -> 2.30.0

gil7788:
- Encountered Missing Zircuit Garfield Testnet Config During Development

Before:
Screenshot From 2025-05-20 21-10-39

After:
Screenshot From 2025-05-20 21-11-38

In principle ZircuitTestnet is deprecated and not use anymore, but that's a contribution that should be done on Viem side.


technophile-04:
Thanks @gil7788!! We actually updated the version in #1099

PR #1106, opened 2025-06-06
shadcn init

technophile-04:
### Description: 

Tinkering with shadCN as discussed in #1079 

#### Things updated: 

- Initalized shadCN and removed daisyUI styles from global.css
- basic home page with all button types 
- Migrated Header and RianbowkitCustombutton and other AddressInfoDropdown to get the gist of the stuff 🙌

#### Problems noticed while using shadCN default UI components: 

- shadCN init (had to use npx yarn init didn't work, users won't face this though) 
- shadCN needs to fix docs for installing components via yarn https://github.com/shadcn-ui/ui/pull/7517 
   - we need to also mention they have to ` cd packages/nextjs` 
- Button's does not have cursor pointer (checkout https://github.com/shadcn-ui/ui/issues/6843 might or might not be fixed for now added a hack at https://github.com/scaffold-eth/scaffold-eth-2/commit/f8e1fcdc7e55b36c61df906b0206601631fd3bab)
- After adding the component it doesn't format it, well when the user commits the changes our commit hooks format it 

#### Resources: 
- Article on `Slot` component and `asChild` inside ShadCN button component https://freedium.cfd/https://javascript.plainenglish.io/understanding-aschild-and-slot-in-react-clean-flexible-component-rendering-247515736cca


technophile-04:
Closing this. 

We also plan to remove daisyUI and have just use tailwindCSS so that we don’t restrict users to use a certain frontend UI library. Now users can easily use a frontend UI library of their own choice, like Shadcn, etc.

Also plan to extract main components and their UI dependent hooks into a different package. This will make SE-2 more minimal, clean and lean. 

Advanced developers who don’t want to use a bit of opinionated components and wants to develop their own web 3 components with their own styling then utilize `scaffold-ui/hooks` which exposes all the necessary stuff required to a specific web3 component. 

This will cater to mostly all kind of users and also make SE-2 more flexible and compatible with any kind of frontend ui library.

PR #1117, opened 2025-06-24
Add `toBlock` param to useScaffoldEventHistory

rin-st:
To test
- Paste this code to `/app/page.tsx`
<details><summary>Details</summary>
<p>

```
"use client";

import { useState } from "react";
import Link from "next/link";
import type { NextPage } from "next";
import { useAccount } from "wagmi";
import { BugAntIcon, MagnifyingGlassIcon } from "@heroicons/react/24/outline";
import { Address } from "~~/components/scaffold-eth";
import { useScaffoldEventHistory, useScaffoldWriteContract } from "~~/hooks/scaffold-eth";

const Home: NextPage = () => {
  const { address: connectedAddress } = useAccount();

  const [newGreeting, setNewGreeting] = useState("");

  const { writeContractAsync: writeSetGreeting } = useScaffoldWriteContract({ contractName: "YourContract" });

  const handleSetGreeting = async () => {
    try {
      await writeSetGreeting({
        functionName: "setGreeting",
        args: [newGreeting],
        value: 0n,
      });
      setNewGreeting(""); // Clear input after successful transaction
    } catch (error) {
      console.error("Error setting greeting:", error);
    }
  };

  const { data: greetingChangeEvents } = useScaffoldEventHistory({
    contractName: "YourContract",
    eventName: "GreetingChange",
    fromBlock: 0n,
    toBlock: 10n,
    watch: true,
  });

  console.log(greetingChangeEvents);

  return (
    <>
      <div className="flex items-center flex-col flex-grow pt-10">
        <div className="px-5">
          <h1 className="text-center">
            <span className="block text-2xl mb-2">Welcome to</span>
            <span className="block text-4xl font-bold">Scaffold-ETH 2</span>
          </h1>
          <div className="flex justify-center items-center space-x-2 flex-col sm:flex-row">
            <p className="my-2 font-medium">Connected Address:</p>
            <Address address={connectedAddress} />
          </div>

          <div className="flex flex-col gap-3 mt-6 mb-6">
            <input
              type="text"
              placeholder="Enter new greeting"
              className="input input-bordered w-full max-w-xs mx-auto"
              onChange={e => setNewGreeting(e.target.value)}
              value={newGreeting}
            />
            <button
              className="btn btn-primary w-full max-w-xs mx-auto"
              onClick={handleSetGreeting}
              disabled={!newGreeting}
            >
              Set Greeting
            </button>
          </div>

          {greetingChangeEvents && greetingChangeEvents.length > 0 ? (
[...]

Issue #1110, opened 2025-06-13
useScaffoldEventHistory() alchemy block range issue


rin-st:
From discussion: https://github.com/scaffold-eth/scaffold-eth-2/discussions/1105#discussion-8414925

1. We need to set a limit to a maximum of 500 blocks per request (when working with alchemy?)
2. Add block range parameter `toBlock`, `blockLimit`, or similar

technophile-04:
Tysm all! Love the discussion! 

Yup, like the idea of having the `toBlock` param and let's call it `toBlock` because `viem` also uses same naming convention. 

Regarding the `batchSize` discussion, you makes sense! I was also checking other rpc providers and they seem to also have limit eg. for quicknode they just allow 5 blocks for free tier. 

<img width="1114" alt="Image" src="https://github.com/user-attachments/assets/8fabe953-d0a3-4e41-962a-5253b73a0d8a" /> 

So it would be nice such that we fetch the logs in batch of certain number and also we can allow this `batchSize` number to be configurable by developer. 

Let's tackle this two in separate PR 🙌

PR #1118, opened 2025-06-26
Add `blocksBatchSize` param to useScaffoldEventHistory

rin-st:
Default value is 500 since we're working mostly with alchemy. If users want to work with quicknode or other providers they can specify `blocksBatchSize` for their case

To test do the same as in https://github.com/scaffold-eth/scaffold-eth-2/pull/1117 but also experiment with parameter `blocksBatchSize` to the useScaffoldEventHistory on `app/page.tsx`. 

PR is to https://github.com/scaffold-eth/scaffold-eth-2/pull/1117 branch (I branched from it), so `toBlock` also works in this PR

Second part of #1110

### DIFF
--- a/packages/hardhat/.env.example
+++ b/packages/hardhat/.env.example
@@ -7,7 +7,7 @@
 
 # To access the values stored in this .env file you can use: process.env.VARIABLENAME
 ALCHEMY_API_KEY=
-ETHERSCAN_MAINNET_API_KEY=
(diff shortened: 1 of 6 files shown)


damianmarti (approved):
Tested using small blocksBatchSize values and everything worked fine! Thanks!

technophile-04 (approved):
Tested and workd nicely!! Thanks Rinat!

PR #1122, opened 2025-07-03
fix blockexplorer address page when contract is empty

technophile-04:
## To reproduce: 

On the main branch, do `yarn chain` and `yarn start` without `yarn deploy `. Go to blockeplorer and click one of the addresses. That page will be broken. 

The problem was that the `contracts` variable was an empty object, and JS gives `Boolean({})` as `true`, and hence it broke up. 

We now do `Object.keys(contracts[chainId]).length === 0` which checks if the object is empty or not.


technophile-04:
Merging this 🙌

Issue #1126, opened 2025-07-11
deploying contracts to mainnet fails


technophile-04:
### Description: 

When trying to deploy a contract `mainnet` it fails, it works for other networks like base or other but when you run: 

```
yarn deploy --network mainnet 
```

It gives error something like this: 

```
deploying "YourContract"invalid address (argument="address", value="", code=INVALID_ARGUMENT, version=address/5.8.0) {"reason":"invalid address","code":"INVALID_ARGUMENT","argument":"address","value":"","checkKey":"to","checkValue":"","transactionHash":"0xc6438d147ac1bc3e26683e9170b975707e41e670b0f6236267d5f41325b1dbd7"} Error: invalid address (argument="address", value="", code=INVALID_ARGUMENT, version=address/5.8.0)
```

portdeveloper:
After trying almost everything, I noticed that the only thing we don't change is the RPC that we are sending the tx to. 

And then I tried to change the RPC. Looks like the alchemy RPC is causing the problem.

Just change the RPC to something like `https://eth.drpc.org`  and the issue goes away magically!

technophile-04:
Tysm Port for digging in 🙏, just created an issue on https://github.com/wighawag/hardhat-deploy/issues/571. Let's see if it gets solved early, until then we can keep this issue open here and people can change the RPC manually to make it work on the mainnet. 

If it takes more time then we could just hardcode drpc URL for mainnet 🙌

carletex:
> If it takes more time then we could just hardcode drpc URL for mainnet 🙌

+1!

Thanks Port!

technophile-04:
Temp fixed at #1130, we can track the main issue on https://github.com/wighawag/hardhat-deploy/issues/571

PR #1130, opened 2025-07-15
update hardhat mainnet rpc url

technophile-04:
## Description

Hot fix for #1126. 

--- 

### Issue encountered with BG RPC: 

For some reason, we are not able to connect to BG RPC via ethers `JSONRpcProvider` while running `yarn accoun`, it errors out saying: 

Screenshot 2025-07-15 at 11 12 22 AM

Seems like ethers is first making a request with an empty request body / payload. Not sure why this is happening only with BG RPC maybe we are not following some standard for `/` route like for now its return `400` maybe it should return something different? I tried reading about sepc but there is not specific instruction on `/` route.  But RPC is not down and it works great you can try at https://ethereum-json-rpc.com/ by inputting custom URL with BG endpoint 

Another issue which I saw was is when you do `yarn deploy --network mainnet` its get stuck: 

Screenshot 2025-07-15 at 10 06 46 AM
 
 
 Although transaction happened and there were 6 block confirmation it was still stuck there 
 
 So, for now, went with the DRPC URL

### DIFF
--- a/packages/hardhat/hardhat.config.ts
+++ b/packages/hardhat/hardhat.config.ts
@@ -55,8 +55,7 @@ const config: HardhatUserConfig = {
       },
     },
     mainnet: {
-      // url: `https://eth-mainnet.alchemyapi.io/v2/${providerApiKey}`,
-      url: "https://mainnet.rpc.buidlguidl.com/",
+      url: "https://eth.drpc.org",
       accounts: [deployerPrivateKey],
     },
     sepolia: {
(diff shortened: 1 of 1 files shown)


packages/hardhat/hardhat.config.ts
```diff
@@ -55,7 +55,7 @@ const config: HardhatUserConfig = {
       },
     },
     mainnet: {
-      url: `https://eth-mainnet.alchemyapi.io/v2/${providerApiKey}`,
+      url: "https://eth.drpc.org",
```
technophile-04, line 58: Was planning to add a comment referencing issue #1126 above this line and also commented Alchemy URL but then thought since this is just hot-fix and when https://github.com/wighawag/hardhat-deploy/issues/571 is solved we can revert back. 

But happy to add a comment 🙌

technophile-04:
Merging this 🙌

PR #1132, opened 2025-07-15
Add deployedOnBlock to deployedContracts on Hardhat

damianmarti:
generateTsAbis adds the blockNumber from the transaction receipt from the deployment transaction, and you can get this information using useDeployedContractInfo hook.

Only implemented on Hardhat for now.

To test it, just deploy the default contract and use the useDeployedContractInfo hook, you will get the new deployedOnBlock attribute.

Feature from the discussion https://github.com/scaffold-eth/scaffold-eth-2/discussions/1129

I can implement this on Foundry too and then I can change the Ponder extension to use this new attribute by default.


carletex:
And maybe we could use this value (if defined) as default on the `fromBlock` prop in events hooks, as Elliott was suggeting? cc @damianmarti

damianmarti:
> And maybe we could use this value (if defined) as default on the `fromBlock` prop in events hooks, as Elliott was suggeting? cc @damianmarti

Yes! I will add this in another PR.

Issue #1133, opened 2025-07-17
Upgrade to node v22


technophile-04:
### Description: 

New node LTS version is v22. Also Vercel by default seems to use that 

https://vercel.com/docs/functions/runtimes/node-js/node-js-versions#setting-the-node.js-version-in-project-settings

rin-st:
Upgrade to Nodejs v22

TODO:
- [ ] Se-2
- [ ] Create-eth
- [ ] Se-2 docs
- [ ] Sre challenges
- [ ] Sre v2

rin-st:
Closing this for now since we decided to keep v20 in se-2 for now and change to node 22 only in live projects https://github.com/scaffold-eth/scaffold-eth-2/pull/1139#issuecomment-3106153274

PR #1139, opened 2025-07-22
Upgrade nodejs to v22

rin-st:
Up nodejs version to v22 for the whole se-2

**Hardhat:**
Support node v22 from [v2.22.6](https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%402.22.6). We're using `^2.22.10`, [code](https://github.com/scaffold-eth/scaffold-eth-2/blob/5dc015c3510c5048dbbdd1a95dc9b535d91f7c9e/packages/hardhat/package.json#L44).
Also, they suggest to use v22 in their docs https://hardhat.org/tutorial/setting-up-the-environment#linux but still allow >=18. As a note, for hardhat v3 required version is [>= 22](https://hardhat.org/hardhat3-alpha#installation)

I remember there was some issues with hardhat and node v22 but looks like we had it when used previous hardhat versions. I tested all the package.json hh scripts, and also tested create-eth with extension installation with both hardhat and foundry, and no issues with node v22

**Nextjs:** 
default deployed version is 22 ([docs](https://vercel.com/docs/functions/runtimes/node-js/node-js-versions#setting-the-node.js-version-in-project-settings)) so no need to set anything explicitly

Part of https://github.com/scaffold-eth/scaffold-eth-2/issues/1133#issuecomment-3101504479

@Pabl0cks could you check if `yarn start` and `yarn deploy` works without warnings on windows with node js 22? 

Note: I think you all have it, mentioning just in case. To switch nodejs version, you can use https://github.com/nvm-sh/nvm


technophile-04:
> this is going to break for a bunch of people that are in node 20, which is still supported, until next year.

Yup was thinking the same 

> So, what's the benefit of updating early? 

If we upgrade to v22, people scaffolding new SE-2 apps from now on will get v22 support which is long term and then don't need to update their apps dependencies even after 1 yr. If we still be on v20 then after 1 yr people who had scaffolded their app now will have to manually update all the things. (this was mentioned by Rinat and also the reason makes sense) 

But there are very less SE-2 projects that are being maintained for 1yr so maybe it's fine? Also looking at changes there seems not a lot of breaking changes. 

So maybe we could wait for 1yr, incase if something big comes in b/w and we need to update the library for eg hardhat v3 requires v22, then we move to node v22 before 1 yr to support it 🙌

carletex:
> If we still be on v20 then after 1 yr people who had scaffolded their app now will have to manually update all the things. (this was mentioned by Rinat and also the reason makes sense)

This is a compelling argument too!

Not sure, I see PROS/CONS on each side.

A couple of random thoughts:

Upgrading to v22 in something like SRE makes a lot of sense (we are already upgrading node version, so why not upgrade to the recommended one and we'll have more years of support without worrying). But SRE is a project...

...as in SE-2, as a "public library" might make more sense to support the lowest version possible (when EOL + dependencies permit).

technophile-04:
> Upgrading to v22 in something like SRE makes a lot of sense (we are already upgrading node version, so why not upgrade to the recommended one and we'll have more years of support without worrying). But SRE is a project...

> ...as in SE-2, as a "public library" might make more sense to support the lowest version possible (when EOL + dependencies + infra permit).

I think we should go with this! Wait in SE-2 but with other "projects" we should up them to v22 🙌

Pabl0cks (commented):
> @Pabl0cks could you check if yarn start and yarn deploy works without warnings on windows with node js 22?

Working nicely on Windows, no warnings 🙏 Thx Rinat!

image

image

PR #1149, opened 2025-08-01
refactor: use `useMemo` to optimize also update in ReadOnlyFunctionForm

crStiv:
Replaced direct call to `transformAbiFunction` with a `useMemo` hook to optimize performance in `WriteOnlyFunctionForm.tsx`.


technophile-04 (commented):
Thanks @crStiv, can we also update it in ReadOnlyFunctionForm? I think you missed that

PR #1150, opened 2025-08-01
fix `fromBlockValue` type or empty contracts

technophile-04:
## Description

Fixes this : https://github.com/scaffold-eth/create-eth/actions/runs/16674616228/job/47198269656?pr=328


rin-st (approved):
Thanks!

PR #1158, opened 2025-08-11
configure porto

technophile-04:
## Description

Seems like porto only works on `baseSepolia` https://porto.sh/sdk/api/chains#supported-chains and it doesn't work with any other network. 

https://github.com/user-attachments/assets/6d97235c-f3f3-45e4-93f8-31a947288b3d

### To test: 

Configure `chains.baseSepolia` in `targetNetworks`


technophile-04:
Closing this will create an extension for this 🙌

PR #1161, opened 2025-08-12
fix: add input length validation to prevent ReDoS attacks in contract utils

crStiv:
Add input length validation to isBigInt and isJsonString functions to prevent potential ReDoS attacks

- Add 1000 character limit to isBigInt function to prevent processing of extremely long numeric strings
- Add 10000 character limit to isJsonString function to prevent processing of extremely long JSON strings  
- Add explanatory comments about security considerations

The functions are used in the debug contract interface for parsing user inputs, making this a security improvement for the development environment.


technophile-04 (commented):
Thanks for the PR! This extra check feels like a paranoia check, since these functions are just used in Debug Contract page (which won't be used by final users of application built)...so skipping it just to avoid noise 🙌

PR #1162, opened 2025-08-12
remove duplicate events from end 

technophile-04:
## Description

Not the best way, but a simple way. I tried tinkering hard at https://github.com/technophile-04/test-arb-history, but its hard to get it properly working because the sometime the interval at which block are mined + some times due to the blockBatchSize goes there are some overlaps in `getNextParams` and because of which this issue might be causing. 

Tried lot of approach, but then thought maybe lets go with simple approach only. 

Also I think the `useScaffoldEventHistory` isn't good for l2s because it will burn rpcs calls alot even with batching.

For example if you look at this contract: 

https://sepolia.arbiscan.io/tx/0x8fc14b8dda901d6beaa929a4e4528b6eaddb2109ad8bd96f5bdd61978e65558f it was deployed at block : [181594184](https://sepolia.arbiscan.io/block/181594184) 

And current blocknumber is : 183246630

So there are total of 16,52,446 block minned in b/w and useScaffoldEvenHistory need to query all of them, even if we query in batches of 500 (default) or 1000 it's still almost 1.5k request (getLogs).

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
@@ -254,7 +254,13 @@ export const useScaffoldEventHistory = <
 
   // Combine historical data from infinite query with live events from watch hook
   const historicalEvents = query.data?.pages || [];
-  const combinedEvents = [...liveEvents, ...historicalEvents] as typeof historicalEvents;
+  const allEvents = [...liveEvents, ...historicalEvents] as typeof historicalEvents;
+
+  // remove duplicates
+  const combinedEvents = allEvents.filter((event, index, self) => {
+    const eventKey = `${event?.transactionHash}-${event?.logIndex}-${event?.blockHash}`;
+    return index === self.findIndex(e => `${e?.transactionHash}-${e?.logIndex}-${e?.blockHash}` === eventKey);
+  }) as typeof historicalEvents;
 
   return {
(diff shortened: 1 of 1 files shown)


technophile-04:
Thanks @escottalexander for the thoughts, love them! 

> I am hopeful that Alchemy will switch back to allowing a 10k batch limit since it is industry standard. Nothing we can do about this short of switching default providers or having clever logic that attempts 10k but readjusts when it fails with Alchemy. The complexity overhead is just not worth it.

Was just reading [alchemy docs](https://www.alchemy.com/docs/deep-dive-into-eth_getlogs#making-a-request-to-eth_getlogs): 

Screenshot 2025-08-13 at 3 06 48 PM

yeah and also it's hard to write that readjustable logic and when people you other RPC url than alchemy again it will cause problem, yup adds lots of complexity :( 

> Perhaps we should think more about making indexer solutions first class extensions, maybe even asking the user if they want to install one when they run create-eth without another extension specified. (Bonus points if we made them installable WITH other extensions, since they are largely self-contained or at least could be added in a self-contained manner) 

Yeah, maybe we should think of this not sure which would be best since on indexer's different dev's have different choices (ponder, subgraph etc) also for newbies they might get overwhelmed (do they need to learn this indexers frameworks too? so use scaffold-eth 2)...but yeah giving them the options to choose use it or not might be good

technophile-04:
Thanks, all! Merging this 🙌

PR #1164, opened 2025-08-13
fix(nextjs): add noopener to external tx link to prevent reverse‑tabnabbing

anim001k:
Description
- What: Add rel="noopener noreferrer" to the block explorer link in TxnNotification.
- Why: Harden against reverse‑tabnabbing, satisfy link linting, and align with AddressLinkWrapper.
- Scope: packages/nextjs/hooks/scaffold-eth/useTransactor.tsx only.
- Impact/Risk: No functional changes; security hardening. Low risk.


technophile-04 (commented):
https://github.com/scaffold-eth/scaffold-eth-2/blob/main/CONTRIBUTING.md#rules  

Hey we don't accept pull request from bots or the account which has closed contribution history. 

Also I don't think blockeExplore links which we are pointing will be exploited

PR #1173, opened 2025-08-20
docs: add production usage warning for useScaffoldEventHistory

### LINKED ISSUE #1166: document `useScaffoldEventHistory` to be used only locally
### Description: 

Although `useScaffoldEventHistory` is a very powerful hook, but it's not optimised at all since it uses `getLogs` and might burn the mainnet's RPC a lot, especially L2s, which have very little block time. It's best for querying local (hardhat / anvil) chains events. There are various complexity to make this hook a bit optimised but instead we should suggest indexers. 

We should expose this information to the users of the hooks that they need to be mindful of using this hook on production. 

Some of the idea which we disussed: 

1. Using JSDoc with deprecate, so that users will get a warning message not to use it in prod in their IDE 
2. Updateing the SE-2 docs mentioning about use of ponder / other indexer for prod

anumukul:
## Description
Addresses issue #1166 by adding documentation warnings about using `useScaffoldEventHistory` in production.

## Changes Made
- ✅ Added JSDoc deprecation warning to the hook
- ✅ Updated documentation to warn about RPC limitations
- ✅ Recommended indexer solutions for production use

## Why This Matters
The hook uses `getLogs` which can burn through RPC limits, especially on L2s with short block times. This change helps developers make informed decisions about when to use this hook.

## Testing
- [x] JSDoc warnings appear in IDE
- [x] Documentation renders correctly
- [x] No build errors

Closes #1166

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,16 @@
 
 ![Debug Contracts tab](https://github.com/scaffold-eth/scaffold-eth-2/assets/55535804/b237af0c-5027-4849-a5c1-2e31495cccb1)
 
+---
+
+## ⚠️ Important Note about useScaffoldEventHistory
+
+> **Warning:** `useScaffoldEventHistory` is intended for local development only.  
+> This hook uses `getLogs` and can overload RPC endpoints, especially on production and L2 networks.  
+> **For production, please use a dedicated indexer (e.g., [Ponder](https://ponder.dev)) to query contract events efficiently.**
+
+---
+
 ## Requirements
 
 Before you begin, you need to install the following tools:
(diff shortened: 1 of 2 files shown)


technophile-04:
Thanks @anumukul! Just added couple of comments, and can you also add this :

```ts
  useEffect(() => {
    if (selectedNetwork.id !== hardhat.id) {
      console.log(
        "⚠️ useScaffoldEventHistory is not optimized for production use. It can overload RPC endpoints (especially on L2s with short block times).",
      );
    }
  }, [selectedNetwork.id]);
``` 

So that we also warn the users of the hook in console, when the chain in not hardhat / anvil?

PR #1174, opened 2025-08-25
fix bg rpc url 

technophile-04:
## Description

We have correct url in `hardhat.config.ts` but here we missed it. Thanks to Austin for noticing it 🙌


damianmarti (approved):
Thanks! This is the right RPC URL and works fine with it!

rin-st (approved):
Lgtm, thanks!

PR #1185, opened 2025-10-03
Bump Viem

gskril:
Apps should be on viem [v2.35.0](https://github.com/wevm/viem/releases/tag/viem%402.35.0) or greater to take advantage of the latest [Universal Resolver](https://docs.ens.domains/resolvers/universal), which will make supporting [ENSv2](https://ens.domains/ensv2) seamless on day 1

For some reason, this only builds if I remove Burner wallet support. Keeping it as a draft until I can resolve that issue

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -33,7 +33,7 @@
     "react-dom": "^19.0.0",
     "react-hot-toast": "^2.4.0",
     "usehooks-ts": "^3.1.0",
-    "viem": "2.34.0",
+    "viem": "2.37.11",
     "wagmi": "2.16.4",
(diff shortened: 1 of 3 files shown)


technophile-04:
Hey thanks @gskril! Can we also update the wagmi and rainwbokit version? We usually do that in the same PR since most of them are dependent and are core libs of SE-2 

 > For some reason, this only builds if I remove Burner wallet support. Keeping it as a draft until I can resolve that issue
 
 Ohh maybe if we update the viem version it shall solve the issue example: https://github.com/scaffold-eth/burner-connector/pull/46  (maybe we create a same version bump pr in burner-connector repo as well) 
 
 Thanks!

PR #1190, opened 2025-10-28
Update Alchemy API key to new rolling key for tracking new/old requests

ZakGriffith:
## Description

Updated Alchemy API key in both scaffold.config.ts and hardhat.config.ts files to a newly created key in a new Alchemy App named Scaffold-ETH October 28 2025.  This will allow us to see new vs old requests within Alchemy, and eventually shut down the old key.


technophile-04 (approved):
Image

Works great! Also randomly tested with arbitrumSepolia contract and works

technophile-04:
>  (or that's done directly on create-eth)  

Yup yup directly in `create-eth` foundry branch here is bit outdated

PR #1192, opened 2025-10-29
Migration to Hardhat v3

### LINKED ISSUE #1191: Hardhat V3
### Description: 

Migrate to Hardhat v3 since, seems like it's already in production. 

https://hardhat.org/migrate-from-hardhat2 

Previous attempts: #567, https://github.com/scaffold-eth/scaffold-eth-2/pull/644

carletex:
Creating this draft so we can start tinkering & discussing:

For now:
- Installed new deps (removed old ones)
- Updated README with new NodeJS requirement (we might need to update in docs, etc)
- `yarn chain` working
- `yarn deploy` working
  - Using Ignition. Let's see if we don't need **hardhat-deploy** anymore (also available for HHv3)
  - Missing: 
    -  [ ] inheritedFunctions
    - [ ] deployedOnBlock

ToDo (will be adding more):
- [ ] Fix GitHub lint workflow
- [ ] Create new _keystore_ system (move to custom one => hardhat-keystore / config Variables)
- [x] Make `yarn deploy` more generic. Before it deployed everything / tag-based. Now only the specified conttract.
- [ ] --network for deploy (+ config networks on hardhat.config)
- [ ] Verify / Etherscan API key
- [ ] Tests (default should be solidity, TS or both?)
- [ ] Mainnet Fork
- [ ] TS Generation: get inherited functions + deployed block

Known Issues:
- If I deploy, make a change to the contract, and deploy again, I get `[ YourContract ] Nothing new to deploy based on previous execution stored in ./ignition/deployments/chain-31337`

Fixes #1191

### DIFF
--- a/packages/hardhat/scripts/generateTsAbis.ts
+++ b/packages/hardhat/scripts/generateTsAbis.ts
@@ -17,7 +17,7 @@ const generatedContractComment = `
 `;
 
 const DEPLOYMENTS_DIR = "./ignition/deployments";
-// const ARTIFACTS_DIR = "./artifacts";
+const ARTIFACTS_DIR = "./artifacts";
 
 function getDirectories(path: string) {
   return fs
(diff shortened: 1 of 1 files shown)


packages/hardhat/ignition/modules/ScaffoldEthDeployModule.ts
carletex: Went with this solution for now (naming it generic, so people can deploy everything here by default.... since the module path is hardcoded on package.json)

e.g. This works:

```ts
export default buildModule("ScaffoldEthDeployModule", m => {
  const yourContract = m.contract("YourContract", ["0x0000000000000000000000000000000000000000"]);
  const yourContract2 = m.contract("YourContract2", ["0x0000000000000000000000000000000000000000"]);

  return { yourContract, yourContract };
});
```

they say in their [docs](https://hardhat.org/ignition/docs/guides/creating-modules)
```
You can create multiple modules in a single file, but each must have a unique ID. To deploy a module, you must export it using module.exports = or export default. As a best practice, we suggest maintaining one module per file, naming the file after the module ID.
```

But I couldn't make it work (creating more modules in the file + exporting both)

---

Note: they also have https://hardhat.org/ignition/docs/guides/scripts, where you can do more "complex" logic (e.g. console.logs of contract read data)

packages/hardhat/hardhat.config.ts
```diff
@@ -38,116 +27,39 @@ const config: HardhatUserConfig = {
...
-        url: `https://eth-mainnet.alchemyapi.io/v2/${providerApiKey}`,
-        enabled: process.env.MAINNET_FORKING_ENABLED === "true",
-      },
-    },
     mainnet: {
+      type: "http",
+      chainType: "l1",
       url: "https://mainnet.rpc.buidlguidl.com",
-      accounts: [deployerPrivateKey],
+      accounts: [configVariable("DEPLOYER_PRIVATE_KEY")],
```
carletex:
```
yarn hardhat keystore set DEPLOYER_PRIVATE_KEY
```

We'd need to abstract this to the user for `yarn generate` etc.

Also: for the keystore you need a 8 character password. Unless you use the `--dev` flag, which doesn't ask for a password at all https://hardhat.org/docs/guides/configuration-variables#improving-ux-when-using-keystore-values-during-the-dev-process

packages/hardhat/hardhat.config.ts
```diff
@@ -1,29 +1,19 @@
...
+import type { HardhatRuntimeEnvironment } from "hardhat/types/hre";
+import type { TaskArguments } from "hardhat/types/tasks";
+import hardhatIgnitionViemPlugin from "@nomicfoundation/hardhat-ignition-viem";
+import { overrideTask } from "hardhat/config";
+import generateTsAbis from "./scripts/generateTsAbis.js";
+import hardhatToolboxViem from "@nomicfoundation/hardhat-toolbox-viem";
 
 // If not set, it uses ours Alchemy's default API key.
 // You can get your own at https://dashboard.alchemyapi.io
 const providerApiKey = process.env.ALCHEMY_API_KEY || "cR4WnXePioePZ5fFrnSiR";
```
carletex: I think we can use regular env vars for non-critical data (like this key) and then the keystore (configVariable) for PKs, etc.

packages/hardhat/ignition/modules/ScaffoldEthDeployModule.ts
technophile-04: In foundry we have similar case. So what we have is insted of hardcoding path infront of `forge deploy <file_name>` we have created a proxy javascript script. 

So if the user run `yarn deploy` => `node parseArgs.js`. And if you have just ran `yarn deploy` it will run default all in one deploy script `hardhat ignition deploy ignition/modules/ScaffoldEthDeployModule.ts` but if you pass extra `--file` arg then it run only that particualr deploy script. 

on this case people could run `yarn deploy --file YourContract.ts` (we tell people not to include the whole path, it shall be relative to `ignition/modules/`). And then we ask people to follow the following structure: 

```shell
ignition/
└── modules/
    ├── MyContract.ts // each module of each contract
    ├── ScaffoldEthDeployModule.ts
    └── YourContract.ts // each module of each contract
```

And in `ScaffoldEthDeployModule.ts` we do: 

```ts
import { buildModule } from "@nomicfoundation/hardhat-ignition/modules";
import YourContractModule from "./YourContract.js";
import MyContractModule from "./MyContract.js";

export default buildModule("ScaffoldEthDeployModule", m => {
  const { yourContract } = m.useModule(YourContractModule);
  const { myContract } = m.useModule(MyContractModule);

  return { yourContract, myContract };
});
```

--- 

Lol also we shall discuss this only if we are sure we are gonna use ignition, if we are gonna use hardhat-deploy, then probably not worth  the discussion

packages/hardhat/hardhat.config.ts
technophile-04: > I think we can use regular env vars for non-critical data (like this key)  

Agree!! Let's have etherscan and alchemy api key in .env only instead of keystore

PR #1193, opened 2025-10-30
Scaffold-ui blockexplorer

rin-st:
Draft PR just to show how it works now

We have 3 blockexplorer pages: `/blockexplorer`, `/blockexplorer/address/[address]` and `/blockexplorer/transaction/[txHash]`. So, to make them work as before, I export from `@scaffold-ui/block-explorer`https://github.com/scaffold-eth/scaffold-ui/pull/63 3 components for that 3 pages + one helper for `Address` server page (I'll try to make it without helper later)

So, to add blockexplorer to the app, user will need at least 3 components, see this PR. But I think it's not a problem if we describe it nicely in the docs. What do you think? Please lmk and if everything is ok, I'll continue to work on it (some refactorings, code cleanups, readmes, examples, styles)

Note: I think we can add flags to create-eth for debug page and block-explorer and with these flags pages will be installed automatically

To test:
- Run `scaffold-ui` from [block-explorer branch](https://github.com/scaffold-eth/scaffold-ui/tree/block-explorer). `pnpm i`, and then `pnpm dev`
- Run `se-2` from this PR branch. `yarn install` -> `yarn chain/deploy/start`


rin-st:
Closing for now

PR #1194, opened 2025-10-31
remove uniswap sdk

technophile-04:
This was found dangling. We don't need this anymore since now we have https://github.com/scaffold-eth/scaffold-ui/blob/main/packages/hooks/src/useFetchNativeCurrencyPrice.ts#L25


rin-st (approved):
Lgtm!

PR #1211, opened 2025-12-05
Upgrade yarn version

technophile-04:
## Description

Ran: 

```
yarn set version berry
yarn install
```

The #1184 seems way too old. Also, I'm not sure what commands they used to migrate because there are some files that need to be added which were not present. 

## TODO: 

- [x] Research about `npmMinimalAgeGate` and see whats the best and standard people are using and add it (check https://github.com/scaffold-eth/scaffold-eth-2/pull/1211#issuecomment-3669297584)


carletex:
> Since we're not in a hurry most of the time, I think it's ok to use "7d", it should be enough to prevent most of the supply chain attacks.

We also have to think about the opposite side of this. When you want to update to a safer version currently released (nextjs / react recently). I guess you just need to add temporarily to `npmPreapprovedPackages`, right?

Also, for our `npmPreapprovedPackages` packages (burner, sui?) we'd need to implement something similar to this? If not, we are still vulnerable (way less, but still)

rin-st:
> We also have to think about the opposite side of this. When you want to update to a safer version currently released (nextjs / react recently). I guess you just need to add temporarily to npmPreapprovedPackages, right?

I think yes, I don't see other solutions

> Also, for our npmPreapprovedPackages packages (burner, sui?) we'd need to implement something similar to this? If not, we are still vulnerable (way less, but still)

Yes, wanted to say it too but forgot

technophile-04:
> I think we need to remove react/next from preapproved packages.
As I remember npmMinimalAgeGate option was added after [this](https://safedep.io/eslint-config-prettier-major-npm-supply-chain-hack/), when patch versions of popular packages was compromised. Hopefully it will not be the case for react/next but we're not 100% sure. 

Ohh man :( I seee yup and agree, the more I think and read it just makes sense to only have our packages (which are controled by us) to have in `npmPreapprovedPackages` rest we can remove and add depending on the need

Kushmanmb (commented):
yarn install

rin-st (approved):
Lgtm, thanks!

technophile-04:
Was having a meeting with Carlos earlier and he flagged a concern, since we have `scaffold-ui` and our other maintained packages on `npmPreapprovedPackages`, what happens if we ship a new `scaffold-ui` version and it ends up pulling a freshly compromised transitive. Does the allowlist just wave the whole subtree through the gate?

Digging a bit, `npmPreapprovedPackages` is a strict name match. It only lets the named package through the gate, it doesn't cascade to that package's dependencies. Every transitive gets evaluated on its own. So if we ship scaffold-ui with a hypothetical bad `axios@1.14.1` published less than 7 days ago, scaffold-ui itself bypasses because we allowlisted it, but axios still hits the 7d gate and gets blocked. Yarn then falls back to the latest version that satisfies the semver range and is at least 7 days old

Also tried running a small verification of both behaviors locally with yarn 4.10.3: 

Setup: fresh project on yarn 4.10.3 with `npmMinimalAgeGate: 525600` (365 days), `axios@^1.0.0` as the only direct dep, and toggled `npmPreapprovedPackages: [axios]` vs `[]` between the two runs to isolate the cascade vs fallback behavior.

1. **`npmPreapprovedPackages` does not cascade.** Allowlisted axios + 365d gate → `follow-redirects@^1.15.11: No candidates found`. The transitive hit the gate even though its parent was allowlisted.

2. **Yarn falls back to the latest semver-compatible version old enough to pass.** No allowlist + 365d gate + `axios@^1.0.0` → resolved to 1.8.4 instead of latest 1.15.0.

<details>

<summary>
    sidquest: 
</summary>

I went to check how scaffold-ui is actually built, because I kept wondering if maybe the deps are bundled into the dist and the consumer never resolves them. Turns out no. All three packages (hooks, components, debug-contracts) just run `tsc` and emit ESM to `dist/esm/`. No webpack, no rollup. The published dist keeps every import statement intact, and the published package.json ships a real `dependencies` field. So when someone runs `yarn install`, yarn reads scaffold-ui's package.json and resolves the deps fresh, which means every one of them goes through the 7d gate at the consumer's install time

</details>

So this is safe to merge 🙌 and yup the behvaour doesn't change for create-eth since while doing yarn install only while creating the yarn.lock it follows all this rules

.yarnrc.yml
```diff
@@ -1,13 +1,19 @@
+compressionLevel: mixed
+
 enableColors: true
 
+enableGlobalCache: false
```
rin-st: why not use a global cache?

rin-st (commented):
> Does the allowlist just wave the whole subtree through the gate?
> ...

Thank you for detailed comment!

Looking good! Added one question

carletex:
Thanks for the research, Shiv! <3

.yarnrc.yml
technophile-04: Ohh actually it was added by internal yarn migration script when I ran: 

```
yarn set version berry
yarn install
```

truns out from their doc, they actually do this because in yarn v3 we didn't had gloabl cache set to true that's why the script also followed the same here. 

But if you scaffold a new yarnv4 proejct its turned on by default. So in [`f899f4b` (this PR)](https://github.com/scaffold-eth/scaffold-eth-2/pull/1211/commits/f899f4b3034baedb008dbbd49b6c8520e6a9a249) I just turned it on. Also research there is no downsides of this. and tested by releasing new beta version at 

```
npx create-eth@0.0.0-beta-20260416083001
```

I initially thought that it might decrease the create-eth install time because of global cache but on my testing it was simlar only :( 

Thanks for pointing out!

rin-st: > I initially thought that it might decrease the create-eth install time because of global cache but on my testing it was simlar only :(

I cleared cache and tried to check install times:

`time npx create-eth@0.0.0-beta-20260416083001 -s hardhat <folder-name>`

First installation:
`129.93s user 71.56s system 105% cpu 3:10.89 total`

Second (packages cached)
`41.92s user 58.37s system 202% cpu 49.534 total`

Description by AI

```
  - user 129.93s — CPU time in userspace (compilation, JS execution)
  - system 71.56s — CPU time in kernel (file operations, network, I/O)
  - cpu 105% — slightly more than 1 core used (parallel processes)
  - total 3:10.89 — real wall-clock time (3 min 11 sec)

  User + system > total because some work ran in parallel across multiple cores.

  The main number is 3:10.89 total — that's how long the user actually waited.
```

rin-st: Also, `enableGlobalCache: true` is default value so it's not necessary to keep it in config file. But I think it's fine to keep it explicit too. Up to you

technophile-04:
Merging this 🙌

PR #1213, opened 2025-12-05
add bg-mainnet-rpc as extra fallback for mainnet

### LINKED ISSUE #1209: use buidlguidl client rpc url for mainnet
### Description: 

Currently, this code governs the RPC endpoint that is being selected. 
https://github.com/scaffold-eth/scaffold-eth-2/blob/f4ce312db2d2f87c3f08e34a73a9b13f36afde9f/packages/nextjs/services/web3/wagmiConfig.tsx#L20-L32

So if the user haven't configured the alchemy api key. We use public rpc first. On mainnet, the public RPC configured by Viem most of the time takes too long, which doesn't allow the fallback default Alchemy RPC to run. 

Since we have a more reliable pulic rpc by buidlguidl it makes sense to use it. 

### Some solutions: 

1. Just uncomment the ovverride: 

https://github.com/scaffold-eth/scaffold-eth-2/blob/f4ce312db2d2f87c3f08e34a73a9b13f36afde9f/packages/nextjs/scaffold.config.ts#L30-L33

We can just uncomment this, it will use mainnet bg RPC for mainnet always. 

Disadvantage: Forms an kind of edge case in genereal flow of configuring RPC urls. If people bring their own Alchemy api key then they specially have to remmber for mainnet then need to uncommet the ovverride. 

Advantage: Simple solution. 

2. Just for mainnet add BG client RPC as first fallback: 

```ts
      if (alchemyHttpUrl) {
        const isUsingDefaultKey = scaffoldConfig.alchemyApiKey === DEFAULT_ALCHEMY_API_KEY;
        // If using default Scaffold-ETH 2 API key, we prioritize the default RPC
        // for mainnet, because default RPC takes time we priortise buidlguidl client first which is more reliable.
        const mainnetFallbackWithDefaultRPC = [http("https://mainnet.rpc.buidlguidl.com")];
        rpcFallbacks = isUsingDefaultKey
          ? [...(chain.id === mainnet.id ? mainnetFallbackWithDefaultRPC : []), http(), http(alchemyHttpUrl)]
          : [http(alchemyHttpUrl), http()];
      }
```

So if mainnet is enabled and while using default Alchemy key. We have 3 fallback for mainnet, first is BG client. 

Disadvantage: Adds a bit of complexity to the code. If user tries to read wagmiConfig he might get lost. 

Advantage: The flow remains same and user don't need to think of this exception. 

cc @carletex  and @rin-st fo this

technophile-04:
## Description

Fixes #1209 

Umm not strong opinion either. Just leaning towards this approach a bit, but happy to go with override RPC url

### DIFF
--- a/packages/nextjs/services/web3/wagmiConfig.tsx
+++ b/packages/nextjs/services/web3/wagmiConfig.tsx
@@ -17,21 +17,22 @@ export const wagmiConfig = createConfig({
   connectors: wagmiConnectors(),
   ssr: true,
   client({ chain }) {
-    let rpcFallbacks = [http()];
-
-    const rpcOverrideUrl = (scaffoldConfig.rpcOverrides as ScaffoldConfig["rpcOverrides"])?.[chain.id];
     // Extra fallback for mainnet.
     const mainnetFallbackWithDefaultRPC = [http("https://mainnet.rpc.buidlguidl.com")];
+    let rpcFallbacks = [...(chain.id === mainnet.id ? mainnetFallbackWithDefaultRPC : []), http()];
+
+    const rpcOverrideUrl = (scaffoldConfig.rpcOverrides as ScaffoldConfig["rpcOverrides"])?.[chain.id];
+
     if (rpcOverrideUrl) {
-      rpcFallbacks = [http(rpcOverrideUrl), ...(chain.id === mainnet.id ? mainnetFallbackWithDefaultRPC : []), http()];
+      rpcFallbacks = [http(rpcOverrideUrl), ...rpcFallbacks];
     } else {
       const alchemyHttpUrl = getAlchemyHttpUrl(chain.id);
       if (alchemyHttpUrl) {
         const isUsingDefaultKey = scaffoldConfig.alchemyApiKey === DEFAULT_ALCHEMY_API_KEY;
         // If using default Scaffold-ETH 2 API key, we prioritize the default RPC
         rpcFallbacks = isUsingDefaultKey
-          ? [...(chain.id === mainnet.id ? mainnetFallbackWithDefaultRPC : []), http(), http(alchemyHttpUrl)]
-          : [http(alchemyHttpUrl), ...(chain.id === mainnet.id ? mainnetFallbackWithDefaultRPC : []), http()];
+          ? [...rpcFallbacks, http(alchemyHttpUrl)]
+          : [http(alchemyHttpUrl), ...rpcFallbacks];
       }
     }
 


carletex (approved):
Looks and works great.

Thanks!

packages/nextjs/services/web3/wagmiConfig.tsx
```diff
@@ -20,14 +20,18 @@ export const wagmiConfig = createConfig({
...
 
     const rpcOverrideUrl = (scaffoldConfig.rpcOverrides as ScaffoldConfig["rpcOverrides"])?.[chain.id];
+    // Extra fallback for mainnet.
+    const mainnetFallbackWithDefaultRPC = [http("https://mainnet.rpc.buidlguidl.com")];
     if (rpcOverrideUrl) {
-      rpcFallbacks = [http(rpcOverrideUrl), http()];
+      rpcFallbacks = [http(rpcOverrideUrl), ...(chain.id === mainnet.id ? mainnetFallbackWithDefaultRPC : []), http()];
     } else {
       const alchemyHttpUrl = getAlchemyHttpUrl(chain.id);
       if (alchemyHttpUrl) {
```
rin-st: if for some reason there is no `alchemyHttpUrl` for mainnet (which is very weird, user need to remove `scaffoldConfig.alchemyApiKey` and/or `RPC_CHAIN_NAMES[chainId]`), then for mainnet there will be no `mainnetFallbackWithDefaultRPC` fallback . 

To avoid it we can change code a bit

```typescript
client({ chain }) {
    const mainnetFallbackWithDefaultRPC = [http("https://mainnet.rpc.buidlguidl.com")];
    let rpcFallbacks = [...(chain.id === mainnet.id ? mainnetFallbackWithDefaultRPC : []), http()];

    const rpcOverrideUrl = (scaffoldConfig.rpcOverrides as ScaffoldConfig["rpcOverrides"])?.[chain.id];

    // Extra fallback for mainnet.
    if (rpcOverrideUrl) {
      rpcFallbacks = [http(rpcOverrideUrl), ...rpcFallbacks];
    } else {
      const alchemyHttpUrl = getAlchemyHttpUrl(chain.id);
      if (alchemyHttpUrl) {
        const isUsingDefaultKey = scaffoldConfig.alchemyApiKey === DEFAULT_ALCHEMY_API_KEY;
        // If using default Scaffold-ETH 2 API key, we prioritize the default RPC
        rpcFallbacks = isUsingDefaultKey
          ? [...rpcFallbacks, http(alchemyHttpUrl)]
          : [http(alchemyHttpUrl), ...rpcFallbacks];
      }
    }
```

But again the case is very rare so you can omit this

PR #1221, opened 2026-01-14
add agents.md 

technophile-04:
### Description: 

Removed the cursor rules and we now have `AGENTS.md` and `CLAUDE.md` file. `Claude.md` file basically references to `AGENTS.md` only because we don't want to maintain two separate similar copies. 

> 1. **Document requirements**: Clearly specify dependencies and commands in your `CLAUDE.md` file. If you have an `AGENTS.md` file, you can source it in your `CLAUDE.md` using `@AGENTS.md` to maintain a single source of truth. 
~ [Calude Code docs](https://code.claude.com/docs/en/claude-code-on-the-web#best-practices)

So this file will basically will be added as system promp by the harness you are using (not all but most of them do it). So for every new conversion the harness will preload this file in context. 

> Your CLAUDE.md file becomes part of Claude's system prompt. Every conversation starts with this context already loaded, eliminating the need to explain basic project information repeatedly.

While researching what are the best practices, there is as such no standard and pepople are actually writing plain egnlish in ths file about project. But a certain pattern I have seen in most of the famous repos is that they use this file as a guardrail for AI so that it does't go too crazy, keep the consistency for all the common things, and know how to look into the repo.

So people mainly include code style guidelines, testing instructions, repository conventions, developer environment setup, and project-specific warnings. Tried following the same flow. 

I initally generated this file by using `/innit` command from calude code (which generated CLAUDE.md and then copied it over to AGENTS.md) and followed the pattern mentioned about. 

Generally everyone suggested to keep the file max 300-600 lines and in some places I even found people suggesting to keep it under 100. Seem like 200 is sweet spot for us since we could mention all the imp things. But we can decrease and refine a bit because I feel the less liness + cripier it is better for LLM. 

An nice line which I liked in Claude documentation for adding all the AI compatibility (like skills, subagents, plans.md etc) to your repo was: 

> **Start simple, expand deliberately**

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,9 @@
 
 🧪 An open-source, up-to-date toolkit for building decentralized applications (dapps) on the Ethereum blockchain. It's designed to make it easier for developers to create and deploy smart contracts and build user interfaces that interact with those contracts.
 
+> [!NOTE]
+> 🤖 Scaffold-ETH 2 is AI-ready! It has everything agents need to build on Ethereum. Check `.agents/`, `.claude/`, `.opencode` or `.cursor/` for more info.
+
 ⚙️ Built using NextJS, RainbowKit, Foundry/Hardhat, Wagmi, Viem, and Typescript.
 
 - ✅ **Contract Hot Reload**: Your frontend auto-adapts to your smart contract as you edit it.


.agents/agents/docs-fetcher-summarizer.md
```diff
@@ -0,0 +1,141 @@
+You are an expert documentation researcher and technical information synthesizer specializing in extracting relevant, actionable information from blockchain development library and framework documentation. Your role is to fetch, analyze, and summarize specific documentation sections that will enable another agent to successfully implement features in a Scaffold-ETH 2 project.
```
rin-st: A lot of people use [context7](https://github.com/upstash/context7) for documentations. Take a look, it could be useful for us too. So instead of finding the info from the whole documentation website we can use that mcp, I think it should be much faster and cheaper token wise. That repo has docs of everything below including se-2. (but just https://ethereum.org/ instead of https://ethereum.org/developers) I'll test it tomorrow if you won't do it before me :)

---

Tried current version and yep, it's kinda slow (probably I was thinking what to answer for some tens of seconds, not sure if it counts) and relatively expensive 

Image

AGENTS.md
```diff
@@ -0,0 +1,184 @@
+# AGENTS.md
```
Pabl0cks: Suggestion: Maybe in a future iteration we could explore the nested agents (having 1 global, 1 for nextjs and one for hardhat/foundry). See https://cursor.com/docs/context/rules#improvements

Image

AGENTS.md
```diff
@@ -0,0 +1,184 @@
...
+  eventName: "GreetingChange",
+  watch: true,
+  fromBlock: 31231n,
+  blockData: true,
+});
+```
+
+SE-2 also provides other hooks to interact with blockchain data: `useScaffoldWatchContractEvent`, `useScaffoldEventHistory`, `useDeployedContractInfo`, `useScaffoldContract`, `useTransactor`.
+
+**IMPORTANT: Alwayes use hooks from `packages/nextjs/hooks/scaffold-eth` for contract interactions. Always refer to the hook names as they exist in the codebase.**
```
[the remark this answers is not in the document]
technophile-04: fixed thanks!

.claude/skills/defi-protocol-templates
```diff
@@ -0,0 +1 @@
+../../.agents/skills/defi-protocol-templates
```
rin-st: nit: no empty line at the end here and in some other files

technophile-04:
Pushed some small tweak mainly including content related to foundry. 

We could have used template and conditionally only show the foundry content when foundry was selected, but while discussing with Carlos we thought AI is smart enough to infer it and follow the instrction mentioned depending on flavour choose. We can also update it later to use template. 

Merging this as v0.1 and we can iterate more since this doesn't have any changes to core logic of SE-2

PR #1227, opened 2026-01-30
Add-extension skill

rin-st:
Still work in progress, but you can already test it and make a general overview

`/add-extension` skill adds extensions to existing SE-2 projects. 

Allows adding multiple extensions one by one. For now works only with `create-eth-extensions` extensions and only with Claude code

---

  Example:

```
  claude /add-extension erc-20
```

Or just open claude and run `/add-extension` with or without params. It suggests you what to do if needed

(I tested mostly with erc-20, erc-721, ponder)

---

How it works, skill + Claude together (generated by Claude)

  The skill script:
  1. Validates SE-2 project and extension name
  2. Fetches extension files from repo
  3. Detects frameworks (Hardhat/Foundry)
  4. Analyzes what needs to change (new files, merges, dependencies)
  5. Auto-merges simple cases (new files, package.json deps)
  6. Outputs summary of changes for Claude to see

  Then Claude takes over:
  1. Reviews the skill's output/summary
  2. Handles complex merges - when files conflict or need manual integration
  3. Suggests changes - explains what the extension does and how to integrate it
  4. Edits files - merges configs, updates imports, adapts extension code to your project
  5. Installs deps - runs yarn/npm install if needed

---

There are multiple things in the script that do the same as create-eth does: fetch extension data, detect framework etc. Since we can't use create-eth here I think it's ok. There's multiple things that I can mention there, but I think it's easier to just read the code. Start from `skill.mjs` and then we can discuss. Or just ask any questions here!

Some notes anyway:

When adding extension, script adds new field to current app `package.json`, so it could understand which extensions already installed when adding new extensions.

```
  "scaffoldEth": {
    "extensions": [
      "erc-721"
    ]
  }
```

The most interesting part is how it allows the use multiple extensions:
- it fetches files of all the extensions
- do regular things for non-conflicting files
- if there's some args files which is in 2 or more extensions, it takes all the args, combine them together, and then use in the template

Solidity framework is chosen by current project structure. So if there's hardhat in the current project and there's both frameworks in extension, hardhat will be taken automatically. There's not so clear for now what to do if the user wants to take opposite framework in the extension.
- prohibit it? (I think I'll do it that way)
- rewrite solidity framework folder and package.json script?

There's some small bugs/AI hallucinations I'm still working on. But feel free to add your thoughts!

### DIFF
--- a/.claude/skills/add-extension/README.md
+++ b/.claude/skills/add-extension/README.md
@@ -2,6 +2,8 @@
 
 Claude Code skill for adding Scaffold-ETH-2 extensions to existing projects post-creation.
 
+**Requirements:** SE-2 project root, Git, Yarn
+
 ## Quick Start
 
 ```bash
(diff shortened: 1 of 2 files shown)


Pabl0cks (commented):
Awesome job Rinat! This is looking good, it's going to be awesome to have this working in prod 🔥

Did a first pass from my Windows machine, executing the `/add-extension` with Claude, installing `Ponder` and then `Drizzle-neon`, will try it from WSL too but writing down the "walls" I faced, which were quickly spoted&fixed by cursor+opus.

During the installation I saw some errors, pasting Ponder and then Drizzle-neon installations:

<details><summary>Screenshots</summary>

Ponder
Image
Drizzle-neon
Image

</details>
 
Then trying to use the SE-2 instance with the extensions installed:
- Missed to add `drizzle-kit` commands to nextjs package.json
- Missed to add the drizzle-neon related dependencies to nextjs `package.json`
- Workspace '@se-2/ponder' not found. Cursor added it

Image

After those fixes, the database and ponder worked nicely. Will update when I test with WSL 🙌

damianmarti:
Wow!! This feature is great!!!

carletex:
> "Let Claude read and Merge the file"

Not only that, but maybe the future is more like:
- No template files / engine / etc
- Just a curated list of well-crafted examples that the LLM can just look and "insert" into your project.

This might be a bigger conversation, but we might be going to a place where _extensions_ don't make sense in the current state, and it's just more like "just tell Claude what you want" (especially if there is a curated place where it can look at examples)

PR #1263, opened 2026-04-08
fix: add hardhat deploy script gas-limit guidance to AGENTS.md

### LINKED ISSUE #1262: yarn deploy fails with gas-cap errors on manual post-deploy contract calls
When a Hardhat deploy script makes manual contract calls after deployment (e.g. transferOwnership, grantRole, initialize), the transaction can fail with a gas-limit error.

This happens because these calls inherit blockGasLimit from the Hardhat network config as their gas cap, which may exceed the practical per-tx limit accepted by the running yarn chain node or a forked environment.

Pabl0cks:
## Description

Adds a note under the Hardhat flavor section of AGENTS.md guiding agents to set explicit gas limits on manual post-deploy calls (transferOwnership, grantRole, initialize, etc.) to prevent gas errors when inheriting the default `hardhat.config.ts` blockGasLimit. 
Another solution would be globally setting a blockGasLimit in `hardhat.config.ts`, but I think this might be a cleaner solution.

Includes two recommended patterns:

- estimateGas + 20% margin for general calls
- Explicit gasLimit for simple, known admin operations

## Additional Information

## Related Issues

Fixes #1262


technophile-04:
Thanks Pablo!

I was thinking since agents are pretty good at picking the right fix when they understand what's going wrong, maybe we just have something like isntead of hardcoded code block: 

"if you hit a gas-cap error on a post-deploy call, the real failure is usually hidden underneath the AutomaticGasProvider fallback, try staticCall first to surface it" 

like giving them a mental model they can apply and let the agent find a better way


<details>

<summary>
    Context: 
</summary>

> When a Hardhat deploy script makes manual contract calls after deployment (e.g. transferOwnership, grantRole, initialize), the transaction can fail with a gas-limit error. 

From #1262, like this PR change actually won't 100% solve that issue, but good, nice changes to have, I feel. 


tried diggin a bit deeper to see what's actually happening, when you call `myContract.myMethod(...)` without a `gasLimit`, hardhat-ethers hands the tx to hardhat core's provider stack. One of those is the [AutomaticGasProvider](https://github.com/NomicFoundation/hardhat/blob/v2/packages/hardhat-core/src/internal/core/providers/gas-providers.ts#L120), whose job is to fill in gas field before the tx goes out. it does that by calling `eth_estimateGas` internally. The problem is the failure path, if that internal estimate throws because the call would revert (say, `"Ownable: caller is not the owner"`), the provider doesn't bubble it up. it [catches the error and silently falls back to `blockGasLimit`](https://github.com/NomicFoundation/hardhat/blob/v2/packages/hardhat-core/src/internal/core/providers/gas-providers.ts#L92-L94), 30M by default. The tx then gets rejected by the node for exceeding the block cap, so what the user sees is a confusing "gas limit too high" error instead of the actual "not the owner" revert becuase it gets swalloed. 

that's why we have the two workarounds in the `AGENTS.md` snippet, and i think they're solving different problems:

- **the explicit `estimateGas + 20%` pattern** short-circuits the fallback by doing the estimate ourself. when the call would revert, our `estimateGas` line throws with the real revert reason and the script stops before any tx goes out. good for surfacing the real errors, no gas wasted.
- **the hardcoded `gasLimit: 100_000` pattern** skips estimation entirely, which helps when `estimateGas` is flaky for non-revert reasons (forked state weirdness, rpc quirks), cases where the first pattern would also throw and kill the script even though the tx would've succeeded on-chain. tradeoff there is that if the call would actually revert, the tx still goes out and the user pays gas up to the revert point.

so they're almost opposites. one surfaces the real error, the other skips estimation to avoid false negatives. they don't stack and neither alone covers both cases.

one shape that might work better is `staticCall` first (pure `eth_call` dry-run, no tx, no gas), then `estimateGas + margin` if the call is valid, then hardcoded `gasLimit` only when estimateGas is demonstrably flaky:

<details>

<summary>
    example snippet
</summary>

```typescript
try {
  await myContract.myMethod.staticCall(arg1, arg2); // this will check if there is actual "Not the owner error" 
  const gas = await myContract.myMethod.estimateGas(arg1, arg2); 
  await myContract.myMethod(arg1, arg2, { gasLimit: (gas * 120n) / 100n });
} catch (e) {
  // real revert surfaces here, or fall back to hardcoded gasLimit if rpc is flaky
}
```

</details>


</details>

Pabl0cks:
Thanks @technophile-04 for the nice research! ♥

I've been testing adding your sentence to let models try to get to the right root cause and fix, with claude code+ sonnet 4.6 and codex + gpt 5.4 and they both end up adding the `{ gasLimit: 100_000 })` to fix the issue, after 2-3 tries of trying to uncover the root problem. 

<details><summary>This is the reasoning of gpt 5.4:</summary>

## Summary

The failing command was:

```powershell
yarn deploy
```

The visible error was:

```text
ProviderError: Transaction gas limit is 60000000 and exceeds transaction gas cap of 16777216
```

The contracts themselves were deploying successfully:

```text
YourContract
StakeCoin
YieldCoin
StakingContract
```

The failure happened after deployment, during this post-deploy setup call in `packages/hardhat/deploy/01_deploy_staking.ts`:

```ts
await yieldCoinInstance.setMinter(stakingContract.address);
```

So the deploy bug was not caused by contract creation. It was caused by the transaction that configures `YieldCoin` so the `StakingContract` can mint reward tokens.

## Path 1: Inspect The Deploy Script

The first file checked was:

```text
packages/hardhat/deploy/01_deploy_staking.ts
```

The original flow was:

```ts
const stakeCoin = await deploy("StakeCoin", ...);
const yieldCoin = await deploy("YieldCoin", ...);
const stakingContract = await deploy("StakingContract", ...);

const yieldCoinInstance = await hre.ethers.getContractAt("YieldCoin", yieldCoin.address);
await yieldCoinInstance.setMinter(stakingContract.address);
```

Reason for checking this first: the stack trace pointed directly at `setMinter`, and the repo instructions mention that gas-cap errors on post-deploy calls often hide the real underlying failure.

Result: this identified the relevant failing operation, but not the exact root cause yet.

## Path 2: Inspect YieldCoin And StakingContract

The next file checked was:

```text
packages/hardhat/contracts/YieldCoin.sol
```

The relevant code was:

```solidity
address public minter;

error NotMinter();
error MinterAlreadySet();

function setMinter(address _minter) external {
    if (minter != address(0)) revert MinterAlreadySet();
    minter = _minter;
}
```

This showed that `setMinter` can only succeed once.

Reason for checking this: a failed retry after a partial deploy could call `setMinter` again against an already configured `YieldCoin`, causing `MinterAlreadySet()`.

Result: this revealed a real deploy-script bug. The script was not idempotent. If deployment partially succeeded once, rerunning `yarn deploy` could fail on the setup transaction.

Why this was not the whole fix: after making the script idempotent, the first retry still failed while trying to send `setMinter`. That meant there was another issue beyond only "minter already set".

## Path 3: Check Whether The Failure Was A Revert Hidden By Gas Estimation

The first patch made the deploy script read the current minter:

```ts
const currentMinter = await yieldCoinInstance.minter();

if (currentMinter === hre.ethers.ZeroAddress) {
  await yieldCoinInstance.setMinter.staticCall(stakingContract.address);
  const tx = await yieldCoinInstance.setMinter(stakingContract.address);
  await tx.wait();
} else if (currentMinter.toLowerCase() === stakingContract.address.toLowerCase()) {
  console.log("YieldCoin minter already set to StakingContract:", stakingContract.address);
} else {
  throw new Error(...);
}
```

Reason for this path: `staticCall` is a good way to expose a Solidity revert before sending a transaction. If `setMinter` was reverting because of `MinterAlreadySet`, `staticCall` would fail with the real contract error instead of the misleading gas-cap message.

Result: `staticCall` succeeded, but the actual transaction still failed with:

```text
Transaction gas limit is 60000000 and exceeds transaction gas cap of 16777216
```

Why this path partially failed: it proved the Solidity call was valid, but it did not address Hardhat or ethers assigning an invalid gas limit to the real transaction.

Why it was still useful: it separated two possible causes:

- contract revert: not the current issue
- bad transaction gas limit: the actual remaining issue

## Path 4: Check Signer Usage

The script originally used:

```ts
const yieldCoinInstance = await hre.ethers.getContractAt("YieldCoin", yieldCoin.address);
```

It was changed to:

```ts
const deployerSigner = await hre.ethers.getSigner(deployer);
const yieldCoinInstance = await hre.ethers.getContractAt("YieldCoin", yieldCoin.address, deployerSigner);
```

Reason for this path: deploy scripts should be explicit about which signer sends post-deploy transactions. If a wrong signer or provider-only contract instance were involved, the transaction could behave unexpectedly.

Result: this made the script clearer and safer, but it did not fix the gas-cap error by itself.

Why it failed as the primary fix: `YieldCoin.setMinter` has no owner check, so signer authorization was not the reason the transaction failed. The transaction was still being given a bad gas limit.

## Path 5: Pin A Small Explicit Gas Limit

The remaining failure happened here:

```ts
const tx = await yieldCoinInstance.setMinter(stakingContract.address);
```

Even though `staticCall` succeeded, ethers or Hardhat still sent the transaction with a `60,000,000` gas limit. The local node cap was:

```text
16777216
```

So the send was changed to:

```ts
const tx = await yieldCoinInstance.setMinter(stakingContract.address, { gasLimit: 100_000 });
```

Reason for this path: `setMinter` is a simple single-storage-write function. It does not need anything close to 60 million gas. A 100,000 gas limit is conservative and safely above the expected cost of setting one storage slot, while staying far below the local transaction gas cap.

Result: `yarn deploy` succeeded.

The command was then run a second time. The second run also succeeded and printed:

```text
YieldCoin minter already set to StakingContract
```

That confirmed both parts of the fix:

- first-time setup works
- repeated deploys do not fail after a partial deployment

## Final Fix

The final deploy script now does three important things:

```ts
const deployerSigner = await hre.ethers.getSigner(deployer);
const yieldCoinInstance = await hre.ethers.getContractAt("YieldCoin", yieldCoin.address, deployerSigner);
const currentMinter = await yieldCoinInstance.minter();

if (currentMinter === hre.ethers.ZeroAddress) {
  await yieldCoinInstance.setMinter.staticCall(stakingContract.address);
  const tx = await yieldCoinInstance.setMinter(stakingContract.address, { gasLimit: 100_000 });
  await tx.wait();
  console.log("YieldCoin minter set to StakingContract:", stakingContract.address);
} else if (currentMinter.toLowerCase() === stakingContract.address.toLowerCase()) {
  console.log("YieldCoin minter already set to StakingContract:", stakingContract.address);
} else {
  throw new Error(`YieldCoin minter is already set to ${currentMinter}, expected ${stakingContract.address}`);
}
```

## Why This Is The Right Fix

This is the right fix because it addresses the actual deploy failure without weakening the contracts or changing their intended behavior.

The contract design says `YieldCoin.minter` should be assigned once. The deploy script now respects that by checking the current value before trying to set it.

The gas-cap error came from the transaction being sent with an invalid gas limit for the local node. The fix pins a reasonable gas limit only for the simple `setMinter` setup call. It does not raise the node gas cap, does not alter Hardhat global config, and does not mask contract failures.

The `staticCall` is also useful because if `setMinter` ever becomes invalid in the future, the deploy script should fail before sending the transaction and expose the real contract-level problem.

The final fix is intentionally narrow:

- no contract behavior changed
- no network-wide gas settings changed
- deploys are now retry-safe
- the post-deploy setup transaction uses a valid gas limit
- future unexpected minter mismatches fail loudly instead of silently continuing

</details>

technophile-04:
Ohh just inspecting your current code snippet, I think it solves it solves the problems 🙌 like handles the `staticCall`  internally since the `estimateGas` will revert if there is actual revert reason. Soo this is just perfect! 

Thanks Pablo!

technophile-04 (approved):

PR #1078, opened 2025-03-25
Tailwind v4 + Daisy v5 migration

carletex:
(Just run the npx codemod TW tool). @technophile-04 will continue with the missing parts (Daisy, etc)

Fixes #1069

PR #1266, opened 2026-04-13
better default for border radius

### LINKED ISSUE #1264: bug: large rounded text areas cut off text

- [x] I have looked through the [existing issues](https://github.com/scaffold-eth/scaffold-eth-2/issues)

npx create-eth@latest

### Current Behavior

gm sers

i keep running into a weird issue where large text areas have a huge border radius thing 

Image

i was going to keep adding it to my skill to fix this but i think it's worth bringing back to you guys first

🫡 this issue was written by a human 

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

carletex:
Fixes #1264 

Changedradius default from 9999rem to 1rem.. which fixes the pill-shaped textarea issue, and it's unnoticeable for shorter elements like buttons and inputs since the browser caps border-radius at half the element's height anyway (and some of them are hardcoded already)

I think nothing changed visually (just textareas), but please take a look!

Before
image

After
image

### DIFF
--- a/packages/nextjs/styles/globals.css
+++ b/packages/nextjs/styles/globals.css
@@ -106,6 +106,7 @@
 
 .btn {
   @apply shadow-md;
+  border-radius: 9999rem;
 }
 
 .btn.btn-ghost {
--- a/packages/nextjs/styles/globals.css
+++ b/packages/nextjs/styles/globals.css
@@ -33,7 +33,7 @@
   --color-warning: #ffcf72;
   --color-error: #ff8863;
 
-  --radius-field: 9999rem;
+  --radius-field: 1rem;
   --radius-box: 1rem;
   --tt-tailw: 6px;
 }
@@ -58,7 +58,7 @@
   --color-warning: #ffcf72;
   --color-error: #ff8863;
 
-  --radius-field: 9999rem;
+  --radius-field: 1rem;
   --radius-box: 1rem;
 
   --tt-tailw: 6px;


technophile-04 (approved):
Thanks Carlos! Was just trying to dig why at first place we had `999rem` in the button, we added it at https://github.com/scaffold-eth/scaffold-eth-2/commit/3222718f7fac8b4fb8a58a2a98fd3bf2d0653d10, so I think the main goal of `999 rem` was to keep **only button** pilled. Rest keep other components as daisyUI defaults (daisyUI pre v5 defaults)

BUT then we migrated to DaisyUI v5 at #1078, we ported 

`--rounded-btn: 9999rem` → `--radius-field: 9999rem`  

they both are not same :( actually `--radius-field: 9999rem` covers more than buttons and it included textareas, inputs as well. And that's why this happend.  

Pushed https://github.com/scaffold-eth/scaffold-eth-2/pull/1266/commits/e244ac96ab689c1923489e220c8547245630f22c so that we only target the button pillnes (preservaing 2023 https://github.com/scaffold-eth/scaffold-eth-2/commit/3222718f7fac8b4fb8a58a2a98fd3bf2d0653d10 intension) without https://github.com/scaffold-eth/scaffold-eth-2/pull/1266/commits/e244ac96ab689c1923489e220c8547245630f22c while increasing the size of the button, it would look a bit boxy: 

```tsx
<button className="h-12 btn btn-primary">Hello world</button>
```
| Before https://github.com/scaffold-eth/scaffold-eth-2/pull/1266/commits/e244ac96ab689c1923489e220c8547245630f22c | After https://github.com/scaffold-eth/scaffold-eth-2/pull/1266/commits/e244ac96ab689c1923489e220c8547245630f22c | 
| ------- | ----- | 
| Screenshot 2026-04-13 at 3 40 20 PM | Screenshot 2026-04-13 at 3 39 46 PM  | 


I thought we could just have https://github.com/scaffold-eth/scaffold-eth-2/pull/1266/commits/e244ac96ab689c1923489e220c8547245630f22c and remove the `--radius-field: 1rem;` completely, but it's actually required since from daisyUI v4 => daisyUI v5, daisyUI v5 moved to sharper corners and having https://github.com/scaffold-eth/scaffold-eth-2/pull/1266/commits/e244ac96ab689c1923489e220c8547245630f22c actually makesure we have a bit of rounded things

rin-st:
Looking good to me!

I think we can skip updating scaffold-ui for now, so sui inputs will have full rounded borders as before. 

But later, after styles change, we'll need to update se-2 components too

rin-st (approved):

PR #1272, opened 2026-04-17
hardhat v3 migration

technophile-04:
### To test:

Make sure to have clean install of se-2 because branching from `main` comes with old gitignored files which might cause type checks failure or deployments error

```shell
git clone --branch hardhat-v3-migration https://github.com/scaffold-eth/scaffold-eth-2.git se-2-hardhat-v3
```

```shell
cd se-2-hardhat-v3
yarn install
```

### Description: 

Migrated hardhat v2 -> hardhat v3 and hardhat-deploy v1 -> hardhat-deploy -> v2. 

All the magical commands and their flags work in the same way, without needing to bend and add our extra glue to make them work . 

I think this is ready to be tested as initial iteration 🙌

#### Some things to know: 

- Everything is ESM. `"type": "module"` in `package.json`, `module: "node16"` in tsconfig, `.js` extensions on every relative import
- `hardhat compile` must run before `hardhat deploy` because deploy scripts statically import from `generated/artifacts/`, and Node resolves those imports before any code runs. Our `yarn deploy` script chains `hardhat compile && tsx scripts/runHardhatDeployWithPK.ts` for exactly this reason.
- Swapped `ts-node` for `tsx` because the`ts-node` was pain with `module: node16` 
- Dropped `hardhat-gas-reporter` for now since it hasn't shipped a v3-compatible release yet (every version through 2.3.0 pins `hardhat: ^2.16.0`)
- `--reset` isn't a built-in CLI flag in v2. v2 exposes `alwaysOverride: true` per-deploy call instead. We added back the flag via a task override in `hardhat.config.ts` + a `deployWithReset` wrapper in `rocketh/config.ts`. Filed [wighawag/hardhat-deploy#592](https://github.com/wighawag/hardhat-deploy/issues/592) asking for native support.
- Etherscan V2 error messages are misleading. "Missing or unsupported chainid parameter" usually means "unsupported," not "missing." Chain 31337 (local hardhat) returns this error. 
- TODO: Get hardhat-verify working, currently the base commands uses `rocketh/verfifier` which is hardhat-deploy compatible

### Some notes: 

<details>

<summary>
    What's rocketh
</summary>

Rocketh is a framework-agnostic deployment engine. In v1, hardhat-deploy was a monolith tightly coupled to hardhat 2 internals. When hardhat 3 rewrote the plugin system, that coupling broke. So v2 pulled the deployment engine out into rocketh (framework-agnostic) and rebuilt hardhat-deploy v2 as a thin adapter that bridges hardhat 3 into rocketh. 

Architecture:

```
hardhat 3 (compilation, networks, tasks)
   └── hardhat-deploy v2 (plugin adapter)
         └── rocketh (deployment engine)
                ├── @rocketh/deploy        (deploy function)
                ├── @rocketh/read-execute  (read/write)
                ├── @rocketh/signer        (signing)
                └── @rocketh/node          (file-based env loading)
```

**Why three files under `packages/hardhat/rocketh/`**

- **`config.ts`** — named accounts (`deployer` at index 0), signer protocols (registers `privateKey` from `@rocketh/signer`), and the extension registry (`deploy` + `read` + `execute`). 
- **`deploy.ts`** — one-line glue: exports `deployScript` from `setupDeployScripts`. Every `deploy/*.ts` file imports this to wrap its script body.
- **`environment.ts`** — exports `loadAndExecuteDeploymentsFromFiles` and `loadEnvironmentFromHardhat`. These are the loaders tests and post-deploy scripts use to get a typed `env` with all deployments resolved.

The end users probably won't be required to touch this files. 

</details>

### Future TODO's 

- Using hardhat's configuration variable + keystore  
- Having the solidity test file examples as well (according to hardhat docs, both `.ts` and `.t.sol` helps in different-different way) 
- Migrating to viem 
- Tinkering with ignition, and see if it's good fit for us


packages/hardhat/hardhat.config.ts
```diff
@@ -36,118 +29,131 @@ const config: HardhatUserConfig = {
...
       accounts: [deployerPrivateKey],
     },
-    polygon: {
-      url: `https://polygon-mainnet.g.alchemy.com/v2/${providerApiKey}`,
-      accounts: [deployerPrivateKey],
-    },
-    polygonAmoy: {
-      url: `https://polygon-amoy.g.alchemy.com/v2/${providerApiKey}`,
-      accounts: [deployerPrivateKey],
-    },
```
technophile-04, line 88: Moved the networks a bit, kept polygon at bottom...Like for `yarn account` it easier to see the balance of network early

packages/hardhat/scripts/runVerify.ts
technophile-04, line 1: The `rocketh-verify` cli uses `-e` flag for the `network`, but to keep the `yarn deploy --network <network>` compatible, added this script

damianmarti:
@technophile-04 Thanks for the migration!!

I started the hardhat chain successfully, but then when I tried to deploy the contract with `yarn deploy`, I'm getting:

```
Compiled 1 Solidity file with solc 0.8.30 (evm target: prague)
An unexpected error occurred:

MalformedAbiError: Not a valid ABI
    at extractAbi (/ssd/workspace/buidlguidl/scaffold-eth-2/packages/hardhat/node_modules/typechain/src/parser/abiParser.ts:379:9)
    at <anonymous> (/ssd/workspace/buidlguidl/scaffold-eth-2/packages/hardhat/node_modules/typechain/src/typechain/io.ts:45:31)
    at Array.filter (<anonymous>)
    at skipEmptyAbis (/ssd/workspace/buidlguidl/scaffold-eth-2/packages/hardhat/node_modules/typechain/src/typechain/io.ts:45:6)
    at runTypeChain (/ssd/workspace/buidlguidl/scaffold-eth-2/packages/hardhat/node_modules/typechain/src/typechain/runTypeChain.ts:24:33)
    at generateTypes (/ssd/workspace/buidlguidl/scaffold-eth-2/packages/hardhat/node_modules/@nomicfoundation/hardhat-typechain/src/internal/generate-types.ts:50:24)
    at Array.build (/ssd/workspace/buidlguidl/scaffold-eth-2/packages/hardhat/node_modules/@nomicfoundation/hardhat-typechain/src/internal/hook-handlers/solidity.ts:54:13)
    at async next (/ssd/workspace/buidlguidl/scaffold-eth-2/packages/hardhat/node_modules/hardhat/src/internal/core/hook-manager.ts:121:13)
    at async runSolidityBuild (/ssd/workspace/buidlguidl/scaffold-eth-2/packages/hardhat/node_modules/hardhat/src/internal/builtin-plugins/solidity/tasks/build.ts:182:19)
    at async Promise.all (index 0)
```

technophile-04:
@damianmarti make sure to run: 

```
yarn install
```

And then: 

```
cd packages/hardhat && yarn clean
```

damianmarti:
> @damianmarti make sure to run:
> 
> ```
> yarn install
> ```
> 
> And then:
> 
> ```
> cd packages/hardhat && yarn clean
> ```

It worked! Thanks!!

damianmarti:
@technophile-04 Found errors running hardhat:check-types:

```
yarn hardhat:check-types
typechain-types/YourContract.ts:24:8 - error TS2835: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean './common.js'?

24 } from "./common";
          ~~~~~~~~~~

typechain-types/factories/YourContract__factory.ts:16:42 - error TS2835: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean '../common.js'?

16 import type { NonPayableOverrides } from "../common";
                                            ~~~~~~~~~~~

typechain-types/factories/YourContract__factory.ts:17:58 - error TS2835: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean '../YourContract.js'?

17 import type { YourContract, YourContractInterface } from "../YourContract";
                                                            ~~~~~~~~~~~~~~~~~

typechain-types/factories/index.ts:4:39 - error TS2835: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean './YourContract__factory.js'?

4 export { YourContract__factory } from "./YourContract__factory";
                                        ~~~~~~~~~~~~~~~~~~~~~~~~~

typechain-types/index.ts:4:35 - error TS2835: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean './YourContract.js'?

4 export type { YourContract } from "./YourContract";
                                    ~~~~~~~~~~~~~~~~

typechain-types/index.ts:5:28 - error TS2834: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

5 export * as factories from "./factories";
                             ~~~~~~~~~~~~~

typechain-types/index.ts:6:39 - error TS2835: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean './factories/YourContract__factory.js'?

6 export { YourContract__factory } from "./factories/YourContract__factory";
                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 7 errors in 4 files.

Errors  Files
     1  typechain-types/YourContract.ts:24
     2  typechain-types/factories/YourContract__factory.ts:16
     1  typechain-types/factories/index.ts:4
     3  typechain-types/index.ts:4
```

It's working ok on main branch.

packages/hardhat/scripts/runVerify.ts
```diff
@@ -0,0 +1,56 @@
+import "dotenv/config";
+import { spawn } from "child_process";
+
+const DEFAULT_ETHERSCAN_KEY = "DNXJA8RX2Q3VZ4URQIWP7Z68CJXQZSC6AW";
```
damianmarti: Maybe moving this to an export const on hardhat.config.ts

packages/hardhat/hardhat.config.ts
```diff
@@ -1,32 +1,25 @@
...
   process.env.__RUNTIME_DEPLOYER_PRIVATE_KEY ?? "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
-// If not set, it uses our block explorers default API keys.
-const etherscanApiKey = process.env.ETHERSCAN_V2_API_KEY || "DNXJA8RX2Q3VZ4URQIWP7Z68CJXQZSC6AW";
 
 // If not set, it uses ours Alchemy's default API key.
 // You can get your own at https://dashboard.alchemyapi.io
 const providerApiKey = process.env.ALCHEMY_API_KEY || "cR4WnXePioePZ5fFrnSiR";
 
-const config: HardhatUserConfig = {
+// TODO: We need to figure out where should the etherscan api key live. Currently we've runVerify.ts
```
damianmarti: Oh, I just added a comment about this :-)

technophile-04:
Ohh @damianmarti I think those files are carry over from `main` branch. When you were on main, the old @typechain/hardhat plugin wrote typechain-types/ with bare v2-style imports 

I think if you do a clean install of SE-2 it shall work: 

```shell
git clone --branch hardhat-v3-migration https://github.com/scaffold-eth/scaffold-eth-2.git se-2-hardhat-v3
```

```shell
cd se-2-hardhat-v3
yarn install
```

damianmarti:
> Ohh @damianmarti I think those files are carry over from `main` branch. When you were on main, the old @typechain/hardhat plugin wrote typechain-types/ with bare v2-style imports
> 
> I think if you do a clean install of SE-2 it shall work:
> 
> ```shell
> git clone --branch hardhat-v3-migration https://github.com/scaffold-eth/scaffold-eth-2.git se-2-hardhat-v3
> ```
> 
> ```shell
> cd se-2-hardhat-v3
> yarn install
> ```

Yes, sorry, you are right! Working now!

packages/hardhat/rocketh/config.ts
```diff
@@ -0,0 +1,39 @@
...
+  },
+  data: {},
+  signerProtocols: {
+    privateKey,
+  },
+} as const satisfies UserConfig;
+
+// TODO: Remove this wrapper when hardhat-deploy adds native --reset support
+// Tracking issue: https://github.com/wighawag/hardhat-deploy/issues/592
+const deployWithReset = ((env: any) => {
```
damianmarti: Can we avoid **any** type here?

packages/hardhat/package.json
```diff
@@ -1,61 +1,67 @@
...
   "name": "@se-2/hardhat",
   "version": "0.0.1",
+  "type": "module",
   "scripts": {
     "account": "hardhat run scripts/listAccount.ts",
     "account:generate": "hardhat run scripts/generateAccount.ts",
     "account:import": "hardhat run scripts/importAccount.ts",
     "account:reveal-pk": "hardhat run scripts/revealPK.ts",
-    "chain": "hardhat node --network hardhat --no-deploy",
+    "chain": "hardhat node --network hardhat",
```
damianmarti, line 10: Removing **--no-deploy** is because of a behavior change when starting the node from the current hardhat to V3, right?

rin-st:
Great job Shiv! I'm still reviewing, but here's some food for thought

1. I tried to copy-paste zk voting challenge to this branch and figured out that artifacts generated differently in v2 and v3 for npm deps. Npm deps artifacts are absent in `/hardhat/generated/artifacts/`  and should be extracted from `/hardhat/artifacts/build-info/**.output.json`. Without it, deployments for deps contracts fails

In case imported dep is inherited, we don't need that artifact, it became part of decendant contract:

```
import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";

contract YourToken is ERC20 {
...
```

But in case like this for zk-voting challenge, 
image

we need artifact and it could be extracted from `build-info`

Update: I see you did similar thing [here](https://github.com/scaffold-eth/scaffold-eth-2/blob/9df187a6015c756fb983ab252fe59db7456d47ed/packages/hardhat/scripts/generateTsAbis.ts#L22) for generating ts abis

```
/**
 * `PoseidonT3` and `LeanIMT` come from npm packages, so Hardhat does not emit
 * standalone artifact files for them in `generated/artifacts`. We load them
 * directly from the solc build-info instead.
 */
function loadBuildInfoArtifact(sourceName: string, contractName: string) {
  const buildInfoDir = join(__dirname, "..", "artifacts", "build-info");
// NOTE: Simplification. It could be different output files.
  const outputFile = readdirSync(buildInfoDir).find(f => f.endsWith(".output.json")); 
  if (!outputFile) {
    throw new Error("No solc build-info output found. Run `yarn compile` first.");
  }
  const { output } = JSON.parse(readFileSync(join(buildInfoDir, outputFile), "utf8"));
  const contract = output.contracts?.[sourceName]?.[contractName];
  if (!contract) {
    throw new Error(`Contract ${contractName} not found in build-info at ${sourceName}`);
  }
  return {
    contractName,
    sourceName,
    abi: contract.abi,
    bytecode: `0x${contract.evm.bytecode.object}`,
    deployedBytecode: `0x${contract.evm.deployedBytecode.object}`,
    linkReferences: contract.evm.bytecode.linkReferences ?? {},
    deployedLinkReferences: contract.evm.deployedBytecode.linkReferences ?? {},
  };
}

const poseidonT3Artifact = loadBuildInfoArtifact("npm/poseidon-solidity@0.0.5/PoseidonT3.sol", "PoseidonT3");
const leanIMTArtifact = loadBuildInfoArtifact("npm/@zk-kit/lean-imt.sol@2.0.1/LeanIMT.sol", "LeanIMT");

...

```

Reasons, from chats with Opus:

<details><summary>From claude</summary>
<p>

**Cursor**

Ah, got it — you're asking why `PoseidonT3` and `LeanIMT` just worked by **name** in the old Hardhat v2 deploy script, but in Hardhat v3 (rocketh) you had to manually load them from `build-info`.

Here's the Hardhat v2 version for reference:

```34:47:/Users/liana/Documents/prog/buidl-guidl/create-eth/qzkhh4/packages/hardhat/deploy/00_deploy_your_voting_contract.ts
  const poseidon3 = await deploy("PoseidonT3", {
    from: deployer,
    log: true,
    autoMine: true,
  });

  const leanIMT = await deploy("LeanIMT", {
    from: deployer,
    log: true,
    autoMine: true,
    libraries: {
      PoseidonT3: poseidon3.address,
    },
  });
```

No artifact, no imports. It just works. Why?

## The difference is in how the two deploy systems resolve artifacts

### Hardhat v2 + `hardhat-deploy` (the old project)

`hre.deployments.deploy("PoseidonT3", ...)` does **name-based artifact lookup**. Internally it calls `hre.artifacts.readArtifact("PoseidonT3")`, which scans **every compiled artifact in the whole project** — including transitive dependencies from `node_modules` like `poseidon-solidity` and `@zk-kit/lean-imt.sol`. As long as the library ended up being compiled (because some local contract imports it), `hardhat-deploy` finds its ABI + bytecode by name and uses it.

So in v2, "compiled" == "deployable by name". Zero ceremony.

### Hardhat v3 + `rocketh` (the new project)

`env.deploy(...)` does **not** do name-based lookup. It requires you to hand it an artifact object explicitly:

```38:44:/Users/liana/Documents/prog/buidl-guidl/qhhv3/se-2-hardhat-v3/packages/hardhat/deploy/00_deploy_your_contract.ts
  await env.deploy("YourContract", {
    account: deployer,
    artifact: artifacts.YourContract,
    args: [deployer],
  });
```

Those `artifacts.*` entries come from `packages/hardhat/generated/artifacts/index.ts`, which is generated by Hardhat v3's `generateTypedArtifacts` step. And that generator **only emits entries for contracts defined directly in your `contracts/` folder** — not for anything pulled in from `node_modules`.

So `artifacts.HonkVerifier` and `artifacts.Voting` exist (they live in your `contracts/`), but `artifacts.PoseidonT3` and `artifacts.LeanIMT` do not — even though they still get compiled and end up in `artifacts/build-info/*.output.json`.

That's exactly the gap the `loadBuildInfoArtifact` helper fills: it reaches into the raw solc output and hand-builds the artifact object that rocketh's `env.deploy` needs.

## TL;DR

- **v2 / hardhat-deploy:** resolve by **name**, scans all compiled artifacts including npm deps → `deploy("PoseidonT3")` just works.
- **v3 / rocketh:** resolve by **artifact object**, and the auto-generated `artifacts` module only covers your local `contracts/` files → npm library contracts have to be loaded manually from `build-info`.

Same compiler output in both cases. Different deploy API ergonomics on top of it.

</p>
</details> 

So, currently for v3:
  - Inherited from npm (OZ ERC20, Ownable, etc.) → no extra artifact, folded into your contract.
  - Standalone contract/library from npm you must deploy → not in generated/artifacts, load from build-info.

Maybe we can make it better some way to avoid importing from `build-info` by users. Maybe try to use helper function I mentioned above under the hood or at least export it. Or maybe there's some other solutions
  
---

2. Before backmerging this to `create-eth`, we need to be prepared for that case: 
- let's say current `create-eth` version is `1.2.3` and SRE challenges use that version. ->
- we backmerge new hardhat v3 and let's say make `create-eth` version `1.3.0` ->
- we found some small bug in some challenge and to solve it we need to make a fix in `create-eth`. We can't make it in `1.3.1` since it will require migrating all the challenges to hardhat v3 before that fix (challenges still use v2 and will break with v3) ->
- so for that case we will need to add the fix to `1.2.4` instead (and 1.3.1 too).

I think it should already be possible with changeset, right? Or do we need some additional steps for that?

packages/hardhat/deploy/00_deploy_your_contract.ts
```diff
@@ -1,44 +1,40 @@
...
-
-// Tags are useful if you have multiple deploy files and only want to run one of them.
-// e.g. yarn deploy --tags YourContract
-deployYourContract.tags = ["YourContract"];
+    // Read back from the deployed contract
+    const greeting = await env.read(yourContract, { functionName: "greeting" });
+    console.log("👋 Initial greeting:", greeting);
+  },
+  {
+    // Tags are useful if you have multiple deploy files and only want to run one of them.
```
rin-st: some of them*

packages/hardhat/rocketh/config.ts
```diff
@@ -0,0 +1,39 @@
...
+  const deployFn = deployExtension.deploy(env);
+  return (name: any, args: any, options?: any) => {
+    const isReset = process.env.HARDHAT_DEPLOY_RESET === "true";
+    return deployFn(name, args, { ...options, alwaysOverride: isReset || options?.alwaysOverride });
+  };
+}) as typeof deployExtension.deploy;
+
+const extensions = {
+  ...deployExtension,
```
rin-st: nit: looks like we can comment out or remove this line for now, since we have `deploy: deployWithReset,` below

packages/hardhat/deploy/00_deploy_your_contract.ts
```diff
@@ -1,44 +1,40 @@
...
-  console.log("👋 Initial greeting:", await yourContract.greeting());
-};
-
-export default deployYourContract;
-
-// Tags are useful if you have multiple deploy files and only want to run one of them.
-// e.g. yarn deploy --tags YourContract
-deployYourContract.tags = ["YourContract"];
+    // Read back from the deployed contract
+    const greeting = await env.read(yourContract, { functionName: "greeting" });
```
rin-st, line 32: this line breaks deployment for `optimismSepolia` (and I think for others too) when `HARDHAT_DEPLOY_RESET` is `true`. 

Reason:

```
The visible error — AbiDecodingZeroDataError: Cannot decode zero data ("0x") at line 32 — came from an RPC read-after-write race on Alchemy OP Sepolia:

  1. Rocketh broadcasts the deploy tx → Alchemy mines it and returns the contract address.
  2. One line later, the script calls env.read(yourContract, { functionName: "greeting" }), which issues eth_call with blockTag: "latest".
  3. That call hits a different Alchemy replica that hasn't yet caught up to the block containing the new contract. The replica sees no code at that address and returns 0x (empty).
  4. viem's decodeFunctionResult can't decode 0x against the string return type → throws.

  I confirmed the deploy itself was fine: a raw eth_getCode / eth_call via curl against the same Alchemy URL returned the full bytecode and decoded the greeting correctly.
```

We can just comment out that line. Another option is to add a timeout before the `env.read` but timeout is different for different chains and also it causes script to be slower

rin-st (commented):
Awesome job Shiv!

Added some comments, and one more thing here:

I'm getting this warning at the end of deploy message

`WARNING: hre.network.connect() is deprecated and will be removed in a future version. Use hre.network.create() or hre.network.getOrCreate() instead.`

damianmarti (approved):
Great work @technophile-04!!

This is working nicely!

Tested local chain, compile, deploy, tests, debug page, deploy to Sepolia, verify contract, check types, flatten, format, lint, etc.

Everything is working fine.

We need to check AGENTS.MD. I think there are some minor things to update, like the deploy tags code.

Thanks!!

technophile-04:
> https://github.com/scaffold-eth/scaffold-eth-2/pull/1272#issuecomment-4308351383 

For Standalone contract/library from npm I created https://github.com/wighawag/hardhat-deploy/issues/599 let's see what wighawag has to say and if we could get the native support for it. 

<details>

<summary>
    Regarding create-eth version handling: 
</summary>

The only thing is that for the smooth migration, we'll need to move the major version. The actual lever is npm dist-tags plus a long-lived `v2` branch, not a changesets feature. Plan would be: 
1. Ship hardhat v3 as `create-eth@3.0.0` so `npx create-eth@latest` flips cleanly to v3 for new users. 
2. Cut a `v2` branch from the last 2.x commit. Frozen except for bug fixes. Add it to `release.yml`'s trigger list.
3. Add a `changeset:release:legacy` script that wraps `changeset publish --tag legacy`. The workflow picks which script to run based on `github.ref_name` (one ternary, ~6 line diff).
4. Bug fix flow for v2-era challenges: PR to `v2` → add a changeset → merge → CI publishes 2.0.x to the `legacy` tag automatically.

Like we dont' give a tag, then if we release a new version of v2 for example 2.0.16 then people doing `npx create-eth@latest` will get the latest v2 version instead of v3. 

Existing challenges keep working either by pinning (`create-eth@2.0.15`) or by tag (`npx create-eth@legacy`). Both paths are stable.


</details>


> I'm getting this warning at the end of deploy message
> WARNING: hre.network.connect() is deprecated and will be removed in a future version. Use hre.network.create() or hre.network.getOrCreate() instead.

https://github.com/wighawag/hardhat-deploy/pull/594 should be fixed when they release the new version 🙌

alcuadrado:
Just a random comment: I highly highly recommend migrating from `yarn` to `npm`, or better yet, `pnpm`. Yarn used to b the best package manager, but it hasn't kept up with the rest. For starters, it doesn't autoinstall peer dependencies, while every other package manager does. 

Note that `npm` and `pnpm` support monorepos as well.

alcuadrado:
> * Dropped `hardhat-gas-reporter` for now since it hasn't shipped a v3-compatible release yet (every version through 2.3.0 pins `hardhat: ^2.16.0`)

hey, if you run with `--gas-stats` you have that functionality builtin

packages/hardhat/package.json
technophile-04: Yes, exactly right. Quick history:

- HH2 + hardhat-deploy v1: hardhat node had a built-in side effect, it auto-ran your deploy scripts at node startup. `--no-deploy` was a hardhat-deploy-provided CLI flag that opted out of that auto-deploy so `yarn chain` could just give you a clean node without committing deployments 
- HH3 + hardhat-deploy v2 / rocketh: that auto-deploy-on-node-start hook is gone. hardhat node only starts a node. Deploys are now strictly an explicit step (yarn deploy)

packages/hardhat/rocketh/config.ts
technophile-04: Removed the wrapper `fa2bd61` , I had it as a hack for `--reset` option but now no more needed

packages/hardhat/rocketh/config.ts
technophile-04: `fa2bd61` no more required...we had wrapper because of the hardhat-deploy didn't ship `--reset`...but not its live

packages/hardhat/deploy/00_deploy_your_contract.ts
technophile-04: https://github.com/wighawag/hardhat-deploy/issues/601 Created the issue here

packages/hardhat/hardhat.config.ts
technophile-04: moved at [`9a6379f` (this PR)](https://github.com/scaffold-eth/scaffold-eth-2/pull/1272/commits/9a6379f1e36c34db1ce866fd977a92edd8af4d4e)

packages/hardhat/deploy/00_deploy_your_contract.ts
technophile-04: They solved the issue, pushed an update to the deps: 

https://github.com/wighawag/hardhat-deploy/issues/601#issuecomment-4419132669

rin-st:
Thanks for all the changes!

For me, almost everything works great. Even `LeanIMT` seems to be resolving and generating right

Test branch changes, comparing to this https://github.com/scaffold-eth/scaffold-eth-2/compare/hardhat-v3-migration...with-zk?expand=1 . (Main is contracts, deployment file, hardhat config and package.json changes)

`npmFilesToBuild` is a bit different from https://github.com/wighawag/hardhat-deploy/issues/599#issuecomment-4343698155, maybe it caused bugs, but not sure

I had only one deployment error, but next time I run `yarn deploy` it worked

<p>

```
yarn deploy --network optimismSepolia
No contracts to compile
✔ Enter password to decrypt private key:
Environment: optimismSepolia 
         Chain: OP Sepolia 
         Tags: testnet
✔ gas price is currently in this range:
slow: 0.00000000000006025 (priority: 0.00000000000006)
average: 0.000000000000480252 (priority: 0.000000000000480002)
fast: 0.000000000000535252 (priority: 0.000000000000535002)

Do you want to proceed (note that gas price can change for each tx) … yes
- Executing /Users/liana/Documents/prog/buidl-guidl/se-2-hardhat-v3/packages/hardhat/deploy/00_deploy_your_contract.ts
  - Deploying YourContract  with tx:
      0x309988b598f84b591a4973a6b66b64a1bf526004b794bf312f3562e4b06c2b64
      (tx not found)
    => 0x22cf0318fa317329857a9931f1d770cde4c08472
👋 Initial greeting: Building Unstoppable Apps!!!


- Executing /Users/liana/Documents/prog/buidl-guidl/se-2-hardhat-v3/packages/hardhat/deploy/00_deploy_your_voting_contract.ts
  - Deploying HonkVerifier  with tx:
      0x24cce5abbde7bc3f5746dd0b5f44ff77fc1fc650fdae1514b85f0c2abfe662cd
      (tx not found)
    => 0x0930d305116f12cac3a9d4b7d60d13fd3fb8e8ec
  - Deploying PoseidonT3  with tx:
      0x296906b1465aa6b2a6db8804fbfb0a00603de646f1b862a965e5ff1f9d0bfd0a
      (tx not found)
    => 0x955c98ef05a2e280e82942e4242a8d3db2a3f9bd
  - Deploying LeanIMT  with tx:
      0x0efa032caef835c67e08c4d85e2a009d0bbf9d2a64c08b8d41d45aa63b193fae
      (type 0x2, maxFeePerGas: 1000316, maxPriorityFeePerGas: 1000000)
    => 0xc28f1298f7e7b5e76f258c5d1130ef18e3514350
An unexpected error occurred:

ProviderError: replacement transaction underpriced
    at HttpProvider.request (/Users/liana/Documents/prog/buidl-guidl/se-2-hardhat-v3/packages/hardhat/node_modules/hardhat/src/internal/builtin-plugins/network-manager/http-provider.ts:146:21)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async TransactionHashTrackerProvider._request (/Users/liana/Documents/prog/buidl-guidl/se-2-hardhat-v3/packages/hardhat/node_modules/@rocketh/core/src/providers/TransactionHashTracker.ts:16:20)
    at async broadcastTransaction (/Users/liana/Documents/prog/buidl-guidl/se-2-hardhat-v3/packages/hardhat/node_modules/rocketh/src/environment/index.ts:846:20)
    at async Object.broadcastDeployment (/Users/liana/Documents/prog/buidl-guidl/se-2-hardhat-v3/packages/hardhat/node_modules/rocketh/src/environment/index.ts:933:18)
    at async <anonymous> (/Users/liana/Documents/prog/buidl-guidl/se-2-hardhat-v3/packages/hardhat/node_modules/@rocketh/deploy/src/index.ts:455:22)
    at async deployScript.tags (/Users/liana/Documents/prog/buidl-guidl/se-2-hardhat-v3/packages/hardhat/deploy/00_deploy_your_voting_contract.ts:42:5)
    at async Object.executeDeployScriptModules (/Users/liana/Documents/prog/buidl-guidl/se-2-hardhat-v3/packages/hardhat/node_modules/rocketh/src/executor/index.ts:514:15)
    at async runScriptWithHardhat (/Users/liana/Documents/prog/buidl-guidl/se-2-hardhat-v3/packages/hardhat/node_modules/hardhat-deploy/src/tasks/deploy.ts:55:2)
    at async next (/Users/liana/Documents/prog/buidl-guidl/se-2-hardhat-v3/packages/hardhat/node_modules/hardhat/src/internal/core/tasks/resolved-task.ts:178:16) {
  code: -32000,
  data: undefined
}

If you think this is a bug in Hardhat, please report it here: https://hardhat.org/report-bug
```

</p>
</details>  

Verification worked well for all the contracts on the first run.

Great job!

rin-st (approved):

alcuadrado:
> For me, almost everything works great. Even `LeanIMT` seems to be resolving and generating right

Just to confirm: there was a bug on our side, which we fixed and released the fixed version.

> `ProviderError: replacement transaction underpriced`

While this other one seems to be a hardhat-deploy issue

technophile-04:
> I had only one deployment error, but next time I run yarn deploy it worked

Probably it might be the some rpc node issue, I ran it twice with `--reset` and it worked always

packages/hardhat/.gitignore
```diff
@@ -8,9 +8,9 @@ node_modules
 coverage
 coverage.json
 
-# typechain
-typechain
-typechain-types
+# generated contract types
+types
+generated
```
carletex, line 13: Why are we committing now the `typechain-types` folder?

technophile-04: > Why are we committing now the typechain-types folder?  

Ohhh I forget to remove that dir after removing it from gitignore [`8f0a9ad` (this PR)](https://github.com/scaffold-eth/scaffold-eth-2/pull/1272/commits/8f0a9ad922c032eb2b85cfa856edd9fa45ca4206)

`typechain-types` dir is no longer generated, and instead new dir is generated `/types`

Pabl0cks (approved):
Awesome job @technophile-04 ! Tested on Windows and everything worked great. 

My agents found a couple of nitpicks, not introduced in this PR, and probably discussed in the past, we can probably ignore or just document somewhere:

1. `yarn deploy --tags X --reset` resets the whole deployment folder, then only runs the tagged scripts. So after `yarn deploy --tags CodexPrProbe --reset`, the frontend contract map only contained the tagged probe contracts until I ran a full `yarn deploy` again.

  This looks like expected hardhat-deploy v2 behavior, not a migration blocker, but it is easy to misread `--reset --tags X` as "reset only X". I would document it somewhere visible:
  
  > `--reset` wipes the deployment folder before scripts run. If combined with `--tags`, only the tagged deployments will be restored. Run a full `yarn deploy` afterward to restore all local deployments.

2. `generateTsAbis` loses inherited-function labels for multi-contract files because it matches Solidity filename to contract name. This is pre-existing and only affects Debug Contracts inheritance tooltips.

packages/hardhat/deploy/00_deploy_your_contract.ts
carletex, line 1: q: Is there a way to bypass the gas confirmation? If so, we probably want to do it by default (to mantain old behav)?

Image

carletex, line 1: When re-deploying, you get:

Image

- No info on contract being reused
- No info of address of previous contract

Which were nice to have

Image

technophile-04:
> yarn deploy --tags X --reset resets the whole deployment folder, then only runs the tagged scripts. So after yarn deploy --tags CodexPrProbe --reset, the frontend contract map only contained the tagged probe contracts until I ran a full yarn deploy again. 

yeah, that was actually hard to handle for hardhat-deploy (in v1 as well the behaviour is same) and was one of the reason why they removed the `--reset` flag completed in hardhat-deploy v2. But we asked them to add the feature back [here](https://github.com/wighawag/hardhat-deploy/issues/592) and they for now keep the working same v1 for simplicity (wiping everthing even though --tags was passed) 

> generateTsAbis loses inherited-function labels for multi-contract files because it matches Solidity filename to contract name. This is pre-existing and only affects Debug Contracts inheritance tooltips. 

fixed at [`745688a` (this PR)](https://github.com/scaffold-eth/scaffold-eth-2/pull/1272/commits/745688a88ed1a83173569e3963fc70081f44767a)

packages/hardhat/deploy/00_deploy_your_contract.ts
technophile-04: https://github.com/wighawag/rocketh/pull/38 created this

technophile-04: fixed at [`71c7ce5` (this PR)](https://github.com/scaffold-eth/scaffold-eth-2/pull/1272/commits/71c7ce558d50b7bc83d28ce89399dcd3c3f8caea) it no more asks you for confirmation

rin-st (approved):
Working great for me!

Approving that prompt and multi-contract issues fixed

packages/hardhat/deploy/00_deploy_your_contract.ts
technophile-04: Merging this, will update the version as soon as it's fixed on hardhat-deploy side 🙌

PR #1275, opened 2026-04-26
Drop local getBlockExplorerAddressLink and per-call blockExplorerAddressLink props

rin-st:
Draft for now but ready for review. Follows https://github.com/scaffold-eth/scaffold-ui/pull/101

Before merge:

- Bump scaffold-ui packages and remove relative paths.
- Undo changes from `next.config.ts`.

## Before

- `packages/nextjs/utils/scaffold-eth/networks.ts` defined a local `getBlockExplorerAddressLink(network, address)` with hardhat → `/blockexplorer/...`, otherwise chain explorer, fallback etherscan.
- 8 `<Address>` call sites passed:
  ```tsx
  blockExplorerAddressLink={
    targetNetwork.id === hardhat.id ? `/blockexplorer/address/${addr}` : undefined
  }
  ```

## After

- Removed the local `getBlockExplorerAddressLink`. `RainbowKitCustomConnectButton` imports it from `@scaffold-ui/hooks` instead.
- All `<Address>` props simplified — drop `blockExplorerAddressLink` entirely; pass `chain={targetNetwork}` so resolution uses the user-selected network instead of `configuredChains[0]`.
- Behavior change: when a chain has no explorer (and isn't `31337`), `<Address>` renders the address as plain text rather than linking to etherscan. This matches the upstream fix and matches user expectations.

## How to test

1. `yarn install` (picks up new `@scaffold-ui/hooks` version).
2. `yarn chain` + `yarn deploy` + `yarn start`.
3. **Home page:** connect a wallet → click the address → `/blockexplorer/address/{addr}` opens in the same tab.
4. **Block Explorer page** (`/blockexplorer`): each address in the transactions table → same-tab navigation to `/blockexplorer/address/{addr}`.
5. **Transaction details page:** `From` / `To` / `Contract Creation` addresses → same-tab `/blockexplorer/...`.
6. **Faucet modal** (only on hardhat): faucet address → same-tab `/blockexplorer/...`.
7. **Connect button dropdown:** "View on Block Explorer" → `/blockexplorer/...`.
8. **QR modal:** address renders without a link (unchanged — uses `disableAddressLink`).
9. **Switch to sepolia** in `scaffold.config.ts` and reconnect:
   - Home + dropdown addresses link to `https://sepolia.etherscan.io/address/{addr}` in a new tab.
10. **Debug Contracts** page (`/debug`): contract header address and addresses inside read methods all link correctly per address (verifies the upstream `<Contract>` fix end-to-end).


packages/nextjs/app/blockexplorer/_components/TransactionsTable.tsx
```diff
@@ -45,39 +44,14 @@ export const TransactionsTable = ({ blocks, transactionReceipts }: TransactionsT
...
                     <td className="w-2/12 md:py-4">
-                      <Address
-                        address={tx.from}
-                        size="sm"
-                        onlyEnsOrAddress
-                        blockExplorerAddressLink={
-                          targetNetwork.id === hardhat.id ? `/blockexplorer/address/${tx.from}` : undefined
-                        }
-                      />
+                      <Address address={tx.from} size="sm" onlyEnsOrAddress chain={targetNetwork} />
```
damianmarti, line 47: @rin-st are the chain params really needed?

rin-st: we have [this line](https://github.com/scaffold-eth/scaffold-ui/blob/e2861ab8647d55dc22ae0ea23fc8af299b45157e/packages/components/src/Address/Address.tsx#L62) in `Address` component, so if we have multiple networks in our app, without passing chain it will always resolve to `configuredChains[0]`.

damianmarti (approved):
Working great and code looks good! Thanks!

technophile-04 (approved):
This is great! Thanks @rin-st!

PR #1280, opened 2026-04-29
Deps: next 15 → 16 and minor frontend updates

rin-st:
Part 1 of #1279

### Summary
Next.js 15.2.8 → 16.2.4 + related frontend dep bumps. Hardhat package untouched.

### Dep bumps (`packages/nextjs`)
- `next` 15.2.8 → 16.2.4
- `eslint-config-next` 15.2.3 → 16.2.4
- `eslint` stays on 9.x (10 not yet supported by `eslint-plugin-react`'s peer)
- `@tailwindcss/postcss` `latest` → `^4.2.4` (pinned), `tailwindcss` 4.1.3 → 4.2.4
- `daisyui` 5.0.9 → 5.5.19
- `@heroicons/react` 2.1.5 → 2.2.0
- `kubo-rpc-client` 5 → 6, `bgipfs` 0.0.12 → 0.0.18
- `next-themes` 0.3 → 0.4, `next-nprogress-bar` 2.3.13 → 2.4.7
- `react`/`react-dom` 19.2.3 → 19.2.5, `@types/react` patched
- `@trivago/prettier-plugin-sort-imports` 4 → 6, `prettier` 3.5.3 → 3.8.3
- `type-fest` 4 → 5, `@scaffold-ui/*` minor bumps, `vercel` ~51 → ~52, `zustand`/`qrcode.react`/`blo`/`react-hot-toast`/`usehooks-ts` patches

### `next.config.ts`
- Removed `eslint` block — option dropped in Next 16; lint is no longer part of `next build`.
- Removed `webpack` config (fs/net/tls fallbacks + `pino-pretty`/`lokijs`/`encoding` externals). Turbopack handles these natively in Next 16; `pino-pretty` is auto-externalized by Next.
- Added empty `turbopack: {}` to silence the "webpack config without turbopack config" error and make Turbopack-by-default explicit.

### ESLint
- Migrated `eslint.config.mjs` off `@eslint/eslintrc`/`FlatCompat` to native flat configs from `eslint-config-next` 16 (`/core-web-vitals`, `/typescript`).
- Disabled new `react-hooks/set-state-in-effect` (from `eslint-plugin-react-hooks` 7) — flags SE-2's standard `setMounted(true)` hydration guards and conditional sync effects; refactoring would be a larger separate change.

### Scripts (`packages/nextjs`)
- `lint`: `next lint` → `yarn eslint .` (`next lint` deprecated in 16; flat config takes over).
[...]

Issue #129, opened 2023-02-03
Show error when sending a TX to a non-existing contract


carletex:
On `useScaffoldContractWrite.ts`: "If the contract is not deployed this "works". I get a "mined successfully" msg"

```ts
    if (wagmiContractWrite.writeAsync && writeTx) {
      try {
        // If the contract is not deployed this "works". I get a "mined successfully" msg.
        await writeTx(wagmiContractWrite.writeAsync());
      } catch (e: any) {
        const message = getParsedEthersError(e);
        toast.error(message);
      }
    } else {
```

I think we need to handle that case in the transactor `useTransactor.tsx`.

PR #1282, opened 2026-05-06
UI tweaks/redesign

carletex:
Opening this as a discussion-starter for a sharper, less-soft visual direction for SE-2.

In this first iteration, I included:

- Removed all the radius and shadows.
- full height menu item (instead of "pill" item)
- /test page so we can see and test different palettes. (we should test way more palettes!). When you select a palette, it's stored in local storage, so you can refresh + navigate to the home, etc.

image

Once we agree on some of this basic stuff, we could ***merge and iterate later.

*** we need to adapt scaffold-ui too

But there is something about "semantics" that we could also fix here (this is something that has been bugging me for a long time). Some examples:

1. **`secondary` has two jobs.** Used both as a button color (`btn-secondary`, correct per DaisyUI) and as a muted surface background (`bg-secondary` in Header active link, `/debug` banner, contract-picker hover, theme toggle base... incorrect per DaisyUI, which intends `base-200`/`base-300` for muted surfaces). The current default palette gets away with it because `bg-secondary` is `#dae8ff` (light) and dark text remains readable. Any palette where secondary isn't a near-white surface immediately breaks readability. Patched the immediate bleed with `text-secondary-content` pairings; the cleaner fix is to migrate surface usages to `bg-base-300`.

2. **`btn-primary` vs `btn-secondary` swapped.** Header CTAs (Connect Wallet, Faucet) are `btn-secondary`; Footer dev tools (price chip, Faucet modal trigger, Block Explorer link) are `btn-primary`. Backwards... primary should be the user-facing CTA, not the dev-tool decoration.

3. **`text-neutral` used as muted-text.** DaisyUI defines `neutral` as a non-saturated dark/light, not a "muted text" role. 

Would love to hear some first impressions!

PS. Be open-minded! We are so used to the SE-2 style that we need to be open-minded to accept a new style (it happened to me haha)


packages/nextjs/app/test/page.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+      <Section title="Backgrounds — base shades">
+        <div className="grid grid-cols-3 gap-3">
+          <div className="bg-base-100 border border-base-content/10 p-6 text-center">bg-base-100</div>
+          <div className="bg-base-200 border border-base-content/10 p-6 text-center">bg-base-200</div>
+          <div className="bg-base-300 border border-base-content/10 p-6 text-center">bg-base-300</div>
+        </div>
+      </Section>
+
+      <Section title="Text colors">
```
rin-st: I think we're not interested in this block, maybe except success/warning/error, since other colors not used for texts. And even for success/warning/error in buttons/badges text color is `--error-warning-content`

Image

packages/nextjs/styles/globals.css
```diff
@@ -33,8 +33,8 @@
   --color-warning: #ffcf72;
   --color-error: #ff8863;
 
-  --radius-field: 1rem;
-  --radius-box: 1rem;
+  --radius-field: 0rem;
+  --radius-box: 0rem;
```
rin-st, line 37: here and for dark theme we can also define `--rounded-selector: 0rem` small sized component won't have radius too https://daisyui.com/docs/utilities/#border-radius

Image

packages/nextjs/styles/globals.css
```diff
@@ -58,8 +58,8 @@
   --color-warning: #ffcf72;
   --color-error: #ff8863;
```
rin-st, line 61: not sure if we need to add --color-(success/warning/error)-content colors

rin-st (commented):
Love how it looks overall, predefined color schemes and how you show all the examples!

1. Colors
I think we need to use latest daisyui version so it will be up to date with current docs. There's an update in https://github.com/scaffold-eth/scaffold-eth-2/pull/1280, but we can also make it here

> But there is something about "semantics" that we could also fix here (this is something that has been bugging me for a long time). Some examples:
> 1. ... the cleaner fix is to migrate surface usages to bg-base-300.
> 2. Backwards... primary should be the user-facing CTA, not the dev-tool decoration.
> 3. text-neutral used as muted-text. DaisyUI defines neutral as a non-saturated dark/light, not a "muted text" role.

Completely agree, let's fix all of these! Let's figure out our inconsistencies with the color docs https://daisyui.com/docs/colors/#list-of-all-daisyui-color-names and refactor colors to match it. For example, accent and info different from primary (or don't use them at all), maybe even use base-100 as default bg color to have better elevations. I didn't find many bugs, though, just one `text-accent`, and I'll add comments in the code for others

2. Themes
I like our current theme, we can leave it with maybe some minor changes. And I like Ethereum light theme, but dark bg is too dark, and dark base shades are mixed up, from darker to lighter base-200 -> base-100 -> base 300

3. Radius/shadows/height/other styles. 
I like how it looks. But first thing I'm thinking of when I see this is what we need to change so it will look good everywhere. scaffold-ui -> se-2/create-eth -> SRE challenges custom non-scaffold-ui styles to match new ones (better with new images) -> other extensions custom styles if needed. Scaffold-ui bump with those changes should be major to not break the scaffold-ui styles of the apps where it's already used (if any). Assuming we're agree to change all of it

> Removed all the radius and shadows.

There are still shadows in the code

---

And since we're continuing to work with daisy, let's close https://github.com/scaffold-eth/scaffold-eth-2/issues/1079?

packages/nextjs/styles/globals.css
carletex: Added `radius-selector` (not rounded-selector! haha)  ca44ad83849a

packages/nextjs/styles/globals.css
carletex: Seems like DaisyUI calculates the foreground / background color like this:

image

And it seems that it has a dynamic calc for these values too:

image
image


Which is usually legible. So I don't think we need to go that far (adding -color-(success/warning/error)-content colors)

packages/nextjs/app/test/page.tsx
carletex: Yeah, maybe I over-added some stuff. But not a big deal since this will be removed before merge!

carletex:
Thanks Rinat. Pushed some tweaks.

> Let's figure out our inconsistencies with the color docs
> maybe even use base-100 as default bg color to have better elevations

Yeah, there are some inconsistencies in the current palette that we should fix!

Maybe I good iteration would be to use the same current palette (+ removing border-radious and shadows + fixing inconsistencies). I guess it'll be easier to fix the surrounding stuff (s-ui, website, etc)

> I think we need to use latest daisyui version 

Updated here just in case too.

> There are still shadows in the code

I left the ones for the dropdown and notifications (which I think make sense). Anything else?

rin-st:
> Maybe I good iteration would be to use the same current palette (+ removing border-radious and shadows + fixing inconsistencies). I guess it'll be easier to fix the surrounding stuff (s-ui, website, etc)

Agree!

> I left the ones for the dropdown and notifications (which I think make sense). Anything else?

Oh, I thought you wanted to remove all the shadows. Makes sense!

> Yeah, there are some inconsistencies in the current palette that we should fix!

What do you think, should we fix those inconsistencies in this branch? I think in that case it will be a bit easier to update s-ui/sre - one time instead of 2 or more. But in the case of N iterations, it will be easier to track and fix bugs

technophile-04:
> But there is something about "semantics" that we could also fix here (this is something that has been bugging me for a long time). 
> ~ Carlos

SAME! Especially with AI, when you ask it to update the theme, some things get a bit out of contrast. Thanks for tackling this, really like the changes! 

>  same current palette (+ removing border-radious and shadows + fixing inconsistencies). I guess it'll be easier to fix the surrounding stuff (s-ui, website, etc)
> ~ Carlos

Agree, while switching to other themes more small inconsistencies in contrast appeared

> I think in that case it will be a bit easier to update s-ui/sre - one time instead of 2 or more 
> ~ Riant

++ would be really nice if we fix the major ones before updating the s-ui, but testing this PR not too deeply seems like we've covered most of the cases ...one which I found was: 

Screenshot 2026-05-14 at 9 49 10 PM

We've `btn-circle` for `Faucet`, `RevealPrivateKeyModal`, etc., we need to remove that

damianmarti:
@carletex This looks great! 

I like sunset palette :-)

Makes sense to be able to select the color palette at the scaffold.config file? So the devs can choose the colors easily...

There are still some rounded boxes, like the notifications and the components on the debug page.

I agree that the classes should be fixed, using better semantics for them, using primary for the real primary components.

Thanks!!

carletex:
cc @rin-st would it make sense to create a beta release or similar for scaffold-ui so we can install it here and see how they would look like with the new non-round/shadow stuff? Or is it a better way?

rin-st:
Yep, but first I'm trying to make the palette consistent with daisy ui colors docs and fixing our small issues we still have (I'll create PR to this branch for those changes). Next step - non-rounded scaffold-ui version.

Btw added small changes here [0af3af3](https://github.com/scaffold-eth/scaffold-eth-2/pull/1282/commits/0af3af38eaed3f7f6ac8917cb4b3af110aafcce4), fixed some rouned modal close buttons, faucet button, notification corners

rin-st:
> Yep, but first I'm trying to make the palette consistent with daisy ui colors docs and fixing our small issues we still have (I'll create PR to this branch for those changes).

#1296 

---

> cc @rin-st would it make sense to create a beta release or similar for scaffold-ui so we can install it here and see how they would look like with the new non-round/shadow stuff? Or is it a better way?

Scaffold-UI PR: https://github.com/scaffold-eth/scaffold-ui/pull/110

To test:

- clone scaffold-ui to the same folder as se-2
- [Next steps](https://github.com/scaffold-eth/scaffold-ui/blob/77b28e0186346783a59ad8eed45a1aa5c8f81ff9/README.md?plain=1#L17) (but don't change nextjs config/webpack stuff): 

image

carletex:
This is looking pretty good to me!

Saw some comments on https://github.com/scaffold-eth/scaffold-ui/pull/110. After tackling those... I think this is ready to merge.

What do you guys think? cc @rin-st @technophile-04 

PS: Maybe we need to "un-round" the logo too?
image

carletex:
PS: Not sure if we should include https://github.com/scaffold-eth/scaffold-eth-2/pull/129 on here... or can be done later. On a first look, I see some changes on the UI colors..

image
image

.. that we might need to discuss.

rin-st:
> Saw some comments on [scaffold-eth/scaffold-ui#110](https://github.com/scaffold-eth/scaffold-ui/pull/110). After tackling those... I think this is ready to merge.

Everything is resolved

> PS: Maybe we need to "un-round" the logo too? image

---

So I think PR is ready to merge. Steps

- Merge [scaffold-eth/scaffold-ui#110](https://github.com/scaffold-eth/scaffold-ui/pull/110), release new versions
- Update scaffold-ui deps in se-2. 
- Remove `/test` page and components
- Quick test if everything works and merge

carletex:
Thanks Rinat!!

Pinging @technophile-04 for a final review and to coordinate the release.

technophile-04 (approved):
Everthing is looking great! Really goood work Carlos and Rinat 🫶

PR #1291, opened 2026-05-27
fix(nextjs): stop excluding contracts/ from format script

technophile-04:
Users paste ABIs into `externalContracts.ts` and expect prettier to format on save / `yarn format`. The `contracts/` exclusion in the format script was meant to skip auto-generated `deployedContracts.ts`, but it also caught `externalContracts.ts` which is user-pasted code that should be formatted.

This change removes `contracts` from the format-ignore pattern so `yarn format` formats `externalContracts.ts` as users expect.

This is also helpful for creat-eth, because if not https://github.com/scaffold-eth/create-eth/actions/runs/26504441866/job/78053025622?pr=430


rin-st (approved):
Lgtm, thanks!

Issue #1293, opened 2026-05-27
Bug: FaucetButton and Notifications are buggy in blockexplorer page


rin-st:
Steps to reproduce:

1)
- open `/blockexplorer` page
- click to faucet button (top right corner). You'll see two notifications first time

<img width="1402" height="252" alt="Image" src="https://github.com/user-attachments/assets/c53210d3-946a-47f1-8952-acc599c9a92c" />

---

2)
- click other pages, for example `/debug` and then return to blockexplorer
- click to faucet button, sometimes it doesnt work at all

damianmarti:
There is something wrong with the pagination using Hardhat v3, when empty blocks are mined, empty pages are added to the block explorer. 

I can take a look at this!

rin-st:
Yes, take a look please! I noticed it too, but then couldn't reproduce it

damianmarti:
Fixed https://github.com/scaffold-eth/scaffold-eth-2/pull/1295

PR #1295, opened 2026-05-27
Fix blockexplorer pagination when auto mining is set

damianmarti:
When the local blockchain is set to auto-mine blocks, the block explorer pagination breaks because a lot of empty blocks are created, and the pagination is based on blocks, not on transactions. 

The pagination is always done based on transactions now. It works well with auto mining set to off and on.

The faucet works ok. The error was related to the pagination issue, not the faucet itself.

_Closes #1293 _

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useFetchBlocks.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useFetchBlocks.ts
@@ -12,7 +12,8 @@ import {
 import { hardhat } from "viem/chains";
 import { decodeTransactionData } from "~~/utils/scaffold-eth";
 
-const BLOCKS_PER_PAGE = 20;
+const TRANSACTIONS_PER_PAGE = 20;
+const BLOCK_BATCH_SIZE = 50;
 
 export const testClient = createTestClient({
   chain: hardhat,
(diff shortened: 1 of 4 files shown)


packages/nextjs/hooks/scaffold-eth/useFetchBlocks.ts
```diff
@@ -84,36 +134,56 @@ export const useFetchBlocks = () => {
   }, [fetchBlocks]);
 
   useEffect(() => {
-    const handleNewBlock = async (newBlock: any) => {
+    const handleNewBlock = async (newBlock: Block) => {
       try {
-        if (currentPage === 0) {
-          if (newBlock.transactions.length > 0) {
+        setTotalTransactions(prevTotal => prevTotal + newBlock.transactions.length);
```
rin-st: is that possible that some transactions have already been counted here? Like in line 165 `setBlocks` you check for intersection of prevBlocks and latest block, do we need it here too?

packages/nextjs/hooks/scaffold-eth/useFetchBlocks.ts
```diff
@@ -84,36 +134,56 @@ export const useFetchBlocks = () => {
...
+            const latestBlockNumber = blockWithTxDetails.number!;
+            const existingBlockIndex = prevBlocks.findIndex(block => block.number === latestBlockNumber);
+
+            const nextBlocks =
+              existingBlockIndex >= 0
+                ? prevBlocks.map((block, index) => (index === existingBlockIndex ? blockWithTxDetails : block))
+                : [blockWithTxDetails, ...prevBlocks];
+
+            const trimmedBlocks = [...nextBlocks];
+            while (
```
rin-st: Can we avoid `reduce` in `while` condition?

rin-st (commented):
Thank you Damu, working great and solves both issues!

Added two questions

packages/nextjs/hooks/scaffold-eth/useFetchBlocks.ts
damianmarti: Nice catch! Fixed!

packages/nextjs/hooks/scaffold-eth/useFetchBlocks.ts
damianmarti: yes, changed!

rin-st (approved):
Lgtm, thank you!

PR #1300, opened 2026-06-11
Fix duplicated logs in useContractLogs

damianmarti:
The block explorer's address logs tab (useContractLogs) showed duplicated log entries. Two overlapping fetch ranges caused this:

  1. The initial fetch retrieved logs from block 0 to "latest".
  2. On every new block, the watchBlockNumber callback fetched from prevBlockNumber to "latest" — a range that overlapped both the initial fetch and previous callback fetches, so boundary blocks were counted twice (and more on fast-polling chains).

  In addition, under React Strict Mode the effect's double-invocation could register a duplicate watcher, since the watcher
  was set up with no guard against the effect having already been cleaned up.

### How to test

  1. Open the block explorer, navigate to the deployed contract's address, and open the Logs tab.
  2. Trigger transactions that emit events (e.g. setGreeting from the Debug Contracts page) and verify each event appears
  exactly once, including with React Strict Mode (dev mode) and while new blocks are being mined.
  3. Navigate between different addresses and verify logs from the previous address never appear in the new list.

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useContractLogs.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useContractLogs.ts
@@ -9,31 +9,67 @@ export const useContractLogs = (address: Address) => {
   const client = usePublicClient({ chainId: targetNetwork.id });
 
   useEffect(() => {
+    setLogs([]);
+    let unwatch: (() => void) | undefined;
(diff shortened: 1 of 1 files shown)


rin-st (approved):
Lgtm! Thanks!

PR #1301, opened 2026-06-12
Cache block explorer transaction total to avoid full-chain rescans on pagination

damianmarti:
Fixes a performance bug where the block explorer walked from latestBlock down to block 0 on every page change and refetch, issuing a getBlock({ includeTransactions: true }) for every block just to compute totalTransactions — even after the current page's 20 transactions were already collected.

With auto-mining (~1 block/s), a day-old local chain is ~86k blocks. Each pagination click previously triggered ~86k RPC calls. On page 0, the scan continued all the way to genesis even after 20 transactions were found, solely to count the total.

This is related to the fix applied at https://github.com/scaffold-eth/scaffold-eth-2/pull/1295

It adds some complexity to the code. We have to take into account whether fixing this is worth the added complexity, since this is only an issue on the local chain.

### DIFF
--- a/packages/nextjs/app/blockexplorer/_components/ContractTabs.tsx
+++ b/packages/nextjs/app/blockexplorer/_components/ContractTabs.tsx
@@ -26,7 +26,7 @@ const publicClient = createPublicClient({
 });
 
 export const ContractTabs = ({ address, contractData }: PageProps) => {
-  const { blocks, transactionReceipts, currentPage, totalTransactions, setCurrentPage } = useFetchBlocks();
+  const { blocks, transactionReceipts, currentPage, hasNextPage, setCurrentPage } = useFetchBlocks(address);
   const [activeTab, setActiveTab] = useState("transactions");
   const [isContract, setIsContract] = useState(false);
 
@@ -38,15 +38,6 @@ export const ContractTabs = ({ address, contractData }: PageProps) => {
 
     checkIsContract();
   }, [address]);
-
-  const filteredBlocks = blocks.filter(block =>
-    block.transactions.some(tx => {
-      if (typeof tx === "string") {
-        return false;
-      }
-      return tx.from.toLowerCase() === address.toLowerCase() || tx.to?.toLowerCase() === address.toLowerCase();
-    }),
-  );
 
   return (
     <>
@@ -84,8 +75,8 @@ export const ContractTabs = ({ address, contractData }: PageProps) => {
       )}
       {activeTab === "transactions" && (
         <div className="pt-4">
-          <TransactionsTable blocks={filteredBlocks} transactionReceipts={transactionReceipts} />
-          <PaginationButton currentPage={currentPage} totalItems={totalTransactions} setCurrentPage={setCurrentPage} />
+          <TransactionsTable blocks={blocks} transactionReceipts={transactionReceipts} />
+          <PaginationButton currentPage={currentPage} hasNextPage={hasNextPage} setCurrentPage={setCurrentPage} />
         </div>
       )}
       {activeTab === "code" && contractData && (
--- a/packages/nextjs/app/blockexplorer/_components/PaginationButton.tsx
+++ b/packages/nextjs/app/blockexplorer/_components/PaginationButton.tsx
@@ -2,15 +2,13 @@ import { ArrowLeftIcon, ArrowRightIcon } from "@heroicons/react/24/outline";
 
 type PaginationButtonProps = {
   currentPage: number;
-  totalItems: number;
+  hasNextPage: boolean;
   setCurrentPage: (page: number) => void;
 };
 
-const ITEMS_PER_PAGE = 20;
-
-export const PaginationButton = ({ currentPage, totalItems, setCurrentPage }: PaginationButtonProps) => {
+export const PaginationButton = ({ currentPage, hasNextPage, setCurrentPage }: PaginationButtonProps) => {
   const isPrevButtonDisabled = currentPage === 0;
-  const isNextButtonDisabled = currentPage + 1 >= Math.ceil(totalItems / ITEMS_PER_PAGE);
+  const isNextButtonDisabled = !hasNextPage;
 
   const prevButtonClass = isPrevButtonDisabled ? "btn-disabled cursor-default" : "btn-primary";
   const nextButtonClass = isNextButtonDisabled ? "btn-disabled cursor-default" : "btn-primary";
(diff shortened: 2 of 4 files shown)


rin-st:
Thanks Damu! I refactored the code a bit, see below

**TL;DR**

Reworked this to *not* cache the transaction total at all. Instead of maintaining a running count — which needs a full-chain scan to seed and can drift on chain restart / HMR — the explorer now fetches **one extra transaction** per page and uses its presence as the "is there a next page?" signal. Result: no full scan ever (not even on first load), no module-level cache to go stale, and less code. Also fixes a pre-existing bug where the address-filtered view showed partial pages.

**Details**

The cached-total approach works, but it carries the complexity cost called out in the original description plus a few sharp edges: the cache is seeded by one full-chain scan and kept in sync by incremental live updates with no reconciliation, so it can drift; and since it's a module-level singleton it survives a `yarn chain` restart / HMR and serves a stale (or doubly-counted) total until a hard page reload.

Key observation: the total is only ever used to enable/disable the **Next** button — it's never displayed. So we don't need a count at all. `fetchPageItems` now collects `TRANSACTIONS_PER_PAGE + 1` items; if the extra one comes back, a next page exists. That removes the cache, the scan, and all the state that could go stale. Cold load of page 0 is now ~21 txs from the top instead of a whole-chain walk.

This also let me fix a separate, **pre-existing** issue (present on `main` too): the address page (`/blockexplorer/address/...`) paginated *all* transactions globally and filtered the current page client-side, so it showed only the matching subset of the global 20 (e.g. 10 rows) with a Next button driven by the global count. (To reproduce the bug click to the addresses in blockexplorer after running the script below, for example `0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC`, you'll see less than 20 txs on the page)

<p>

image

</p>
</details> 

`fetchPageItems` is now address-aware — it skips non-matching transactions while scanning — so the address view paginates the address's own transactions and shows full pages. The redundant client-side filter in `ContractTabs` is removed.

Tradeoff: for an address with few transactions, filling a page scans deeper (down to genesis for a very sparse address on a large chain). Active addresses stay fast. There's no cheap middle ground without a real tx index, which is out of scope for a local explorer.

**How to test**

Spin up the stack:

```bash
yarn chain      # terminal 1
yarn deploy     # terminal 2
yarn start      # terminal 3
```

Seed a mix of transactions — the helper script sends txs involving a target account interleaved with noise between other accounts, which is exactly the case that exposed the partial-page bug:

Script to add to `scripts/sendTestTxs.ts`

<p>

```
// TEMP test helper: seed the local chain with transactions so the block explorer —
// and especially its address-filtered view — has multiple pages to page through.
//
// Sends a mix: transactions involving a TARGET account interleaved with "noise"
// transactions between other accounts. Filtering the explorer by the target should
// then show full pages of 20 despite the noise — the exact case of the address bug.
//
// Local chain only — uses the well-known deterministic Hardhat dev keys (public,
// funded only on the local chain). Remove before merge.
//
// Usage (from packages/hardhat):
//   yarn tsx scripts/sendTestTxs.ts [total] [targetIndex]
//     total        total transactions to send (default 60)
//     targetIndex  which dev account is the filter target (default 1 → 0x7099…79C8)
import { createPublicClient, createWalletClient, http, parseEther } from "viem";
import { privateKeyToAccount } from "viem/accounts";
import { hardhat } from "viem/chains";

// Hardhat/Anvil deterministic accounts #0–#3 ("test test … junk" mnemonic).
const DEV_KEYS = [
  "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
  "0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d",
  "0x5de4111afa1a4b94908f83103eb1f1706367c2e68ca870fc3fb9a804cdab365a",
  "0x7c852118294e51e653712a81e05800f419141751be58f605c371e15141b007a6",
] as const;

const total = Number(process.argv[2] ?? 60);
const targetIndex = Number(process.argv[3] ?? 1);

const transport = http("http://127.0.0.1:8545");
const publicClient = createPublicClient({ chain: hardhat, transport });
const wallets = DEV_KEYS.map(pk => createWalletClient({ account: privateKeyToAccount(pk), chain: hardhat, transport }));
const addressOf = (i: number) => wallets[i].account.address;

const send = async (fromIdx: number, toIdx: number) => {
  const hash = await wallets[fromIdx].sendTransaction({
    type: "eip1559",
    to: addressOf(toIdx),
    value: parseEther("0.0001"),
  });
  await publicClient.waitForTransactionReceipt({ hash });
};

async function main() {
  // Two accounts that are NOT the target carry the noise traffic.
  const noiseFrom = DEV_KEYS.findIndex((_, i) => i !== targetIndex);
  const noiseTo = DEV_KEYS.findIndex((_, i) => i !== targetIndex && i !== noiseFrom);

  console.log(`Target ${addressOf(targetIndex)} — sending ${total} txs (target + noise interleaved)...`);

  let targetTxs = 0;
  for (let i = 0; i < total; i++) {
    if (i % 2 === 0) {
      // Target transaction, alternating direction so the target appears as both from and to.
      if (i % 4 === 0) await send(targetIndex, noiseFrom);
      else await send(noiseFrom, targetIndex);
      targetTxs++;
    } else {
      // Noise: neither side is the target.
      await send(noiseFrom, noiseTo);
    }
    if ((i + 1) % 10 === 0) console.log(`  ${i + 1}/${total}`);
  }

  console.log(`Done. Latest block: ${await publicClient.getBlockNumber()}`);
  console.log(`${targetTxs} txs involve the target → ~${Math.ceil(targetTxs / 20)} page(s) when filtered.`);
  console.log(`Open: http://localhost:3000/blockexplorer/address/${addressOf(targetIndex)}`);
}

main().catch(err => {
  console.error(err);
  process.exit(1);
});

```

</p>
</details> 

```bash
cd packages/hardhat
yarn tsx scripts/sendTestTxs.ts          # 60 txs, ~30 involve account #1
# or: yarn tsx scripts/sendTestTxs.ts 100 1   # [total] [targetIndex]
```

Verify:
- **Global explorer** `/blockexplorer` — pages through all txs 20 at a time; Next disables on the last page.
- **Address view** (the script prints the URL, e.g. `/blockexplorer/address/0x7099…79C8`) — shows a full 20 of *that address's* txs per page with Next enabled while more exist, instead of a partial page.
- **Deep page / last page** — Next correctly disables when no further txs exist (the `+1` probe avoids an empty trailing page).

damianmarti:
Thanks @rin-st for the refactor and the script! Nice improvement!

Working great!!

technophile-04 (approved):
Thanks Rinat really nice approach 🫶, and thanks damu for bringing this up and first iteration! Merging this, everything loooks great!

PR #1304, opened 2026-06-24
New ui tweaks

rin-st:
Working on migrating challenges to new ui and found small issues
1. If top menu buttons in the second half of the screen, you can't click in the middle of the button, you can just click on small top and bottom part of it
2. Buttons text become two-line when there's not enough space in first half

https://github.com/user-attachments/assets/37ec405f-a553-4b20-a194-743cb21e77d8

3. Theme-switcher has different color on the top of it, which is not visible sometimes and looks like styles are broken

Sre
Screenshot 2026-06-24 at 16 14 54

Se-2
Screenshot 2026-06-24 at 17 24 36

This PR fixes all of it.

To reproduce and check the fixes

```
npx create-eth@2.0.21 -e https://github.com/scaffold-eth/se-2-challenges/tree/challenge-tokenization-new-ui
```

And then add all changes from this PR to the instance

### DIFF
--- a/packages/nextjs/components/SwitchTheme.tsx
+++ b/packages/nextjs/components/SwitchTheme.tsx
@@ -26,10 +26,11 @@ export const SwitchTheme = ({ className }: { className?: string }) => {
 
   return (
     <div className={`flex space-x-2 h-8 items-center justify-center text-sm ${className}`}>
+      {/* [--depth:0] removes DaisyUI's glossy top highlight so the indicator is a single flat color */}
       <input
         id="theme-toggle"
         type="checkbox"
-        className="toggle bg-secondary toggle-primary hover:bg-accent transition-all"
+        className="toggle bg-secondary toggle-primary hover:bg-accent transition-all [--depth:0]"
         onChange={handleToggle}
         checked={isDarkMode}
       />
--- a/packages/nextjs/components/Header.tsx
+++ b/packages/nextjs/components/Header.tsx
@@ -41,7 +41,7 @@ export const HeaderMenuLinks = () => {
               passHref
               className={`${
                 isActive ? "bg-base-300" : ""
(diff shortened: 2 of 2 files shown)


damianmarti (approved):
All fixes working fine now! Thanks!!

packages/nextjs/components/SwitchTheme.tsx
```diff
@@ -26,10 +26,11 @@ export const SwitchTheme = ({ className }: { className?: string }) => {
 
   return (
     <div className={`flex space-x-2 h-8 items-center justify-center text-sm ${className}`}>
+      {/* [--depth:0] removes DaisyUI's glossy top highlight so the indicator is a single flat color */}
       <input
         id="theme-toggle"
         type="checkbox"
-        className="toggle bg-secondary toggle-primary hover:bg-accent transition-all"
+        className="toggle bg-secondary toggle-primary hover:bg-accent transition-all [--depth:0]"
```
technophile-04, line 33: I was just researching a bit on the `--depth` flag, and it seems like it's been used only for interactive surfaces. For example, button, input text area, table, things where the user interacts. Since we have gone with the flattened theme like having the `.btn {box-shadow: none;}` set manuallly and globally. Do you feel that maybe we should just set this flag at global level as well? 

<details>

<summary>
    Example: 
</summary>

```css
/*
  DaisyUI 5 ignores --depth/--noise set inside @plugin "daisyui/theme" — it re-emits
  its own default --depth:1, so setting it in the theme block above does nothing.
  Override it here in plain (unlayered) CSS instead: this beats daisyUI's @layer base
  output, and since custom properties inherit, it flattens every component that reads
  var(--depth) — toggle, btn, inputs, select, tabs, etc.
*/
:root,
[data-theme] {
  --depth: 0;
}

```

</details>

technophile-04 (approved):
Thanks Rinat! added a small comment, but probably we can handle / disucss it later.  But this looks good!

technophile-04 (approved):

PR #1281, opened 2026-05-04
feat: web3 deps update

rin-st:
## Summary

Part 2 of #1279 

Updates web3 dependencies in `packages/nextjs`:

- `viem` 2.39.0 → 2.48.4
- `@rainbow-me/rainbowkit` 2.2.9 → 2.2.10
- `wagmi` stays on `2.19.5`

### Why wagmi stays on v2

Wagmi v3 is out, but **RainbowKit doesn't support it yet** — it still pins wagmi v2. Bumping wagmi alone breaks the connector types.

> Note: ConnectKit and Reown AppKit (ex-Web3Modal) also haven't shipped wagmi v3 support, so swapping the kit wouldn't unblock the upgrade.

### Code changes from the viem bump

- `publicClient.getBytecode(...)` → `publicClient.getCode(...)` (renamed in viem) in `ContractTabs.tsx` and `useDeployedContractInfo.ts`
- Enabled `@typescript-eslint/no-deprecated` in `eslint.config.mjs` to catch future viem/wagmi deprecations early
- Fixed stale JSDoc `useContractRead` → `useReadContract` in `useScaffoldReadContract.ts`

## Test plan

- [ ] `yarn next:check-types` clean
- [ ] `yarn next:lint` clean
- [ ] `yarn start` + connect via MetaMask, WalletConnect, burner — read/write contract works
- [ ] Block explorer detects EOA vs contract correctly (verifies `getCode` swap)

---

Waits for
- #1280
- https://github.com/scaffold-eth/burner-connector/pull/54. `burner-connector` is currently a local path link — switch back to a published version before merging.

~~⚠️ Before merge change burner-connector dependency. in package.json~~

PR #1310, opened 2026-07-28
fix: pin @coinbase/cdp-sdk to 1.52.0 to avoid @x402 module-not-found

rin-st:
⚠️ `create-eth@latest` instances (both hh and foundry) have this bug and couldn't start the frontend. Need to backmerge to `create-eth` after this merged

<p>

image

</p>
</details> 

Fresh installs resolve `@coinbase/cdp-sdk` to >=1.53.0 (via RainbowKit → `@wagmi/connectors` baseAccount → `@base-org/account`), which imports `@x402/*` packages declared as optional peer deps. They aren't installed, so `yarn start` fails with `Module not found: Can't resolve '@x402/core/client'`.

Pin cdp-sdk to 1.52.0 (last version without `@x402` imports) via root resolutions. Also protects new create-eth instances.

Remove the pin once Coinbase ships a fix.

--

just mentioning here #1281, maybe let's merge it too. It's becoming outdated

🤖 Generated with [Claude Code](https://claude.com/claude-code)


technophile-04 (approved):
Thanks Rinat! Merging this 🙌

PR #1316, opened 2026-08-22
fix: check both bytecode prefixes in decodeTxData for contract creation

### LINKED ISSUE #1246: Bug: decodeTxData.ts checks wrong bytecode prefix (0x60e06040 vs 0x60806040)

sanskar717:
Fixes #1246

The original check only matched `0x60e06040`, missing the standard
Solidity init code prefix `0x60806040`. Tested locally and found that
different compiler/optimizer configs produce different prefixes
(0x60806040 vs 0x60e06040), so this fix checks for both instead of
replacing one with the other — keeping the existing prefix as
suggested in the issue, while adding the standard one.

Note: A previous PR on this issue was closed by @technophile-04 with
feedback that the correct standard prefix is `0x60806040` (not
`0x6080604`, which is too broad and matches non-standard prefixes like
`0x60806041`). This PR uses the exact `0x60806040` prefix per that
guidance, and additionally keeps `0x60e06040` as a secondary check
since local testing (solc 0.8.30, via-ir) showed that config producing
that prefix instead of the standard one.

Verified locally: before the fix, contract creation txs showed
"⚠️ Unknown" in the block explorer. After the fix, they correctly
show as Contract Creation with no false Unknown label.

**Before:**
Screenshot 2026-08-22 171832

**After:**
Screenshot 2026-08-22 173014


technophile-04 (approved):
Thanks! Looks good to me!

sanskar717:
Thanks for the review and merge, @technophile-04 — appreciate the feedback on this, my first merged PR here!

PR #1317, opened 2026-08-24
fix(blockexplorer): trim live blocks by transaction, not by whole block

pucedoteth:
## The bug

The `watchBlocks` handler in `useFetchBlocks` keeps page 0 down to one page of transactions by popping **whole blocks** off the end:

```ts
while (transactionsInTrimmedBlocks > TRANSACTIONS_PER_PAGE && trimmedBlocks.length > 0) {
  const removedBlock = trimmedBlocks.pop();
  if (removedBlock) {
    transactionsInTrimmedBlocks -= removedBlock.transactions.length;
  }
}
```

Removing a whole block overshoots whenever the block crossing the boundary carries more than the overflow. The worst case is a block with more than `TRANSACTIONS_PER_PAGE` transactions: it is popped in a single step, the running total drops straight to zero, the loop exits, and the handler returns an **empty array** — so the block explorer shows nothing.

If that block arrives while the page is already full, it takes the twenty transactions that were on screen down with it. And because `fetchBlocks` only re-runs on `currentPage` / filter changes (not on new blocks), the page **stays blank** until the user navigates away and back.

Counting the same scenarios through the current and proposed logic:

| case | before | after |
| --- | --- | --- |
| automine: 20 × 1-tx blocks, new 1-tx block | 20 | 20 |
| interval mining: blocks of 3, new 3-tx block | 18 | 20 |
| 21-tx block arrives on an empty page | **0** | 20 |
| 25-tx block arrives onto a full page | **0** | 20 |
| oldest block is large (18 txs) | **3** | 20 |

(transactions visible on page 0 after the new block is handled)

The default **automine path is unaffected** — one transaction per block never straddles the boundary, which is why this isn't visible in the standard `yarn chain` flow. It shows up with interval mining (`anvil --block-time`, hardhat interval) or any script that lands several transactions in one block — the same non-automine setup #1295 dealt with.

## The fix

Trim by transaction instead of by block: keep whole blocks while they fit, and keep the straddling block with its newest transactions only.

That is already how the fetched page renders — `fetchPageItems` slices to `TRANSACTIONS_PER_PAGE` and `groupTransactionsByBlock` can emit a block holding a subset of its transactions — so a partially-shown block is not a new rendering state for `TransactionsTable`.

`hasNextPage` is computed from the untrimmed total, exactly as before.

## Testing

The repo has no test runner in `packages/nextjs` (no vitest/jest and no `test` script), so there is nowhere to add a unit test. The table above comes from running the old and new trimming logic side by side, extracted verbatim from this handler, over those five block layouts.

`yarn check-types` and `yarn lint` are clean, and the changed file passes `prettier --check`.

Happy to fold in a small test setup if you'd want one for this hook.

---
🤖 Written with [Claude Code](https://claude.com/claude-code).


technophile-04:
Closing this, lets first create an issue with evidence like image and reproducible example

